PR #66, opened 2026-03-06
feat(integrations): OpenClaw gateway integration + Anthropic top_p fix

vdk888:
## OpenClaw Gateway Integration

Adds a thin CLI bridge so Ouroboros Socratic interviews can run over any async chat platform (Telegram, WhatsApp, Discord, etc.) via OpenClaw agents.

---

### `src/ouroboros/integrations/openclaw_bridge.py` *(new)*
- Moved from repo root into proper package location
- Removed `sys.path.insert` hack — now invoked as `python -m ouroboros.integrations.openclaw_bridge`
- Added docstring explaining `litellm.modify_params = True` (cross-provider compatibility)
- Sessions dir defaults to `$SOCRATIC_SESSIONS_DIR` (no hardcoded paths)

### `integrations/openclaw/skills/socratic-spec/SKILL.md` *(new location)*
- Moved from `skills/socratic-spec/` into `integrations/openclaw/`
- Replaced hardcoded `~/openclaw/workspace/ouroboros-ref` with `$OUROBOROS_PATH` env var
- Updated all commands to use `python -m ouroboros.integrations.openclaw_bridge`

### `docs/integrations/openclaw.md` *(195-line integration guide)*
- Fixed `PR #XXX` placeholder → PR #82

### `README.md`
- Added OpenClaw to integrations section

---

## What's NOT in this PR

- **`top_p` fix for Anthropic** → now in PR #82 (`fix/anthropic-top-p`), merged independently

---

## Architecture

The delegation pattern is intentional: the bridge is a thin wrapper — it calls Ouroboros's actual `InterviewEngine`, `AmbiguityScorer`, and `SeedGenerator` directly. No reimplementation of logic.

Each CLI command does ONE step and exits. State persists to JSON files between turns, so the stateless chat agent can call commands sequentially across user messages.

---

## Tests

46 tests — `tests/unit/test_openclaw_bridge.py` covers all commands (start, respond, score, seed, status, complete) with mocked LLM calls.

### DIFF
--- a/src/ouroboros/integrations/openclaw_bridge.py
+++ b/src/ouroboros/integrations/openclaw_bridge.py
@@ -62,11 +62,7 @@ from ouroboros.providers.litellm_adapter import LiteLLMAdapter
 litellm.modify_params = True
 
 # --- Config ---
-SESSIONS_DIR = Path(
-    os.environ.get(
-        "SOCRATIC_SESSIONS_DIR", Path.home() / "socratic-sessions"
-    )
-)
+SESSIONS_DIR = Path(os.environ.get("SOCRATIC_SESSIONS_DIR", Path.home() / "socratic-sessions"))
 DEFAULT_MODEL = os.environ.get("SOCRATIC_MODEL", "claude-sonnet-4-5")
 
 


Q00:
Great work addressing all the previous feedback — the code quality is solid and CI is clean. Before we merge, a few questions:

**1. Use case & motivation**
What was the original motivation for building this? How do you see users actually using it — is there a specific workflow or audience you had in mind?

**2. API key requirement vs Max Plan**
Our core user base is mostly Claude Code + Max Plan users who don't use API keys directly. This bridge requires a separate API key via LiteLLM. How do you think about that gap? Is there a path to supporting Max Plan users, or is this targeting a different audience?

**3. Maintenance ownership**
As Ouroboros core APIs evolve (`InterviewEngine`, `AmbiguityScorer`, `SeedGenerator`), this integration layer will need to stay in sync. Would you be open to owning the `integrations/openclaw/` and `src/ouroboros/integrations/` folders as CODEOWNER? That way you'd get notified on any breaking changes and can keep things updated.

**4. Demo**
Do you have any screenshots or a short recording of this running in an actual chat platform (Telegram, Discord, etc.)? Would love to see it in action!

PR #84, opened 2026-03-07
feat(eval): make Stage 1 mechanical verification language-agnostic

aaddrick:
## Summary

Stage 1 (Mechanical Verification) was hardcoded to Python -- ruff, pytest,
mypy. If you ran `ooo evaluate` on a Zig, Rust, Go, or Node project, it
would fail immediately trying to find tools that weren't relevant.

This PR fixes that. It auto-detects the project language from marker files
and runs the appropriate commands for that language.

Detection scans `working_dir` for files like `build.zig`, `Cargo.toml`,
`go.mod`, `package-lock.json`, etc. Priority ordering matters here: `uv.lock`
is checked before `pyproject.toml` so a uv-managed project doesn't get
misidentified. Falls back gracefully when nothing is detected -- skips the
checks rather than erroring.

Supported languages out of the box: Python (uv and pip), Zig, Rust, Go, and
Node (npm, pnpm, bun, yarn). Projects with non-standard setups can drop a
`.ouroboros/mechanical.toml` override to specify their own commands.

**Files changed:**

1. `src/ouroboros/evaluation/languages.py` (new) -- `LanguageConfig` dataclass
   and `detect_language()` implementation
2. `src/ouroboros/evaluation/mechanical.py` -- refactored to pull commands from
   `LanguageConfig` instead of hardcoded strings
3. `src/ouroboros/mcp/tools/definitions.py` -- added `working_dir` param to
   `ouroboros_evaluate` so callers can pass the project root
4. `src/ouroboros/evaluation/__init__.py` -- exports the new types
5. `docs/guides/language-support.md` (new) -- user-facing docs for supported
   languages and override format
6. `tests/unit/evaluation/test_languages.py` (new) -- detection tests for all
   supported languages, custom overrides, and edge cases
7. `tests/unit/evaluation/test_mechanical.py` -- updated for the new approach

I tested this manually against Selkie (a Zig project). Stage 1 now passes
where it was previously failing.

## Test Plan

- [x] All existing tests pass
- [x] New unit tests cover detection logic for each supported language
- [x] Manually verified against a real Zig project (Selkie, built with `build.zig`)

### DIFF
--- a/docs/guides/language-support.md
+++ b/docs/guides/language-support.md
@@ -49,6 +49,14 @@ coverage_threshold = 0.5    # minimum coverage ratio (default: 0.7)
 2. **Auto-detected preset** — based on marker files
 3. **Skip** — if neither exists, checks are skipped
 
+### Security: Executable Allowlist
+
+Commands from `.ouroboros/mechanical.toml` are validated against an allowlist of known build/test/lint executables before execution. If a command uses an executable not on the list, it is silently skipped and a warning is logged.
+
+This prevents untrusted repositories from running arbitrary commands when evaluated in CI/CD environments. Hardcoded language presets bypass this check since they are trusted.
+
+If your tool is blocked, check the `_ALLOWED_EXECUTABLES` set in `src/ouroboros/evaluation/languages.py` and submit a PR to add it.
+
 ### Examples
 
 **Zig project with custom build flags:**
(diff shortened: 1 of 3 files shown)


Q00 (changes requested):
Two HIGH issues need fixing: (1) TOML parse error handling, (2) MechanicalConfig default change regression check. See detailed review comment.

aaddrick:
> Two HIGH issues need fixing: (1) TOML parse error handling, (2) MechanicalConfig default change regression check. See detailed review comment.

Roger doger

PR #85, opened 2026-03-07
fix(eval): use structured JSON output for LLM evaluation responses

aaddrick:
## Summary

I ran into this while testing the language-agnostic mechanical verification (#84). Stage 1 started passing for Zig projects, but Stage 2 semantic evaluation kept failing with "Could not find JSON in response." The LLM was returning prose preamble before the JSON, and the `extract_json_payload` parser couldn't find valid JSON in the response.

The root cause: the evaluation pipeline relies on prompt instructions to get JSON back, but nothing enforces it. The model starts with "Looking at the artifact..." or "I need to evaluate..." and either never produces JSON, or the JSON comes after the `max_tokens` cutoff.

This PR adds structured JSON output enforcement across all three evaluation pathways.

## What changed

**`CompletionConfig` (base.py)**
- Added `response_format` field. Defaults to `None` so existing callers are unaffected.

**`ClaudeCodeAdapter` (claude_code_adapter.py)**
- When `response_format` has type `json_schema`, passes `output_format` to the Agent SDK's `ClaudeAgentOptions`. The SDK handles structured output natively and returns `structured_output` on the `ResultMessage`.
- Extracts `structured_output` from `ResultMessage` when present.
- Strips OpenRouter provider prefixes from model names before passing to the Agent SDK. The SDK uses Anthropic's API directly and doesn't understand `openrouter/anthropic/claude-opus-4-6`. Non-Anthropic models (like `openrouter/openai/gpt-4o`) fall back to the SDK default.

**`LiteLLMAdapter` (litellm_adapter.py)**
- Passes `response_format` through to `litellm.acompletion()` when set. LiteLLM supports this natively for OpenAI and compatible providers.

**`AnthropicAdapter` (anthropic_adapter.py)**
- Anthropic's API doesn't support `response_format` directly. Uses assistant prefill with `{` to nudge JSON output, then prepends `{` to the response since it's not echoed back.

**Evaluation schemas**
- `semantic.py`: Added `SEMANTIC_RESULT_SCHEMA` with score, ac_compliance, goal_alignment, drift_score, uncertainty, reasoning. Strengthened the prompt to say "Respond with ONLY a JSON object."
- `consensus.py`: Added `VOTE_SCHEMA` and `JUDGMENT_SCHEMA` for the three consensus CompletionConfig calls.
- `qa.py`: Added `QA_VERDICT_SCHEMA` for the QA handler.

**Tests**
- 3 new tests: `test_evaluate_passes_json_response_format` (semantic), `test_includes_response_format_when_set` and `test_omits_response_format_when_none` (LiteLLM adapter).

## How it works

The `ClaudeCodeAdapter` path (used by the MCP server) now passes the schema to the Agent SDK via `output_format`. The SDK enforces the JSON structure at the API level. The existing `extract_json_payload` fallback parsers remain in place for safety.

For LiteLLM users, `response_format` passes through natively. For Anthropic direct users, the prefill technique nudges JSON output without hard enforcement.

## Test plan

- [x] All 2267 unit tests pass
- [x] `ruff check` and `ruff format --check` pass on all files
- [x] Tested end-to-end with a Zig project (Selkie): Stage 2 semantic evaluation now returns valid structured JSON instead of failing with parse errors
- [ ] Should be tested with Python projects to confirm no regressions on the standard path

### DIFF
--- a/src/ouroboros/evaluation/consensus.py
+++ b/src/ouroboros/evaluation/consensus.py
@@ -24,6 +24,7 @@ import json
 from ouroboros.core.errors import ProviderError, ValidationError
 from ouroboros.core.ontology_aspect import AnalysisResult
 from ouroboros.core.types import Result
+from ouroboros.evaluation.json_utils import extract_json_payload
 from ouroboros.evaluation.models import (
     ConsensusResult,
     DeliberationResult,
@@ -135,53 +136,6 @@ def build_consensus_prompt(context: EvaluationContext) -> str:
 ```
 
 Cast your vote as a JSON object with: approved (boolean), confidence (0-1), and reasoning."""
-
-
-def extract_json_payload(text: str) -> str | None:
-    """Extract JSON object from text using bracket-matching approach.
-
-    Uses brace counting to find the first complete JSON object,
-    avoiding issues with multiple disjoint brace blocks (e.g., code snippets).
-
-    Args:
-        text: Raw text potentially containing JSON
-
-    Returns:
-        Extracted JSON string or None if not found
-    """
-    start = text.find("{")
-    if start == -1:
-        return None
-
-    # Count braces to find matching closing brace
-    depth = 0
-    in_string = False
-    escape_next = False
-
-    for i, char in enumerate(text[start:], start=start):
-        if escape_next:
-            escape_next = False
-            continue
-
-        if char == "\\":
-            escape_next = True
-            continue
-
-        if char == '"' and not escape_next:
-            in_string = not in_string
-            continue
-
-        if in_string:
-            continue
-
-        if char == "{":
-            depth += 1
-        elif char == "}":
-            depth -= 1
-            if depth == 0:
-                return text[start : i + 1]
-
-    return None
 
 
 def parse_vote_response(response_text: str, model: str) -> Result[Vote, ValidationError]:
--- a/src/ouroboros/evaluation/json_utils.py
+++ b/src/ouroboros/evaluation/json_utils.py
@@ -1,0 +1,60 @@
+"""Shared JSON extraction utilities for evaluation modules.
+
+Provides a robust bracket-matching JSON extractor used by semantic,
+consensus, and QA evaluation stages.
+"""
+
+import re
+
+
+def extract_json_payload(text: str) -> str | None:
+    """Extract the first complete JSON object from text.
+
+    Uses brace-depth counting to find the first balanced ``{...}`` block,
+    correctly handling nested objects, strings with escaped characters,
+    and code-fenced responses.
+
+    Args:
+        text: Raw text potentially containing a JSON object
+
+    Returns:
+        Extracted JSON string, or None if no complete object is found
+    """
+    # Strip code fences first (```json ... ```)
+    fence_match = re.search(r"```(?:json)?\s*(\{[\s\S]*?\})\s*```", text)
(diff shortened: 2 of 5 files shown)


Q00 (changes requested):
## Code Review — PR #85

Solid approach to enforcing structured JSON output across evaluation pathways. The problem diagnosis (LLM returning prose before JSON) is spot-on, and the three-adapter strategy is well thought out. A few issues need addressing before merge.

---

### 🔴 Must Fix

**1. `AnthropicAdapter` — JSON prefill prepended before truncation produces broken JSON**

In `_parse_response()`, the `{` prefill is prepended to content *before* the `InputValidator.validate_llm_response()` truncation check runs. If the response hits `MAX_LLM_RESPONSE_LENGTH`, the JSON gets cut mid-object (e.g. `{"score": 0.9, "ac_comp...`), and callers receive silently broken JSON.

Fix: run truncation on raw content **before** prepending the `{` prefill.

**2. `semantic.py` and `consensus.py` have divergent `extract_json_payload` implementations**

`semantic.py` uses `text.rfind("}")` (naive, fails when response contains embedded code with `}` characters). `consensus.py` uses correct bracket-depth counting. Same function name, different logic — DRY violation and correctness bug.

Fix: extract the depth-counting version to a shared utility (e.g. `ouroboros.evaluation.utils`) and reuse it in both files.

**3. `ClaudeCodeAdapter` — verify `output_format` is a valid `ClaudeAgentOptions` field**

`options_kwargs["output_format"]` is passed to `ClaudeAgentOptions(**options_kwargs)`. If the Agent SDK doesn't have this field, every structured-output call via `ClaudeCodeAdapter` will raise `TypeError` at construction time — retried 5 times before returning a cryptic error. Please verify against the SDK version in use.

---

### 🟡 Should Fix

**4. `json_schema` response format missing required wrapper for OpenAI-compatible endpoints**

The OpenAI spec requires:
```json
{
  "type": "json_schema",
  "json_schema": {
    "name": "vote",
    "strict": true,
    "schema": { ... }
  }
}
```

Currently `VOTE_SCHEMA`, `SEMANTIC_RESULT_SCHEMA`, and `QA_VERDICT_SCHEMA` are passed directly as the `json_schema` value without the `name`/`schema` wrapper. OpenAI-compatible APIs will reject with 400.

**5. `languages.py` — command injection risk via `.ouroboros/mechanical.toml`**

TOML-sourced command strings are passed through `shlex.split` then to `subprocess_exec` without sanitization. In CI/CD environments evaluating untrusted repos, this is a shell injection vector. Consider validating command executables against an allowlist, or at minimum document the trust boundary.

---

### 🟠 Scope

**6. `languages.py`, `mechanical.py` changes, and `docs/guides/language-support.md` appear to belong in PR #84**

The `MechanicalConfig` default behavior changed from "run all checks" (hardcoded `uv` commands) to "skip all checks" (`None` defaults). Code that constructs `MechanicalConfig()` directly will silently pass all checks. Consider splitting these into #84 to keep this PR focused on structured JSON output.

---

### Strengths

- Clean `response_format` opt-in via `CompletionConfig` — existing callers unaffected
- Anthropic prefill technique is pragmatic given API limitations
- Good test coverage for the new paths (3 new test files)
- Keeping `extract_json_payload` fallback as safety net is wise defensive design

Q00 (approved):
LGTM!

Issue #121, opened 2026-03-11
[Bug] MCP server crashes on Windows: os.kill(pid, 0) raises WinError 87


formicoidea:
## Description

  The MCP server fails to start on Windows when a stale PID file exists at `~/.ouroboros/mcp-server.pid`. The `_check_stale_instance()` function uses `os.kill(pid, 0)` to check if an old process is still running, but **signal 0 is a
  POSIX-only convention** and is not supported on Windows.

  ## Error

  ```
  OSError: [WinError 87] Paramètre incorrect
  ```

  Full traceback points to `ouroboros/cli/commands/mcp.py`, line 56:

  ```python
  os.kill(old_pid, 0)  # Signal 0 = check existence
  ```

  ## Steps to reproduce

  1. Run `ouroboros mcp serve` on Windows (starts successfully, creates PID file)
  2. Stop the server (Ctrl+C, crash, or session end)
  3. Run `ouroboros mcp serve` again → crash due to stale PID file

  ## Environment

  - OS: Windows 11
  - Python: 3.14.3
  - ouroboros-ai: 0.21.0
  - mcp SDK: 1.26.0

  ## Workaround

  Manually delete `~/.ouroboros/mcp-server.pid` before restarting.

  ## Suggested fix

  Replace `os.kill(pid, 0)` with a Windows-compatible process existence check:

  ```python
  import sys

  def _is_process_running(pid: int) -> bool:
      if sys.platform == "win32":
          import ctypes
          kernel32 = ctypes.windll.kernel32
          PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
          handle = kernel32.OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, False, pid)
          if handle:
              kernel32.CloseHandle(handle)
              return True
          return False
      else:
          try:
              os.kill(pid, 0)
              return True
          except ProcessLookupError:
              return False
          except PermissionError:
              return True  # process exists but we can't signal it
  ```

  Or more simply, wrap the existing `os.kill` call in a `try/except OSError` that treats `WinError 87` as "process not found".

PR #123, opened 2026-03-11
fix: handle Windows OSError in _check_stale_instance() (#121)

Q00:
## Summary

- `os.kill(pid, 0)` in `_check_stale_instance()` is POSIX-only — on Windows it raises `OSError (WinError 87)` instead of `ProcessLookupError`, crashing MCP server startup when a stale PID file exists.
- Added a fallback `except OSError` clause that treats the error as "process not found" and cleans up the stale PID file, consistent with the existing `ProcessLookupError` handling.

## Test plan

- [x] `uv run pytest tests/unit/cli/ -x -q` — 154 passed
- [x] `uv run pytest tests/ --ignore=tests/unit/mcp --ignore=tests/integration/mcp --ignore=tests/e2e -x -q` — 2314 passed

### DIFF
--- a/src/ouroboros/cli/commands/mcp.py
+++ b/src/ouroboros/cli/commands/mcp.py
@@ -60,6 +60,11 @@ def _check_stale_instance() -> bool:
         return True
     except PermissionError:
         return False  # Process exists but we can't signal it
+    except OSError:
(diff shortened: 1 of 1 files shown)


Q00:
closes #121

PR #141, opened 2026-03-14
feat(tui): add native SLT backend as alternative TUI monitor

subinium:
## What

Adds `ouroboros-tui` — a native TUI monitor built with [SuperLightTUI](https://github.com/subinium/SuperLightTUI) — as an opt-in alternative to the existing Python/Textual TUI.

**Default behavior is unchanged.** The Python TUI remains the default when running `ouroboros tui monitor`.

## How to use

```bash
# Option 1: via CLI flag (requires binary in PATH)
ouroboros tui monitor --backend slt

# Option 2: standalone binary
ouroboros-tui
ouroboros-tui --mock    # demo mode without DB
ouroboros-tui --help
```

### Install the binary

```bash
cd crates/ouroboros-tui
cargo install --path .
```

## What's in the PR

**New: `crates/ouroboros-tui/`** (Rust, ~2,600 lines)
- Reads `~/.ouroboros/ouroboros.db` (same EventStore, same schema)
- 6 screens matching the Python TUI: Dashboard, Execution, Logs, Debug, Lineage, Session Selector
- Same keybindings: `q`, `p`/`r`, `1-5`, `l`, `d`, `e`, `s`, `Ctrl+P`
- Logs view uses SLT 0.6.1 Table widget (sortable headers, filtering, pagination)
- 0.5s DB polling, 60fps rendering
- `--mock` mode for demo without a running workflow
- Cargo.lock committed (binary crate best practice)

**Modified: `src/ouroboros/cli/commands/tui.py`** (+30 lines)
- Adds `--backend slt` option to `ouroboros tui monitor`
- Locates `ouroboros-tui` binary via `shutil.which()`, passes `--db-path`
- Clear error message with install instructions if binary not found
- No change to default behavior

## Why SLT

- Single static binary (~2.7 MB), no Python runtime dependency
- Built on [SuperLightTUI](https://github.com/subinium/SuperLightTUI): immediate-mode Rust TUI with flexbox layout, 30+ widgets
- 2 runtime dependencies (crossterm + unicode-width)

## Not in scope

- Pre-built release binaries (can add CI workflow later)
- Replacing the Python TUI as default
- Changes to EventStore schema or event format

### DIFF
--- a/crates/ouroboros-tui/Cargo.lock
+++ b/crates/ouroboros-tui/Cargo.lock
@@ -348,9 +348,9 @@ checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 
 [[package]]
 name = "superlighttui"
-version = "0.6.1"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d10b817ff82b3f76387e073316640bc4a0b6b62561a1c3e16c6d64847f0c4652"
+checksum = "355035a12e04a280e46508bcd8aec980c8a5e61adcfa18939cb1da7cff3f6ad6"
 dependencies = [
  "crossterm",
  "unicode-width",
--- a/crates/ouroboros-tui/Cargo.toml
+++ b/crates/ouroboros-tui/Cargo.toml
@@ -13,6 +13,6 @@ name = "ouroboros-tui"
 path = "src/main.rs"
 
 [dependencies]
-superlighttui = "0.6.1"
+superlighttui = "0.7.1"
 rusqlite = { version = "0.33", features = ["bundled"] }
 serde_json = "1"
--- a/crates/ouroboros-tui/src/main.rs
+++ b/crates/ouroboros-tui/src/main.rs
@@ -189,18 +189,20 @@ fn main() -> std::io::Result<()> {
 }
 
 fn handle_global_keys(ui: &mut Context, state: &mut AppState) {
+    let on_logs = state.tabs.selected == 2;
+
     if ui.key('q') {
         ui.quit();
     }
     if ui.key_mod('p', KeyModifiers::CONTROL) {
         state.command_palette.open = !state.command_palette.open;
     }
-    if ui.key('p') && !state.command_palette.open {
+    if ui.key('p') && !state.command_palette.open && !on_logs {
         state.is_paused = true;
         state.status = ExecutionStatus::Paused;
         state.add_log(LogLevel::Info, "tui", "Execution paused by user");
     }
-    if ui.key('r') && !state.command_palette.open {
+    if ui.key('r') && !state.command_palette.open && !on_logs {
         state.is_paused = false;
         state.status = ExecutionStatus::Running;
         state.add_log(LogLevel::Info, "tui", "Execution resumed");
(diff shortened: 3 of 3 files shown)


Q00 (commented):
Hey @subinium, thanks for putting this together — really cool to see a native TUI option for Ouroboros.

The SLT immediate-mode API keeps things impressively concise. 6 screens with tree + table + sparklines + command palette in ~2,600 lines is no joke. I especially like that the integration is fully opt-in — zero changes to default behavior. The `--mock` mode is a nice touch for demos too.

A few things I spotted while going through the code:

---

**1. UTF-8 panic in `short_payload` (db.rs) and `truncate` (state.rs)**

Both functions slice strings by byte index:

```rust
// db.rs
format!("{}...", &s[..117])

// state.rs
format!("{}...", &s[..max.saturating_sub(3)])
```

Rust panics if the slice lands in the middle of a multi-byte character. Our interview responses and seed goals can contain Korean text — a 40+ character Korean sentence easily exceeds 120 bytes, and the 117th byte will likely split a 3-byte hangul character. This would crash the TUI on load for any session with Korean content in the payload.

Safe alternative:

```rust
fn truncate(s: &str, max: usize) -> String {
    if s.chars().count() <= max {
        s.to_string()
    } else {
        let t: String = s.chars().take(max.saturating_sub(3)).collect();
        format!("{t}...")
    }
}
```

**2. Log filter silently overwrites text filter (logs.rs)**

```rust
if !filter_text.is_empty() {
    state.log_table.set_filter(&filter_text);
}
if let Some(level) = state.log_level_filter {
    state.log_table.set_filter(level.label()); // overwrites the above
}
```

When both are active, the level filter wins and the text filter is silently dropped. Users would expect them to stack. Could either combine them or make the UI clarify they're mutually exclusive.

**3. Page counter always shows last page (logs.rs)**

```rust
let page = if total_pages > 0 {
    total_pages.min(total_pages)  // this is always total_pages
} else { 1 };
```

Think you meant `current_page` here — right now it always shows "Page N/N".

**4. Docs update (can be a follow-up PR)**

The TUI docs (`docs/guides/tui-usage.md`, `docs/cli-reference.md`, `docs/guides/cli-usage.md`) don't mention `--backend slt` yet. Happy for this to land in a separate PR if you prefer — just flagging so it doesn't get forgotten.

---

Item 1 is the main one I'd want addressed before merge — the rest are quick fixes. Happy to re-review once updated. Great work on this!

subinium:
Thanks for the thorough review! Here's the status on each item:

**1. UTF-8 panic ✅** — Switched both `short_payload` (db.rs) and `truncate` (state.rs) from byte-index slicing to `chars().count()` / `chars().take()`. Korean payloads won't crash anymore. (`a817f9d`)

**2. Log filter stacking 🚧** — Restructured the filter logic to combine text + level filters instead of overwriting. However, this requires an upstream SLT change — `set_filter` currently treats the entire string as a single needle, so multi-token matching (e.g. `"ERROR deploy"`) doesn't work yet. Working on a [multi-token AND filter](https://github.com/subinium/SuperLightTUI) update for SLT; will bump the dep and confirm once that lands.

**3. Page counter ✅** — Fixed to use `state.log_table.page + 1` (0-based field) instead of `total_pages.min(total_pages)`. (`a817f9d`)

**4. Docs ✅** — Added `--backend slt` to `tui-usage.md`, `cli-reference.md`, and `cli-usage.md`. (`a817f9d`)

Will update once #2 is fully resolved!

PR #142, opened 2026-03-14
feat(eval): add Java Maven language preset

### LINKED ISSUE #135: Add Java Maven to the languages supported out of the box
You already support some number of languages out the box, adding Maven support would be relatively straight forward as it already has simple reliable compile and test phases (although I'm not sure what the lint would be).

hackertaco:
## Summary
- Add `java-maven` language preset with `mvn clean compile` (build) and `mvn test` (test)
- Lint/static/coverage set to None — users opt-in via `.ouroboros/mechanical.toml`
- Detection via `pom.xml` marker file, placed after Go and before Node in priority order
- Docs updated with Maven example and toml override example

## Design Decision
Followed Zig's minimal pattern (build + test only) because Maven's checkstyle/spotbugs require explicit pom.xml plugin config unlike Rust's `cargo clippy` or Go's `go vet` which work out of the box. Validated by contrarian + architect agent review.

## Test Plan
- [x] `test_detect_java_maven` — pom.xml detection
- [x] `test_pom_xml_detected_before_package_json` — Maven > Node priority
- [x] `test_pom_xml_detected_before_node_lockfiles` — Maven > Node lockfiles
- [x] `test_go_mod_detected_before_pom_xml` — Go > Maven priority
- [x] `test_auto_detect_java_maven` — config building with all commands verified
- [x] `test_java_maven_preset_has_build_and_test_only` — preset values
- [x] `test_java_maven_preset_no_quiet_flags` — no -q flags
- [x] `test_java_maven_preset_is_frozen` — immutability
- [x] All 40 language tests pass

Closes #135

### DIFF
--- a/docs/guides/language-support.md
+++ b/docs/guides/language-support.md
@@ -11,6 +11,7 @@ Ouroboros Stage 1 (Mechanical Verification) auto-detects your project's language
 | Zig | `build.zig` | — | `zig build` | `zig build test` | — | — |
 | Rust | `Cargo.toml` | `cargo clippy` | `cargo build` | `cargo test` | — | — |
 | Go | `go.mod` | `go vet ./...` | `go build ./...` | `go test ./...` | — | `go test -cover ./...` |
+| Java (Maven) | `pom.xml` | — | `mvn clean compile` | `mvn test` | — | — |
 | Node (npm) | `package-lock.json` | `npm run lint` | `npm run build` | `npm test` | — | — |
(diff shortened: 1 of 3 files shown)


Q00 (approved):
Clean addition — follows the existing preset pattern exactly, detection priority makes sense (Go > Maven > Node), and the design decision to exclude checkstyle/spotbugs from defaults is the right call since they need explicit pom.xml plugin config unlike cargo clippy or go vet.

Tests are thorough. LGTM.

Minor note for a future PR: many Java projects use `./mvnw` (Maven Wrapper) instead of `mvn` — could be worth detecting `mvnw` and preferring it when present. But that's a separate enhancement.

Issue #98, opened 2026-03-08
Inspired by OctopusGarden


Q00:
Patterns and ideas from [OctopusGarden](https://github.com/foundatron/octopusgarden) by [@foundatron](https://github.com/foundatron) that would complement Ouroboros.

Ouroboros kills ambiguity at the front (Socratic interview → seed). OctopusGarden validates at the back (holdout scenarios + LLM-as-judge). Together, they close the loop.

## Sub-issues

- [ ] #91 — **Holdout Validation (LLM-as-Judge)**: Information barrier where the generator never sees test scenarios
- [ ] #93 — **Adaptive Feedback Fidelity**: Progressive detail scaling based on iteration count and stall status
- [ ] #96 — **Gene Transfusion**: Extract coding pattern DNA from exemplar projects and inject into new ones
- [ ] #97 — **Scenario Preflight**: 4-dimension AC quality gate (coverage, feasibility, isolation, chains)

## Credit

All ideas above are inspired by OctopusGarden's implementation. Each sub-issue references the specific source files.

## Key OG source files

- `internal/scenario/judge.go` — holdout validation + LLM judge
- `internal/attractor/convergence.go` — feedback fidelity tiers
- `internal/gene/` — pattern DNA extraction/injection
- `internal/preflight/scenario.go` — scenario quality gate

Issue #97, opened 2026-03-08
feat(eval): scenario preflight — 4-dimension quality gate before execution


Q00:
## Summary

Before executing against acceptance criteria, run a preflight check that scores scenario/AC quality across four dimensions: coverage, feasibility, isolation, and logical chaining.

## Motivation

Bad acceptance criteria waste entire execution cycles. If ACs don't cover all seed behaviors, the generated code will have blind spots. If ACs are interdependent (not isolated), a single failure cascades. Catching these issues BEFORE execution saves time and tokens.

This is the validation-side complement to ouroboros's AmbiguityScorer (which checks spec quality). Together they ensure both the spec AND the test criteria are solid before any code runs.

## Design (from OctopusGarden)

Four scoring dimensions (each 0.0–1.0):

| Dimension | Question |
|-----------|----------|
| **Coverage** | Do ACs cover all behaviors described in the seed? |
| **Feasibility** | Can each AC actually be verified mechanically or by LLM judge? |
| **Isolation** | Are ACs independent? Can each be evaluated without others passing first? |
| **Chains** | For multi-step ACs, do the steps logically flow? |

```python
@dataclass
class ScenarioPreflightResult:
    coverage: float
    feasibility: float
    isolation: float
    chains: float
    aggregate: float  # simple average
    passed: bool      # aggregate >= threshold (default 0.6)
    issues: list[ACIssue]  # specific problems per AC

@dataclass
class ACIssue:
    ac_id: str
    dimension: str  # "coverage" | "feasibility" | "isolation" | "chains"
    detail: str     # human-readable explanation
```

## Integration with Ouroboros

- Run automatically after `ooo seed` generates acceptance criteria.
- Gate `ooo run` — refuse to execute if preflight score < threshold.
- Surface issues in `ooo evaluate` output so they can be fixed via `ooo evolve`.
- The existing 3-stage evaluation pipeline could add this as "Stage 0" (pre-execution quality gate).
- Works alongside AmbiguityScorer: ambiguity checks the spec, preflight checks the ACs.

## Inspired by

[OctopusGarden](https://github.com/foundatron/octopusgarden) by [@foundatron](https://github.com/foundatron) — specifically `internal/preflight/scenario.go` with its 4-dimension scoring system and `ScenarioIssue` structure.

Issue #96, opened 2026-03-08
feat(seed): gene transfusion — extract coding pattern DNA from exemplar projects


Q00:
## Summary

Extract coding patterns, invariants, and architectural structure from a completed reference project, then inject them as contextual guidance when generating new code from a seed. Like transferring DNA from a healthy organism to a new one.

## Motivation

When ouroboros generates code, it starts from scratch each time. But most teams have existing projects with established patterns (error handling style, project structure, testing conventions, API design). Gene transfusion captures these patterns once and reuses them across all future seeds — ensuring consistency without manual style guides.

## Design (from OctopusGarden)

**Extraction Phase** (run once per exemplar):

File selection strategy within 20,000 token budget:
1. `marker`: `pyproject.toml`, `go.mod`, `package.json` (highest priority)
2. `readme`: README.md (first 100 lines)
3. `dockerfile`: Dockerfile
4. `entrypoint`: `main.py`, `cmd/*/main.go`, `app.ts`
5. `handler`: Largest file in `handlers/`, `routes/`, `api/`
6. `model`: Largest file in `models/`, `types/`, `schemas/`

LLM extracts structured gene guide with sections:
- **PATTERN**: Architectural patterns (e.g., "hexagonal architecture", "repository pattern")
- **INVARIANTS**: Hard rules (e.g., "all errors wrapped with context", "no raw SQL")
- **EDGE CASES**: Known gotchas the exemplar handles
- **STACK**: Technology choices with versions
- **STRUCTURE**: Directory layout convention
- **BOOT**: Initialization sequence
- **BUILD**: Build and deployment pipeline

**Injection Phase** (during `ooo run`):
```
PROVEN PATTERNS (extracted from a working exemplar — synthesize equivalent 
behavior adapted to the specification above. Preserve the structural approach 
and invariants. The SPECIFICATION always takes precedence over these patterns 
on any conflict):
{gene_guide}
```

Cross-language support: If gene is from Go but seed targets Python, the injection prompt explicitly notes this and asks for idiomatic translation.

## Integration with Ouroboros

- New command: `ooo gene extract <path>` → saves gene guide as YAML/JSON
- Gene reference in seed: `METADATA.gene: ./genes/my-backend-gene.yaml`
- `ooo run` injects gene into system prompt when present
- Gene library: `~/.ouroboros/genes/` for reusable patterns across projects
- Could integrate with `ooo interview` — "Do you have an existing project to base patterns on?"

## Inspired by

[...]

Issue #93, opened 2026-03-08
feat(eval): adaptive feedback fidelity — progressive detail scaling


Q00:
## Summary

Dynamically adjust the detail level of evaluation feedback based on iteration count and stall status. Start compact (save tokens), escalate to full detail when stuck.

## Motivation

In the `ooo ralph` loop or `ooo evolve`, early iterations often succeed with minimal feedback. Sending full evaluation details from iteration 1 wastes tokens. But when stuck at iteration 5+, the LLM needs maximum context to break through.

## Design (from OctopusGarden)

Three fidelity tiers:

| Tier | When | Content | Max Size |
|------|------|---------|----------|
| **Compact** | Iterations 1-2 | AC summary only, pass/fail | ~4KB |
| **Standard** | Iterations 3-4 | Failed AC details, skip passing ones | ~12KB |
| **Full** | Iterations 5+ | All AC details, full evaluation output | ~24KB |

Auto-escalation rule: if stall count ≥ 2, bump fidelity one tier early.

```python
def determine_fidelity(iteration: int, stall_count: int) -> FeedbackFidelity:
    if iteration <= 2:
        base = FeedbackFidelity.COMPACT
    elif iteration <= 4:
        base = FeedbackFidelity.STANDARD
    else:
        base = FeedbackFidelity.FULL
    
    if stall_count >= 2 and base < FeedbackFidelity.FULL:
        base = FeedbackFidelity(base.value + 1)
    
    return base
```

## Integration with Ouroboros

- Apply in `reflect.py` when constructing the evaluation context for Wonder/Reflect.
- Apply in Ralph loop when feeding QA results back into the next iteration.
- The `ooo evaluate` output could include a `fidelity` field indicating what level was used.
- Estimated token savings: 40-60% on successful runs (most converge in iterations 1-3).

## Inspired by

[OctopusGarden](https://github.com/foundatron/octopusgarden) by [@foundatron](https://github.com/foundatron) — specifically `internal/attractor/convergence.go` feedback fidelity system.

PR #174, opened 2026-03-21
feat(eval): add reward_hacking_risk to Stage 2 semantic evaluation

Q00:
## Summary
- Add `reward_hacking_risk` (0.0-1.0) to `SemanticResult` — detects artifacts that game the evaluator instead of solving the real task
- Add Anti-Gaming Verification to Stage 2 prompt — zero additional LLM calls, zero cost
- Add **Wonder Scope Guard** — prevents ontology over-expansion by bounding Wonder questions to the seed's goal and constraints

## Context
Inspired by [OctopusGarden #98](https://github.com/Q00/ouroboros/issues/98). Reward hacking was observed multiple times where generated code passed AC checks by gaming the evaluation.

### What we tried first: Ralph loop (7 generations)
We ran a full Ralph loop to implement reward-hacking prevention + compact feedback tiers. The result: **+12,970 LOC**, 9 bugs (2 critical), and massive over-engineering. Wonder engine expanded ontology from 4 to 9 fields in 3 generations, then oscillated (9→8→7→8) as it tried and failed to prune — never converging.

**Root cause discovered**: Wonder treated ontological incompleteness as ambiguity. Since every ontology is inherently incomplete (Gödel), Wonder always found something to add — infinite recursion. The ontology is a shadow of the seed (Plato's cave); making the shadow more detailed doesn't bring it closer to the Form.

### What we did instead: +64 LOC
- **Layer 1** (commit 1): Deepen the existing Stage 2 prompt to ask "does this code actually work, or does it just look like it satisfies the AC?" — same LLM call, $0 cost (Socrates: "know thyself" applied to the evaluator)
- **Wonder Scope Guard** (commit 2): Bound Wonder's questions to the seed's goal and constraints. "An ontology is ALWAYS incomplete — that is normal, not a gap to fill."

### Commit 1: `feat(eval): add reward_hacking_risk` (+26 LOC)
- `models.py`: `SemanticResult.reward_hacking_risk: float = 0.0` with range validation
- `semantic.py`: JSON schema + anti-gaming prompt section + parsing
- `events/evaluation.py`: `reward_hacking_risk` in Stage 2 completion event

### Commit 2: `fix(evolve): Wonder scope guard` (+38 LOC)
- `wonder.py`: SCOPE GUARD in system prompt + seed context in user prompt
- `loop.py`: Pass `seed` to `WonderEngine.wonder()`

## Checks passed
- [x] ruff lint
- [x] ruff format
- [x] mypy
- [x] pytest (258 passed)

## Test plan
- [x] Existing 258 evaluation tests pass — backward compatible (default 0.0)
- [x] Manual: Run `ooo evaluate` on known reward-hacking artifacts to verify detection
- [x] Manual: Run `ooo ralph` and verify Wonder no longer adds unbounded ontology fields
- [ ] Monitor: Track `reward_hacking_risk` distribution via PostHog

## Deferred to 0.27+
- Layer 2: Independent probe system (if Layer 1 proves insufficient)
- FeedbackTier (Compact/Standard/Full) for token savings
- OctopusGarden sub-issues: #91 holdout, #93 feedback fidelity, #96 gene transfusion, #97 preflight

### DIFF
--- a/src/ouroboros/core/lineage.py
+++ b/src/ouroboros/core/lineage.py
@@ -147,12 +147,14 @@ class EvaluationSummary(BaseModel, frozen=True):
 
     @model_validator(mode="after")
     def _reconcile_approval_with_ac_results(self) -> "EvaluationSummary":
-        """Eagerly reconcile approval_status with AC results at construction time."""
+        """Eagerly reconcile approval fields with AC results at construction time."""
         if self.ac_results:
             ac_passed = all(ac.passed for ac in self.ac_results)
-            expected = "approved" if ac_passed else "rejected"
-            if self.approval_status != expected:
-                object.__setattr__(self, "approval_status", expected)
+            expected_status = "approved" if ac_passed else "rejected"
+            if self.approval_status != expected_status:
+                object.__setattr__(self, "approval_status", expected_status)
+            if self.final_approved != ac_passed:
+                object.__setattr__(self, "final_approved", ac_passed)
         return self
 
     @property
--- a/tests/unit/core/test_lineage.py
+++ b/tests/unit/core/test_lineage.py
@@ -188,6 +188,7 @@ class TestEvaluationSummary:
         # Deliberately do NOT access run_verdict_passed before serializing
         payload = summary.model_dump(mode="json")
         assert payload["approval_status"] == "approved"
+        assert payload["final_approved"] is True
 
 
 class TestACResult:


Q00:
## Review Response — Round 9

Commit: `d0ee644`

### Blocker fixed

**Finding**: `_reconcile_approval_with_ac_results` reconciled `approval_status` but left `final_approved` stale → split-brain in serialized payload.

**Fix**: After-validator now also reconciles `final_approved` when AC results override it:
```python
if self.final_approved != ac_passed:
    object.__setattr__(self, "final_approved", ac_passed)
```

**Serialization test** now asserts both fields:
```python
payload = summary.model_dump(mode="json")
assert payload["approval_status"] == "approved"
assert payload["final_approved"] is True  # ← new
```

No field in the serialized payload can disagree with any other. Single authoritative truth.

```
2743 passed, 7 skipped
```

PR #237, opened 2026-03-27
feat: add additional_tools parameter to execute_seed (#181)

### LINKED ISSUE #181: Enhancement: Inherit parent MCP tools in direct MCP-initiated ooo run executions
## Summary

Follow-up to #176 and the excellent fix in PR #180. The nested delegation inheritance works perfectly — child `ooo run` executions inherit parent MCP tools via `PreToolUse` hook injection and session forking. However, **direct** MCP-initiated executions (calling `ouroboros_execute_seed` / `ouroboros_start_execute_seed` from the main Claude session) do not inherit the parent session's MCP tools.

## Current Behavior

When a user calls `ooo run` from their Claude Code session (which has MCP tools like `mcp__github__*`, `mcp__chrome-devtools__*`), the `ExecuteSeedHandler.handle()` in `definitions.py` creates an `OrchestratorRunner` without `mcp_manager` and without inherited tools:

```python
# definitions.py:238-246
runner = OrchestratorRunner(
    adapter=agent_adapter,
    event_store=event_store,
    console=console,
    debug=False,
    enable_decomposition=True,
    inherited_runtime_handle=inherited_runtime_handle,  # None for direct calls
    inherited_tools=inherited_effective_tools,           # None for direct calls
)
```

The `inherited_runtime_handle` and `inherited_effective_tools` are only populated when the internal delegation fields (`_ooo_parent_claude_session_id`, `_ooo_parent_effective_tools`) are present in the tool arguments — which only happens when a parent orchestrator's `PreToolUse` hook injects them during nested execution.

For direct calls from the main Claude session, these fields are absent, so the runner starts with only `DEFAULT_TOOLS = ["Read", "Write", "Edit", "Bash", "Glob", "Grep"]`.

## Evidence

Smoke test on v0.25.1: a seed with AC "use `mcp__github__issue_read` to read issue 176" completed successfully, but the subagent's own output noted:

> "The `mcp__github__issue_read` MCP tool was not registered in this environment's tool set, so I used the GitHub REST API via `curl` as a fallback."

The subagent used Bash + curl (available in DEFAULT_TOOLS) instead of the MCP tool.

## Desired Behavior

Direct MCP-initiated `ooo run` executions should inherit the parent Claude session's effective MCP tool set, just as nested delegated runs do after PR #180.

## Possible Approaches

The `ExecuteSeedHandler` runs inside the Ouroboros MCP server process, which doesn't have direct access to the parent Claude session's MCP tool registry. The delegation mechanism in PR #180 works because the parent orchestrator (running inside Claude) can inject metadata via hooks. For direct calls, the MCP server would need a way to discover which MCP tools the calling Claude session has available.

Some ideas (for maintainer consideration):
1. **Claude Agent SDK session info** — if `claude_agent_sdk` exposes the parent session's tool list, the handler could read it
2. **MCP server configuration** — allow the Ouroboros MCP server config to specify additional tools that should be available to all executions
3. **Tool list as explicit parameter** — add an optional `additional_tools` parameter to `execute_seed` for the caller to pass their available tools

## Context

- PR #180 fixed nested delegation perfectly — this is a follow-up enhancement, not a regression
- Workaround: run MCP-dependent tasks at the main conversation level instead of via `ooo run`
- Related: #176 (original bug report)

## Environment

- Ouroboros v0.25.1
- Claude Code with GitHub MCP + Chrome DevTools MCP configured in `~/.claude.json`

jcfernandez-890825:
## Summary

Add an optional `additional_tools` parameter to `ouroboros_execute_seed` and `ouroboros_start_execute_seed` that allows callers to explicitly pass MCP tool names from the parent Claude session. These tools are merged with the existing default tool set when creating the `OrchestratorRunner`.

**Replaces #223** (clean single-commit rebase on current main).

## Changes

**`src/ouroboros/mcp/tools/execution_handlers.py`:**
- Add `additional_tools` parameter (type: `ARRAY`, `items={type: string}`, optional)
- Merge with `inherited_effective_tools` using `dict.fromkeys()` for dedup
- Guarded by `not is_resume` — resumed sessions preserve original tool set
- `StartExecuteSeedHandler` inherits automatically via shared parameter list

**`tests/unit/mcp/tools/test_definitions.py`:**
- `test_definition_includes_additional_tools_parameter` — schema validation including `items`
- `test_additional_tools_merged_into_inherited_tools` — direct-call merge
- `test_no_additional_tools_preserves_default_behavior` — default `None`
- `test_additional_tools_ignored_on_resume` — resume regression test

## Design

- **Explicit opt-in** — caller must pass tool names; no session introspection
- **Merge, not replace** — works alongside PR #180 delegation mechanism
- **Resume-safe** — `additional_tools` ignored when `session_id` is provided

Closes #181

### DIFF
--- a/src/ouroboros/mcp/tools/execution_handlers.py
+++ b/src/ouroboros/mcp/tools/execution_handlers.py
@@ -170,6 +170,19 @@ class ExecuteSeedHandler:
                     required=False,
                     default=False,
                 ),
+                MCPToolParameter(
+                    name="additional_tools",
+                    type=ToolInputType.ARRAY,
+                    description=(
+                        "Optional list of additional tool names to make available during "
+                        "execution. These are merged with the default tool set (Read, Write, "
+                        "Edit, Bash, Glob, Grep). Use this to pass MCP tools from the parent "
+                        "Claude session (e.g. ['mcp__github__issue_read', "
+                        "'mcp__chrome-devtools__click'])."
+                    ),
+                    required=False,
+                    items={"type": "string"},
+                ),
             ),
         )
 
@@ -257,6 +270,21 @@ class ExecuteSeedHandler:
         inherited_effective_tools = (
             None if is_resume else _extract_inherited_effective_tools(arguments)
         )
+
+        # Merge caller-supplied additional_tools (only for new executions)
+        additional_tools = arguments.get("additional_tools") if not is_resume else None
+        if isinstance(additional_tools, list) and additional_tools:
+            extra = list(dict.fromkeys(t for t in additional_tools if isinstance(t, str) and t))
+            if extra:
+                if inherited_effective_tools:
+                    seen = set(inherited_effective_tools)
+                    inherited_effective_tools = list(inherited_effective_tools)
+                    for tool in extra:
+                        if tool not in seen:
+                            inherited_effective_tools.append(tool)
+                            seen.add(tool)
+                else:
+                    inherited_effective_tools = extra
 
         log.info(
             "mcp.tool.execute_seed",
--- a/tests/unit/mcp/tools/test_definitions.py
+++ b/tests/unit/mcp/tools/test_definitions.py
@@ -1015,6 +1015,173 @@ metadata:
 """
 
 
+class TestAdditionalToolsParameter:
+    """Tests for the additional_tools parameter on ExecuteSeedHandler."""
+
+    def test_definition_includes_additional_tools_parameter(self) -> None:
+        """ExecuteSeedHandler exposes additional_tools in its tool schema."""
+        handler = ExecuteSeedHandler()
+        param_names = {p.name for p in handler.definition.parameters}
+        assert "additional_tools" in param_names
+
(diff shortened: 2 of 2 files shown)


jcfernandez-890825:
@Q00 Apologies for the late reply — addressed your feedback but forgot to follow up here!

**Done:** Rebased to a clean single commit on current `main` (`04892f7`, +195 lines across 2 files). The worktree imports, verification_artifacts, and task_workspace argument are all gone — this PR now contains only the `additional_tools` parameter + 4 tests.

That said, we've since noticed that #262 / #264 / #265 by @shaun0927 takes a more comprehensive approach (wiring `mcp_manager` directly through the handler for automatic tool discovery). We even tested the #265 branch locally and confirmed it works end-to-end with Chrome DevTools MCP in `ooo run` subagents — [test report on PR #264](https://github.com/Q00/ouroboros/pull/264#issuecomment-4162128189).

**Your call:** happy to keep this PR as a simpler complementary mechanism, or close it if #264 is the preferred path. Either way, our use case (#181) gets solved, which is what matters. Thanks for reviewing!

PR #303, opened 2026-04-02
feat(providers): add Kiro CLI adapters for LLM completion and agent execution

BangShinChul:
Implements `KiroCodeAdapter` and `KiroAgentAdapter` — new provider and orchestrator adapters that shell out to `kiro-cli chat --no-interactive` for LLM completions and `--trust-all-tools` for autonomous agent execution.

This lets Ouroboros leverage a local Kiro CLI session, following the same subprocess pattern as `ClaudeCodeAdapter` and `CodexCliLLMAdapter`.

### New adapters

| File | Description |
|------|-------------|
| `providers/kiro_adapter.py` | `KiroCodeAdapter` — single-response LLM completion via subprocess, with retry, timeout, model-name mapping |
| `orchestrator/kiro_adapter.py` | `KiroAgentAdapter` — autonomous agent execution with line-by-line streaming, retry logic, LLMAdapter bridge |

### Config & factory wiring

| File | Description |
|------|-------------|
| `config/models.py` | Add `"kiro"` to `runtime_backend` Literal, add `kiro_cli_path` field |
| `config/loader.py` | Add `get_runtime()`, `get_kiro_cli_path()` (`OUROBOROS_RUNTIME`, `OUROBOROS_KIRO_CLI_PATH` env vars) |
| `config/__init__.py` | Export new helpers |
| `providers/factory.py` | Register `"kiro"` / `"kiro_cli"` backend, wire into `create_llm_adapter()` |
| `orchestrator/factory.py` | Wire `KiroAgentAdapter` into `create_agent_runtime()` |
| `providers/__init__.py` | Export `KiroCodeAdapter` |

### Docs & tests

| File | Description |
|------|-------------|
| `README.md`, `README.ko.md` | Add Kiro CLI quick-start guide |
| `tests/unit/test_kiro_adapters.py` | 25 unit tests — command building, streaming, timeout, retry, env flags, prompt building, process cleanup |

## How to use

Set runtime in `.env`:

```bash
OUROBOROS_RUNTIME=kiro
```

Or via environment variable:
```bash
OUROBOROS_RUNTIME=kiro ooo interview "Build a task CLI"
```


Q00 (changes requested):
I see three blocking issues here. First, the new Kiro quick-start only tells users to set `OUROBOROS_RUNTIME=kiro`, but `src/ouroboros/cli/commands/init.py` now calls `create_llm_adapter()` without an explicit backend, and `create_llm_adapter()` still resolves from `llm.backend` rather than `OUROBOROS_RUNTIME`. That means interview/seed/QA flows can stay on the old LLM backend even though the docs say runtime selection is enough. 

Second, `src/ouroboros/mcp/tools/qa.py` now hardcodes `model="claude-sonnet-4-20250514"` and drops the backend-aware `get_qa_model(...)` path, so non-Claude backends ignore configured QA models and the Kiro path gets a Claude model id. 

Third, the `kiro` branch in `src/ouroboros/providers/factory.py` discards `cwd`, `timeout`, `max_retries`, and related adapter options, and `src/ouroboros/providers/kiro_adapter.py` always launches the subprocess with `cwd=os.getcwd()`. That makes Kiro behave differently from the existing adapters and breaks callers that rely on per-request working-directory control for brownfield interview / repo-aware flows.

Q00 (changes requested):
Excited to see Kiro CLI integration coming! The adapter structure and factory wiring look solid. However, comparing against the existing `ClaudeCodeAdapter` and `CodexCliRuntime` implementations, there are several safety patterns that need to be ported over before this is production-ready.

Also, please rebase onto `main` — there are recent changes that may affect the config/factory layers.

---

## Critical

### C1: No timeout on stdout reading loop (`orchestrator/kiro_adapter.py`)

```python
async for raw_line in proc.stdout:   # ← no idle timeout — blocks forever if kiro-cli hangs
    ...
await asyncio.wait_for(proc.wait(), timeout=_DEFAULT_TIMEOUT)  # ← never reached
```

`CodexCliRuntime` applies `first_chunk_timeout_seconds` (60s) + `chunk_timeout_seconds` (300s) on each `stream.read()` call. If `kiro-cli` hangs silently, this adapter will block indefinitely.

### C2: No `asyncio.CancelledError` handling — subprocess orphaned on task cancellation

When the orchestrator aborts an AC, `CancelledError` propagates through the async generator. Without explicit handling, the subprocess may be orphaned. `CodexCliRuntime` explicitly catches `CancelledError`, terminates the process, then re-raises.

### C3: No child environment sanitization

```python
env = {**os.environ, "OUROBOROS_SUBAGENT": "1"}  # inherits everything
```

`CodexCliLLMAdapter._build_child_env()` strips `CLAUDECODE`, `OUROBOROS_AGENT_RUNTIME`, `OUROBOROS_LLM_BACKEND` to prevent nested session detection and recursive MCP server startup. Without this, invoking Kiro from inside a Claude Code session will inherit `CLAUDECODE=1` and potentially conflict.

---

## High

### H1: No `_OUROBOROS_DEPTH` recursion guard

If Kiro calls back into Ouroboros (which calls Kiro again), there's no depth ceiling. Existing adapters enforce `_MAX_OUROBOROS_DEPTH = 5`.

### H2: No concurrent stderr drain in agent adapter

Stdout is read line-by-line while stderr is only read after `proc.wait()`. If `kiro-cli` fills the stderr pipe buffer (~64KB), the process deadlocks. `CodexCliRuntime` runs `asyncio.create_task(self._collect_stream_lines(process.stderr))` concurrently.

### H3: No `response_format` / JSON enforcement

When `config.response_format` requests `json_schema`, `ClaudeCodeAdapter` injects the schema as a prompt instruction and uses `extract_json_payload()` with retry. Without this, structured-output callers (seed architect, evaluator, QA judge) will receive raw prose and fail silently.

### H4: `allowed_tools` parameter ignored

Interview mode passes `allowed_tools=[]` to block all tool use. The Kiro adapter accepts this but never applies it (neither as a CLI flag nor as a prompt constraint). `CodexCliLLMAdapter` embeds tool restrictions directly in the prompt text.

---

## Medium

### M1: `"kiro"` missing from `_RUNTIME_HANDLE_BACKEND_ALIASES`

`orchestrator/adapter.py` has `_RUNTIME_HANDLE_BACKEND_ALIASES` for `RuntimeHandle` construction. Without a `"kiro"` entry, `RuntimeHandle(backend="kiro")` will raise `ValueError`.

### M2: Yielded `AgentMessage`s don't carry `resume_handle`

Both `ClaudeAgentAdapter` and `CodexCliRuntime` attach `resume_handle=current_handle` to every yielded message. The session manager inspects this for persistence. Without it, `ooo status` and the TUI dashboard won't show Kiro runtime state.

### M3: `"kiro_cli"` alias missing from runtime backend map

`adapter.py` adds `"kiro": "kiro"` but not `"kiro_cli": "kiro"`, inconsistent with the `_KIRO_BACKENDS = {"kiro", "kiro_cli"}` sets in `runtime_factory.py` and `providers/factory.py`.

---

## Low

### L1: `<system>` XML tags as CLI positional argument

The system prompt is wrapped in `<system>` tags and passed as a CLI arg. `CodexCliLLMAdapter` uses `## System Instructions` markdown headers instead. Worth verifying that `kiro-cli` actually interprets XML system tags.

### L2: `max_turns` not enforced

The parameter is accepted but not passed to the CLI. Interview's `max_turns=1` has no effect.

---

## Recommendation

The existing adapters (`claude_code_adapter.py`, `codex_cli_adapter.py`, `codex_cli_runtime.py`) have accumulated these safety patterns through real production incidents. I'd recommend using `CodexCliRuntime` as the primary reference since it's the closest architectural match (subprocess + streaming). The key patterns to port:

1. `_build_child_env()` with env stripping + depth guard
2. Concurrent stderr drain via `asyncio.create_task`
3. Per-read idle timeout in the stdout loop
4. `CancelledError` handling with process termination
5. JSON enforcement layer for structured output

Looking forward to seeing Kiro fully integrated! 🎉

shaun0927:
I like the direction here. A Kiro backend is consistent with where this project has been going: broader multi-runtime support, shared orchestration surfaces, and backend-specific adapters behind a common contract.

That said, I would not merge this PR in its current form.

The main issue is **runtime-contract parity**.

Right now the branch gets most of the way there on subprocess safety, config wiring, and LLM/runtime routing, but two gaps still matter at the maintainer bar for this repo:

1. **Dispatcher-backed runtime behavior is still not really preserved.**
   The factory now passes `skill_dispatcher`, but the adapter still does not use it during task execution. So selecting `OUROBOROS_RUNTIME=kiro` still drops part of the non-Claude runtime behavior instead of being a true drop-in backend.

2. **Targeted resume still degrades silently.**
   Accepting `resume_session_id` and then only passing a bare `--resume` is not the same thing as honoring the requested session. If Kiro cannot resume a specific session id, that needs to fail fast explicitly rather than silently resuming "something."

There is also a process concern:

3. **The branch is stale / conflicting with `main`.**
   Even if the remaining contract issues were fixed, I would want this rebased before merge.

So my recommendation would be:

- keep the multi-runtime direction,
- rebase onto current `main`,
- either implement full dispatcher + targeted-resume parity,
- **or** narrow the backend surface explicitly and reject unsupported behaviors instead of silently degrading them.

In other words: **directionally yes, current PR shape no**. The backend is promising, but this repo has been moving toward stricter contracts, not backend-specific silent exceptions.

PR #311, opened 2026-04-03
perf(#310): defer orphan scan and skip terminal sessions at MCP startup

### LINKED ISSUE #310: MCP server startup blocked by synchronous orphan session scan — causes connection timeout with large EventStore
## Summary

When the EventStore database grows large, the MCP server's synchronous `cancel_orphaned_sessions()` call during startup takes too long to complete, causing MCP clients (e.g., Claude Code) to time out before the JSON-RPC `initialize` handshake finishes.

## Reproduction

- **Environment:** ouroboros-ai 0.27.0, Claude Code CLI, Linux
- **DB:** ~/.ouroboros/ouroboros.db — ~1.4 GB, 165K events, 109 sessions
- **Symptom:** `claude mcp list` → `plugin:ouroboros:ouroboros — ✗ Failed to connect`
- **Root cause:** Server takes ~38s to produce first JSON-RPC response, exceeding the client's connection timeout.

### Measured startup times (time to first JSON-RPC response)

| DB size | Events | Sessions | Startup time |
|---------|--------|----------|-------------|
| 1.4 GB | 165,187 | 109 | **~38 s** |
| 547 MB | 57,184 | 22 | ~17 s |
| 121 MB | 12,129 | 5 | ~5 s |

## Bottleneck analysis

The hot path in `_run_mcp_server()` (`cli/commands/mcp.py`):

```python
await event_store.initialize()
repo = SessionRepository(event_store)
cancelled = await repo.cancel_orphaned_sessions()   # ← blocks serving
await server.serve()                                  # ← clients can connect only after this
```

`find_orphaned_sessions()` (`orchestrator/session.py:866`) performs:

1. `get_all_sessions()` — loads all `orchestrator.session.*` events
2. For **each** session → `replay("session", session_id)` — loads **every** event for that session into memory
3. Iterates all events to determine status via `_status_from_event()`

This is **O(S × E)** where S = number of sessions and E = average events per session. The vast majority of sessions are already terminal (COMPLETED/CANCELLED/FAILED), yet they are still fully replayed every time.

### 1. Defer orphan scan to after server starts serving (highest impact, simplest change)

Move `cancel_orphaned_sessions()` to a background task so the MCP handshake completes immediately:

```python
async def _deferred_cleanup(repo):
    await asyncio.sleep(2)
    await repo.cancel_orphaned_sessions()

asyncio.create_task(_deferred_cleanup(repo))
await server.serve()  # clients can connect immediately
```

### 2. SQL-level filtering to skip terminal sessions (avoids unnecessary replay)

Before calling `replay()` per session, exclude sessions that already have a terminal event with a single query:

```sql
SELECT DISTINCT aggregate_id FROM events
WHERE event_type = 'orchestrator.session.started'
  AND aggregate_id NOT IN (
    SELECT aggregate_id FROM events
    WHERE event_type IN (
      'orchestrator.session.completed',
      'orchestrator.session.failed',
      'orchestrator.session.cancelled'
    )
  )
```

Only truly RUNNING/PAUSED sessions (typically 0–1) would need full replay.

### 3. (Longer term) Snapshot table for session status

Maintain a lightweight read-model table updated on each status-changing event, eliminating the need for full event replay during orphan detection entirely.

## Current workaround

Manually trim old events and vacuum the database:

```python
import sqlite3
conn = sqlite3.connect(os.path.expanduser("~/.ouroboros/ouroboros.db"))
conn.execute("DELETE FROM events WHERE timestamp < '2025-01-01'")
conn.commit()
conn.execute("VACUUM")
conn.close()
```

isac322:
## Summary
- Defer `cancel_orphaned_sessions()` to an `asyncio.create_task()` background task so the MCP JSON-RPC handshake completes immediately
- Add `EventStore.get_non_terminal_session_ids()` — a single SQL subquery that skips sessions with terminal events, avoiding expensive full-replay for completed/failed/cancelled sessions

## Measured improvement (165K events, 109 sessions, 1.4 GB DB)
| | Before | After |
|---|---|---|
| Time to first JSON-RPC response | **~38 s** | **< 1 s** |
| Orphan scan | Blocks `server.serve()` | Runs in background |

## Changes
- `cli/commands/mcp.py` — move orphan cleanup after `server.serve()` via background task
- `orchestrator/session.py` — use `get_non_terminal_session_ids()` instead of `get_all_sessions()` + per-session replay
- `persistence/event_store.py` — new `get_non_terminal_session_ids()` method with index-only subquery

Closes #310

### DIFF
--- a/src/ouroboros/cli/commands/mcp.py
+++ b/src/ouroboros/cli/commands/mcp.py
@@ -138,22 +138,11 @@ async def _run_mcp_server(
     else:
         event_store = EventStore()
 
-    # Auto-cancel orphaned sessions on startup.
-    # Sessions left in RUNNING/PAUSED state for >1 hour are considered orphaned
-    # (e.g., from a previous crash). Cancel them before accepting new requests.
-    # NOTE: find_orphaned_sessions now checks for active runtime processes first,
-    # so sessions with live claude/codex agents won't be cancelled even if stale.
-    try:
-        await event_store.initialize()
-        repo = SessionRepository(event_store)
-        cancelled = await repo.cancel_orphaned_sessions()
-        if cancelled:
-            _stderr_console.print(
-                f"[yellow]Auto-cancelled {len(cancelled)} orphaned session(s)[/yellow]"
-            )
-    except Exception as e:
-        # Auto-cleanup is best-effort — don't prevent server from starting
-        _stderr_console.print(f"[yellow]Warning: auto-cleanup failed: {e}[/yellow]")
+    # Initialize EventStore eagerly — required by tool handlers.
+    # Orphan cleanup is deferred to a background task (see below) so that the
+    # MCP JSON-RPC handshake is not blocked by an expensive full-replay scan
+    # of the event history.  See https://github.com/Q00/ouroboros/issues/310
+    await event_store.initialize()
 
     # Auto-discover and connect MCP bridge for server-to-server communication
     from ouroboros.mcp.bridge import create_bridge_from_env
@@ -218,8 +207,23 @@ async def _run_mcp_server(
 
     _write_pid_file()
 
+    # Schedule orphan cleanup as a background task so it does not block
+    # the MCP handshake.  The scan is best-effort — failures are logged
+    # but never prevent the server from operating.
+    async def _deferred_orphan_cleanup() -> None:
+        try:
+            repo = SessionRepository(event_store)
+            cancelled = await repo.cancel_orphaned_sessions()
+            if cancelled:
+                _stderr_console.print(
+                    f"[yellow]Auto-cancelled {len(cancelled)} orphaned session(s)[/yellow]"
+                )
+        except Exception as e:
+            _stderr_console.print(f"[yellow]Warning: auto-cleanup failed: {e}[/yellow]")
+
     # Start serving
     try:
+        asyncio.create_task(_deferred_orphan_cleanup())
         await server.serve(transport=transport, host=host, port=port)
     finally:
         _cleanup_pid_file()
--- a/src/ouroboros/orchestrator/session.py
+++ b/src/ouroboros/orchestrator/session.py
@@ -892,12 +892,19 @@ class SessionRepository:
         orphaned: list[SessionTracker] = []
 
         try:
-            # Get all session start events to enumerate sessions
-            start_events = await self._event_store.get_all_sessions()
+            # Pre-filter at the SQL level: only replay sessions that do NOT
+            # have a terminal event (completed/failed/cancelled).  This avoids
+            # the expensive O(S×E) full-replay for the common case where most
+            # sessions are already finished.
+            # See https://github.com/Q00/ouroboros/issues/310
+            candidate_ids = await self._event_store.get_non_terminal_session_ids()
 
-            for start_event in start_events:
-                session_id = start_event.aggregate_id
+            log.info(
+                "orchestrator.orphan_detection.candidates",
+                candidate_count=len(candidate_ids),
+            )
 
+            for session_id in candidate_ids:
                 # Replay all events for this session
                 try:
                     events = await self._event_store.replay("session", session_id)
@@ -927,9 +934,9 @@ class SessionRepository:
                 last_activity = last_event.timestamp
                 if last_activity is None:
                     # If no timestamp, use start_time from event data as fallback
-                    start_time_str = start_event.data.get("start_time")
-                    if start_time_str:
-                        last_activity = datetime.fromisoformat(start_time_str)
+                    start_ts = events[0].data.get("start_time") if events else None
+                    if start_ts:
+                        last_activity = datetime.fromisoformat(start_ts)
                     else:
                         continue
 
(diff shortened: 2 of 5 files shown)


shaun0927:
The performance goal is absolutely the right one. Improving MCP startup latency on large event stores fits the current direction of the repo very well, and the SQL-side prefilter is especially promising because it attacks the real replay-cost problem instead of just masking it.

That said, I would not merge the current shape of this PR.

The reason is that the optimization now changes a correctness contract, not just a performance characteristic:

1. **Moving orphan cleanup into a fire-and-forget background task introduces a real startup race.**
   The server can now begin serving requests before cleanup finishes, which means stale-session cancellation can race with legitimate immediate traffic. That is a materially different behavior from the previous "cleanup before serving" contract.

2. **The background task lifecycle is not managed.**
   The task handle is discarded, so there is no coordinated shutdown / cancellation / await path. In this codebase, once a task shares mutable persistence state, I want its lifetime to be explicit.

3. **The updated tests do not fully prove the new control flow.**
   Yielding once with `await asyncio.sleep(0)` is not a strong synchronization point for background work, and several of the session tests look weaker against the new SQL-prefilter path than they were against the old replay-based path.

So my maintainer take is:

- **keep** the SQL prefilter direction,
- but do **not** land the uncoordinated deferred-cleanup behavior as-is.

I would be much more comfortable with one of these shapes:

- a coordinated startup task whose lifecycle is tracked explicitly and shut down cleanly,
- a narrower pre-serve cleanup that only touches the prefitered candidate set,
- or a more explicit "cleanup scheduled / cleanup complete" contract if serving-before-cleanup is truly intended.

In other words: **good performance direction, but the current implementation trades too much correctness clarity for latency.**

shaun0927:
@isac322 @Q00 — heads up: while this PR was open, **#326** landed on `main` with a different fix for the same hot path. The new approach:

- `EventStore.get_session_activity_snapshots` (`src/ouroboros/persistence/event_store.py:519+` on `main`) pre-aggregates per-session start / last-activity / latest-status in a **single SQL query using window functions** (`ROW_NUMBER() OVER PARTITION BY aggregate_id`) — no per-session replay.
- `SessionRepository.find_orphaned_sessions` (`src/ouroboros/orchestrator/session.py:945-959` on `main`) now detects the snapshot method and **bypasses the old replay loop entirely** when it's available.
- Released as **`v0.30.0`** (`2243836`).

That subsumes both halves of this PR:
- The SQL prefilter (`get_non_terminal_session_ids`) is strictly weaker than the snapshot query (still needs per-candidate replay; snapshot does it in one pass).
- The deferred-cleanup change is no longer needed at all — with the snapshot path, `cancel_orphaned_sessions()` runs in milliseconds, so the synchronous pre-serve contract holds with no race and no untracked task lifecycle.

Proposed close path, unless either of you objects:

1. **@isac322** — if you can spare a moment, would you reproduce against your 1.4 GB / 165K-event DB on `v0.30.0` and report the new time-to-first-JSON-RPC? If it's back under your MCP client timeout, I'll close this PR with credit and close #310 too. If you'd rather not, I'll run it on a synthetic dataset and report back.
2. **@Q00** — final-call review only on the close path; no code change needed.

Thanks again @isac322 for surfacing the same hot path and the well-measured root-cause analysis — sorry for the missed signal during your review cycle (both #326 and your PR were in flight simultaneously).

shaun0927:
Closing as **superseded by #326 (released in v0.30.0)**.

I went through main carefully to confirm before closing — both halves of this PR are present in stronger form, and the Medium-severity lifecycle concern from the bot's second review is also fixed there:

| Concern in #311 | Status on main (v0.30.0) |
|---|---|
| Per-session replay during orphan scan | Replaced by `EventStore.get_session_activity_snapshots()` (`src/ouroboros/persistence/event_store.py:445`) — single SQL pass using `ROW_NUMBER() OVER PARTITION BY aggregate_id` for start / last-activity / latest-status. Strictly stronger than `get_non_terminal_session_ids` (no per-candidate replay). |
| Deferred cleanup blocking handshake | `cleanup_task = asyncio.create_task(_run_startup_cleanup, name=...)` (`src/ouroboros/cli/commands/mcp.py:228`) — same pattern. |
| **Untracked task lifecycle (bot review #2, Medium)** | Resolved: `cleanup_task.cancel(); await cleanup_task` on shutdown (`src/ouroboros/cli/commands/mcp.py:300-303`). |
| `find_orphaned_sessions` integration | `src/ouroboros/orchestrator/session.py:959` detects the snapshot method and bypasses the legacy replay loop entirely (with replay as fallback for older stores). |

Net effect on the #310 hot path is the same direction this PR was pushing, with the lifecycle gap closed.

@isac322 — thank you for the well-measured root-cause work; the 38s → <1s breakdown in your reproduction was the critical signal that drove #326 in the same window. Apologies for the missed cross-link during your review cycle. If you do get a chance to re-measure on your 1.4 GB / 165K-event DB against `v0.30.0`, please drop the number on #310 — I'll close that issue once we have your confirmation (or after a reasonable wait if you're busy).

I'll also file a separate follow-up for the \`orchestrator.session.started\`-anchored candidate set (Low #5 from the bot review), since that sidecase still applies to the snapshot query on main.

PR #337, opened 2026-04-07
fix: enforce security checks on FastMCP transport path

### LINKED ISSUE #288: security: FastMCP tool wrapper bypasses SecurityLayer checks
## Problem

The FastMCP tool dispatch wrapper at `adapter.py:466-494` calls `handler.handle()` directly, bypassing the `SecurityLayer.check_request()` pipeline. This means authentication, rate limiting, authorization, and input validation are **not applied** to tool calls received via the FastMCP stdio/SSE transport — the primary production transport.

The `SecurityLayer` is only invoked through the `MCPServerAdapter.call_tool()` method (`adapter.py:296-352`), which is the programmatic API path, not the FastMCP-served path.

## Evidence

1. **FastMCP wrapper** (`adapter.py:466-494`):
   ```python
   async def tool_wrapper(**kwargs: Any) -> str:
       # ... kwargs unwrapping ...
       result = await handler.handle(kwargs)  # direct call, no security check
       return result.text_content
   ```

2. **Programmatic path with security** (`adapter.py:296-352`):
   ```python
   async def call_tool(self, tool_name: str, arguments: dict) -> MCPToolResult:
       if self._security:
           await self._security.check_request(request)  # security applied here
       # ... then handler.handle()
   ```

3. **SecurityLayer pipeline** (`security.py:576-623`):
   - Authentication (`security.py:593`) — defaults to `AuthMethod.NONE`
   - Rate limiting (`security.py:600`) — disabled by default
   - Authorization (`security.py:614`)
   - Input validation (`security.py:619`) — scans string values for dangerous patterns (path traversal, command injection)

## Current Impact

**Low in current configuration**: Auth is `NONE`, rate limiting is disabled. The input validation scanner (`security.py:474-525`) that checks for path traversal and injection patterns is the most relevant check being bypassed.

**High if security is enabled for production**: Any future deployment that enables auth or rate limiting via `SecurityConfig` would have a false sense of security — the checks only apply to the programmatic API, not the actual MCP transport.

## Trade-offs

Adding security checks to the FastMCP path adds latency to every tool call:
- Auth check: negligible (in-memory comparison)
- Rate limiting: negligible (token bucket check)
- Input validation: ~1-5ms (regex scanning of all string values in kwargs)

The input validation scan is the most expensive but still sub-millisecond for typical payloads.

## Questions for Maintainers

1. Is the security bypass intentional for the stdio transport (trusted local process), with security only intended for SSE/remote transports?
2. If security should apply to all transports, should it be added to `tool_wrapper` or should `tool_wrapper` route through `call_tool()`?
3. Is the input validation scanner (`security.py:474-525`) considered necessary for stdio mode where the caller is the local Claude Code process?

Sathvik-1007:
Fixes #288

## Problem

The FastMCP tool wrapper called handler.handle() directly, bypassing the SecurityLayer.check_request() pipeline. This meant authentication, rate limiting, authorization, and input validation were never applied to tool calls from the stdio or SSE transports.

The SecurityLayer was only enforced on the call_tool() method, which is the programmatic API path used for testing and internal calls, not the actual MCP transport.

## Impact

Current impact is low because auth defaults to NONE and rate limiting is disabled. However, the input validation scanner that checks for path traversal patterns, dangerous code patterns, and shell metacharacters was being bypassed on all production transport paths.

If auth or rate limiting were enabled in future deployments, the security would be completely ineffective since the checks only applied to the test API.

## Solution

Changed the FastMCP tool_wrapper to route through call_tool() instead of calling handler.handle() directly. This adds approximately 1-5ms latency per tool call due to input validation regex scanning, but ensures all security checks are consistently applied regardless of transport.

## Testing

Added test_fastmcp_path_enforces_security to verify that path traversal input is rejected when passed through the FastMCP wrapper. All existing unit and integration tests pass.

## Changes

- src/ouroboros/mcp/server/adapter.py: tool_wrapper now calls self.call_tool()
- tests/unit/mcp/server/test_adapter.py: added security enforcement test

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -583,9 +583,10 @@ class MCPServerAdapter:
                         normalized_kwargs.setdefault(key, value)
 
                     # Route through call_tool() to enforce security checks.
-                    # Note: FastMCP does not provide credentials, so auth will use
-                    # default AuthMethod.NONE behavior and rate limiting will not
-                    # apply (requires client_id). Input validation is enforced.
+                    # FastMCP does not provide credentials, so:
+                    # - Input validation is enforced
+                    # - Auth/authorization will reject if any auth method configured
+                    # - Rate limiting cannot apply (requires client_id)
                     result = await self.call_tool(h.definition.name, normalized_kwargs)
                     if result.is_ok:
                         # Convert MCPToolResult to FastMCP format
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -23,7 +23,7 @@ from ouroboros.mcp.errors import (
     MCPToolError,
 )
 from ouroboros.mcp.server.protocol import PromptHandler, ResourceHandler, ToolHandler
-from ouroboros.mcp.server.security import AuthConfig, RateLimitConfig, SecurityLayer
+from ouroboros.mcp.server.security import AuthConfig, AuthMethod, RateLimitConfig, SecurityLayer
 from ouroboros.mcp.types import (
     MCPCapabilities,
     MCPPromptDefinition,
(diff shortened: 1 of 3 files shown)


Sathvik-1007:
@Q00 — bot approved PR #337 (commit cd2ef9c). No blocking findings.

**What fix do:**
- FastMCP `tool_wrapper` now routes through `call_tool()` — no more security bypass on stdio transport
- Startup guards reject incompatible auth/rate-limit configs early (clear error > silent failure)
- `AuthMethod.NONE` always returns `authenticated=True` regardless of `required` flag (was backwards — `required=True` accidentally set `authenticated=False`)

**Bot note:** fix couples transport guard + auth semantic change. Both intentional — FastMCP can't supply credentials, so NONE is the only valid auth, and NONE must always authenticate.

4466 tests pass. Ready for review.

Q00 (approved):
LGTM — security path unified, startup guards solid, auth semantics fixed. Ship it.

PR #347, opened 2026-04-09
perf(#346): non-blocking event emissions on interview hot path

### LINKED ISSUE #346: perf(interview): non-blocking event emissions on interview hot path
## Problem

The `InterviewHandler` emits diagnostic events (interview_started, response_recorded, interview_completed, etc.) via `_emit_event()`, which performs a synchronous `await` on every SQLite INSERT. These events are emitted at 6 points on the interview hot path.

Each emission involves:
1. `_ensure_initialized()` — lazy SQLite engine + WAL mode setup
2. `_event_store.append(event)` — SQLite INSERT in a transaction

The existing code already wraps emissions in `try/except` and swallows errors (line 492), confirming events are **non-critical diagnostics** that should not block the interview flow. Yet the `await` forces each emission to complete before proceeding.

## Proposed Solution

Add a `_emit_event_bg()` method that wraps `_emit_event()` in `asyncio.create_task()` for fire-and-forget execution. Task references are stored in a `_bg_tasks: set[asyncio.Task]` with `done_callback` cleanup to prevent garbage collection — following the [Python docs recommendation](https://docs.python.org/3/library/asyncio-task.html#creating-tasks).

```python
def _emit_event_bg(self, event: Any) -> None:
    task = asyncio.create_task(self._emit_event(event))
    self._bg_tasks.add(task)
    task.add_done_callback(self._bg_tasks.discard)
```

All 6 call sites are updated from `await self._emit_event(...)` to `self._emit_event_bg(...)`.

## Expected Impact

- **-2~10ms per round** (1-2 SQLite writes removed from critical path)
- Eliminates a serial blocking point before the main LLM calls
- No data loss risk — events still execute, just asynchronously

## Files Changed

- `src/ouroboros/mcp/tools/authoring_handlers.py`

## Test Plan

- All 31 interview handler tests pass
- Event store assertions updated to account for async emission timing

shaun0927:
## Summary

- Replace synchronous `await self._emit_event()` with fire-and-forget `self._emit_event_bg()` at all 6 call sites in `InterviewHandler`
- Store task references in `_bg_tasks: set[asyncio.Task]` with `done_callback` cleanup to prevent GC ([Python docs recommendation](https://docs.python.org/3/library/asyncio-task.html#creating-tasks))
- Events are diagnostic-only and already wrapped in `try/except` error swallowing — they should not block the interview hot path

## Motivation

Each `_emit_event()` call performs a synchronous `await` on a SQLite INSERT, adding 2-10ms of serial latency per round. Since events are non-critical diagnostics (the existing error handler silently swallows failures), blocking the interview flow on event persistence is unnecessary overhead.

Closes #346

## Test plan

- [x] All 37 existing interview handler tests pass
- [x] Full unit test suite: 3906 passed, 9 skipped
- [ ] Manual verification: interview round-trip time measured before/after

### DIFF
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -541,8 +541,22 @@ class InterviewHandler:
             await self._event_store.initialize()
             self._initialized = True
 
+    async def _drain_bg_tasks(self, timeout: float = 5.0) -> None:
+        """Await all pending background event tasks before shutdown."""
+        if not self._bg_tasks:
+            return
+        tasks = list(self._bg_tasks)
+        done, pending = await asyncio.wait(tasks, timeout=timeout)
+        for t in pending:
+            t.cancel()
+        # Await cancelled tasks so CancelledError propagates cleanly
+        if pending:
+            await asyncio.gather(*pending, return_exceptions=True)
+        self._bg_tasks.clear()
+
     async def close(self) -> None:
-        """Close the event store if this handler owns it."""
+        """Drain pending event tasks, then close the event store if owned."""
+        await self._drain_bg_tasks()
         if self._owns_event_store:
             await self._event_store.close()
             self._initialized = False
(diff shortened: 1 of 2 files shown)


Q00 (changes requested):
Requesting changes because this turns event persistence into fire-and-forget without a drain point. The handler still closes its owned event store in finally, so pending background append tasks can race with shutdown and silently lose events. Please add explicit task draining or cancellation semantics before close().

Q00 (approved):
LGTM

PR #349, opened 2026-04-09
perf(#345): optimize ambiguity scorer token usage and enforce format error limit

### LINKED ISSUE #345: perf(interview): optimize ambiguity scorer token usage and enforce format error limit
## Problem

The ambiguity scorer has two inefficiencies:

### 1. Oversized initial token allocation

`initial_max_tokens` defaults to 2048, but typical scoring responses are 200-350 tokens (structured JSON with 3-4 dimension scores + justifications). The LLM reserves capacity for 2048 tokens on every call, even though it almost never uses more than 512.

### 2. Unused format error retry limit

The `AmbiguityScorer` defines `max_format_error_retries = 5` (line 153) but **never checks it** in the retry loop. When the LLM produces malformed JSON (not truncated, just wrong format), the scorer retries with the same token budget until the global `max_retries` limit is hit. This wastes 2-5 seconds per futile retry.

Unlike truncation errors (which benefit from doubling `max_tokens`), format errors indicate a prompt problem — retrying with identical parameters rarely helps.

### Token optimization
- Reduce `initial_max_tokens` from 2048 to 512
- The existing adaptive retry mechanism already handles truncation: if `finish_reason == "length"`, tokens double automatically (512 → 1024 → 2048)
- Net effect: most calls succeed on the first attempt with lower token allocation

### Format error early exit
- Add a `format_error_count` counter in the retry loop
- Increment only on non-truncation parse failures (`finish_reason != "length"`)
- Break when `format_error_count >= max_format_error_retries`
- This activates the previously-defined but unused `max_format_error_retries` field

## Expected Impact

- **-0~0.5s per scoring call** from reduced initial token allocation
- **Worst-case prevention**: eliminates 6-15 seconds of wasted retries on persistent format errors
- No scoring quality impact — successful responses are identical

## Files Changed

- `src/ouroboros/bigbang/ambiguity.py` — token default + format error counter
- `src/ouroboros/bigbang/interview.py` — matching token default

## Test Plan

- Existing ambiguity scorer tests (60 tests) pass
- Format error early exit tested via mock adapter returning malformed JSON

shaun0927:
## Summary

- Reduce `initial_max_tokens` from 2048 to 512 in `AmbiguityScorer` and `InterviewEngine`
- Enforce the existing but unused `max_format_error_retries` field with a dedicated counter in the retry loop

### Token over-allocation
Typical scoring responses are 200-350 tokens (structured JSON with 3 dimension scores + justifications), yet the scorer reserves 2048 tokens on every call. Reducing to 512 matches actual usage. The adaptive retry mechanism already handles truncation: if `finish_reason == "length"`, tokens double automatically (512 → 1024 → 2048).

### Unused format error limit
`AmbiguityScorer` defines `max_format_error_retries = 5` but never checks it in the retry loop. When the LLM produces malformed JSON (not truncated — wrong format), the scorer retries with identical parameters until the global `max_retries` is exhausted. Unlike truncation errors (which benefit from doubling tokens), format errors indicate a prompt problem that retrying cannot fix. This wastes 2-5 seconds per futile retry.

The fix adds a `format_error_count` counter that increments only on non-truncation parse failures and breaks when `max_format_error_retries` is reached.

Closes #345

## Test plan

- [x] All 60 ambiguity scorer tests pass
- [x] Default value assertion updated in `test_scorer_default_values`
- [ ] Manual verification: scoring latency measured before/after

### DIFF
--- a/src/ouroboros/bigbang/ambiguity.py
+++ b/src/ouroboros/bigbang/ambiguity.py
@@ -196,7 +196,7 @@ class AmbiguityScorer:
     llm_adapter: LLMAdapter
     model: str = field(default_factory=get_clarification_model)
     temperature: float = SCORING_TEMPERATURE
-    initial_max_tokens: int = 2048
+    initial_max_tokens: int = 512
     max_retries: int | None = 10  # Default to 10 retries (None = unlimited)
     max_format_error_retries: int = 5  # Stop after N format errors (non-truncation)
 
@@ -259,6 +259,7 @@ class AmbiguityScorer:
         last_error: Exception | ProviderError | None = None
         last_response: str = ""
         attempt = 0
+        format_error_count = 0
 
         while True:
             # Check retry limit if set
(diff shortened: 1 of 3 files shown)


Q00 (approved):
LGTM. The lower default token budget and the explicit format-error retry boundary both look sound, and the adaptive retry path still preserves a safe fallback when 512 tokens are not enough.

PR #350, opened 2026-04-09
perf(#343): parallelize ambiguity scoring and question generation

### LINKED ISSUE #343: perf(interview): parallelize ambiguity scoring and question generation
## Problem

During each interview round, the MCP handler calls two independent LLM operations **sequentially**:

1. `_score_interview_state()` — evaluates ambiguity across 3 dimensions (~3-8s)
2. `engine.ask_next_question()` — generates the next Socratic question (~3-8s)

These operations share the same input state but produce independent outputs. Running them serially means the user waits **6-16 seconds per round**, which degrades the interactive interview experience significantly.

## Proposed Solution

Use `asyncio.gather()` to run scoring and question generation concurrently. Since both operations:

- Read from the same `InterviewState` snapshot
- Write to independent outputs (score → `state.ambiguity_score`, question → return value)
- Are pure LLM calls with no shared mutable state during execution

They can safely execute in parallel. The expected latency becomes `max(scoring, question_gen)` instead of `scoring + question_gen`.

## Expected Impact

- **~50% latency reduction per round** (6-16s → 3-8s)
- No change to interview quality — identical LLM calls, just concurrent
- When scoring determines the interview is complete (`is_ready_for_seed`), the parallel question result is discarded (acceptable tradeoff — saves time on all non-final rounds)

## Files Changed

- `src/ouroboros/mcp/tools/authoring_handlers.py` — `asyncio.gather()` for parallel execution with `return_exceptions=True`

## Test Plan

- Unit tests verify parallel execution path and fallback on scoring failure
- Benchmark test confirms ≥2.0x speedup

shaun0927:
## Summary

- Use `asyncio.gather()` to run ambiguity scoring and question generation concurrently during interview rounds where completion is possible
- Add `return_exceptions=True` with explicit error unwrapping for safe parallel execution
- Fall back to sequential question generation if the parallel call fails

## Motivation

During each interview round (when `answered_rounds >= MIN_ROUNDS_BEFORE_EARLY_EXIT`), two independent LLM operations run **sequentially**:

1. `_score_interview_state()` — evaluates ambiguity across 3 dimensions (~3-8s)
2. `engine.ask_next_question()` — generates the next Socratic question (~3-8s)

These share the same input state but produce independent outputs. Running them serially means the user waits **6-16 seconds per round**. With `asyncio.gather()`, latency becomes `max(scoring, question_gen)` — approximately **50% reduction**.

When scoring determines the interview is complete (`qualifies_for_seed_completion` + streak requirement met), the parallel question result is discarded. This is an acceptable tradeoff: it wastes one LLM call on the final round but saves 3-8 seconds on every non-final round.

Closes #343

## Test plan

- [x] 5 new parallel scoring tests (concurrent execution, early exit, scoring failure fallback)
- [x] Updated existing auto-complete test assertion (`assert_not_called` → `assert_called_once`)
- [x] All 42 interview handler tests pass
- [ ] Benchmark test confirms ≥2.0x speedup

### DIFF
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -5,7 +5,6 @@ Contains handlers for interview and seed generation tools:
 - InterviewHandler: Manages interactive requirement-clarification interviews.
 """
 
-import asyncio
 from dataclasses import dataclass, field
 import os
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


Q00 (changes requested):
Requesting changes because scoring and question generation are not actually independent in the current design. _score_interview_state mutates state.ambiguity_score and completion_candidate_streak, and ask_next_question uses that snapshot to shape the next prompt. Running them in parallel means question generation can use stale routing state on the rounds where closure-mode behavior matters most. This needs an immutable scored snapshot or equivalent restructuring first.

src/ouroboros/mcp/tools/authoring_handlers.py
```diff
@@ -1019,35 +1018,14 @@ async def handle(
...
-                            _par_question = None
-
+                        # Scoring must complete before question generation:
+                        # _score_interview_state mutates state.ambiguity_score,
+                        # completion_candidate_streak, and ambiguity_breakdown.
+                        # ask_next_question reads those fields to build the
+                        # system prompt (closure mode, seed-ready, streak).
+                        # Running them in parallel would give the question
+                        # generator stale routing context.
+                        live_score = await self._score_interview_state(llm_adapter, state)
```
Q00, line 1028: LGTM

Q00 (approved):

PR #352, opened 2026-04-09
fix(#351): filter non-builtin tools from inherited delegation context

### LINKED ISSUE #351: fix(orchestrator): phantom tool entries from inherited bridge tools in delegated sessions
## Problem

When a parent session delegates execution to a child via `execute_seed`, the parent's full `effective_tools` list — including bridge/MCP tools — is injected into the child's arguments through the `_ooo_parent_effective_tools` hidden arg. The child then appends **all** inherited tool names to its `base_tools` list before any bridge discovery occurs.

For bridge tools (e.g., `mcp__chrome-devtools__click`), this creates **phantom catalog entries**: the tool name is normalized into an `MCPToolDefinition` with no backing handler or server connection, because `normalize_runtime_tool_definition()` accepts any string and always produces a definition.

### Data flow

```
Parent:  effective_tools = ["Read", "Edit", "Bash", "mcp__chrome__click"]
                                                     ^^^^^^^^^^^^^^^^
Hook:    _ooo_parent_effective_tools = [same list]    injected as hidden arg

Child:   _extract_inherited_effective_tools() → ["Read", "Edit", "Bash", "mcp__chrome__click"]
         ↓
Runner:  _resolve_tools() line 993-996
         base_tools.append("mcp__chrome__click")      ← added as builtin
         ↓
         assemble_session_tool_catalog(base_tools)
         ↓
         normalize_runtime_tool_definition("mcp__chrome__click")
         → MCPToolDefinition(name="mcp__chrome__click", description="mcp__chrome__click", parameters=())
           ^^ phantom entry: name exists, no handler, no server
         ↓
MCPToolProvider.get_tools()                            ← bridge tools discovered HERE
         → real "mcp__chrome__click" with handler      (if mcp_manager available)
```

The phantom entry from step 3 and the real entry from step 4 collide. If `mcp_manager` is not available in the child, only the phantom remains.

### Impact

1. **LLM confusion**: Claude sees `mcp__chrome__click` in `allowed_tools`, attempts to call it, but no handler exists → tool execution error → retries → wastes 5-10 LLM calls ($0.05-1.00) per phantom tool
2. **Silent failure**: No warning that a phantom entry was created — errors surface deep in execution as generic "tool execution failed"
3. **Prefix mismatch**: Parent `tool_prefix="mcp_"` vs child `tool_prefix=""` causes the same tool to appear under two names in the catalog

### Root cause

Bridge tool names are passed through the same `list[str]` channel as builtin tools, but they require fundamentally different resolution:
- **Builtins** (Read, Edit, Bash): Always available, resolved by name against `_RUNTIME_TOOL_DESCRIPTIONS`
- **Bridge tools**: Only available when a live MCP server connection exists, resolved by `MCPToolProvider.get_tools()` via `mcp_manager`

The child already receives `mcp_manager` through `BridgeAwareMixin` → `ExecuteSeedHandler` → `OrchestratorRunner`. Bridge tools will be discovered naturally by `MCPToolProvider.get_tools()` when `mcp_manager` is set. Passing their names via `inherited_tools` is redundant when the bridge is available, and harmful when it isn't.

## Proposed fix

In `_resolve_tools()`, filter inherited tools to only include known runtime builtins. Non-builtin tools (bridge/MCP) are logged and deferred to the `MCPToolProvider` discovery step that already follows.

```python
known_builtins = {d.name for d in enumerate_runtime_builtin_tool_definitions()}
for tool_name in self._inherited_tools:
    if tool_name in known_builtins and tool_name not in base_tools:
        base_tools.append(tool_name)
    elif tool_name not in known_builtins:
        log.info("orchestrator.runner.inherited_tool_deferred_to_bridge", tool=tool_name)
```

This ensures:
- Builtin tools (WebFetch, WebSearch, etc.) are still inherited correctly
- Bridge tools are only present in the catalog when backed by a live connection
- The structural guarantee: **tool name and tool handler always travel together**

shaun0927:
## Summary

- Filter inherited tools in `OrchestratorRunner._resolve_tools()` to only include known runtime builtins
- Non-builtin tools (bridge/MCP) are logged and deferred to `MCPToolProvider` discovery
- Update two existing tests to verify the new behavior

Closes #351

## Problem

When a parent session delegates to a child via `execute_seed`, the parent's full `effective_tools` — including bridge/MCP tools like `mcp__chrome-devtools__click` — is passed as `inherited_tools`. The child appends **all** these names to `base_tools` before bridge discovery occurs, creating phantom catalog entries: `MCPToolDefinition` objects with a name but no backing handler or server connection.

`normalize_runtime_tool_definition()` accepts any string and always produces a definition, so phantom entries are indistinguishable from real ones until the LLM tries to call them and gets a runtime error.

### Impact

- LLM sees phantom tools in `allowed_tools`, attempts calls, gets repeated failures (5-10 wasted LLM calls per phantom tool)
- Errors surface as generic "tool execution failed" with no indication the tool was a phantom
- Prefix mismatches (`mcp_chrome_navigate` vs `chrome_navigate`) cause the same tool to appear twice

## Root cause

Bridge tools were passed through the same `list[str]` channel as builtins, but they require fundamentally different resolution:

| Tool type | Resolution | Always available? |
|-----------|-----------|-------------------|
| **Builtin** (Read, Edit, Bash) | Name lookup in `_RUNTIME_TOOL_DESCRIPTIONS` | Yes |
| **Bridge/MCP** (chrome_navigate) | `MCPToolProvider.get_tools()` via live `mcp_manager` | Only with active connection |

The child already receives `mcp_manager` through `BridgeAwareMixin` → `ExecuteSeedHandler` → `OrchestratorRunner`. Bridge tools are discovered by `MCPToolProvider.get_tools()` (runner.py line 1010) which returns real definitions tied to live connections. Passing their names via `inherited_tools` is **redundant** when the bridge is available, and **harmful** when it isn't.

## Fix

In `_resolve_tools()`, check each inherited tool against `enumerate_runtime_builtin_tool_definitions()`. Only known builtins (Read, Write, Edit, Bash, Glob, Grep, WebFetch, WebSearch, NotebookEdit) are added to `base_tools`. Non-builtins are logged with `inherited_tool_deferred_to_bridge` and left for `MCPToolProvider` to discover.

```python
known_builtins = {d.name for d in enumerate_runtime_builtin_tool_definitions()}
for tool_name in self._inherited_tools:
    if tool_name in known_builtins and tool_name not in base_tools:
        base_tools.append(tool_name)
    elif tool_name not in known_builtins:
        log.info("orchestrator.runner.inherited_tool_deferred_to_bridge",
                 tool=tool_name, has_mcp_manager=self._mcp_manager is not None)
```

### Structural guarantee

After this change, a tool name can only enter the session catalog through two paths:
1. **Builtin resolution**: name is in `_RUNTIME_TOOL_DESCRIPTIONS` → always has a handler
2. **MCP discovery**: `MCPToolProvider.get_tools()` returns definitions from live server connections → always has a handler

There is no path where a name enters without a backing handler. Phantom entries are structurally impossible.

## Test plan

- [x] Updated `test_execute_seed_uses_inherited_runtime_handle`: verifies builtin `WebFetch` is inherited while bridge `mcp__chrome-devtools__click` is excluded
- [x] Updated `test_get_merged_tools_includes_inherited_tools`: same verification at the `_get_merged_tools()` level
- [x] All 662 orchestrator tests pass
- [x] Full unit suite: 4320 passed, 2 skipped
- [x] Ruff lint clean
- [x] MyPy clean

### DIFF
--- a/src/ouroboros/orchestrator/adapter.py
+++ b/src/ouroboros/orchestrator/adapter.py
@@ -305,15 +305,22 @@ def _runtime_handle_lifecycle_state(
 
 def runtime_handle_tool_catalog(
     runtime_handle: RuntimeHandle | None,
-) -> list[dict[str, Any]] | None:
-    """Return a copy of the serialized startup tool catalog when present."""
+) -> dict[str, Any] | list[dict[str, Any]] | None:
+    """Return a copy of the serialized startup tool catalog when present.
+
+    Accepts both the legacy ``list`` format and the ``dict`` format produced
+    by :func:`serialize_tool_catalog` when ``inherited_capabilities`` are
+    present.
+    """
     if runtime_handle is None:
         return None
 
     tool_catalog = runtime_handle.metadata.get("tool_catalog")
-    if not isinstance(tool_catalog, list):
-        return None
-    return list(tool_catalog)
+    if isinstance(tool_catalog, list):
+        return list(tool_catalog)
+    if isinstance(tool_catalog, dict):
+        return dict(tool_catalog)
+    return None
 
 
 def runtime_handle_capability_graph(
--- a/src/ouroboros/orchestrator/mcp_tools.py
+++ b/src/ouroboros/orchestrator/mcp_tools.py
@@ -202,13 +202,20 @@ class SessionToolCatalogEntry:
 
 @dataclass(frozen=True, slots=True)
 class SessionToolCatalog:
-    """Deterministic merged tool catalog for a single runtime session."""
+    """Deterministic merged tool catalog for a single runtime session.
+
+    ``inherited_capabilities`` tracks MCP/bridge tool names that a parent
+    session authorized for the child **without** synthesizing catalog entries.
+    These names are preserved for observability and downstream authorization
+    but do **not** appear in ``tools`` / ``entries`` (no phantom definitions).
+    """
 
     tools: tuple[MCPToolDefinition, ...] = field(default_factory=tuple)
     attached_tools: tuple[MCPToolDefinition, ...] = field(default_factory=tuple)
     entries: tuple[SessionToolCatalogEntry, ...] = field(default_factory=tuple)
     attached_entries: tuple[SessionToolCatalogEntry, ...] = field(default_factory=tuple)
     conflicts: tuple[ToolConflict, ...] = field(default_factory=tuple)
+    inherited_capabilities: frozenset[str] = field(default_factory=frozenset)
 
 
 def _infer_tool_input_type(value: Any) -> ToolInputType:
@@ -1179,17 +1186,35 @@ def _restore_tool_definition_for_catalog_source(
 
 
 def normalize_serialized_tool_catalog(
-    tool_catalog: Sequence[Mapping[str, Any]] | None,
+    tool_catalog: Sequence[Mapping[str, Any]] | Mapping[str, Any] | None,
     *,
     tool_prefix: str = "",
 ) -> SessionToolCatalog | None:
-    """Rehydrate a serialized startup/session catalog into `SessionToolCatalog`."""
+    """Rehydrate a serialized startup/session catalog into `SessionToolCatalog`.
+
+    Accepts both the legacy list format and the dict format produced by
+    :func:`serialize_tool_catalog` when ``inherited_capabilities`` are present.
+    """
     if not tool_catalog:
         return None
 
+    # Handle the dict format with inherited_capabilities
+    inherited_capabilities: frozenset[str] = frozenset()
+    entries: Sequence[Mapping[str, Any]]
+    if isinstance(tool_catalog, Mapping):
+        raw_entries = tool_catalog.get("entries", ())
+        entries = raw_entries if isinstance(raw_entries, Sequence) else ()
+        raw_inherited = tool_catalog.get("inherited_capabilities", ())
+        if isinstance(raw_inherited, Sequence) and not isinstance(raw_inherited, str):
+            inherited_capabilities = frozenset(str(n) for n in raw_inherited)
+    else:
+        entries = tool_catalog
+
     builtin_tools: list[MCPToolDefinition] = []
     attached_tools: list[MCPToolDefinition] = []
-    for entry in tool_catalog:
+    for entry in entries:
+        if not isinstance(entry, Mapping):
+            continue
         definition = _normalize_session_catalog_tool_definition(entry)
         if definition is None:
             continue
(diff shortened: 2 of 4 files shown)


Q00 (changes requested):
Requesting changes because this fixes phantom MCP tool definitions by dropping inherited non-builtin tools, but it also changes the delegated-session capability contract. When the child does not rediscover MCP tools locally, inherited MCP capabilities can disappear entirely rather than being represented safely. Please separate allowed-capability inheritance from catalog normalization or discovery so valid inherited MCP tools are preserved without synthesizing fake definitions.

Q00 (changes requested):
@shaun0927 
Requesting changes because this still does not satisfy the previous review requirement: preserve inherited MCP capabilities without synthesizing phantom definitions.

  I reproduced two runtime regressions locally on commit `34ec578`:

  1. `tool_catalog` metadata shape changed, but the runtime-handle reader was not updated.

  `serialize_tool_catalog()` now returns a dict when `inherited_capabilities` is present (`src/ouroboros/orchestrator/mcp_tools.py`), but `runtime_handle_tool_catalog()`
  still only accepts a list (`src/ouroboros/orchestrator/adapter.py`). A direct round-trip through `RuntimeHandle.metadata["tool_catalog"]` now fails:

  ```python
  serialized = serialize_tool_catalog(catalog_with_inherited_capabilities)
  type(serialized).__name__   # "dict"
  runtime_handle_tool_catalog(handle)   # None
```
  So the new metadata shape is not actually consumable at runtime.

  2. Inherited MCP capabilities are still lost once MCP discovery succeeds.

  _get_merged_tools() correctly separates inherited non-builtin tools into session_catalog.inherited_capabilities before discovery (src/ouroboros/orchestrator/runner.py).
  But when discovery succeeds, it replaces that catalog with provider.session_catalog, which drops the inherited capabilities again.

  I reproduced this with:

  - inherited tool: mcp__chrome-devtools__click
  - discovered MCP tool: an unrelated external tool

  and the returned catalog ended up with:

  tool_catalog.inherited_capabilities == frozenset()

  So the delegated-session capability contract is still not preserved in the success path.

  Because of these two issues, this is not just a missing-test problem. The current implementation still has runtime regressions in the exact area this PR is trying to fix.

  Please update the runtime-handle path to consume the new serialized shape, and preserve/merge inherited MCP capabilities after discovery instead of overwriting them.

shaun0927:
**Labels**: `bug` `orchestrator` — Fixes #351

Q00 (approved):
LGTM. Latest head closes the earlier inherited-capability regressions and the runtime-handle round-trip is consistent on the current SHA.

Follow-up:
Add one end-to-end regression that serializes a SessionToolCatalog with inherited capabilities through the exact delegated-session runtime-handle path, so future catalog-shape changes cannot silently drop that metadata again.

PR #353, opened 2026-04-10
Teach the runtime engine what tools mean before it schedules them

Q00:
Adopt the first architecture slice from the Claude runtime adoption analysis by adding a capability graph, a policy plane, control-plane hints, and a derived coordinator envelope while preserving the existing tool-catalog seam.

image

New architecture provide the hints and description of tools and mcp. 
It is like
```

hammer 
safety: normal
parallel: true
approval: normal

saw
safety: low
parallel: false
approval: needs approval
```

https://claude.com/blog/best-practices-for-using-claude-opus-4-7-with-claude-code

The model calls tools less often and  reasons more. This produces better results in many cases.

### DIFF
--- a/src/ouroboros/claude_permissions.py
+++ b/src/ouroboros/claude_permissions.py
@@ -1,0 +1,58 @@
+"""Shared Claude Code permission policy helpers.
+
+Translation table between the engine-owned
+:class:`~ouroboros.orchestrator.policy.SandboxClass` vocabulary and the Claude
+Agent SDK's ``permission_mode`` values.  Mirrors the role of
+``codex_permissions.py`` for Claude so every provider adapter reads the same
+sandbox enum and resolves it through its own flat lookup table.  The module
+deliberately does not make policy decisions — those live in
+``orchestrator/policy.py``.
+"""
+
+from __future__ import annotations
+
+from typing import Literal
+
+import structlog
+
+from ouroboros.sandbox import SandboxClass
+
+log = structlog.get_logger(__name__)
+
+# Claude SDK accepts a fixed vocabulary for ``permission_mode``.  Exposed as
+# a type alias so downstream adapter code carries the narrow type rather than
+# an opaque ``str``.
+ClaudePermissionMode = Literal["default", "acceptEdits", "bypassPermissions"]
+
+# Engine SandboxClass → Claude SDK permission_mode.
+#
+# The mapping happens to be identity for the legacy mode strings because
+# Ouroboros previously adopted Claude's vocabulary as the shared lingua
+# franca.  Introducing the enum flips the ownership: the SandboxClass is
+# now authoritative and these strings are Claude-specific translations.
+_SANDBOX_TO_CLAUDE_MODE: dict[SandboxClass, ClaudePermissionMode] = {
+    SandboxClass.READ_ONLY: "default",
+    SandboxClass.WORKSPACE_WRITE: "acceptEdits",
+    SandboxClass.UNRESTRICTED: "bypassPermissions",
+}
+
+
+def claude_permission_mode_for_sandbox(sandbox: SandboxClass) -> ClaudePermissionMode:
+    """Translate an engine sandbox class into a Claude SDK permission_mode.
+
+    Raises ``KeyError`` if the enum grows and Claude's table was not updated —
+    failing loudly beats silently defaulting to a possibly-permissive mode.
+    """
+    mode = _SANDBOX_TO_CLAUDE_MODE.get(sandbox)
+    if mode is None:
+        msg = f"No Claude SDK permission_mode registered for sandbox class {sandbox!r}"
+        raise KeyError(msg)
+    if sandbox is SandboxClass.UNRESTRICTED:
+        log.warning("permissions.bypass_activated", sandbox=sandbox.value)
+    return mode
+
+
+__all__ = [
+    "ClaudePermissionMode",
+    "claude_permission_mode_for_sandbox",
+]
--- a/src/ouroboros/codex_permissions.py
+++ b/src/ouroboros/codex_permissions.py
@@ -1,8 +1,12 @@
 """Shared Codex CLI permission policy helpers.
 
-This module centralizes how Ouroboros maps internal permission modes onto the
-currently supported Codex CLI flags. Both the agent runtime and the Codex-based
-LLM adapter use the same policy so permission behavior stays predictable.
+This module is the *translation table* between the engine-owned
+:class:`~ouroboros.orchestrator.policy.SandboxClass` vocabulary and the Codex
+CLI's native ``--sandbox`` / ``--full-auto`` flags.  It deliberately does not
+make policy decisions; those live in ``orchestrator/policy.py``.  Both the
+Codex agent runtime and the Codex-based LLM adapter go through this module,
+so Codex-side behavior stays consistent and is derived from the same sandbox
+enum every other provider maps.
 """
 
 from __future__ import annotations
@@ -11,11 +15,33 @@ from typing import Literal
 
 import structlog
 
+from ouroboros.sandbox import SandboxClass
+
 log = structlog.get_logger(__name__)
 
 CodexPermissionMode = Literal["default", "acceptEdits", "bypassPermissions"]
 
 _VALID_PERMISSION_MODES = frozenset({"default", "acceptEdits", "bypassPermissions"})
+
+# Legacy permission-mode vocabulary → engine SandboxClass.  The ``default``
+# mode is read-only, ``acceptEdits`` maps to workspace-write (the Codex
+# ``--full-auto`` flag), and ``bypassPermissions`` removes both the sandbox
+# and the approval gate.  External callers that still speak the string
+# vocabulary funnel through this mapping on their way to the sandbox enum.
+_PERMISSION_MODE_TO_SANDBOX: dict[CodexPermissionMode, SandboxClass] = {
+    "default": SandboxClass.READ_ONLY,
+    "acceptEdits": SandboxClass.WORKSPACE_WRITE,
+    "bypassPermissions": SandboxClass.UNRESTRICTED,
+}
+
+# Engine SandboxClass → Codex CLI invocation flags.  This is the only place
+# Codex-specific flag names should appear for sandbox selection; new sandbox
+# classes must add an entry here or the invariant test fails.
+_SANDBOX_TO_CODEX_ARGS: dict[SandboxClass, list[str]] = {
+    SandboxClass.READ_ONLY: ["--sandbox", "read-only"],
+    SandboxClass.WORKSPACE_WRITE: ["--full-auto"],
+    SandboxClass.UNRESTRICTED: ["--dangerously-bypass-approvals-and-sandbox"],
+}
 
 
 def resolve_codex_permission_mode(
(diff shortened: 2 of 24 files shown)


shaun0927:
Great work — this is the architecture slice I was hoping to see. I've been tracing the tool delegation path for the performance PRs (#347–#350) and ran into a concrete problem that this PR structurally addresses. Some notes from that experience:

## What this PR solves well

The `origin` field in `CapabilitySemantics` (BUILTIN / ATTACHED_MCP / PROVIDER_NATIVE / FUTURE_RUNTIME) is exactly the missing piece I hit when debugging phantom tool entries in #351. The engine couldn't distinguish "Read" (always available) from "mcp__chrome-devtools__click" (requires live bridge connection) because both were plain strings in `inherited_tools`. With `origin` as a first-class field, that distinction becomes structural.

The `scope` field (KERNEL / SIDECAR / ATTACHMENT / SHELL_ONLY) is a clean way to express what I was trying to solve with a blunter instrument — filtering against `enumerate_runtime_builtin_tool_definitions()`. Scope is more expressive and forward-compatible.

The policy-derived coordinator envelope (`derive_coordinator_tools()`) replacing `COORDINATOR_TOOLS` hardcoding is the right move. This makes the coordinator's tool access auditable and consistent with the capability graph.

## Merge conflict with #352

My PR #352 modifies `runner.py:993-996` (inherited tools filtering), and this PR modifies `runner.py:998+` (catalog → capability graph → policy). The changes are adjacent but on different lines — they should rebase cleanly. I'll rebase #352 onto main after this merges.

The two PRs are complementary layers:
- **#352**: prevents non-builtin tools from entering `base_tools` before catalog assembly (input validation)
- **#353**: assigns semantic meaning after catalog assembly and filters by policy (output validation)

Together they guarantee: a tool name cannot reach the LLM's `allowed_tools` unless it (a) is a known builtin or discovered by a live bridge, AND (b) passes the policy plane for the current session role.

## Gaps I noticed — possible follow-ups

**1. Policy decision events**

The policy plane evaluates `visible` / `executable` per capability, but these decisions aren't persisted to the event store. For audit ("did any session bypass approval for a destructive tool?") and debugging ("why was my tool hidden?"), emitting `policy.capability.evaluated` events would be valuable. The event store already supports JSON payloads, so this is low-cost.

**2. Interview/evaluation handler wiring**

`PolicySessionRole.INTERVIEW` and `PolicySessionRole.EVALUATION` profiles are defined with `max_mutation_class=READ_ONLY`, but the actual `InterviewHandler` and `EvaluateHandler` don't consume policy yet — they still use whatever tools the runner provides. Wiring these handlers through `allowed_capability_names()` with the appropriate role would enforce the read-only constraint at the engine level rather than relying on handler-level self-discipline.

**3. Inherited tools + origin awareness**

The `inherited_tools` path in `_resolve_tools()` still operates on plain `list[str]` before the capability graph is built. If `_build_delegated_tool_context_update()` (adapter.py:174) could attach `origin` metadata to delegated tools, the child runner could make smarter decisions — e.g., skip ATTACHED_MCP tools when `mcp_manager` is None, instead of relying on name-based filtering.

**4. `_infer_attached_semantics()` coverage**

The keyword-matching heuristic ("delete" → DESTRUCTIVE, "read" → READ_ONLY) is a reasonable starting point, but tools like `chrome_navigate` or `puppeteer_screenshot` won't match any keyword and will fall through to the `EXTERNAL_SIDE_EFFECT / SERIALIZED / ELEVATED` default. A user-override config path (`~/.ouroboros/tool_capabilities.yaml`) could let users correct misclassified tools without code changes.

Happy to pick up any of these as follow-up PRs.

Q00:
## Maintainer note: intent behind this capability/policy slice

The intent of this PR is to stop treating tools as plain strings.

As agent runtimes become more capable, they also become more selective about when to call tools, when to reason internally, and when to fan out work. That makes the framework boundary more important: the engine needs to communicate not only *which* tools exist, but also *what those tools mean*.

This PR introduces the first architecture slice for that:

- a capability graph derived from the session tool catalog
- explicit capability semantics such as origin, scope, mutation class, parallel safety, interruptibility, and approval class
- a policy plane that turns those semantics into visible/executable decisions
- control-plane metadata that can guide runtime execution shape
- coordinator tool envelopes derived from policy instead of hardcoded lists
- audit events for policy decisions

The larger design goal is runtime-agnostic control. The engine should not keep adding provider-specific permission branches every time Claude, Codex, OpenCode, Gemini, or an MCP bridge exposes a different tool shape. Instead, provider runtimes should be able to consume a shared engine-owned description of capability semantics and policy decisions.

This matters especially for newer agent behavior where models may use larger context windows and rely more on internal reasoning instead of tool calls. In that world, the harness must make implicit human knowledge explicit: which tools are read-only, which tools mutate the workspace, which tools require approval, which tools are unsafe to run in parallel, and which tools are only valid when a live bridge/provider connection exists.

This PR is deliberately scoped as a foundation rather than a full scheduler rewrite. It preserves the existing tool-catalog seam and keeps behavioral risk moderate while giving follow-up PRs a stable layer to build on.

## Follow-up PRs

The follow-up stack turns this PR from metadata into enforcement and operational structure:

1. #448 — Enforce read-only policy envelopes for MCP handlers  
   Wires `InterviewHandler` and `EvaluateHandler` through the INTERVIEW/EVALUATION read-only policy profiles. These handlers now consume the policy plane directly instead of relying on convention.

2. #449 — Batch capability policy audit events  
   Replaces per-capability policy event writes with a batched `policy.capabilities.evaluated` event. This keeps auditability while avoiding event-store write amplification as tool catalogs grow.

3. #450 — Support user capability overrides  
   Adds `~/.ouroboros/tool_capabilities.yaml` / `OUROBOROS_TOOL_CAPABILITIES` overrides for attached MCP tool semantics. This addresses the fact that keyword-based inference is intentionally conservative and can misclassify tools such as browser navigation or screenshots.

4. #451 — Select role policy by origin and scope  
   Moves policy matching beyond simple name whitelists. Provider-native and future-runtime capabilities can now be admitted by semantic selectors such as origin, scope, and stable-id prefix.

5. #452 — Consume control-plane hints in batch scheduling  
   Makes the parallel executor actually consume control-plane hints. When executable tools require serialized or isolated execution, same-batch AC fanout is constrained accordingly.

Together, these PRs move Ouroboros toward a model where tools are not just callable functions, but typed capabilities with policy, auditability, and scheduling semantics.

That is the core architectural bet here: as models become stronger and more adaptive, the framework must become more explicit about tool meaning, execution context, and policy boundaries.

Q00:
## Layering update: scheduler decoupled from capability classification

Following up on the reviewer's note that the executor was consuming control-plane hints at the wrong granularity (session-level tool availability being read as proof that every AC in the batch must run serially), this PR now separates *what a capability means* from *when it forces scheduling decisions*. The capability graph, policy decisions, and control-plane state remain a declarative audit surface; the batch scheduler no longer inspects them.

### Why the old layering was wrong

Treating "tool X is in the session catalog" as evidence for "AC₀ and AC₁ must not run concurrently" conflates three different questions:

1. Is this tool declared in the session?
2. Will this specific AC actually invoke it?
3. Would two concurrent invocations race with each other?

Only (1) is knowable at batch-start. (2) is not knowable until the model issues a tool call. (3) is a runtime property, not a scheduling property. Collapsing them made the ``parallel_safety`` enum dishonest — a capability classified as ``SERIALIZED`` is meant to declare *tool-call-level* serialization (same runtime session cannot invoke it concurrently), not *batch-level* serialization.

### New layering

```
┌──────────────────────────────────────────────────────────────────────────┐
│ DECLARATIVE (metadata / audit only)                                      │
│                                                                          │
│   session_catalog                                                        │
│        │                                                                 │
│        ▼                                                                 │
│   capability_graph  ── user overrides from                               │
│        │              ~/.ouroboros/tool_capabilities.yaml                │
│        │              (mtime-cached, merge-on-inferred)                  │
│        ▼                                                                 │
│   policy_decisions  ── per role                                          │
│        │              (IMPLEMENTATION / COORDINATOR /                    │
│        │               INTERVIEW / EVALUATION)                           │
│        ▼                                                                 │
│   control_plane_state                                                    │
│        │                                                                 │
│        └──▶ runtime_handle.metadata  +  policy.capabilities.evaluated    │
│             (persisted for replay / debugging / auditability)            │
└──────────────────────────────────────────────────────────────────────────┘
                                  │
                                  │ (no longer drives scheduling)
                                  ▼
┌──────────────────────────────────────────────────────────────────────────┐
│ STATIC cross-AC safety                                                   │
│                                                                          │
│   LevelCoordinator.detect_file_conflicts(batch_indices)                  │
│        └─ prevents same-file races across ACs before they run            │
└──────────────────────────────────────────────────────────────────────────┘
                                  │
                                  ▼
┌──────────────────────────────────────────────────────────────────────────┐
│ EXECUTION (scheduler is now tool-catalog-agnostic)                       │
│                                                                          │
│   parallel_executor._execute_ac_batch(batch_indices):                    │
│       async with anyio.create_task_group() as tg:                        │
│           for ac_idx in batch_indices:                                   │
│               tg.start_soon(_run_ac, ac_idx)                             │
│                                                                          │
│   Cross-AC parallelism is the default.  Mixed catalogs                   │
│   (Read + Edit + Bash) still fan out in parallel, pinned by              │
│   ``test_batch_fans_out_in_parallel_regardless_of_tool_catalog``.        │
└──────────────────────────────────────────────────────────────────────────┘
                                  │
                                  ▼
┌──────────────────────────────────────────────────────────────────────────┐
│ DYNAMIC tool-call safety (correct home for the enum)                     │
│                                                                          │
│   Each AC runs in its own provider runtime session (worker pool),        │
│   so cross-AC invocations of ISOLATED_SESSION_REQUIRED capabilities      │
│   are naturally isolated.  Within a single session, the provider         │
│   runtime is responsible for serializing concurrent invocations —        │
│   which is the layer ``parallel_safety`` was always meant to live at.    │
└──────────────────────────────────────────────────────────────────────────┘
```

### Supporting cleanups shipped with this layering change

These came out of the same elegance pass and land in separate commits so the scheduler change can be reviewed on its own:

- **Single override integration path** — ``build_capability_graph`` no longer accepts a pre-coerced ``capability_overrides=`` kwarg. Raw YAML from ``OUROBOROS_TOOL_CAPABILITIES`` (or ``~/.ouroboros/tool_capabilities.yaml``) is merged *onto* inferred semantics per-field. The legacy wholesale-replacement path and the public ``load_tool_capability_overrides`` helper are removed; they had no external consumers.
- **Resilient override loader** — YAML parse errors, unreadable paths, and invalid enum values all log a structured warning and return an empty mapping. The override layer is an optional enhancement; a single bad line in user config must never take down interview / evaluation / execution.
- **Single-responsibility coercion** — ``_coerce_capability_semantics`` no longer both logs and re-raises. It logs and returns ``None``; the one caller decides what to do on failure. The duplicate ``_default_attached_semantics()`` factory and ``_INHERITED_CAPABILITY_SEMANTICS`` constant are collapsed into a single ``_DEFAULT_ATTACHED_SEMANTICS``.
- **YAGNI removal in policy** — ``RoleCapabilityProfile.allowed_stable_id_prefixes`` had no profile populating it and no consumer reading it; removed along with the dead clause in ``_matches_role_selector``.
- **Batched audit events** — per-capability ``policy.capability.evaluated`` events are replaced with a single batched ``policy.capabilities.evaluated`` event per evaluation, eliminating event-store write amplification as tool catalogs grow.
- **Test tightening** — the partial-override test now pins every dimension (the declared one to the user's value, the undeclared ones to the exact inferred values for a read-fingerprinted attached tool), so a regression back to wholesale replacement would fail fast.

### Net effect

- ``parallel_safety`` classifications stay truthful: they describe tool-call-level semantics and nothing more.
- Batch scheduling is now tool-catalog-agnostic and does not degrade to serial on false signals.
- The capability/policy/control-plane layer keeps its value as a declarative audit surface for replay and debugging, exactly as intended, and is ready for the real enforcement layer (tool-call-level provider guards) to be wired up in a follow-up PR without rework.

Issue #269, opened 2026-03-30
evolve agent subprocesses (claude_agent_sdk) never terminate after completion, causing unbounded memory growth


isac322:
## Environment

- **ouroboros-ai**: v0.26.4
- **Claude Code**: 2.1.86
- **OS**: Linux 6.18.20-1-MANJARO (x86_64)
- **Runtime**: Claude (claude_agent_sdk)
- **System RAM**: 64GB

## Summary

During an `evolve` run (Generation 3), Ouroboros spawns Claude CLI subprocesses via `claude_agent_sdk` for each AC/sub-AC. **These agent processes never terminate after completing their work**, accumulating indefinitely and consuming unbounded memory.

## Observed behavior

- Session: `orch_8edc447a4229`, Execution: `exec_ca754fdf799a`
- 10 ACs executing in parallel, decomposed into sub-ACs
- **47 agent subprocesses** accumulated over ~3.5 hours (16:45–19:57)
- Oldest agents running for **4+ hours** after their work completed
- Each agent consumes 115–323 MB RSS; **total 7.3 GB**
- `session_status` reports `runtime_signal: session_completed` and `runtime_status: completed`, yet the underlying OS processes remain alive

## Process tree

```
Claude Code (pts/7, PID 805952)
  └── uv ouroboros mcp serve (PID 805992)
       └── python ouroboros (PID 806090)
            ├── claude agent (PID 1359111) — started 16:45, RSS 119MB, 4h+
            ├── claude agent (PID 1359174) — started 16:45, RSS 116MB, 4h+
            ├── ... (45 more)
            └── claude agent (PID 1956419) — started 19:57, RSS 323MB
```

All agents have ppid=806090 (the ouroboros python process). No orphaned processes (ppid=1) were found. This indicates ouroboros spawns child processes but does not terminate or reap them after completion.

## Additional observation: status inconsistency

The `session_status` response shows contradictory state:
- `runtime_status: completed`, `runtime_signal: session_completed`
- But top-level `status: running`, `completed_count: 0`
- `ac_tracking.completed: [3]`, yet AC 3 is listed as `executing` in the AC array

This may be a separate issue, but it suggests the orchestrator is not properly propagating completion signals from finished agents.

## Expected behavior

- Agent processes should be terminated (or terminate themselves) once their AC/sub-AC work is complete and results have been collected by the orchestrator
- At minimum, the orchestrator should kill completed agent processes after collecting their output

## Impact

[...]

Issue #199, opened 2026-03-24
v0.26.0b5: post-run QA fails because decomposed AC summaries discard verification evidence


srausser:
## Summary

In `v0.26.0b5`, `ouroboros run` can produce successful executions that still fail post-run QA because the generated `test_output` artifact is not auditable.

The recurring failure pattern is that decomposed acceptance criteria are summarized as placeholder text like `Decomposed into N Sub-ACs` instead of concrete verification evidence.

This causes QA to return `REVISE` even when the underlying implementation and targeted tests appear to pass.

## Version

- `ouroboros --version` -> `Ouroboros version 0.26.0b5`

## Observed Behavior

Across multiple runs, post-execution QA reports the same structural problems:

- PASS claims without exact test names, assertions, or outputs
- AC sections reduced to `Decomposed into N Sub-ACs`
- summarized/meta-report output instead of auditable per-AC evidence
- truncated or garbled sections for larger runs
- lack of authoritative command normalization in the final artifact

Example QA sessions:
- `qa-a104af1a`
- `qa-3f15ce73`

Representative QA complaints:
- "most criteria only show a status line like `Decomposed into N Sub-ACs`"
- "the artifact is a summarized meta-report rather than raw or linked test output"
- "PASS claims are unsupported by concrete evidence"

## Expected Behavior

When an AC is decomposed, the final execution artifact should still contain auditable parent-level evidence, for example:

- AC text
- PASS/FAIL
- exact test command(s)
- exact test file or test case names
- pass counts / outputs
- concise assertion-level evidence
- sub-AC evidence surfaced under the parent AC

The artifact sent to QA should be a strict per-AC verification report, not a placeholder summary.

## Likely Cause

The problem appears to be in the execution summary/reporting path:

- decomposed ACs return a parent `final_message` like `Decomposed into N Sub-ACs`
- downstream summary generation uses that placeholder instead of aggregating `sub_results`
- some summary/display paths also clip evidence aggressively, which makes larger reports non-auditable

## Impact

This blocks QA signoff for otherwise-successful runs, especially on larger seeds with many decomposed ACs. The issue appears generic and not repo-specific.

## Suggested Fix

- aggregate sub-AC evidence into the parent AC `final_message`
- emit a structured per-AC evidence block for both atomic and decomposed ACs
- stop using placeholder `Decomposed into N Sub-ACs` as the durable artifact
- avoid naive character-tail truncation in the final QA artifact
- preserve one authoritative verification command/result block per AC

[...]

PR #368, opened 2026-04-11
fix(mcp): evaluator must not reuse max_turns=1 shared adapter (fixes #305)

### LINKED ISSUE #305: ooo evaluate fails: Claude Agent SDK 'Command failed with exit code 1' in nested session
## Summary

`ouroboros_evaluate` MCP tool fails with `Claude Agent SDK request failed: Command failed with exit code 1` when invoked from within an active Claude Code session.

## Environment

- **Ouroboros**: v0.27.0 (latest as of 2026-04-03)
- **Claude Code**: 2.1.91
- **claude-agent-sdk**: 0.1.55
- **Python**: 3.14.3 (Homebrew, macOS)
- **OS**: macOS Darwin 25.3.0

## Steps to Reproduce

1. Start a Claude Code session
2. Run `ooo ralph` on a seed (completes successfully — execution + QA pass)
3. Run `ooo evaluate` on the same session

## Expected Behavior

3-stage evaluation pipeline (mechanical → semantic → consensus) runs and returns results.

## Actual Behavior

```
Error executing tool ouroboros_evaluate: Evaluation failed: Claude Agent SDK request failed: Command failed with exit code 1 (exit code: 1)
Error output: Check stderr output for details (details: {'error_type': 'Exception'})
```

## Root Cause Analysis

Traced the error to `EvaluateHandler.handle()` in `evaluation_handlers.py:415`:

```python
llm_adapter = self.llm_adapter or create_llm_adapter(
    backend=self.llm_backend,
    max_turns=20,
)
```

This creates a `ClaudeCodeAdapter` → calls `claude_agent_sdk.query()` → spawns a **nested Claude CLI subprocess** inside the already-active Claude Code session.

The nested CLI process exits with code 1. The existing `CLAUDECODE` env var workaround at `claude_code_adapter.py:548-550` is insufficient — the nested CLI still fails to start.

`ouroboros_start_evolve_step` (ralph) works because it uses a different execution path that doesn't spawn a nested semantic evaluator via the Agent SDK.

## Additional Notes

- **Validation step** in the ralph result also reported: `Validation skipped: [Errno 2] No such file or directory: 'python'` (fixed by symlinking `python3` → `python`, but evaluate still fails after the fix)
- `ooo ralph` and `ooo run` work correctly — only `ooo evaluate` is affected
- Manual 3-stage verification works fine as a workaround

## Workaround

Perform evaluation manually or use the QA verdict from `ooo ralph` (which includes its own QA pass with scoring).

minsing-jin:
## Summary

Fixes the `ooo evaluate` / `ouroboros_evaluate` MCP tool failure described
in #305 (`Claude Agent SDK request failed: Command failed with exit code 1`).

Root cause analysis on `main` (v0.28.3) shows the bug has a different
underlying mechanism than #305 originally hypothesised. The nested Claude
CLI subprocess exits with `exit code 1` because the SDK query hits
`error_max_turns` on the very first tool call — not because the nested CLI
itself fails to start.

## Root Cause

`build_mcp_server` constructs a single shared `llm_adapter` with
`max_turns=1`, which is correct for interview and seed-generation paths
(those are single-shot completions):

```python
# src/ouroboros/mcp/server/adapter.py:805
llm_adapter = create_llm_adapter(
    backend=llm_backend,
    max_turns=1,     # correct for interview / seed generation
    cwd=Path.cwd(),
)
```

That adapter is then injected into every handler including
`EvaluateHandler`. The handler's adapter lookup uses the `or` short-circuit
pattern:

```python
# src/ouroboros/mcp/tools/evaluation_handlers.py:405
llm_adapter = self.llm_adapter or create_llm_adapter(
    backend=self.llm_backend,
    max_turns=20,
)
```

Because `self.llm_adapter` is *always truthy* in production (`build_mcp_server`
injected it), the `or` branch never runs and the `max_turns=20` fallback is
dead code. Stage 2 semantic evaluation inherits the `max_turns=1` ceiling
and fails on the very first `Read` tool call it issues to inspect a spec
file, surfacing as `error_max_turns` in the SDK and `exit code 1` at the
MCP tool boundary.

The existing `CLAUDECODE` env stripping workaround at
`claude_code_adapter.py:557-564` does not help here — it is a separate
issue about nested session detection.

## Fix

Always construct a fresh adapter in `EvaluateHandler.handle`, regardless of
whether one was injected. The shared adapter is still appropriate for
interview / seed generation; only evaluation needs the higher budget.

**`src/ouroboros/mcp/tools/evaluation_handlers.py`**
- Drop the `or` short-circuit; always call `create_llm_adapter(max_turns=20)`.
- Expand inline comment to document the invariant and reference #305.

**`src/ouroboros/mcp/server/adapter.py`**
- No code change. Add a block comment pointing future readers at the
  evaluator path so the interaction between the shared `max_turns=1` and
  the evaluator's private adapter is obvious. Prevents a well-meaning
  refactor from \"bumping max_turns from 1 to 20\" and inadvertently
  breaking interview latency.

## Tests

**`tests/unit/mcp/tools/test_evaluation_handler.py`**

- `test_injected_adapter_skips_create_llm_adapter` — renamed to
  `test_injected_adapter_is_ignored_for_evaluation`. The old test
  *guaranteed* the buggy short-circuit; the new assertion flips the
  invariant: even when an adapter is injected, `create_llm_adapter` must
  still be called with a sufficient `max_turns` budget.
- New regression test `test_injected_max_turns_1_adapter_does_not_leak`:
  constructs an adapter with `_max_turns = 1`, injects it, and asserts
  that the `EvaluationPipeline` receives a *different* adapter instance.
  Guards against a future refactor re-introducing the short-circuit
  pattern.
- `test_creates_adapter_with_sufficient_max_turns` unchanged — still
  asserts `max_turns >= 10` for the adapter used inside the pipeline.

Full suite:

```
.venv/bin/python -m pytest tests/unit/mcp/ -q
584 passed in 4.29s
```

## Verification (external reproducer)

Before the fix, running the `EvaluationPipeline` directly on an
in-the-wild seed:

```
warning mcp.tool.evaluate.pipeline_failed
  error=\"Claude Agent SDK request failed: Command failed with exit code 1
         ... subtype='error_max_turns' max_turns=1 ...\"
```

After patching the same paths on my local install, 32 evaluation calls
across 17 task-level ACs + 5 retroactive audits + 1 holistic aggregate
returned `final_approved=True` (aggregate score 0.88). The same code
also runs cleanly under the existing unit tests.

## Compatibility

- Backwards compatible for all non-test callers: the shared adapter
  stays on `max_turns=1`, interview latency unchanged.
- External callers passing `llm_adapter=` to `EvaluateHandler(...)`
  should note that the injected adapter is no longer consulted for the
  evaluation pipeline (it was broken anyway for `max_turns<10`).
- Interview / seed generation paths untouched.

## Related

- Fixes #305 (evaluator exit code 1 in nested session)
- Adjacent but separate: #269 (evolve subprocess leak), #310 (MCP server
  startup scan), #199 (decomposed AC summary evidence loss)

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -802,6 +802,15 @@ def create_ouroboros_server(
     )
 
     # Create shared LLM adapter for interview/seed/evaluation paths.
+    #
+    # NOTE: ``max_turns=1`` is appropriate for interview question generation
+    # and seed synthesis, where the adapter returns a single-shot response.
+    # Evaluation (Stage 2 semantic) DOES NOT use this adapter — see
+    # ``EvaluateHandler.handle`` in ``mcp/tools/evaluation_handlers.py`` which
+    # constructs a fresh adapter with ``max_turns=20`` so the semantic
+    # evaluator can issue tool calls when reading spec files. Do not bump
+    # this value without first auditing every caller of ``self.llm_adapter``
+    # — it affects interview latency and token usage.
     llm_adapter = create_llm_adapter(
         backend=llm_backend,
         max_turns=1,
--- a/src/ouroboros/mcp/tools/evaluation_handlers.py
+++ b/src/ouroboros/mcp/tools/evaluation_handlers.py
@@ -402,7 +402,23 @@ class EvaluateHandler:
             # Use injected or create services.
             # Evaluation reads multiple spec files (one Read call per AC), so
             # max_turns must be well above 1.
-            llm_adapter = self.llm_adapter or create_llm_adapter(
+            #
+            # IMPORTANT: The shared ``self.llm_adapter`` that may have been
+            # injected by ``build_mcp_server`` (see ``mcp/server/adapter.py``)
+            # is constructed with ``max_turns=1`` for interview / seed-generation
+            # use cases. That value is fatal for evaluation because the Stage 2
+            # semantic evaluator issues at least one ``Read`` tool call per AC to
+            # inspect spec files. With ``max_turns=1`` the Agent SDK returns
+            # ``error_max_turns`` on the very first tool call, surfacing as
+            # ``Command failed with exit code 1`` at the MCP tool boundary.
+            #
+            # To ensure the evaluator always has enough turns, construct a
+            # fresh adapter here rather than relying on the injected instance.
+            # The old ``self.llm_adapter or create_llm_adapter(...)`` pattern
+            # short-circuited on the injected adapter and silently preserved
+            # the ``max_turns=1`` ceiling — see issue #305 for the user-facing
+            # failure it produced.
+            llm_adapter = create_llm_adapter(
                 backend=self.llm_backend,
                 max_turns=20,
             )
--- a/tests/unit/mcp/tools/test_evaluation_handler.py
+++ b/tests/unit/mcp/tools/test_evaluation_handler.py
@@ -135,14 +135,91 @@ class TestEvaluateHandlerAdapterCreation:
             "at least one turn per AC file read. Use max_turns >= 10."
         )
 
-    async def test_injected_adapter_skips_create_llm_adapter(self):
-        """When an adapter is injected (e.g. in tests), create_llm_adapter is not called."""
-        handler = EvaluateHandler(llm_adapter=_make_mock_adapter())
+    async def test_injected_adapter_is_ignored_for_evaluation(self):
+        """Evaluation ALWAYS creates a fresh adapter, even when one is injected.
+
+        Regression for issue #305 / related max_turns fix:
+
+        The shared adapter wired up in ``build_mcp_server``
+        (``mcp/server/adapter.py``) is constructed with ``max_turns=1`` because
+        interview and seed-generation paths only need a single-shot response.
+        If the evaluator reuses that adapter, the very first ``Read`` tool
+        call issued by the Stage 2 semantic evaluator hits ``error_max_turns``
+        and surfaces as ``Command failed with exit code 1`` at the MCP tool
+        boundary.
+
+        To prevent that, the handler now ignores ``self.llm_adapter`` and
+        always constructs a fresh adapter with ``max_turns=20``.
+        """
+        captured: dict = {}
+
+        def _capture(**kwargs):
+            captured.update(kwargs)
+            return _make_mock_adapter()
+
+        # Simulate the shared adapter that build_mcp_server would inject.
+        shared_adapter = _make_mock_adapter()
+        handler = EvaluateHandler(llm_adapter=shared_adapter)
 
         with (
-            patch("ouroboros.mcp.tools.evaluation_handlers.create_llm_adapter") as mock_create,
+            patch(
+                "ouroboros.mcp.tools.evaluation_handlers.create_llm_adapter",
+                side_effect=_capture,
+            ),
             _patch_pipeline(),
         ):
             await handler.handle(_BASE_ARGUMENTS)
 
-        mock_create.assert_not_called()
+        # Fresh adapter was created with a sufficient max_turns budget, even
+        # though one was already injected on the handler.
+        assert "max_turns" in captured, (
+            "create_llm_adapter was not called — the handler must build a "
+            "fresh adapter for evaluation instead of reusing the shared "
+            "interview-tuned adapter."
+        )
+        assert captured["max_turns"] >= 10, (
+            f"max_turns={captured['max_turns']} is too low."
+        )
+
+    async def test_injected_max_turns_1_adapter_does_not_leak(self):
+        """Regression for issue #305: injecting a max_turns=1 adapter must not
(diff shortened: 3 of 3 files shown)


jcfernandez-890825:
Independent smoke test from a downstream consumer — **patch verified end-to-end** 🎉

## Environment

- MCP runtime: `ouroboros-ai==0.27.1` in a uv tool venv (`~/.local/share/uv/tools/ouroboros-ai/`, spawned via `uvx --from ouroboros-ai[mcp,claude]`)
- Python 3.14.2 via uv-managed interpreter
- Downstream skill calling `ouroboros_evaluate` via the MCP tool

Since v0.27.1 has the same buggy code pattern as the PR target (just at slightly different line numbers — `evaluation_handlers.py:405` and `adapter.py:691`), I manually backported both hunks from this PR to the local uv tool venv copies, broke the hardlinks with the shared uv archive cache so the shared cache stays pristine, and restarted the MCP server via `/mcp` reconnect.

## Verification

- `python -m py_compile` — both files compile cleanly
- `python -c "from ouroboros.mcp.tools import evaluation_handlers; from ouroboros.mcp.server import adapter"` — both modules import cleanly
- No remaining `self.llm_adapter or create_llm_adapter(...)` short-circuit in code (only in the explanatory comment block you added)
- New MCP process (verified via `ps`) started 30 minutes after the patch — Python pycache mtime invalidation picks up the new bytecode on first import

## Runtime smoke test

Minimal `ouroboros_evaluate` call:

```
session_id = "test-pr368-smoke-001"
artifact = 'print("hello world")'
artifact_type = "code"
acceptance_criterion = "The code should print hello world"
seed_content = '''
goal: Print a hello world greeting
acceptance_criteria:
  - The script outputs "hello world"
'''
```

Result:

```
Final Approval: APPROVED
Highest Stage Completed: 2

Stage 1 (Mechanical): PASSED — 5/5 checks skipped cleanly (no commands configured)
Stage 2 (Semantic):   Score 1.00 | AC Compliance YES | Goal Alignment 1.00
                      Drift 0.00 | Uncertainty 0.02
                      Reasoning: "The artifact is a single Python statement
                      print('hello world') that directly and correctly
                      satisfies the acceptance criterion..."
```

Before the patch, the same call would have died at Stage 2 with `Claude Agent SDK request failed: Command failed with exit code 1` / `error_max_turns max_turns=1` on the first `Read` tool call — exactly the failure mode the PR body describes. The LLM quoting the artifact in its reasoning proves multi-turn tool calling is working, which is the code path the fix is targeting.

## Notes for downstream backporters (if anyone else is on v0.27.1 via uvx)

1. The `~/.claude/plugins/cache/ouroboros/ouroboros/0.28.2/` directory is the Claude Code plugin's reference source bundle, NOT the MCP runtime. Patch the uv tool venv at `~/.local/share/uv/tools/ouroboros-ai/lib/python3.14/site-packages/ouroboros/mcp/` instead.
2. The `uvx` install path is hardlinked against the shared uv archive cache. Editing the files via a tool that breaks the hardlink (most editors do) isolates the patch to only your venv copy — good for reproducibility, the global cache stays clean.
3. `/mcp` reconnect is sufficient to pick up the new code — no need to reinstall the plugin or `uv tool install --reinstall`.

Hoping this lands soon — looking forward to dropping the manual backport. Thanks @minsing-jin for the clean RCA and @ouroboros-agent[bot] for the thorough review rounds.

Q00:
Great RCA and surgical fix — the `or` short-circuit dead-code analysis is spot-on, and the regression tests properly guard against re-introduction. Thanks for tracking this down, @minsing-jin.

And thanks @jcfernandez-890825 for the thorough end-to-end verification on v0.27.1 — the backport notes are genuinely useful for anyone running into the same issue before this ships.

### Non-blocking: DI dead wire follow-up

This PR correctly makes `EvaluateHandler` always construct a fresh `max_turns=20` adapter. However, `build_mcp_server` (`adapter.py:1390`) still injects the shared `max_turns=1` adapter into `EvaluateHandler`, and `EvaluateHandler.llm_adapter` is now a dead field — it's set but never consulted by `handle()`.

I'll file a follow-up PR after this merges to clean up the dead wiring:

- Remove `llm_adapter` parameter from `EvaluateHandler`
- Remove the `llm_adapter=llm_adapter` injection at `adapter.py:1390`
- Keep `llm_backend` (already passed) as the only configuration input
- Update tests to match

This keeps the current PR's scope minimal while addressing the architectural drift in a separate change.

**LGTM — approve with non-blocking follow-up.**

Q00 (approved):

Q00:
Follow-up PR is up: #375

**Dead wire cleanup** (as discussed) + **large artifact evaluation fix**:

The directory scan fallback in `ArtifactCollector` addresses the "58KB code inline" limitation — when no file paths are extractable from the artifact text, it walks `working_dir` and collects source files. The artifact summary is written to a file and picked up naturally. No more truncation.

PR #381, opened 2026-04-14
fix(codex): unify wrapper-safe CLI resolution across runtime and provider paths

mpfo0106:
## Summary

- extract a shared `cli_policy` module for Codex CLI path resolution and child-process environment sanitation
- apply the shared policy to both `CodexCliRuntime` and `CodexCliLLMAdapter`
- pass the runtime's already-resolved Codex CLI path into nested dependency-analysis adapter construction so top-level and nested Codex calls converge on the same terminal binary
- add parity-focused tests for runtime/provider fallback behavior, child-env policy, and nested analyzer wiring

## Problem

Ouroboros had an inconsistency in how it launched Codex:

1. **Runtime path** (`CodexCliRuntime`) already recognized compiled wrapper binaries and fell back to the first real `codex` executable on `PATH`.
2. **Provider path** (`CodexCliLLMAdapter`) did **not** share that logic. It resolved `get_codex_cli_path()` directly and used the configured candidate as-is.

That left a missing parity point in nested execution flows.

A typical failure shape looked like this:

- top-level AC execution used the runtime path and correctly bypassed a wrapper binary
- nested provider-backed work (for example dependency analysis via `create_llm_adapter(...)`) could still resolve a different `codex` binary than the runtime had already selected
- the system therefore had two different Codex launch policies depending on which code path happened to fire

Even when the top-level execution was healthy, nested Codex calls could diverge into a different binary and a different subprocess environment.

## Why this fix is structured this way

This PR fixes the inconsistency at the policy boundary instead of patching a single callsite.

### Before
- wrapper-safe resolution logic lived only inside `CodexCliRuntime`
- child env sanitation was duplicated between runtime and provider code
- nested dependency-analysis calls rebuilt a provider adapter without reusing the runtime's resolved CLI path

### After
- `src/ouroboros/codex/cli_policy.py` owns the shared Codex launch policy:
  - wrapper-binary detection
  - fallback to the terminal `codex` binary
  - shared child-process environment sanitation
- both runtime and provider paths use the same policy
- `OrchestratorRunner._build_dependency_analyzer()` reuses the runtime's resolved `cli_path` when constructing the nested provider adapter

This keeps the fix architectural rather than tactical:

> one Codex launch policy, reused everywhere

### 1. Shared wrapper-safe resolution
`resolve_codex_cli_path(...)` now centralizes the existing runtime behavior:

- inspect the configured/explicit candidate
- treat compiled binaries as wrapper candidates
- fall back to the first non-wrapper `codex` on `PATH`
- preserve the same structured logging (`cli_wrapper_detected`, `cli_resolved_via_fallback`, `cli_no_fallback`)

### 2. Shared child env policy
`build_codex_child_env(...)` now centralizes the recursion-prevention behavior:

- strip Ouroboros MCP env markers
- strip inherited Codex session markers
- strip `CLAUDECODE`
- increment `_OUROBOROS_DEPTH`
- preserve caller-specific exception behavior through an injected error factory

### 3. Nested analyzer parity
`OrchestratorRunner` now passes the runtime's resolved `cli_path` into `create_llm_adapter(...)` for dependency analysis. That prevents nested provider calls from drifting back to a different configured binary after the runtime has already resolved the correct terminal CLI.

## Why this matters

This change makes Codex invocation behavior deterministic across:

- top-level runtime execution
- provider-backed nested calls
- dependency-analysis paths that create fresh Codex adapters mid-run

It removes a class of bugs where the system could look healthy at the top level but still diverge in deeper execution phases.

### Focused parity and regression tests
- `PYTHONPATH=src ./.venv/bin/pytest -q tests/unit/orchestrator/test_codex_cli_runtime.py tests/unit/orchestrator/test_codex_recursion_guard.py tests/unit/orchestrator/test_runner.py tests/unit/providers/test_codex_cli_adapter.py`
- `PYTHONPATH=src ./.venv/bin/pytest -q tests/unit/codex/test_cli_policy.py tests/unit/providers/test_codex_cli_adapter.py::TestCodexCliLLMAdapter::test_resolve_cli_path_falls_back_from_wrapper tests/unit/orchestrator/test_codex_cli_runtime.py::TestCodexCliRuntime::test_resolve_cli_path_falls_back_from_wrapper tests/unit/orchestrator/test_runner.py::TestOrchestratorRunner::test_build_dependency_analyzer_reuses_resolved_codex_cli_path`
- `PYTHONPATH=src ./.venv/bin/pytest -q tests/unit/orchestrator/test_runtime_factory.py`

### Static checks
- `PYTHONPATH=src ./.venv/bin/ruff check src/ouroboros/codex/cli_policy.py src/ouroboros/orchestrator/codex_cli_runtime.py src/ouroboros/providers/codex_cli_adapter.py src/ouroboros/orchestrator/runner.py tests/unit/codex/test_cli_policy.py tests/unit/orchestrator/test_codex_cli_runtime.py tests/unit/orchestrator/test_codex_recursion_guard.py tests/unit/orchestrator/test_runner.py tests/unit/providers/test_codex_cli_adapter.py`
- `PYTHONPATH=src ./.venv/bin/mypy src/ouroboros/codex/cli_policy.py src/ouroboros/orchestrator/codex_cli_runtime.py src/ouroboros/providers/codex_cli_adapter.py src/ouroboros/orchestrator/runner.py`

### Smoke validation
- runtime/provider smoke parity check with a compiled wrapper candidate + real terminal `codex` script confirming both paths resolve to the same final binary

## Notes

I reviewed the current open PR list before opening this. The existing open PRs are focused on ambiguity milestones, inherited tool filtering, startup performance, and other unrelated areas; I did not find another open PR addressing Codex runtime/provider launch parity.

### DIFF
--- a/src/ouroboros/codex/cli_policy.py
+++ b/src/ouroboros/codex/cli_policy.py
@@ -1,0 +1,160 @@
+"""Shared Codex CLI launch policy helpers for runtime and provider callers."""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Mapping, Sequence
+from dataclasses import dataclass
+import os
+from pathlib import Path
+
+DEFAULT_CODEX_CLI_NAME = "codex"
+DEFAULT_MAX_OUROBOROS_DEPTH = 5
+DEFAULT_CODEX_CHILD_ENV_KEYS = ("OUROBOROS_AGENT_RUNTIME", "OUROBOROS_LLM_BACKEND")
+DEFAULT_CODEX_CHILD_SESSION_ENV_KEYS = ("CODEX_THREAD_ID",)
+_WRAPPER_MAGIC_HEADERS = (
+    b"\xcf\xfa\xed\xfe",  # Mach-O 64-bit
+    b"\xce\xfa\xed\xfe",  # Mach-O 32-bit
+    b"\x7fELF",  # ELF
+)
+
+
+@dataclass(frozen=True, slots=True)
+class CodexCliResolution:
+    """Resolved Codex CLI selection metadata."""
+
+    cli_path: str
+    candidate_path: str
+    wrapper_path: str | None = None
+    fallback_path: str | None = None
+
+
+def resolve_codex_cli_path(
+    *,
+    explicit_cli_path: str | Path | None,
+    configured_cli_path: str | None,
+    default_cli_name: str = DEFAULT_CODEX_CLI_NAME,
+    logger: object,
+    log_namespace: str,
+) -> CodexCliResolution:
+    """Resolve the safest Codex CLI path for nested automation.
+
+    When the configured candidate is a compiled wrapper (for example a Zeude
+    shim), prefer the next real ``codex`` binary on ``PATH`` instead.
+    """
+    if explicit_cli_path is not None:
+        candidate = str(Path(explicit_cli_path).expanduser())
+    else:
+        candidate = configured_cli_path or _which(default_cli_name) or default_cli_name
+
+    path = Path(candidate).expanduser()
+    if not path.exists():
+        return CodexCliResolution(cli_path=candidate, candidate_path=candidate)
+
+    resolved = str(path)
+    if not is_wrapper_binary(resolved):
+        return CodexCliResolution(cli_path=resolved, candidate_path=resolved)
+
+    logger.warning(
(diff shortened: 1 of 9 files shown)


Q00 (approved):
LGTM

Issue #358, opened 2026-04-11
feat(interview): batch up to 10 questions at once to reduce round-trips


Q00:
## Context

From team meeting (2026-04-10): Currently the interview asks 3 questions at a time, but this could be increased significantly.

## Problem

With only 3 questions per batch, the interview requires many round-trips with the user. Each round-trip costs latency (MCP reasoning time + user response time).

## Proposal

- Increase batch size from 3 to up to 10 questions per round
- When there are many ambiguity points to clarify, batch them together
- Users indicated they're willing to wait and answer more questions at once: "한꺼번에 열 개 한꺼번에 다 하면 너무 베스트일 것 같은데"
- Consider adaptive batching: fewer questions when nearing completion, more when starting

## Expected Impact

- Fewer round-trips → lower total interview time
- Users prefer answering many questions at once over many small batches
- Reduces cumulative MCP reasoning overhead per question

## Source

Meeting discussion: "질문을 한꺼번에 여러 개 물어보면 어때요? ... 지금도 세 개씩 물어보긴 하거든요 ... 한꺼번에 열 개 다 하면 너무 베스트일 것 같은데"

Q00:
Consolidated into new parent issue.

Issue #357, opened 2026-04-11
feat(interview): reduce human interaction by auto-confirming AI-determinable facts


Q00:
## Context

From team meeting (2026-04-10): Users struggle with the interview process because too many questions are asked that AI could determine on its own.

## Problem

Currently the interview agent asks users to confirm everything, even things that AI can reasonably infer or determine independently. This creates friction and makes the tool feel "weak" to users.

## Proposal

- Before calling MCP, have the agent attempt to auto-confirm facts it can determine with high confidence
- Only escalate genuinely ambiguous points to the user
- If ambiguity is low enough, skip confirmation and proceed
- Consider a confidence threshold: if AI confidence > threshold, auto-resolve without asking

## Expected Impact

- Reduced interview latency (fewer round-trips with the user)
- Better UX — users feel the AI is smart, not just asking them to do the work
- Maintains accuracy by only auto-confirming high-confidence facts

## Source

Meeting discussion: "AI가 충분히 다 확정지울 수 있는 것도 많거든요. 그래서 보호성을 줄이는 건 좋은데..."

Q00:
Consolidated into new parent issue.

PR #382, opened 2026-04-14
feat(interview): add auto-confirm routing for high-confidence code answers (#357)

shaun0927:
## Summary

- Split PATH 1 (Code Confirmation) into **PATH 1a** (auto-confirm) and **PATH 1b** (confirmation)
- PATH 1a: high-confidence facts from manifests/configs are sent to MCP immediately with `[from-code][auto-confirmed]` prefix, user sees a non-blocking notification
- PATH 1b: medium/low confidence answers preserve the existing confirmation-question flow
- Updated Dialectic Rhythm Guard to count auto-confirms toward the 3-consecutive limit
- Updated example session to demonstrate auto-confirm vs. confirmation routing

## Why this is needed

From team meeting (2026-04-10):
> "AI가 충분히 다 확정지울 수 있는 것도 많거든요. 그래서 보호성을 줄이는 건 좋은데..."

Currently PATH 1 always blocks on user confirmation, even for unambiguous facts like "What language is this project?" when `pyproject.toml` clearly states Python 3.12. This creates unnecessary friction and makes the tool feel "weak".

**The key insight**: auto-confirm also naturally reduces user-facing round-trips (the goal of #358 batch questions), because factual questions are resolved instantly. The user only sees questions that genuinely require their judgment.

## Design decisions

- **SKILL.md only, no MCP code changes**: The MCP server remains a pure question generator (Ouroboros P3 principle). Auto-confirm is a routing decision made by the main session, which has codebase access — the MCP does not.
- **Strict criteria for PATH 1a**: Only exact manifest/config matches qualify. Inferred answers go through PATH 1b with user confirmation.
- **Dialectic Rhythm Guard preserved**: After 3 consecutive non-user answers (1a + 1b + PATH 4), the next question is forced to PATH 2 (human judgment). This prevents the interview from becoming "AI talking to itself."
- **User override**: Users can always say "that's wrong" to correct an auto-confirmed answer.

## How this addresses batch questions (#358) without breaking Socratic integrity

Rather than generating multiple questions simultaneously (which breaks the Socratic chain where each question builds on the previous answer), auto-confirm achieves the same UX goal:

```
Before: 10 user-facing questions × (read + judge + respond) = slow
After:  4 auto-confirmed (instant) + 6 user questions = ~40% fewer user interactions
```

The MCP still generates questions one at a time, each informed by previous answers. The user just encounters fewer blocking interactions.

## Test plan

- [x] SKILL.md is a documentation file — no unit test changes needed
- [x] Verified `test_job_manager` failure is pre-existing on main (unrelated)
- [x] All bigbang + mcp/tools + plugin/skills tests pass

### DIFF
--- a/skills/interview/SKILL.md
+++ b/skills/interview/SKILL.md
@@ -105,12 +105,41 @@ MCP (question generator) ←→ You (answerer + router) ←→ User (human judgm
    ```
    Returns a session ID and the first question.
 
-2. **For each question from MCP, apply 3-Path Routing:**
+2. **For each question from MCP, apply the routing paths below:**
 
-   **PATH 1 — Code Confirmation** (describe current state, user confirms):
+   **PATH 1 — Code Answer** (describe current state from codebase):
    When the question asks about existing tech stack, frameworks, dependencies,
    current patterns, architecture, or file structure:
    - Use Read/Glob/Grep to find the factual answer
+   - **Description, not prescription**: "The project uses JWT" is fact.
+     "The new feature should also use JWT" is a DECISION — route to PATH 2.
+   - Evaluate confidence and choose sub-path:
+
+   **PATH 1a — Auto-confirm** (high-confidence factual, no user block):
+   When ALL of the following are true:
+   - The answer is found as an **exact match** in a manifest or config file
+     (e.g., `pyproject.toml`, `package.json`, `Dockerfile`, `go.mod`, `.env.example`)
+   - The answer is **purely descriptive** — it describes what exists, not what
+     the new feature should do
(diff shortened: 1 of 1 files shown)


Q00 (approved):
@shaun0927 

This direction looks good to me and matches the meeting intent: reduce unnecessary user-facing confirmations for high-confidence factual code answers while preserving the   Socratic rhythm with the guardrail.

Follow-up needed:
  - align the broader interview protocol docs/prompts with the new `[from-code][auto-confirmed]` path, since the current interviewer docs and system prompt still only
  describe `[from-code]`, `[from-user]`, and `[from-research]`.

PR #384, opened 2026-04-14
feat(evaluate): add per-AC checklist aggregator (#366)

shaun0927:
## Summary

- Add a new module `ouroboros.evaluation.checklist` with:
  - `ACCheckItem` — per-AC state (text, passed, reasoning, evidence, questions_used, failure_reason)
  - `ACChecklistResult` — aggregated view with `all_passed`, `pass_rate`, `failed_items`
  - `aggregate_results(ac_texts, results)` — builds a checklist from parallel AC evaluation results
  - `format_checklist(checklist)` — plain-text rendering with `[x]` / `[ ]` markers and Next Steps
  - `build_run_feedback(checklist)` — extracts per-AC feedback for Run re-attempts
- 16 unit tests covering construction, aggregation, ordering, formatting, and feedback generation

## Why this is needed

From the team meeting (2026-04-10):

> \"까고 다시 만들고 까고 다시 만들고\" — evaluation should feed specific AC failures back to Run so the next attempt targets the gap.

Today's evaluation pipeline produces pass/fail verdicts for each AC individually, and orchestration already dispatches one evaluation per AC. **What is missing** is a single aggregated view the user (and Run stage) can act on:

- \"3 of 5 ACs pass, here are the 2 that failed and why\"
- \"Here is the exact feedback to send back to Run for each failure\"
- \"The checklist is incomplete until every AC passes — this is the quality gate the milestones promise\"

This PR adds exactly that view, as a small reusable module.

## Design decisions

- **Purely additive**: this module does NOT modify `SemanticResult`, `EvaluationContext`, `EvaluationResult`, or any existing orchestration path. Existing callers remain unchanged. The aggregator is opt-in — callers route results through it when they want the checklist view.
- **Empty checklist is not ready**: `all_passed` returns `False` for a zero-AC checklist. A Seed with no ACs is definitionally incomplete (see Phase 1 READY milestone), and the gate must refuse to pass it.
- **Phase 1 tie-in** (#363 milestones, PRs #379/#380): when Interview reaches READY (ambiguity ≤ 0.2), every AC should pass when routed through this checklist. That is the concrete quality gate the milestones promise.
- **Forward-compatible with #367** (PR #383): the aggregator reads `questions_used` and `evidence` off `SemanticResult` via `getattr` so it works whether #367 has landed or not. Two tests skip gracefully on code that predates #367 and activate automatically once it lands.
[...]

Issue #366, opened 2026-04-11
feat(evaluate): add Success Criteria evaluator — verify Run output against AC during evaluation


Q00:
## Summary

From #359. Add explicit Success Criteria (Acceptance Criteria) to the seed/brownfield spec, and have the evaluator agent verify Run output against them.

## Problem

- Currently no concrete success criteria in seeds — evaluator has no objective checklist
- Run stage produces output but there's no systematic way to say "this meets the user's actual goals"
- Ambiguity score (0.5→0.2) lacks connection to concrete deliverables

### 1. Add `success_criteria` field to Brownfield / Seed schema
- Can be flat list or nested (grouped by aspect)
- Filled during interview hardening process
- Tied to ambiguity milestones: by score 0.2, all criteria should be concrete and testable

### 2. Evaluator checks AC during Run
- After each Run cycle, evaluator verifies output against each success criterion
- Like filling checkboxes — pass/fail for each criterion
- If criteria not met: send back to Run with specific feedback ("까고 다시 만들고 까고 다시 만들고")
- If all criteria met: guide user to next step (evolve, QA, etc.)

### 3. Runtime verification where possible
- For code: run tests, type checks
- For services: verify it actually runs
- For concepts: validate with examples

## Expected Impact

- Objective quality gate — no more "looks correct but doesn't meet goals"
- Run stage becomes self-correcting against concrete criteria
- Stronger interview → seed pipeline

PR #385, opened 2026-04-14
feat(evaluate): wire checklist into EvaluateHandler for multi-AC (#366)

shaun0927:
## Summary

- Add a new `acceptance_criteria` (ARRAY) parameter to the `ouroboros_evaluate` MCP tool
- When 2+ ACs are provided, fan out one evaluation per AC via `asyncio.gather`, then aggregate through the checklist module (#366 part 1)
- Return a rendered checklist as text and a structured `checklist` + `run_feedback` in `meta`
- Preserve the existing single-AC path unchanged — 1-element lists and callers that don't pass `acceptance_criteria` keep the original behaviour

## Why this is needed

The checklist aggregator from PR #384 exists but has no callable surface — users can't exercise the per-AC view without writing Python directly. This PR plumbs the aggregator into the MCP boundary so:

1. **Users see the whole picture in one call**: \"3 of 5 ACs pass, here are the 2 that failed and exactly why.\"
2. **Run stage gets targeted feedback**: `meta.run_feedback` provides pre-built feedback strings for each failed AC, ready to be injected into a re-attempt prompt.
3. **Ouroboros Physics principle**: the MCP server (Python) now **enforces** the checklist semantics deterministically. An LLM cannot decide to skip the per-AC fan-out — the MCP does it itself.

## Depends on

- #384 (`feat/evaluate-per-ac-checklist`) — provides the checklist module this PR wires in. This PR includes PR #384's commit via cherry-pick so it can be reviewed and merged independently, but ideally #384 lands first.

## Design decisions

- **Single-AC path untouched**: backward compatibility is the top priority. Existing tests and callers pass through without any change.
- **Fan-out via `asyncio.gather`**: AC evaluations are independent, so we evaluate them in parallel. This keeps the total latency close to single-AC.
- **Fail-fast on pipeline error**: if any AC's pipeline evaluation returns `Err`, the whole call returns `Err`. A partial checklist would mislead the Run stage about what actually failed.
- **Blank-entry filtering**: whitespace-only or empty strings in `acceptance_criteria` are dropped during normalization. Callers can safely pass the list from their own data sources without defensive cleanup.
- **Structured meta + text render**: we give the caller both a human-readable checklist (text) and machine-readable data (`meta.checklist`, `meta.run_feedback`) so both humans and tooling can consume it.

## Response shape for multi-AC calls

```json
{
  \"text\": \"Acceptance Criteria Checklist [INCOMPLETE] (1/2 passed, 50%)\\n...\",
  \"meta\": {
    \"multi_ac\": true,
    \"ac_count\": 2,
    \"passed_count\": 1,
    \"pass_rate\": 0.5,
    \"final_approved\": false,
    \"checklist\": [
      {\"ac_text\": \"Charge processed\", \"passed\": true, \"reasoning\": \"...\", ...},
      {\"ac_text\": \"Webhook validated\", \"passed\": false, \"failure_reason\": \"...\", ...}
    ],
    \"run_feedback\": [\"AC not met: Webhook validated — ...\"]
  }
}
```

## Test plan

- [x] 9 new tests in `tests/unit/mcp/tools/test_evaluate_multi_ac.py`:
  - Parameter schema exposes `acceptance_criteria` as ARRAY
  - 1-element list → single-AC path (no `multi_ac` meta flag)
  - 2 passing ACs → `ALL PASSED`, `final_approved=True`, `run_feedback=[]`
  - Mixed outcomes → `INCOMPLETE`, per-AC `failure_reason`, `run_feedback` populated
  - Whitespace/empty entries filtered out
  - Pipeline error propagates as handler error (no partial checklist)
  - Invalid parameter values (None, non-list, int) fall back to single-AC
- [x] 663 passed, 2 skipped — no regressions across `tests/unit/mcp/tools/` and `tests/unit/evaluation/`

### DIFF
--- a/src/ouroboros/mcp/tools/evaluation_handlers.py
+++ b/src/ouroboros/mcp/tools/evaluation_handlers.py
@@ -279,6 +279,17 @@ class EvaluateHandler:
                     required=False,
                 ),
                 MCPToolParameter(
+                    name="acceptance_criteria",
+                    type=ToolInputType.ARRAY,
+                    description=(
+                        "Multiple acceptance criteria for checklist evaluation. "
+                        "When two or more items are provided, each AC is evaluated "
+                        "independently and the results are aggregated into a "
+                        "pass/fail checklist (#366). Overrides acceptance_criterion."
+                    ),
+                    required=False,
+                ),
+                MCPToolParameter(
                     name="artifact_type",
                     type=ToolInputType.STRING,
                     description="Type of artifact: code, docs, config. Default: code",
@@ -349,13 +360,25 @@ class EvaluateHandler:
 
         seed_content = arguments.get("seed_content")
         acceptance_criterion = arguments.get("acceptance_criterion")
+        acceptance_criteria_raw = arguments.get("acceptance_criteria")
         artifact_type = arguments.get("artifact_type", "code")
         trigger_consensus = arguments.get("trigger_consensus", False)
+
+        # Normalize the optional multi-AC list (#366): filter out empty/blank
+        # entries so callers can safely pass `[""]` or an accidental null.
+        acceptance_criteria: tuple[str, ...] = ()
+        if isinstance(acceptance_criteria_raw, list):
+            acceptance_criteria = tuple(
+                str(item).strip()
+                for item in acceptance_criteria_raw
+                if isinstance(item, (str, int, float)) and str(item).strip()
+            )
 
         log.info(
             "mcp.tool.evaluate",
             session_id=session_id,
             has_seed=seed_content is not None,
+            multi_ac_count=len(acceptance_criteria),
             trigger_consensus=trigger_consensus,
         )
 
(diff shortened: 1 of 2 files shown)


src/ouroboros/mcp/tools/evaluation_handlers.py
```diff
@@ -440,6 +463,33 @@ async def handle(
...
+                semantic=SemanticConfig(model=get_semantic_model(self.llm_backend)),
+            )
+            pipeline = EvaluationPipeline(llm_adapter, config)
+
+            # Multi-AC checklist path (#366):
+            # When the caller provides >= 2 acceptance criteria we run the
+            # pipeline once per AC and aggregate the results into a
+            # checklist.  Single-AC callers keep the original single-pass
+            # behaviour — no extra cost or behaviour change for them.
+            if len(acceptance_criteria) >= 2:
```
Q00, line 491: Requesting changes because the new public contract is not implemented correctly yet.

  I reproduced a runtime regression locally on commit `f833415`:

  If `ouroboros_evaluate` is called with a single-item `acceptance_criteria` list, the handler does **not** evaluate that AC. Instead, it falls back to the generic default:

  ```python
  current_ac == "Verify execution output meets requirements"
```
  So the API shape now suggests “single or many ACs”, but the implementation only handles the “many” case correctly. A 1-item list is silently ignored as an evaluated AC.

  This is not just a missing-test issue. It is a correctness bug in the request-to-evaluation mapping.

  There is also a design issue in the multi-AC path: the implementation parallelizes the full EvaluationPipeline.evaluate() per AC, which duplicates Stage 1 mechanical
  verification (lint/build/test) for every AC instead of only parallelizing AC-specific semantic evaluation. That may be acceptable temporarily, but it is the wrong
  execution layer for this feature.

  Please fix the single-item acceptance_criteria behavior first:

  - either treat a 1-item list as the effective current_ac,
  - or reject it explicitly and require callers to use acceptance_criterion.

  Right now the handler accepts the input shape but does not honor its meaning.

Q00 (changes requested):

shaun0927:
Rebased onto upstream/main. The per-AC checklist aggregator commit (`6f568a1`) was auto-detected by git as already merged via PR #384 (commit `4497356` on main) and skipped. Remaining unique commits:

- `c4fc314` fix(evaluate): honour 1-item acceptance_criteria list as effective AC (addresses Q00 review feedback)
- `f1bb116` feat(evaluate): wire checklist into EvaluateHandler for multi-AC

Local test results (847 passed in tests/unit/mcp + tests/unit/evaluation). CI re-running.

shaun0927:
**Labels**: `enhancement` `evaluate` — Refs #366

Q00 (approved):
LGTM. Latest head fixes the 1-item `acceptance_criteria` contract hole and the current patch is mergeable.

Follow-up:
If this lands before the Stage 1 dedup follow-up, keep an eye on the multi-AC path re-running mechanical checks per AC. That is not a blocker for correctness, but it is wasted work.

PR #394, opened 2026-04-15
fix(codex): delegate _which() to shutil.which for cross-platform support (#391)

### LINKED ISSUE #391: bug(codex): custom _which() in cli_policy.py lacks PATHEXT support — regression for Windows
## Summary

PR #381 replaced `shutil.which()` with a custom `_which()` implementation in the new shared `cli_policy.py` module. The custom implementation does not handle Windows `PATHEXT` environment variable, which means it cannot find executables like `codex.exe` or `codex.cmd` on Windows.

This is a **regression** specifically for the **provider path** (`CodexCliLLMAdapter`), which previously used `shutil.which()` directly.

Introduced in v0.28.6 by PR #381 (CLI resolution unification).

## The difference

**`shutil.which()`** (stdlib):
- Checks `PATHEXT` on Windows (`.exe`, `.cmd`, `.bat`, `.com`, etc.)
- Handles CWD-on-PATH Windows behavior
- Resolves symlinks properly

**Custom `_which()`** (`cli_policy.py`):
```python
def _which(name: str) -> str | None:
    for directory in os.environ.get("PATH", "").split(os.pathsep):
        candidate = os.path.join(directory, name)
        if os.path.isfile(candidate) and os.access(candidate, os.X_OK):
            return candidate
    return None
```
- No PATHEXT handling
- Exact name match only (`codex`, not `codex.exe`)

## Impact

| Platform | Before (v0.28.5) | After (v0.28.6) |
|----------|-------------------|------------------|
| macOS/Linux | `shutil.which("codex")` → works | `_which("codex")` → works (same behavior) |
| Windows | `shutil.which("codex")` → finds `codex.exe` via PATHEXT | `_which("codex")` → **returns None** |

**Note**: Issue #387 reports MCP server failure on Windows (v0.28.5). While that issue predates this change, the `_which()` regression in v0.28.6 could **worsen** Windows support.

## Suggested fix

Delegate to `shutil.which()`:
```python
import shutil

def _which(name: str) -> str | None:
    return shutil.which(name)
```

Or, if custom PATH walking is needed for wrapper detection, use `shutil.which()` as the primary resolver and only custom-walk when needed.

## Affected code

- `src/ouroboros/codex/cli_policy.py` — `_which()`
- Consumers: `resolve_codex_cli_path()` → `CodexCliRuntime`, `CodexCliLLMAdapter`

## Environment

- Version: v0.28.6
- Introduced by: PR #381

shaun0927:
## Summary

- Replace custom `_which()` implementation in `cli_policy.py` with delegation to `shutil.which()`
- Restores Windows PATHEXT support that was lost when PR #381 introduced the custom helper

## Problem

PR #381 (v0.28.6) extracted a shared `cli_policy.py` module with a custom `_which()`:

```python
def _which(name: str) -> str | None:
    for directory in os.environ.get("PATH", "").split(os.pathsep):
        candidate = os.path.join(directory, name)
        if os.path.isfile(candidate) and os.access(candidate, os.X_OK):
            return candidate
    return None
```

This does **not** handle Windows `PATHEXT` (`.exe`, `.cmd`, `.bat`), so `_which("codex")` cannot find `codex.exe` on Windows. The **provider path** (`CodexCliLLMAdapter`) previously used `shutil.which()` directly, so this is a regression for Windows users.

## Fix

```python
def _which(name: str) -> str | None:
    import shutil
    return shutil.which(name)
```

`shutil.which()` handles PATHEXT, CWD-on-PATH (Windows), and is the stdlib standard for executable resolution.

## Test plan

- [x] `TestResolveCodexCliPath::test_falls_back_from_wrapper_to_real_cli` — passes
- [x] `TestResolveCodexCliPath::test_keeps_wrapper_when_no_real_cli_exists` — passes
- [x] `TestBuildCodexChildEnv` tests — pass
- [x] All 4 cli_policy tests pass

Closes #391

### DIFF
--- a/src/ouroboros/codex/cli_policy.py
+++ b/src/ouroboros/codex/cli_policy.py
@@ -139,12 +139,14 @@ def build_codex_child_env(
 
 
 def _which(name: str) -> str | None:
-    """Small local ``which`` helper to keep path resolution pure/testable."""
-    for directory in os.environ.get("PATH", "").split(os.pathsep):
-        candidate = os.path.join(directory, name)
-        if os.path.isfile(candidate) and os.access(candidate, os.X_OK):
-            return candidate
-    return None
+    """Locate an executable on ``PATH``, delegating to :func:`shutil.which`.
+
+    Using the stdlib implementation ensures correct behavior on all
+    platforms, including Windows ``PATHEXT`` resolution.
+    """
+    import shutil
+
+    return shutil.which(name)
 
 
 __all__ = [


shaun0927:
**Labels**: `bug` `codex` `platform:windows` — Fixes #391

Q00 (approved):
LGTM. Latest head is green and matches the current bot-reviewed SHA. I didn't find a blocker in the current diff.

PR #413, opened 2026-04-15
fix(security): sanitize CheckpointStore seed_id to prevent path traversal

### LINKED ISSUE #397: security(persistence): CheckpointStore seed_id path traversal allows arbitrary file write/delete
## Summary

`CheckpointStore._get_checkpoint_path()` interpolates `seed_id` directly into the filename without any sanitization. Since `seed_id` is user-provided (via Seed spec), a crafted value containing `../` can escape the checkpoint base directory to write, overwrite, or delete arbitrary files on the filesystem.

This code path was **dead code until v0.28.3**, when commit `6b2cdda` (#338) activated `CheckpointStore` in `ExecuteSeedHandler`.

## Root Cause

```python
# persistence/checkpoint.py:327-340
def _get_checkpoint_path(self, seed_id: str, level: int = 0) -> Path:
    filename = f"checkpoint_{seed_id}.json"   # ← no sanitization
    return self._base_path / filename
```

## PoC

```python
store = CheckpointStore(base_path=Path("/app/data/checkpoints"))
store.initialize()

seed_id = "x/../../PWNED"
checkpoint = CheckpointData.create(seed_id, "exploit", {"pwned": True})
store.save(checkpoint)  # writes to /app/data/PWNED.json (outside base dir)
```

## Impact

- **File write**: `save()` creates/overwrites files at arbitrary paths
- **File delete**: `_rotate_checkpoints()` calls `unlink()` on traversed paths
- **File read**: `load()` reads JSON files from traversed paths, leaking content via error messages

In remote MCP deployments, this enables Remote Code Execution (RCE) via writing to cron/systemd paths.

## Suggested Fix

1. Reject `seed_id` containing `/`, `\`, `..`, null bytes
2. After resolving the path, verify it starts with `base_path.resolve()`
3. Add max length limit for `seed_id`

## Introduced In

v0.28.3 — commit `6b2cdda` (#338)

shaun0927:
## Summary

- **Fixes #397** — `CheckpointStore._get_checkpoint_path()` directly interpolated `seed_id` into filenames without sanitization, allowing crafted IDs like `x/../../PWNED` to read, write, or delete files outside the checkpoint directory.
- Adds `_sanitize_seed_id()` static method that strips null bytes, removes `..` traversal sequences, replaces `/` and `\` with underscores, limits length to 255 chars, and raises `ValueError` for empty results.
- Adds `_validate_path_containment()` that verifies the resolved checkpoint path stays within `self._base_path`.
- Both guards are applied in `_get_checkpoint_path()`, which is used by `save()`, `load()`, and all rollback operations.

## Test plan

- [x] Path traversal with forward slashes is sanitized (file stays in base dir)
- [x] Path traversal with backslashes is sanitized
- [x] No file is created outside the checkpoint directory
- [x] Normal seed_ids continue to work (save + load roundtrip)
- [x] Empty seed_id raises `ValueError`
- [x] Null-byte-only seed_id raises `ValueError`
- [x] `..`-only seed_id raises `ValueError`
- [x] Null bytes are stripped but remaining content preserved
- [x] Seed IDs longer than 255 chars are truncated
- [x] `load()` with traversal seed_id resolves consistently with `save()`
- [x] All 34 existing + new tests pass
- [x] `ruff check` passes clean

### DIFF
--- a/src/ouroboros/persistence/checkpoint.py
+++ b/src/ouroboros/persistence/checkpoint.py
@@ -17,6 +17,7 @@ import hashlib
 import json
 import os
 from pathlib import Path
+import re
 from typing import Any
 
 from ouroboros.core.errors import PersistenceError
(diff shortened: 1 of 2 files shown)


shaun0927:
**Labels**: `security` `persistence` — Fixes #397

Q00 (changes requested):
Requesting changes because the filename-length hardening is still incomplete.

`_sanitize_seed_id()` truncates the raw seed id to 255 characters, but `_get_checkpoint_path()` then wraps it in `checkpoint_<seed>.json` / `.json.<level>`. That still produces basenames longer than 255 bytes on common filesystems (`len("checkpoint_") + 255 + len(".json") = 271`), so long seed ids can still fail with `ENAMETOOLONG` on save/load.

Please cap the sanitized id to the remaining filename budget after the prefix/suffix, and add a regression test that actually writes a long-id checkpoint instead of only asserting on the generated path string.

Q00 (approved):
LGTM

PR #420, opened 2026-04-15
fix(orchestrator): add timeout to rate limit wait loop to prevent starvation (#398)

### LINKED ISSUE #398: bug(orchestrator): _wait_for_shared_rate_limit_budget has no timeout — worker starvation under contention
## Summary

`ClaudeAgentAdapter._wait_for_shared_rate_limit_budget()` (v0.28.5, `adapter.py:865-895`) contains a `while True` loop with **no timeout, no max retries, and no fairness mechanism**. Under contention, workers can starve indefinitely.

## Root Cause

`SharedRateLimitBucket.acquire()` returns a positive wait time when budget is exhausted but **does not make a reservation**. The caller sleeps and retries, but faster workers can repeatedly claim slots first.

```python
# adapter.py:875-895
while True:  # No timeout, no max_retries
    wait_seconds, snapshot = await self._rate_limit_bucket.acquire(estimated_tokens)
    if wait_seconds <= 0:
        return  # Only exit
    await asyncio.sleep(sleep_seconds)  # Retry forever
```

## Confirmed Scenarios

1. **Starvation**: 100 consecutive `acquire()` calls with a full bucket — zero reservations made
2. **No timeout**: Source confirmed: no `asyncio.timeout`, no `max_iter`, no `max_wait`
3. **Thundering herd**: 5 workers × 20 attempts = 100 total calls, zero successes, zero progress
4. **Token starvation**: Same pattern with `token_limit`

## Impact

- A single worker can block indefinitely, hanging the entire orchestrator execution
- User sees infinite "Shared Anthropic budget saturated; waiting..." messages
- No automatic recovery — only killing the process resolves it

## Suggested Fix

- Add `asyncio.timeout(max_wait_seconds)` around the while loop
- Implement fair queuing (FIFO reservation queue) in `acquire()`
- Add `max_retries` counter with exponential backoff

## Test Evidence

10/10 tests passing: `tests/unit/orchestrator/test_rate_limit_starvation.py`

Introduced in: v0.28.5 (commit `44c0383`)

shaun0927:
## Summary
- `_wait_for_shared_rate_limit_budget` previously had an unbounded `while True` loop that could starve workers indefinitely under rate limit contention
- Added `total_waited` tracking and `max_wait_seconds` parameter (default 120s) so the loop exits with a warning after the timeout, allowing the worker to proceed
- Added `RATE_LIMIT_MAX_WAIT_SECONDS` constant to `rate_limit.py` and exported it in `__all__`
- Heartbeat messages now include `total_waited` and `max_wait_seconds` in their data payload for observability

## Test plan
- [x] All 64 existing tests in `test_adapter.py` and `test_rate_limit.py` pass
- [ ] Verify under real contention that workers proceed after 120s timeout instead of hanging
- [ ] Confirm the `orchestrator.adapter.rate_limit_timeout` warning log fires when timeout is reached

Fixes #398


shaun0927:
Addressed critical review finding: the timeout escape hatch was bypassing the rate limit reservation, which would cause an N× RPM burst under concurrent saturation. Added `force_reserve()` to preserve budget accounting during timeout fallback.

shaun0927:
**Labels**: `bug` `orchestrator` — Fixes #398

Q00 (approved):
LGTM. Latest head is green and matches the current bot-reviewed SHA. I didn't find a blocker in the current diff.

Issue #387, opened 2026-04-15
[Bug] Ouroboros MCP server fails in Claude Code CLI


lmiguelgato:
### Summary

After completing the interview, it says "Ouroboros is now executing the seed in the background. You can check progress with ooo status or ooo poll"

I checked for the status and it said "The execution is still running and hasn't produced its first acceptance criteria tree update yet. This is expected for a complex seed — it's likely still in the planning/exploration phase."

Then I check again and "The Ouroboros MCP server has disconnected — the ouroboros_channel_workflow tool and all related tools are no longer available. I can't check the execution status right now"

I tried reconnecting the MCP server, and all the progress is lost, I try `ooo seed` and it starts the interview all over again, so Claude Code answers the questions based on the initial interview, and again the process dies ("The Ouroboros MCP server disconnected again — I can't reach it to poll the job status.") and there's never a seed and never measurable progress in the implementation of the feature.

### Impact

I cannot use this tool beyond the interview part.

### Environment

- Ouroboros version 0.28.5
- OS: Windows 11
- Python 3.11.6
- Installation method: curl -fsSL https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.sh | bash
- Model: Opus 4.6

### Steps to reproduce

1. Install with curl -fsSL https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.sh | bash
2. Open claude code and run `ooo interview`, complete the interview.
3. After it starts running the job, check for status `ooo status` several times to see if it makes progress.

### Expected behavior

It should succeed

### Actual behavior

MCP disconnects every time

### Logs or error output

```text

```

### Checklist

- [x] I searched existing issues before opening this report.
- [x] I can reproduce this on the latest main branch or latest release.
- [x] I included clear reproduction steps.
- [x] I included enough detail for someone else to verify the fix.

PR #432, opened 2026-04-15
fix(install): enforce Python >=3.12 for uv and pipx installers

### LINKED ISSUE #429: install: enforce Python >=3.12 when using uv/pipx installers
## Context

Related to #387 — user reported MCP failures on **Windows 11 with Python 3.11.6**, while `pyproject.toml` requires `>=3.12`.

## Problem

`scripts/install.sh` has a Python version check (`MIN_PYTHON="3.12"`), but the check is **only applied when falling back to `pip`**. When `uv` or `pipx` is available, the installer skips the check and invokes them directly:

```bash
# install.sh (simplified)
if has_uv; then
  uv tool install "$INSTALL_SPEC"      # no --python flag
elif has_pipx; then
  pipx install "$INSTALL_SPEC"          # no --python flag
else
  # only here is MIN_PYTHON checked
```

On Windows where users frequently have Python 3.11 in PATH alongside `pipx`, this results in ouroboros being installed into a 3.11 interpreter it doesn't support — leading to silent MCP failures at runtime.

## Suggested fix

- Pass `--python ">=3.12"` to `uv tool install` (uv supports version specifiers and will download a matching Python if needed)
- Resolve a 3.12+ interpreter for `pipx install --python <path>` or error out with a clear message
- Add the same guard to any PowerShell installer (`scripts/install.ps1`) if it exists

## Expected outcome

Users on systems with only Python 3.11 installed should see a clear error directing them to upgrade, not a broken ouroboros install.

shaun0927:
## Summary

Closes #429 · Refs #387

The Python version check in `scripts/install.sh` was only applied on the pip-fallback path. When `uv` or `pipx` was present, installations could proceed into a Python 3.11 interpreter and cause silent MCP failures at runtime (as reported in #387 on Windows 11 + Python 3.11.6).

## Changes

- `uv tool install` now always passes `--python ">=3.12"` so uv auto-downloads a matching Python if needed
- `pipx install` probes `python3.14`, `python3.13`, `python3.12` first, then falls back to `python3`/`python` with a version check, and passes the resolved interpreter via `--python <path>`
- When no 3.12+ interpreter is found (for the pipx path), the installer now errors out with a clear remediation message
- The existing pip-fallback path's check is untouched

## Test plan

- [x] `bash -n scripts/install.sh` — syntax check passes
- [ ] Manual: fresh install on macOS with only Python 3.11 → clear error
- [ ] Manual: fresh install on macOS with Python 3.12+ → succeeds via uv
- [ ] Manual: Windows Git Bash with Python 3.11 + pipx → clear error

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -39,27 +39,50 @@ elif command -v pipx &>/dev/null; then
   echo "  pipx:   $(pipx --version)"
 fi
 
-# Python check: only required when falling back to pip (no uv, no pipx)
-if [ "$HAS_UV" = false ] && [ "$HAS_PIPX" = false ]; then
-  for cmd in python3 python; do
-    if command -v "$cmd" &>/dev/null; then
-      ver=$("$cmd" -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')" 2>/dev/null || true)
-      if [ -n "$ver" ] && [ "$(printf '%s\n' "$MIN_PYTHON" "$ver" | sort -V | head -n1)" = "$MIN_PYTHON" ]; then
+# Helper: check whether a Python executable meets MIN_PYTHON
+_python_ok() {
+  local cmd="$1"
+  local ver
+  ver=$("$cmd" -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')" 2>/dev/null || true)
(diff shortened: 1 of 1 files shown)


shaun0927:
**Labels**: `enhancement` `install` — Refs #429

Q00 (approved):
LGTM. The installer now enforces Python >=3.12 across the uv/pipx/pip branches instead of only the pure-pip fallback.

Follow-up:
This path is shell-only, so I would still run real smoke installs in uv and pipx environments before release.

PR #433, opened 2026-04-15
feat(cli): add 'ooo resume' to recover in-flight sessions after MCP disconnect

### LINKED ISSUE #430: cli: recover in-flight session after MCP disconnect — 'ooo resume' command
## Context

Related to #387 — when the MCP server disconnects during a long-running seed execution, the user has no way to re-attach to the in-flight session. Running \`ooo seed\` or \`ooo interview\` starts a fresh interview from scratch, silently abandoning the running job.

## Current behaviour

1. User completes \`ooo interview\` → seed execution begins
2. MCP disconnects (various causes: network, stdio pipe, subprocess lifecycle)
3. User runs \`ooo seed\` → interview starts over
4. The previous execution's state (events, AC progress) is still in \`~/.ouroboros/ouroboros.db\` but inaccessible

## Proposed fix

Add a new CLI command that lists non-terminal sessions directly from the EventStore (no MCP dependency) and lets the user re-attach:

\`\`\`
$ ooo resume
Active sessions:
  1. exec_25ca66c14705   started 5m ago   3/7 ACs complete
  2. exec_c875b0452bf9   started 1h ago   stuck, needs cancellation
Select [1-2, c to cancel, q to quit]:
\`\`\`

Additional changes:
- Before \`ooo seed\` / \`ooo interview\` starts, warn if a non-terminal session exists and offer to resume
- \`ooo status\` should work even when the MCP server is unreachable, by reading directly from the EventStore

## Implementation outline

1. New \`src/ouroboros/cli/commands/resume.py\` — DB-only session listing and selection
2. Fallback path in \`status.py\` that bypasses MCP when the server is unreachable
3. Update \`skills/interview/SKILL.md\` and \`skills/seed/SKILL.md\` to detect in-flight sessions before proceeding

## Expected outcome

MCP disconnects become recoverable — user never loses interview answers or execution progress due to transport issues.

shaun0927:
## Summary

Closes #430 · Refs #387

When the MCP server disconnects during long-running seed execution, users previously had no way to re-attach to the in-flight session. `ooo seed` would start a fresh interview, discarding interview answers and execution progress.

This adds a standalone CLI command that reads the SQLite EventStore directly (no MCP dependency), lists non-terminal sessions, and prompts the user to pick one to re-attach to via `ooo status <exec_id>`.

## Changes

- **New**: `src/ouroboros/cli/commands/resume.py` — reads `~/.ouroboros/ouroboros.db` directly, renders a Rich table of running/paused sessions, interactive picker that surfaces `exec_id` for `ooo status` re-attachment
- **New**: `skills/resume/SKILL.md` — skill documentation
- **New**: `tests/unit/cli/test_resume.py` — 14 tests (empty state, corrupted DB, selection flow, quit, out-of-range input)
- **Updated**: `src/ouroboros/cli/main.py` — registered the new `resume` subcommand
- **Updated**: `CLAUDE.md` — added `ooo resume` to the ooo commands table

## Test plan

- [x] `ruff format --check .` and `ruff check .` pass
- [x] `mypy src/ouroboros/cli/commands/resume.py` passes
- [x] `pytest tests/unit/cli/test_resume.py` — 14 tests pass
- [x] `pytest tests/unit/cli/` — 411 tests pass (no regression)
- [ ] Manual: run `ooo resume` with 0, 1, and multiple in-flight sessions
- [ ] Manual: run `ooo resume` with a corrupted DB path → graceful error

## Design notes

- **No MCP dependency by design**: the whole point is that this works when the MCP server is unreachable.
- **Read-only**: `resume` never modifies the EventStore or creates events.
- **Doesn't auto-reconnect**: just surfaces the `exec_id` and suggests the next command. Keeps this PR focused and reviewable; auto-reconnect flow can be a follow-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,6 +32,7 @@ When the user types any of these commands, read the corresponding SKILL.md file 
 | `ooo pm` or `ooo pm ...` | Read `skills/pm/SKILL.md` and follow it |
 | `ooo brownfield` or `ooo brownfield ...` | Read `skills/brownfield/SKILL.md` and follow it |
 | `ooo publish` or `ooo publish ...` | Read `skills/publish/SKILL.md` and follow it |
+| `ooo resume` | Read `skills/resume/SKILL.md` and follow it |
 
 **Important**: Do NOT use the Skill tool. Read the file with the Read tool and execute its instructions directly.
 
--- a/skills/resume/SKILL.md
+++ b/skills/resume/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: resume
+description: "List in-flight sessions and re-attach after MCP disconnect"
+---
+
+# /ouroboros:resume
+
+Recover in-flight sessions after an unexpected MCP server disconnect.
+
+## Usage
+
+```
+ooo resume
+```
+
+**Trigger keywords:** "resume session", "re-attach", "mcp disconnected", "lost session", "in-flight"
+
+## How It Works
+
+`ooo resume` reads the EventStore directly (no MCP server required) and lists
+every session that is still in a `running` or `paused` state. You can then
+pick one from the interactive prompt to receive re-attach instructions.
+
+## Instructions
+
+When the user invokes this skill:
+
+1. Run the CLI command:
+
+   ```
+   ouroboros resume
+   ```
+
+   This reads `~/.ouroboros/ouroboros.db` directly — the MCP server does **not**
+   need to be running.
+
+2. If sessions are listed, enter the number corresponding to the session you
+   want to re-attach to. The command will print the `exec_id`.
+
+3. Re-attach using:
+
+   ```
+   ooo status <exec_id>
+   ```
+
+## Fallback (No sessions found)
+
+If the command reports "No in-flight sessions found", the execution either
+completed, failed, or was already cancelled. Check with:
+
+```
+ouroboros status executions
+```
+
+## Example
+
+```
+User: ooo resume
+
+┌─────────────────────── In-Flight Sessions ───────────────────────┐
+│  #  Session ID          Execution ID        Status    Started    │
+│  1  sess-abc123         exec-xyz789         running   2026-04-15 │
+└───────────────────────────────────────────────────────────────────┘
+
+Enter number to re-attach (1-1), or 'q' to quit: 1
+
+╭─ Re-attach ──────────────────────────────────────────╮
+│ Session selected: sess-abc123                        │
+│ Execution ID:     exec-xyz789                        │
+│                                                      │
+│ Re-attach by running:                                │
+│                                                      │
+│     ooo status exec-xyz789                           │
+╰──────────────────────────────────────────────────────╯
+```
+
+## Next Steps
+
+After re-attaching:
+- `ooo status <exec_id>` — Check current execution status and drift
+- `ooo evaluate` — Evaluate results once execution completes
+- `ooo cancel execution <exec_id>` — Cancel if the session is stuck
--- a/skills/resume/SKILL.md
+++ b/skills/resume/SKILL.md
@@ -46,11 +46,17 @@ When the user invokes this skill:
 
 3. Pick the right re-attach path:
 
-   - **Inspect only** (no execution resumes):
+   - **Inspect only** (read-only interactive monitor):
 
      ```
-     ouroboros status execution <exec_id>
+     ouroboros tui monitor
      ```
+
+     Launches the TUI and lets you pick the session to inspect. The
+     `ouroboros status execution <exec_id>` command is *registered* but its
+     handler is still a placeholder in `src/ouroboros/cli/commands/status.py`
+     (it only prints "Would show details for execution: …"), so it is
+     intentionally not surfaced here. Follow-up tracked as a separate issue.
 
    - **Resume execution** (requires the original seed file):
 
@@ -70,10 +76,10 @@ When the user invokes this skill:
 
 If the command reports "No in-flight sessions found", the execution either
 completed, failed, was cancelled, or the EventStore has never been created.
-Check with:
+To browse historical sessions interactively, use the TUI monitor:
 
 ```
-ouroboros status executions
+ouroboros tui monitor
 ```
 
 ## Example
@@ -92,8 +98,8 @@ Enter number to re-attach (1-1), or 'q' to quit: 1
 │ Session ID:   sess-abc123                                                  │
 │ Execution ID: exec-xyz789                                                  │
 │                                                                            │
-│ Inspect (read-only status):                                                │
-│     ouroboros status execution exec-xyz789                                 │
+│ Inspect (read-only interactive monitor):                                   │
+│     ouroboros tui monitor                                                  │
 │                                                                            │
 │ Resume execution (requires the original seed file):                        │
 │     ouroboros run workflow --orchestrator --resume sess-abc123 seed-001    │
@@ -104,7 +110,7 @@ Enter number to re-attach (1-1), or 'q' to quit: 1
 
 After you have the identifiers:
 
-- `ouroboros status execution <exec_id>` — inspect current execution status
+- `ouroboros tui monitor` — launch the TUI and pick the session to inspect
 - `ouroboros run workflow --orchestrator --resume <session_id> <seed.yaml>` — resume execution
 - `ooo evaluate` — evaluate results once the execution completes
 - `ooo cancel execution <exec_id>` — cancel if the session is stuck
--- a/skills/resume/SKILL.md
+++ b/skills/resume/SKILL.md
@@ -40,7 +40,7 @@ When the user invokes this skill:
 3. Re-attach using:
 
    ```
-   ooo status <exec_id>
+   ouroboros status execution <exec_id>
    ```
 
 ## Fallback (No sessions found)
@@ -70,13 +70,13 @@ Enter number to re-attach (1-1), or 'q' to quit: 1
 │                                                      │
 │ Re-attach by running:                                │
 │                                                      │
-│     ooo status exec-xyz789                           │
+│     ouroboros status execution exec-xyz789             │
 ╰──────────────────────────────────────────────────────╯
 ```
 
 ## Next Steps
 
 After re-attaching:
-- `ooo status <exec_id>` — Check current execution status and drift
+- `ouroboros status execution <exec_id>` — Check current execution status and drift
 - `ooo evaluate` — Evaluate results once execution completes
 - `ooo cancel execution <exec_id>` — Cancel if the session is stuck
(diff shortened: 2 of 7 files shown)


shaun0927:
**Labels**: `enhancement` `cli` — Refs #430

Q00 (changes requested):
Requesting changes because the command's primary recovery instruction points users at the wrong CLI.

`src/ouroboros/cli/commands/resume.py` currently prints `ooo status <exec_id>` as the re-attach command, and the docstring/tests lock in that same string. But this is a terminal CLI command, and the installed console script is `ouroboros`, not `ooo`, so the happy-path output directs users to a non-existent command.

Please print the actual supported CLI equivalent (for example `ouroboros status execution <id>` if that is the intended re-attach path), and update the tests/docs accordingly.

Q00 (changes requested):
## Review Verdict: Changes Requested

The direction is right — we do need a CLI that finds in-flight sessions from the local DB when MCP is down. But the current implementation doesn't fully satisfy the two core promises in the PR description: **read-only recovery** and **re-attach guidance**. I'd hold off on merge until the two blockers below are resolved.

---

### 1. "Read-only" is not actually read-only — it can create the DB and schema

`src/ouroboros/cli/commands/resume.py:34-37`

```python
db_path = os.path.expanduser("~/.ouroboros/ouroboros.db")
os.makedirs(os.path.dirname(db_path), exist_ok=True)
event_store = EventStore(f"sqlite+aiosqlite:///{db_path}")
await event_store.initialize()
```

- `os.makedirs(...)` creates `~/.ouroboros/` if it doesn't exist.
- `EventStore.initialize()` runs `metadata.create_all` (see `src/ouroboros/persistence/event_store.py:163`), which creates a new SQLite file and tables when none exist, and adds missing tables to an existing DB.

The PR description says *"Read-only: resume never modifies the EventStore or creates events"*, but in practice it only guarantees "no events appended" — the filesystem and DB schema are still mutated. A diagnostic/recovery command shouldn't change home-directory state when the user runs it.

**Suggested fix**
- If `~/.ouroboros/ouroboros.db` does not exist, return `No EventStore found` / `No in-flight sessions found` immediately.
- Remove `os.makedirs(...)`.
- Prefer a read-only SQLite URI, or add a `EventStore.initialize(create_schema=False)` path for read-only consumers.
- Add a test that points `HOME` at `tmp_path`, runs `ouroboros resume`, and asserts no `.ouroboros/` directory or DB file is created.

---

### 2. The printed "re-attach" command does not actually work

`src/ouroboros/cli/commands/resume.py:152-159` prints:

```
ouroboros status execution <exec_id>
```

But `src/ouroboros/cli/commands/status.py:48-67` is still a placeholder:

```python
print_info(f"Would show details for execution: {execution_id}")
```

So following the guidance literally does nothing. On top of that, the real resume path in `src/ouroboros/cli/commands/run.py` takes a `session_id`, not an `execution_id`:

```
ouroboros run workflow --orchestrator --resume <session_id> seed.yaml
```

This PR surfaces `exec_id` as the primary handle, which doesn't match the existing `--resume` contract. If the feature is really "recover an in-flight session", the user must be given a command they can actually run next.

**Suggested fix** — pick one:
- (a) Surface both `session_id` and `execution_id`, split "inspect" vs "resume execution" guidance, and note that resume also requires the seed file; or
- (b) Implement `ouroboros status execution <id>` as a real EventStore read inside this PR so the printed command resolves.

---

### 3. O(N sessions × replay) on large EventStores

`src/ouroboros/cli/commands/resume.py:52-73`

Current flow:
- `get_all_sessions()` returns every `orchestrator.session.*` lifecycle event.
- De-duplicate by `session_id`.
- For each session, call `repo.reconstruct_session(session_id)` — which replays events and queries related execution events.

Fine for a small DB, but on long-running users this makes a recovery tool sluggish at exactly the moment the user wants to see in-flight sessions fast. Worse, `src/ouroboros/persistence/event_store.py:445` already provides `get_session_activity_snapshots()`, whose docstring explicitly says *"avoids replaying every event for every session just to detect stale active sessions"* — i.e., the purpose-built API for this exact case.

**Suggested fix**
- Narrow candidates with `get_session_activity_snapshots()` first.
- Reconstruct only `running`/`paused` candidates.
- Sort by `started_at` / last activity and cap the default display (e.g., most recent N) with an opt-in flag to show more.

---

### 4. Tests don't cover the contracts that matter

`tests/unit/cli/test_resume.py`

Most tests mock `_get_event_store`, so they never exercise the behaviors that caused Findings 1 and 2:
- "DB does not exist → command must not create it"
- "initialize must not create schema"

And the corrupted-DB assertions use `assert result.exit_code in (0, 1)`, which is effectively untested error policy. For a recovery CLI, the exit code and message for missing / corrupted DBs should be pinned.

**Tests to add**
- Missing DB: command runs without creating `.ouroboros/` or a DB file.
- Corrupted DB: pinned exit code + pinned error message.
- Printed command matches the real `resume` / `status` contract.
- Multiple active sessions: sorted by most recent.
- Session without `execution_id`: fallback wording is safe and accurate.

---

## Non-blocking notes

- `skills/resume/SKILL.md` is fine, but the word **"re-attach"** overstates current behavior — it's closer to "list and print the next command".
- `CLAUDE.md` lists `ooo resume`; `.claude-plugin/plugin.json` points `"skills": "./skills/"` at the repo-root `skills/` dir, so packaging should pick `skills/resume/` up — worth a quick manual verify.
- In `_get_event_store()`, if `initialize()` raises, `event_store` is never bound and `close()` in the outer `finally` is skipped. Low-impact for a one-shot CLI but worth tightening.

---

## Maintainer recommendation

Holding on merge is the right call. Two things to fix:

1. Make it **genuinely read-only** (no directory, no schema creation).
2. Print a command the user **can actually run** to inspect/resume.

With those landed this is a quick merge — the support/operability value is real, especially paired with `#434 mcp doctor`, which together give users a self-serve recovery layer when MCP dies.

shaun0927:
Thanks for the detailed review — addressed all four blockers in the two commits just pushed (`6e86ce4`, `c49b1b4`).

**1. Genuinely read-only.** Dropped `os.makedirs()` and added `EventStore.initialize(create_schema=False)` so the command opens the store without running `metadata.create_all`. Missing DB short-circuits to "No in-flight sessions found" and exits 0 with no filesystem side-effects. New tests pin this by pointing `HOME` at a `tmp_path` and asserting the `.ouroboros/` dir and DB file never appear.

**2. Re-attach contract.** Output now surfaces both identifiers with the real CLI commands:
- `ouroboros status execution <exec_id>` (inspect)
- `ouroboros run workflow --orchestrator --resume <session_id> <seed.yaml>` (resume execution — takes `session_id` per `run.py`, with the seed file noted)

Session without a recorded `seed_id` prints a `<seed.yaml>` placeholder plus an explanatory hint.

**3. Performance.** Switched the candidate scan to `get_session_activity_snapshots()`; terminal snapshots short-circuit before any replay, and `reconstruct_session()` only runs on the non-terminal set. Results are sorted by `last_activity` desc and capped at 20 rows by default, with `--all` for the full list.

**4. Tests.** Replaced the loose `exit_code in (0, 1)` assertions with pinned exit codes (`EXIT_CORRUPTED_DB = 2` for open/read failures, `0`/`1` elsewhere) and a pinned error message check. Added regression tests for: missing-DB no-side-effect, missing-seed-id fallback wording, most-recent-first sort, display-limit cap, and both the `status execution` and `run workflow --resume` contracts in the printed output.

**Non-blocking notes.** `_get_event_store` now disposes a partially constructed engine before re-raising (so the outer `close()` is no longer required for cleanup in that path). `skills/resume/SKILL.md` softened to "list and print the next command" framing.

Full suite: `tests/unit/cli/` + `tests/unit/persistence/` — 558 passed locally. Ready for re-review.

shaun0927:
Pushed `877cf11` addressing the ouroboros-agent[bot] design note on `137132a` — the `resume` implementation itself is fine (truly read-only, snapshot-backed, pinned contracts), but the surrounding `ooo` skill surfaces still pointed at CLI paths that don't match real handler behavior. This commit closes that gap and adds a contract test so the drift can't silently reappear.

## What changed

**`skills/resume/SKILL.md`** — removed every reference to `ouroboros status execution <exec_id>`. The subcommand is registered but `src/ouroboros/cli/commands/status.py::execution` still only prints `"Would show details for execution: …"`, so surfacing it misleads users. Every inspect-only instruction now points at `ouroboros tui monitor` (the genuinely functional interactive inspection path), with an inline note explaining that the `status execution` placeholder is intentionally not surfaced until the handler is implemented.

**`tests/unit/cli/test_doc_commands.py`** — new contract test suite. Every runnable `ouroboros …` string we emit to users is walked through the installed Typer tree and asserted parseable via `--help`. Scope:

1. Every command emitted by `resume._format_reattach_guidance` (inspect + resume paths).
2. Every `ouroboros …` directive documented in `skills/resume/SKILL.md`.
3. The CLAUDE.md `ooo` table entries that shell out to a real CLI subcommand (`run`, `init`, `config`, `status`, `cancel`, `mcp`, `setup`, `tui`, `pm`, `resume`, `uninstall`).
4. The `cancel execution` path referenced from the skill's "Next Steps" block.
5. A pinned negative check: re-attach guidance must NEVER surface `ouroboros status execution`.

If a new documented command string is added, the test suite is deliberately strict — it ignores nothing, so any future drift of the kind this review caught surfaces immediately.

## Verification

- `uv run ruff format --check .` → 524 files clean
- `uv run ruff check .` → All checks passed
- `uv run mypy src/ouroboros/cli/` → Success, no issues
- `PYTHONPATH=src uv run pytest tests/unit/cli/ -q` → **452 passed** (18 new `test_doc_commands.py` tests + full existing suite)

## Scope

Stays narrow — no changes to `resume.py` semantics or to the EventStore read-only enforcement landed in `137132a`. Just aligns the user-facing surfaces with the real CLI and pins the alignment with a regression test. The `status execution` handler remains a placeholder; implementing it belongs to a separate follow-up, not this PR's recovery scope.

Q00:
@shaun0927 please rebase this one too !

shaun0927:
Rebased onto `main` (top of `a606500`, now at head `432ff77`). All 6 branch commits replayed; one textual conflict in `src/ouroboros/cli/main.py` resolved by keeping both `detect` (added on main) and `resume` (this PR) imports.

Local validation on the rebased HEAD:
- `uv run ruff format --check .` — 555 files clean
- `uv run ruff check .` — all checks passed
- `uv run mypy src/ouroboros/cli/` — 24 files, no issues
- `PYTHONPATH=src uv run pytest tests/unit/cli/ -q` — **574 passed**

No scope-widening changes: the ouroboros-agent[bot] non-blocking notes on `877cf11` (seed-001 placeholder in tests/SKILL) are intentionally left for a follow-up to keep this rebase narrow.

PR #442, opened 2026-04-17
feat(opencode): bridge plugin runtime + mode split (plugin|subprocess)

### LINKED ISSUE #331: feat(orchestrator): OpenCode session pollution — child sessions from ouroboros subprocess calls clutter while picking a session
## Summary

Every ouroboros LLM call spawns `opencode run --format json` as a subprocess. Each call creates a new session visible in OpenCode's session picker. A single `ooo evolve` run can create 10+ orphan sessions cluttering the UI. Need a strategy to either hide, reparent, or clean up these child sessions.

Screenshot_20260406_122944

## Context

- OpenCode sessions are stored in SQLite at `~/.local/share/opencode/opencode.db`
- A `parent_id` column already exists — sessions with `parent_id` set are hidden from the session picker (TUI filters `parentID === undefined`, API filters `isNull(parent_id)` when `roots: true`)
- OpenCode's internal `task` tool already uses this mechanism for subagent sessions
- MCP servers are spawned once at startup, not per-session — no per-call session context exists in the MCP protocol **until now** (see upstream PR below)
- Ouroboros already tracks `server_session_id` per runtime handle, so any reparenting logic hooks into existing infrastructure

## Key Discoveries

After investigating OpenCode internals in depth:

1. **`opencode run` boots its own server** (`run.ts:680-687` via `bootstrap()`). It does NOT connect to the TUI's HTTP server. REST API pre-create from ouroboros is impossible — there's no way to reach the TUI server reliably.

2. **MCP protocol supports `_meta`**: `CallToolRequestParamsSchema` extends `BaseRequestParamsSchema` which has `_meta: Record<string, unknown>` as an optional field. This is the standard mechanism for per-call context.

3. **OpenCode's `callTool()` never passed `_meta`**: In `mcp/index.ts:148`, only `name` and `arguments` were sent. No session context reached MCP servers.

4. **4 duplicate upstream issues requesting this**: #6279, #15117, #9292, #15739 — all asking for session context in MCP/subprocess calls.

5. **Upstream PR opened**: [anomalyco/opencode#21624](https://github.com/anomalyco/opencode/pull/21624) adds `_meta: { sessionID }` to every `callTool()` request. 3 files, +10/-4 lines, all CI green, mergeable. Closes opencode#6279 and opencode#17084.

6. **Both TUI and `opencode run` share the same SQLite DB** (`~/.local/share/opencode/opencode.db` via WAL mode). Direct reparenting via SQL is safe for concurrent writes.

7. **MCP `sampling/createMessage` not implemented in OpenCode** — the MCP protocol has a built-in mechanism for servers to request LLM completions from the host client. The SDK fully supports it (`CreateMessageRequestSchema`, `CreateMessageResultSchema`). However, OpenCode declares zero client capabilities during MCP init and registers no request handlers. A server sending `sampling/createMessage` gets a JSON-RPC method-not-found error.

8. **OpenCode HTTP server only runs in specific modes** — `opencode serve`, `opencode web`, `opencode --port`, or `opencode acp`. In normal CLI/TUI mode, there is NO HTTP server. The plugin SDK uses in-process `app.fetch()`, not HTTP. Any approach relying on HTTP (bridge, REST API) is dead in default CLI/TUI mode.

Sathvik-1007:
## TL;DR

Add **in-process OpenCode plugin runtime** for subagent dispatch. Keep subprocess path for CI. User pick one at setup.

Closes the UX pain from #331: dispatching subagents via `opencode run` subprocess spawned 10+ orphan sessions cluttering the TUI picker. Plugin path runs inside the active OC session — **zero new sessions, zero subprocess boot, live inline Task panes**.

refs #331

---

## the problem (before — from #331)

Every subagent dispatch = new `opencode run` child process = new session row. After one `ooo evolve` with fan-out, picker looks like this:

picker-pollution-before

Subprocess runs = headless. No live UI. Parent LLM blocked ~minutes waiting on children. Orphan sessions stay forever.

## the fix (after — plugin path)

Plugin lives inside active OC session. Intercepts MCP tool call → PATCHes `/session/{id}/prompt` for each subtask → **N live Task panes render in same active pane**, prompts + tool calls + streaming all visible:

inline-task-panes-after

No subprocess. No picker pollution. Parent sees children work in real time.

---

## how plugin path works (for new contributors)

1. User run MCP tool (e.g. `ouroboros_execute_seed`) from inside OC TUI
2. Tool want to fan out N subagents. Instead of spawn `opencode run` N times, tool return **envelope** `{_subagents: [...]}`  (or `_subagent` single)
3. Bridge plugin (`ouroboros-bridge.ts`) register `tool.execute.before` hook. See envelope → intercept
4. For each subtask:
   - `POST /session` → create child session (await, ~50ms)
   - `PATCH /session/{id}` → attach as widget-child of parent pane (await, ~50ms)
   - `PATCH /session/{id}/prompt` → fire subtask prompt (**fire-and-forget**, no await)
5. Parent LLM unblock immediate. Children stream in widget panes. `.then/.catch` settle widget state background.
6. Parallel fanout via **direct PATCH** bypass OC single-flight queue. `MAX_FANOUT=10`.

### key concepts explained

| term | meaning |
|------|---------|
| **bridge plugin** | TS file OC loads at startup. Registers hooks. Lives next to config. |
| **`tool.execute.before`** | OC plugin hook fired before any tool return propagate |
| **envelope** | Magic JSON `{_subagent: {...}}` — signal to plugin "dispatch this" |
| **widget / Task pane** | Inline child UI inside parent session — not a new session in picker |
| **fire-and-forget** | `fetch(...).then(...)` no `await` — parent LLM not block on children |
| **single-flight queue** | OC's default "one prompt at a time per session" serialization. We bypass via direct PATCH. |
| **`--pure`** | `opencode run` flag that disable plugin loading. Stop double-dispatch inside subprocess. |

---

## subprocess path — kept for CI

Run via `opencode run --pure --format json`:
- `--pure` = no plugins load → no double-dispatch risk
- pollute picker (known #331 trade) — acceptable for headless/scriptable use
- this PR **not close #331** — subproc still ship, plugin = opt-in escape hatch

---

## mutual exclusion (why user pick one)

Both wired at once → MCP tool fire inside subprocess → globally-registered plugin trigger → **double dispatch** → 2× tokens + duplicate panes.

User pick via setup:
```
ouroboros setup --runtime opencode --opencode-mode plugin
ouroboros setup --runtime opencode --opencode-mode subprocess
```

Switch = rerun setup.

---

## 10 tools now dispatch via plugin

`qa` · `lateral_think` · `interview` · `pm_interview` · `generate_seed` · `execute_seed` · `start_execute_seed` · `evolve_step` · `start_evolve_step` · `evaluate`

Each handler gated by `should_dispatch_via_plugin(agent_runtime_backend, opencode_mode)`. Gate `False` → fall back to subprocess or inline iteration.

---

## envelope contract (the `response_shape` merge — round 3 fix)

**Problem**: plugin dispatch originally return sole-key `{_subagent: {...}}` envelope → broke tools like `start_execute_seed` that callers chain via `job_id`:

```js
// before round-3: job_id gone → downstream break
{_subagent: {prompt: "...", callID: "..."}}

// after round-3: both coexist
{
  _subagent: {prompt: "...", callID: "..."},
  job_id: "job_abc",
  session_id: "sess_xyz",
  status: "delegated_to_subagent"
}
```

Plugin parser do `obj._subagent` lookup → ignore extra keys. Zero plugin changes. Backward compat: `response_shape=None` → legacy sole-key envelope.

---

## review hardening — 4 rounds, all APPROVE

| round | commit | fix |
|-------|--------|-----|
| 1 | `af81fb5` | `ChannelWorkflowHandler` used `copy.copy` to clone inner handlers before pin `opencode_mode='subprocess'` — no more mutate caller state |
| 2 | `d74d219` | `LateralThinkHandler` multi-persona fan-out now gated (was unconditional envelope → silent drop on subproc); share `_strip_jsonc` helper with setup |
| 3 | `1118919` | preserve public tool response shape via `response_shape` kwarg merge — fix chain-call break |
| 4 | `771c9da` | uninstall predicate-filter bridge entries via `is_bridge_plugin_entry` — remove stale XDG / root / Windows paths not just current install |

---

### new files
- `src/ouroboros/opencode/plugin/ouroboros-bridge.ts` — 444 lines, v28
- `src/ouroboros/opencode/plugin/ouroboros-bridge.test.ts` — 77 bun tests, 1312 `expect()`, ~120ms
- `docs/guides/opencode-subagent-bridge.md` — plugin-specific guide

### touched
- `cli/commands/setup.py` — `--opencode-mode` flag, interactive picker, atomic content-hashed plugin write, cross-platform XDG dir
- `cli/commands/uninstall.py` — predicate filter via `is_bridge_plugin_entry`
- `cli/opencode_config.py` — public `BRIDGE_PLUGIN_SUBDIR`, `BRIDGE_PLUGIN_FILENAME`, `is_bridge_plugin_entry` (round 4)
- `mcp/tools/subagent.py` — `build_subagent_result` / `build_multi_subagent_result` accept `response_shape` kwarg (round 3)
- `mcp/tools/*_handlers.py` — 10 handlers gated via `should_dispatch_via_plugin(runtime, mode)` + pass `response_shape`
- `mcp/tools/channel_workflow_handler.py` — `copy.copy` non-mutating clone (round 1)
- `docs/runtime-guides/opencode.md` — 2-mode Configuration section

---

## tests

| suite | result |
|-------|--------|
| unit pytest | **4406 / 4407 pass** (1 pre-existing `pip-fallback` fail on `origin/main`, unrelated) |
| bun bridge | **77 / 77 pass**, 1312 `expect()`, ~120ms |
| mcp-scoped | 597 pass |
| cli-scoped | 505 pass |
| ruff format + check | clean |
| mypy | clean |
| smoke: `mode=plugin` / `mode=subprocess` / `mode=both` | `ValueError` raised ✓ |

## CI at HEAD `771c9da`

| check | status |
|-------|--------|
| MyPy Type Check | pass 48s |
| Ruff Lint | pass 12s |
| Test Python 3.12 | pass 1m39s |
| Test Python 3.13 | pass 1m44s |
| Test Python 3.14 | pass 1m48s |

**5/5 green.**

---

## runtime knobs (plugin mode)

| env var | default | purpose |
|---------|---------|---------|
| `OUROBOROS_CHILD_TIMEOUT_MS` | `1200000` (20min) | max wall time per child before timeout |
| `OUROBOROS_SUB_RETRIES` | `2` | retry count for failed PATCH |
| `OUROBOROS_MAX_FANOUT` | `10` | max parallel children |

---

## out of scope — follow-ups

- **`HandlerFactory` abstraction** — reviewer round-4 advisory. Handlers default `opencode_mode=None` → legacy-safe subprocess fallback → not a bug, just architectural future work.
- **`resolveMid` / `RESOLVE_RETRIES` race tests** — need `export` from `ouroboros-bridge.ts`. Deferred to avoid touching bridge in this PR.

---

## checklist

- [x] one logical change (opencode runtime path)
- [x] tests included (bun + pytest)
- [x] docs updated (runtime guide + bridge guide)
- [x] conventional commits
- [x] ruff clean
- [x] mypy clean
- [x] 4 review rounds resolved (all APPROVE)
- [x] CI 5/5 green
- [x] Ready to merge

refs #331


Q00:
after fix the lint, then will merge this! @Sathvik-1007 
great work!

Sathvik-1007:
@Q00 , just wait a sec, I am making it more robust...I will let you know when it is ready to merge. I will click the PR checkbox when it is ready.

Sathvik-1007:
@Q00 Ready to merge.

Q00 (changes requested):
Took a pass through the PR branch locally and reproduced all three concerns below. Blocking on the first one; the other two should land before merge but are smaller.

## 1. Plugin-mode `start_*` tools break the background-job contract (blocker)

`ouroboros_start_execute_seed` and `ouroboros_start_evolve_step` are advertised as background-job APIs: callers are expected to chain `ouroboros_job_status(job_id)` / `ouroboros_job_result(job_id)`. In plugin mode that contract is silently broken.

The plugin dispatch branch in `src/ouroboros/mcp/tools/execution_handlers.py:872-882` returns:

```python
response_shape={
    "job_id": None,
    "session_id": arguments.get("session_id"),
    "execution_id": None,
    "status": "delegated_to_subagent",
    ...
}
```

The comment says this "preserves public response shape," but it doesn't really survive the trip:

- `src/ouroboros/mcp/server/adapter.py:634` returns only `tool_result.text_content` to FastMCP. The structured `meta` dict never reaches the MCP wire.
- Even the text JSON is overwritten downstream: `src/ouroboros/opencode/plugin/ouroboros-bridge.ts:497` calls `stamp(out, notify(...))`, which wipes `r.content` and replaces it with the dispatch banner (see `stamp` at line 153).

Net effect: callers receive a banner string, no `job_id`, and no way to poll. `ouroboros_job_status(None)` is not a useful handle.

Pick one and commit to it:
- Register a real `JobManager` record before plugin dispatch and have the bridge drive it to completion when the child finishes, so `job_id` stays meaningful end-to-end, **or**
- Keep the `start_*` path on the existing `JobManager` runner and let the child do its thing inside that job, **or**
- Change the contract: rename / re-document these as fire-and-forget dispatch calls and stop claiming `job_id` is returned.

Relatedly, `channel_workflow_handler` chains `start_execute_seed_handler` → `job_wait_handler` internally (`definitions.py:388-396`). In plugin mode the wait step is going to receive `job_id=None` from the start call, so the channel workflow needs to be walked through too.

## 2. `get_ouroboros_tools` doesn't wire all plugin-capable handlers

`get_ouroboros_tools(runtime_backend="opencode", opencode_mode="plugin")` is the factory the rest of the codebase uses. Three handlers there are constructed bare:

```
src/ouroboros/mcp/tools/definitions.py:376  LateralThinkHandler(),
src/ouroboros/mcp/tools/definitions.py:377  EvolveStepHandler(),
src/ouroboros/mcp/tools/definitions.py:378  StartEvolveStepHandler(),
```

I checked on this branch: the resulting instances have `agent_runtime_backend=None, opencode_mode=None`, so `should_dispatch_via_plugin(...)` returns false and they fall back to subprocess-style responses. The per-tool helpers at `definitions.py:224-262` don't even accept those kwargs, so there's no way to pass them through.

`create_ouroboros_server` (adapter.py:1340-1464) wires them correctly, but this is the surface actually used by:
- `OpenCodeRuntime._get_builtin_mcp_handlers` (`orchestrator/opencode_runtime.py:817-826`) caches handlers from this factory.
- The module-level `OUROBOROS_TOOLS` export (`definitions.py:408`).

So the PR description's "10 plugin-dispatched tools" is not true in practice for the runtime's local MCP path. Fix is small: thread `agent_runtime_backend` and `opencode_mode` into the three constructors in `get_ouroboros_tools`, and add the same kwargs to the three helper factories so external callers can pass them through.

## 3. TypeScript bridge has no CI coverage

The bridge has a solid test file, which is great, but it's only green locally:

- `bun test` in `src/ouroboros/opencode/plugin` passes (77 pass).
- `bunx tsc --noEmit` fails: `ouroboros-bridge.ts:1` imports `type { Plugin } from "@opencode-ai/plugin"`, and `@opencode-ai/plugin` isn't declared in `package.json`.
- `.github/workflows/` runs pytest, Ruff, and MyPy. No `bun` or `tsc` step anywhere.

If this bridge is going to ship as maintained source under a strict `tsconfig.json`, either add `@opencode-ai/plugin` to `devDependencies` or drop the external type import and inline the minimal shape you use. Then add a `bun test` (and ideally `bunx tsc --noEmit`) step to CI so the TS side can't silently rot.

## What I ran

- `uv run pytest tests/unit/mcp/tools/test_lateral_think_handler.py tests/unit/mcp/tools/test_subagent.py -q` → 62 passed on this branch.
- `bun test` in `src/ouroboros/opencode/plugin` → 77 passed.
- `bunx tsc --noEmit` → fails on the missing `@opencode-ai/plugin` module.
- GitHub checks (MyPy, Ruff, Python 3.12/3.13/3.14) are green, which matches the CI gap noted above.

Happy to look again once (1) is resolved. (2) and (3) are straightforward follow-ups.

Sathvik-1007:
@Q00, this bot is malfunctioning. I thought of closing and reopening, but this is happening in all PR's. You gotta fix that asap cause this is spamming and causing a lot of confusion

Q00:
@Sathvik-1007 Sorry for duplicated review. the bot has some error and it cause infinite reviews
will fix it.

Q00:
Thanks for the thorough back-and-forth on this PR — I can see how much effort has gone into chasing the plugin-mode edge cases, and it really shows. 💪

Quick note for anyone catching up: the three latest reviews all target the same commit (`43f2378`), and the findings converge, so **the last review (#4137894913) is the canonical one to act on**. The earlier two are effectively subsumed by it.

### Actionable summary (from the final review)

**Blocking #1 — `src/ouroboros/cli/commands/setup.py:956`**
`_setup_opencode()` always prints `Installed OpenCode bridge plugin and registered MCP entry`, even when `_install_opencode_bridge_plugin()` / `_ensure_opencode_mcp_entry()` / `_ensure_opencode_plugin_entry()` returned early on missing source, parse, or write errors. Users end up with `opencode_mode=plugin` in config and a false-positive success, while subagent dispatch silently fails.
→ **Fix:** make the helpers return a success signal and fail closed; only emit the success message when all three succeed.

**Blocking #2 — `src/ouroboros/mcp/tools/pm_handler.py:457`**
In the `action == "start"` branch, `_save_pm_meta()` persists only `state.codebase_paths` derived from `cwd`, not the caller's `selected_repos`. Subsequent `resume`/`generate` turns rebuild the subagent payload from current args only (lines 566–573), so if a follow-up turn omits `selected_repos`, the chosen repos vanish from context — breaking the documented 1-step `initial_context + selected_repos` contract.
→ **Fix:** persist `selected_repos` (and/or `brownfield_repos`) in `pm_meta` during `start`, and restore them on later turns when not re-supplied.

**Nice-to-have:** regression test — start plugin-mode session with `initial_context + selected_repos`, resume/generate without resending, assert repo context still present.

Really appreciate the careful reviews here — the direction (runtime gating + isolated bridge) is solid, and once these two boundaries fail closed we're in a good place. 🙏

Sathvik-1007:
## Summary of all changes — PR #442

Addressing Q00's [review comment](https://github.com/Q00/ouroboros/pull/442#issuecomment-2814826693) and 12 rounds of bot review. Final bot verdict: **APPROVE** (`4abdbe7`).

---

### Q00's 3 actionable findings (from review #4137894913)

| # | Finding | Fix | Commit |
|---|---------|-----|--------|
| **Blocking 1** | `_setup_opencode()` always prints success even when bridge/MCP/plugin install fails — user left in plugin mode with broken bridge | Helpers return bool success signal; config.yaml write **deferred until all 3 succeed**; failure → no config written → user stays safe. Recovery hint corrected to `--runtime opencode --opencode-mode plugin` | `799dfcf`, `4abdbe7` |
| **Blocking 2** | PM `start` branch persists only cwd-derived `codebase_paths`, not caller's `selected_repos` — repos vanish on resume/generate turns | Persist `selected_repos` in `pm_meta` during start; restore on later turns when not re-supplied; use `is not None` (not truthy) so explicit `[]` round-trips correctly | `ea6cba0`, `799dfcf`, `b0f6cef`, `4abdbe7` |
| **Nice-to-have** | Regression test for selected_repos persistence | `test_1step_start_persists_selected_repos` + `TestPMBrownfieldReposPersistence` class | `b0f6cef` |

### Bot review rounds (R5–R12)

| Round | Key findings | Resolution |
|-------|-------------|------------|
| **R5** | Interview/PM handlers dispatch without validating args | Arg validation runs BEFORE plugin dispatch; 15 tests added | 
| **R5.4** | Plugin path doesn't persist InterviewState | Created `_plugin_save_state`/`_plugin_load_state` helpers (no litellm dep); transcript passed to subagent |
| **R5.5** | Plugin path imports litellm via `create_llm_adapter()` | Eliminated entirely — plugin uses direct file I/O only |
| **R5.7** | PM plugin path doesn't persist `pm_meta` on start | `_save_pm_meta` called on start + select_repos; `_load_pm_meta` restores on resume |
| **R6–R8** | Mode isolation, setup fail-closed, stale docstrings, response_shape in bridge dedupe, generate_seed gates | Each addressed individually across `c98bdb8`–`43f2378` |
| **R9** | `generate_seed` gates too strict for plugin mode (is_complete never set by child) | Relaxed to round-count gate; child session does real validation |
| **R10** | setup fail-closed incomplete, brownfield repos lost on resume, selected_repos not persisted in 1-step start | Config write deferred; repos restored from pm_meta; 3 regression tests |
| **R11** | Plugin child sessions can't write questions back to server-side state — all questions show as placeholder | Added `last_question` param to both interview tool schemas; parent LLM relays child's question text; `next_turn_hint` guides parent; 8 regression tests |
| **R12** | Config still written before install (fail-open); `selected_repos=[]` conflated with missing | Config write fully deferred; `is not None` checks throughout |

### Architecture decisions

- **Plugin is plugin, subprocess is subprocess** — clean separation enforced by `should_dispatch_via_plugin()` single gate
- **No litellm in plugin path** — `_plugin_save_state`/`_plugin_load_state` do direct Pydantic JSON I/O
- **Bridge unmodified** — all fixes are server-side; bridge TypeScript untouched after initial feature commits
- **Channel workflow isolation** — `_isolate()` pins nested handlers to subprocess mode, preventing dead `_subagent` envelopes
- **Fail-closed everywhere** — setup, generate gates, save_state Result checking

### Test coverage

- **641 MCP unit tests** passing (40 new in `test_round5_fixes.py`)
- **90 bridge TypeScript tests** passing (CI job added this PR)
- **6/6 CI checks green** on final commit

### Stats

57 files changed, +8372 / -6607

Sathvik-1007:
@Q00 , waiting for review..

Q00 (approved):
Thanks, I checked the latest head as well. @Sathvik-1007 

The core plugin/runtime fixes look good enough to merge: CI is green, the latest bot review is approved, and the previously blocking runtime issues around plugin dispatch,
response shape, PM repo persistence, and setup config write appear resolved.

I did find one remaining setup failure-path issue: if plugin-mode setup fails during bridge/MCP/plugin registration, `_setup_opencode()` returns early but the top-level setup command can still print `Setup complete!` and exit successfully. 

There is also a macOS-only test portability issue where OpenCode config tests hard-code `~/.config/opencode` while the implementation correctly uses `~/Library/Application Support/OpenCode`.

I don't think either should block this PR at this point. I'll merge this and take a follow-up to:
  1. propagate plugin setup failure to the top-level CLI exit status/success message, and
  2. make the OpenCode config tests platform-aware.

PR #423, opened 2026-04-15
fix(mcp): replace remaining Path.cwd() calls with safe fallback in create_ouroboros_server (#400)

shaun0927:
## Summary

v0.28.2 (`9890bfb`) fixed the MCP crash when `cwd=/` by changing `state_dir` from `Path.cwd()` to `Path.home()`, but left **5 other `Path.cwd()` calls** in `create_ouroboros_server()` untouched. These cause runtime failures when any tool is invoked under OpenClaw gateways that spawn with `cwd=/`.

### Changes

- **Added `_safe_cwd()` helper**: returns `Path.cwd()` if writable and non-root, otherwise falls back to `Path.home()`
- **Computed `effective_cwd` once** at the top of `create_ouroboros_server()` and replaced all 5 occurrences:
  - `create_agent_runtime(cwd=...)` — initial runtime validation
  - `create_llm_adapter(cwd=...)` — shared LLM adapter
  - `cwd=task_cwd or ...` — evolution executor fallback (kept guard)
  - `_looks_like_project_root(...)` — project dir extraction
  - `default_repo=str(...)` — workflow manager
- **Removed stale** `from pathlib import Path` inside `_extract_project_dir()` (no longer needed)

Closes #400

## Test plan

- [x] `test_returns_cwd_when_writable_and_not_root` — normal directory returned as-is
- [x] `test_falls_back_to_home_when_cwd_is_root` — `cwd=/` falls back to `Path.home()`
- [x] `test_falls_back_to_home_when_cwd_not_writable` — non-writable cwd falls back
- [x] All 33 adapter tests pass

PR #419, opened 2026-04-15
fix(security): validate transport URLs and replace private mcp API import

shaun0927:
## Summary

- **SSRF prevention**: `MCPServerConfig.__post_init__()` now validates URL schemes, rejecting `file://`, `gopher://`, `ftp://`, etc. -- only `http://` and `https://` are allowed for SSE/HTTP transports
- **Private API removal**: Replaced `from mcp.shared._httpx_utils import create_mcp_http_client` with the public `httpx.AsyncClient` in the STREAMABLE_HTTP/HTTP branch of `_raw_connect()`
- **Tests**: Added parametrized tests for scheme rejection (file, gopher, ftp) across SSE and STREAMABLE_HTTP transports, acceptance tests for http/https, and a source-level assertion that the private import is absent

Closes #402

## Test plan

- [x] `MCPServerConfig` rejects `file://`, `gopher://`, `ftp://` URLs with `ValueError`
- [x] `MCPServerConfig` accepts `http://` and `https://` URLs
- [x] Source inspection test verifies `_httpx_utils` and `create_mcp_http_client` are absent from adapter.py
- [x] All 78 existing MCP tests pass (`uv run pytest tests/unit/mcp/test_types.py tests/unit/mcp/client/ -v -x`)
- [x] `ruff check` clean on both modified source files

Issue #402, opened 2026-04-15
security(mcp): SSRF via unvalidated transport URL + fragile private API import


shaun0927:
## Summary

Two related issues in the MCP client transport layer introduced in v0.28.3 (#339):

### 1. SSRF — No URL scheme/host validation

`MCPServerConfig.__post_init__()` only checks that `url` is truthy — it does not validate the scheme, host, or port. This allows SSRF attacks via internal network URLs, cloud metadata endpoints, or exotic schemes.

```python
# Accepted without validation:
url = "http://169.254.169.254/latest/meta-data/iam/security-credentials/"  # AWS metadata
url = "http://kubernetes.default.svc/api/v1/secrets"  # K8s API
url = "gopher://internal-redis:6379/_SET pwned true"  # Redis via gopher
url = "file:///etc/passwd"  # local file
```

### 2. Private API dependency

```python
# adapter.py:261
from mcp.shared._httpx_utils import create_mcp_http_client
```

`_httpx_utils` is a private module (underscore prefix), not part of `mcp.shared`'s public API. Any minor/patch update to the `mcp` package can remove or rename it, breaking the HTTP transport entirely.

## Impact

- **SSRF**: In cloud environments, attackers can access IAM credentials via metadata endpoints, or probe internal services
- **Breakage**: `mcp` package update can cause `ImportError`, disabling HTTP/streamable-http transport

## Suggested Fix

1. Validate URL scheme is `http://` or `https://` only
2. Block RFC 1918 / link-local / loopback addresses (configurable allowlist)
3. Replace `mcp.shared._httpx_utils.create_mcp_http_client` with `httpx.AsyncClient` directly

## Introduced In

v0.28.3 — commit `ded5542` (#339)

shaun0927:
**Labels**: `security` `mcp`

PR #447, opened 2026-04-17
fix(mcp): reject transport hostnames that resolve to local/private IPs

### LINKED ISSUE #446: security(mcp): hostname aliases can bypass transport URL validation for local/private targets
## Summary

PR #419 significantly improved MCP transport URL validation by blocking non-HTTP schemes, literal private/loopback IPs, and `localhost`.

However, the current validator still returns early for non-IP hostnames. That leaves a follow-up gap: hostnames that resolve to loopback, link-local, or private IPs are still accepted.

## Current behavior

In `src/ouroboros/mcp/types.py`, `_validate_transport_url()` currently does this for non-IP hostnames:

```python
try:
    ip = ipaddress.ip_address(host_literal)
except ValueError:
    # Not an IP literal -- a DNS name. Static validation ends here.
    return
```

That means validation stops before checking what the hostname actually resolves to.

### Real hostname alias on current `main`

On current `main`, this succeeds even though the hostname resolves to loopback:

- `127.0.0.1.nip.io` resolves to `127.0.0.1`
- `MCPServerConfig(name="demo", transport=TransportType.HTTP, url="http://127.0.0.1.nip.io/")` is accepted

At the same time:

- `http://localhost/` is rejected
- `http://127.0.0.1/` is rejected

So the literal/local-host checks are working, but the hostname-alias class is not covered.

### Metadata-style alias class

I also validated the same gap with a targeted local test by resolving a hostname to `169.254.169.254` via `socket.getaddrinfo()` mocking. Current behavior accepts the hostname; a minimal patch that resolves the hostname before allowing it blocks it correctly.

## Why this matters

This is the same transport-hardening problem family addressed by #402 / #419: the URL validator is intended to be a security boundary for MCP transports. Right now that boundary is still syntax-only for hostnames, so public-looking DNS aliases can bypass the local/private target checks.

I am **not** claiming this fully solves all DNS-rebinding or TOCTOU cases. The narrower point is that obvious hostname aliases to blocked address classes should not pass the current validator.

## Proposed fix

Best-effort follow-up hardening:

1. For non-IP hostnames, resolve with `socket.getaddrinfo()`
2. If any resolved address is loopback / link-local / private / reserved / unspecified, reject it
3. Keep `OUROBOROS_ALLOW_LOCAL_TRANSPORT=1` as the local-dev escape hatch
4. Treat DNS resolution failures as inconclusive (non-fatal), preserving current behavior for temporarily unresolvable public hosts

## Acceptance Criteria

- [ ] Hostnames resolving to loopback IPs are rejected by default
- [ ] Hostnames resolving to link-local metadata IPs (for example `169.254.169.254`) are rejected by default
- [ ] `OUROBOROS_ALLOW_LOCAL_TRANSPORT=1` still allows intentional local development
- [ ] DNS resolution failures remain non-fatal
- [ ] Tests cover hostname-alias cases in addition to literal IP / `localhost` cases

## Related

- #402
- #419

shaun0927:
## Summary

Closes #446.

This is a narrow follow-up hardening pass after #419.

The MCP transport validator now resolves non-IP hostnames once and rejects them when they map to blocked address classes (loopback, link-local, private, reserved, or unspecified IPs).

## Problem

`_validate_transport_url()` already blocks:

- non-HTTP schemes
- literal local/private IPs
- `localhost`
- userinfo / empty-host cases

But for any other hostname it currently returns early after `ipaddress.ip_address()` fails, which means public-looking DNS aliases to local/private targets still pass validation.

Concrete examples on current `main`:

- `http://127.0.0.1.nip.io/` resolves to `127.0.0.1` and is accepted
- a hostname resolving to `169.254.169.254` is accepted unless the caller uses a literal IP

This keeps the validator in the same problem family already addressed by #402 / #419, just one level narrower: the hardening is complete for literals but not for hostname aliases.

### `src/ouroboros/mcp/types.py`

- add `_is_blocked_transport_ip()` to centralize the blocked-address predicate
- add `_resolved_blocked_transport_ips()` to resolve hostnames with `socket.getaddrinfo()`
- for non-IP hostnames:
  - resolve once
  - reject if any resolved address is loopback / link-local / private / reserved / unspecified
- keep `OUROBOROS_ALLOW_LOCAL_TRANSPORT=1` as the local-dev escape hatch
- keep DNS resolution failures non-fatal so temporarily unresolvable public hosts do not become invalid configs

### `tests/unit/mcp/test_types.py`

Add regression coverage for:

- hostname alias resolving to loopback
- hostname alias resolving to `169.254.169.254`
- DNS resolution failure staying non-fatal
- escape hatch still allowing intentional local aliases

## Scope / non-goals

This PR is intentionally narrow.

It does **not** claim to solve every DNS-rebinding or TOCTOU case. The goal is to close the obvious hostname-alias gap while preserving the existing connect-time redirect hardening and the local-development override.

## Test plan

- [x] `ruff check src/ouroboros/mcp/types.py tests/unit/mcp/test_types.py`
- [x] `PYTHONPATH=src pytest tests/unit/mcp/test_types.py -q`
- [x] `PYTHONPATH=src pytest tests/unit/mcp/test_types.py tests/unit/mcp/client/test_adapter_transport_lifecycle.py -q`
- [x] Manual runtime repro: `127.0.0.1.nip.io` accepted on current `main`, blocked with this patch
- [x] Manual runtime repro: mocked hostname resolving to `169.254.169.254` is blocked with this patch

## Why this shape matches recent merged fixes

Recent merged hardening PRs in this area (#419, #423) have been:

- small and surgical
- explicit about the exact boundary being fixed
- paired with focused regression tests
- careful not to over-claim broader security guarantees

This PR follows the same pattern.

### DIFF
--- a/src/ouroboros/mcp/types.py
+++ b/src/ouroboros/mcp/types.py
@@ -125,11 +125,18 @@ def _validate_transport_url(url: str, transport: str) -> None:
     # Strip IPv6 brackets (urlparse already does, but be defensive).
     host_literal = hostname.strip("[]")
 
+    # Normalize canonical DNS form before matching known loopback names: DNS is
+    # case-insensitive and a single trailing dot marks an absolute FQDN, so
+    # ``LOCALHOST.`` and ``localhost`` denote the same host.  ``urlparse``
+    # already lowercases, but it does not strip the trailing dot, which let
+    # variants like ``http://localhost./`` slip past the well-known check.
+    host_literal = host_literal.rstrip(".") or host_literal
+
     # Check for well-known loopback hostnames before attempting IP parsing.
     # These bypass the IP-literal checks because they are DNS names, but
     # they resolve to loopback addresses and must be blocked unless the
     # dev escape hatch is enabled.
-    if host_literal in _LOOPBACK_HOSTNAMES:
+    if host_literal.lower() in _LOOPBACK_HOSTNAMES:
         if allow_local:
             return
         msg = (
--- a/src/ouroboros/mcp/types.py
+++ b/src/ouroboros/mcp/types.py
@@ -86,12 +86,25 @@ def _validate_transport_url(url: str, transport: str) -> None:
         * URLs carrying userinfo (``user:pass@host``) used for credential
           smuggling / host confusion.
         * URLs with an empty hostname (e.g. bare ``http://``).
-        * Literal IPs resolving to loopback, link-local, or private ranges,
-          unless the ``OUROBOROS_ALLOW_LOCAL_TRANSPORT=1`` dev escape is set.
+        * Well-known canonical loopback hostnames (``localhost`` and its
+          trailing-dot / mixed-case variants).
+        * Literal IPs inside loopback, link-local, private, multicast,
+          reserved, or unspecified ranges.
+        * DNS hostnames whose ``getaddrinfo()`` result resolves to any
+          address in those same blocked ranges (e.g. ``*.nip.io`` aliases,
+          DNS rebinding targets, metadata-IP aliases).
 
-    DNS resolution is intentionally out of scope for this static validation;
-    runtime code that follows redirects should be hardened separately (the
-    adapter sets ``follow_redirects=False``).
+    The ``OUROBOROS_ALLOW_LOCAL_TRANSPORT=1`` dev escape re-enables every
+    range above for local development only.
+
+    Boundary note: two distinct normalizations are used here. Canonical
+    matching against ``_LOOPBACK_HOSTNAMES`` uses a ``rstrip('.').lower()``
+    form so that ``LOCALHOST.``/``localhost.``/``localhost`` are recognized
+    as the same name. The DNS / IP-literal checks, however, resolve the
+    exact host the MCP client will connect to (the unmodified value
+    returned by ``urllib.parse``, minus IPv6 brackets), because a stripped
+    trailing dot can change how some resolvers answer. Canonicalization is
+    for identity matching, not for picking the lookup target.
 
     Args:
         url: The transport URL to validate.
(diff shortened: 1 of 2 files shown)


shaun0927:
Pushed `8ffdfcd` addressing the ouroboros-agent design note on canonical localhost normalization.

## What changed

`src/ouroboros/mcp/types.py:_validate_transport_url`
- Strip a single trailing dot (FQDN absolute form) and lowercase before matching `_LOOPBACK_HOSTNAMES`, so `http://localhost./` and `https://LOCALHOST./` hit the well-known loopback branch instead of falling through to DNS resolution.
- `host_literal.rstrip(".") or host_literal` preserves the bare-dot edge case safely.

`tests/unit/mcp/test_types.py::test_rejects_localhost`
- Added parametrized variants covering the three canonical forms flagged in the design note:
  - `http://localhost./`
  - `http://localhost.:3000/`
  - `https://LOCALHOST./`

## Verified locally

- `ruff check src/ouroboros/mcp/types.py tests/unit/mcp/test_types.py` → clean
- `PYTHONPATH=src pytest tests/unit/mcp/test_types.py -q` → 65 passed

## Scope

Still intentionally narrow — this closes the canonical FQDN gap called out by the bot's non-blocking suggestion at `test_types.py:322` without expanding into broader DNS rebinding or IPv4-mapped IPv6 territory. Those remain worth follow-ups but belong to separate PRs.

shaun0927:
Thanks for the follow-up review on 8ffdfcd. Pushed 998cde9 addressing both non-blocking items:

**1. Docstring drift (Suggestion 1, `src/ouroboros/mcp/types.py:92`).**
Removed the stale "DNS resolution is intentionally out of scope for this static validation" claim. The docstring now describes what `_validate_transport_url` actually does: canonical well-known loopback match, IP-literal range check, and `getaddrinfo()`-backed public-alias resolution against the blocked ranges. It also enumerates the escape hatch semantics.

**2. Boundary between canonical matching and the lookup target.**
Previously the validator did `host_literal.rstrip('.').lower()` once and used that mutated string for both well-known-name matching *and* as the DNS / IP-literal lookup target. As you noted, that breaks the boundary with the HTTP client: if the client library preserves the trailing dot and dials `localhost.` (absolute FQDN), a resolver can legitimately answer differently than it would for the stripped relative form, and our check would be validating a different name than the one actually connected to.

Refactored so the validator carries two distinct strings:

- `canonical_host = (lookup_host.rstrip('.') or lookup_host).lower()` — used *only* for matching `_LOOPBACK_HOSTNAMES`. Identity-matching form.
- `lookup_host = hostname.strip('[]')` — the exact host as returned by `urllib.parse` (brackets stripped defensively). Used for `ipaddress.ip_address()` and for the `getaddrinfo()` call, so the SSRF check sees what the HTTP client will see.

Existing behavior is preserved: `http://localhost./`, `LOCALHOST./`, and `http://localhost.:3000/` still fail closed through the canonical match (covered by the existing `test_rejects_localhost` parametrization).

Added two regression tests that pin the boundary in `tests/unit/mcp/test_types.py`:

- `test_dns_lookup_uses_unnormalized_host_for_trailing_dot` — asserts that when a non-loopback host with a trailing dot falls through to DNS (e.g. `http://127.0.0.1.nip.io./`), the `getaddrinfo()` call receives `"127.0.0.1.nip.io."` verbatim (not the stripped form).
- `test_canonical_match_does_not_short_circuit_non_loopback_aliases` — asserts DNS is still attempted for non-loopback canonical hosts, so a future collapse of the two strings into a single normalized value is caught.

Verification from the worktree:

```
uv run ruff format --check . && uv run ruff check .          # clean
uv run mypy src/ouroboros/mcp/types.py                        # clean
PYTHONPATH=src uv run pytest tests/unit/mcp/test_types.py -q  # 67 passed
```

PR #449, opened 2026-04-17
Batch capability policy audit events

Q00:
## Summary

- Add `policy.capabilities.evaluated` as a batched audit event carrying all per-capability decisions
- Keep the existing single-capability event helper for compatibility/tests
- Change runner policy emission from N appends per catalog to one append per policy evaluation

## Rationale

Per-capability events are maximally explicit but create avoidable event-store write amplification as MCP catalogs grow. A single batch event preserves the same decision details while keeping startup/audit overhead bounded.

## Test plan

- `uv run ruff check src/ouroboros/orchestrator/events.py src/ouroboros/orchestrator/runner.py tests/unit/orchestrator/test_events.py tests/unit/orchestrator/test_runner.py`
- `uv run mypy src/ouroboros/orchestrator/events.py src/ouroboros/orchestrator/runner.py`
- `uv run pytest tests/unit/orchestrator/test_events.py tests/unit/orchestrator/test_runner.py -q`


Q00:

PR #450, opened 2026-04-17
Support user capability overrides

Q00:
## Summary

- Load attached-tool capability overrides from `~/.ouroboros/tool_capabilities.yaml`
- Support `OUROBOROS_TOOL_CAPABILITIES` for an explicit override path
- Match overrides by stable id, source/tool key, original name, or normalized tool name
- Add coverage for correcting an attached browser tool to read-only semantics

## Rationale

The default attached MCP classifier is intentionally conservative, but keyword inference misclassifies tools such as browser navigation or screenshots. This lets users correct local MCP capability semantics without code changes while keeping the safe default.

## Test plan

- `uv run ruff check src/ouroboros/orchestrator/capabilities.py tests/unit/orchestrator/test_capabilities.py`
- `uv run mypy src/ouroboros/orchestrator/capabilities.py`
- `uv run pytest tests/unit/orchestrator/test_capabilities.py -q`


Q00:

PR #451, opened 2026-04-17
Select role policy by origin and scope

Q00:
## Summary

- Extend role capability profiles with origin/scope/stable-id selectors
- Keep existing built-in tool names as explicit compatibility selectors
- Allow provider-native and future-runtime read-only capabilities without adding them to name whitelists
- Cover provider-native allowance and attached-MCP exclusion in policy tests

## Rationale

Name whitelists do not scale to provider-native and future-runtime capabilities. The policy plane should be able to admit capabilities based on semantic provenance and scope while preserving the conservative built-in envelope.

## Test plan

- `uv run ruff check src/ouroboros/orchestrator/policy.py tests/unit/orchestrator/test_policy.py`
- `uv run mypy src/ouroboros/orchestrator/policy.py`
- `uv run pytest tests/unit/orchestrator/test_policy.py -q`


Q00:

PR #452, opened 2026-04-17
Consume control-plane hints in batch scheduling

Q00:
## Summary

- Build control-plane hints from the executable tool surface inside `ParallelACExecutor`
- Serialize same-batch AC execution when executable tools require `serialized` or `isolated` execution modes
- Add a regression test proving `Read` + `Edit` batches execute one AC at a time

## Rationale

#353 made control-plane hints visible, but the scheduler did not consume them. This is the narrowest enforcement step: when the tool envelope contains write/isolated capabilities, the executor keeps stage boundaries but avoids fanout inside that batch.

## Test plan

- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py`
- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py`
- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q`
- `uv run pytest tests/unit/orchestrator -q`


Q00:

PR #453, opened 2026-04-18
fix(evaluate): refine node-* presets against package.json before running

Q00:
## Summary

- Stage 1 ran `npm run lint` / `npm test` / `npm run build` blindly for every node project. When `scripts.lint` pointed to `eslint` without a config (exit 254) or `scripts.test` pointed to `vitest` without the dep, evaluation rejected perfectly good code for a tooling mismatch — not a real defect.
- Stage 1 is supposed to be a zero-cost, zero-false-positive gate. This PR makes node preset detection honor the actual manifest so misconfigured scripts downgrade to **skip**, not **fail**.

## Behavior

After language detection, `_refine_node_preset` inspects `package.json`:

| Check | Kept when… | Skipped when… | Fallback |
|---|---|---|---|
| lint | `scripts.lint` exists AND referenced tool (eslint/biome) has a config file (or `eslintConfig` in package.json) | script missing; tool config missing | — |
| test | `scripts.test` exists AND referenced runner (vitest/jest/mocha/ava/playwright) is in deps | script missing; npm "no test specified" stub; runner token without dep | — |
| build | `scripts.build` exists | — | `npx --no-install tsc --noEmit` when `tsconfig.json` exists |

`.ouroboros/mechanical.toml` remains the authoritative escape hatch (covered by a test).

## Test plan

- [x] `uv run pytest tests/unit/evaluation/` — 267 passed
- [x] `uv run pytest tests/ --ignore=tests/unit/mcp --ignore=tests/integration/mcp --ignore=tests/e2e` — 3963 passed, 2 skipped (pre-existing)
- [x] `uv run mypy src/ouroboros/evaluation/languages.py` — clean
- [x] `uv run ruff check` + `ruff format --check` — clean
- [x] 17 new unit tests cover: empty scripts, eslint with/without config (flat + legacy `.eslintrc` + `eslintConfig`), biome with/without config, vitest/jest without dep, npm stub, `node --test` without a dep requirement, tsconfig build fallback, all four package managers (npm/pnpm/yarn/bun), malformed `package.json`, `.ouroboros/mechanical.toml` still overrides refinement.

### DIFF
--- a/src/ouroboros/evaluation/languages.py
+++ b/src/ouroboros/evaluation/languages.py
@@ -10,8 +10,10 @@ Usage:
 """
 
 from dataclasses import dataclass
+import json
 import os
 from pathlib import Path
+import re
 import shlex
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Q00:
Closing in favor of a structural fix.

This PR only addressed the Node subset of a broader class of bug: **hardcoded language presets in `evaluation/languages.py` produce phantom Stage 1 failures** whenever a project's real build/test/lint contract diverges from the preset (stale scripts, missing runners, monorepo layouts, polyglot stacks, custom Taskfiles, etc.).

Rather than keep refining per-ecosystem heuristics, we will:

1. **Delete `languages.py` entirely** — no hardcoded presets.
2. **One-shot AI detect** reads the repo and writes `.ouroboros/mechanical.toml` — each command validated against an existing entry point (package.json script, Makefile target, binary on PATH). Hallucinated commands are dropped.
3. **`.ouroboros/mechanical.toml` becomes the sole contract.** Stage 1 reads it and runs deterministically. Missing toml → auto-detect once on first run; if detect fails, Stage 1 skips (never phantom-fails).
4. **`ooo brownfield`** exposes detect explicitly so users can review/edit the toml before Stage 1 trusts it.

This reviewer's design note (build fallback may not exist; test substring matching too broad) becomes moot — we delete both code paths.

PR #462, opened 2026-04-21
fix(build): remove redundant opencode plugin force-include causing wheel duplicates

Q00:
## Summary
- PyPI dev-publish has been failing (e.g. [run 24700587826](https://github.com/Q00/ouroboros/actions/runs/24700587826)) with `400 Invalid distribution file. ZIP archive not accepted: Duplicate filename in local headers.`
- Root cause: `src/ouroboros/opencode/plugin/` is shipped twice — once via `packages = ["src/ouroboros"]` (hatchling treats it as a package because of `__init__.py`) and again via an explicit `force-include` pointing to the same wheel target. Every file in that directory ended up with two ZIP local headers, which PyPI now rejects.
- Fix: drop the redundant `force-include` line. The `skills` entry stays because that directory lives at the repo root, outside `src/ouroboros/`.

## Verification
Built the wheel locally before and after:

| | duplicate local headers | opencode/plugin files shipped |
|---|---|---|
| before | 6 files x 2 | 6 |
| after | 0 | 6 |

```
$ uv build --wheel
Successfully built dist/ouroboros_ai-0.28.9.dev78-py3-none-any.whl
# no "Duplicate name" warnings
Local-header duplicates: {}
Total local-header entries: 309
```

All six files (`__init__.py`, `opencode-plugin.d.ts`, `ouroboros-bridge.ts`, `ouroboros-bridge.test.ts`, `package.json`, `tsconfig.json`) remain in the wheel at `ouroboros/opencode/plugin/`.

## Test plan
- [x] `uv build --wheel` emits no `Duplicate name` warnings
- [x] Parsed ZIP local headers — 0 duplicates
- [x] All opencode plugin assets still present in the wheel
- [ ] Confirm Dev Publish workflow succeeds after merge

### DIFF
--- a/.github/workflows/dev-publish.yml
+++ b/.github/workflows/dev-publish.yml
@@ -70,6 +70,18 @@ jobs:
           print(f'Runtime version: {v}')
           assert 'dev' in v or 'unknown' not in v, f'Unexpected version: {v}'
           "
+          # Packaging contract: opencode bridge assets must be reachable via
+          # importlib.resources in the installed wheel. Guards against
+          # regressions where hatch include rules stop shipping non-Python
+          # assets (see PR #462).
+          python -c "
+          from importlib.resources import files
+          root = files('ouroboros.opencode.plugin')
+          required = ['ouroboros-bridge.ts', 'package.json', 'tsconfig.json']
+          missing = [n for n in required if not (root / n).is_file()]
+          assert not missing, f'Missing opencode plugin assets in wheel: {missing}'
+          print('opencode plugin assets present:', required)
+          "
 
       - name: Publish to PyPI
         if: steps.check.outputs.skip == 'false'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,6 @@ packages = ["src/ouroboros"]
 
 [tool.hatch.build.targets.wheel.force-include]
 "skills" = "ouroboros/skills"
-"src/ouroboros/opencode/plugin" = "ouroboros/opencode/plugin"
 
 [dependency-groups]
 dev = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,14 +52,18 @@ version-file = "src/ouroboros/_version.py"
 exclude = [
     "/.smoke-home",
     "**/node_modules",
-    # opencode bridge assets are shipped via force-include below so the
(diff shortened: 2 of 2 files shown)


Q00:
Thanks for the catch, @ouroboros-agent — the implicit-inclusion reliance was a real hole. Pushed `0cc07351`:

1. **Restored the explicit `force-include` for `src/ouroboros/opencode/plugin`** so the `importlib.resources` contract is declared in the packaging config, not inferred from hatchling defaults.
2. **Added `src/ouroboros/opencode/plugin/**` to `[tool.hatch.build].exclude`** so the `packages=` sweep no longer emits its own copy. Hatch's force-include paths bypass `exclude`, so every asset ships exactly once — no duplicate ZIP local headers.
3. **Added a packaging smoke check to `dev-publish.yml`**: after `uv pip install dist/*.whl`, assert that `ouroboros-bridge.ts`, `package.json`, and `tsconfig.json` resolve via `importlib.resources.files("ouroboros.opencode.plugin")`. This turns the release-only regression path you flagged into a CI failure.

Verified locally in an isolated `uv run --python 3.12 --with dist/*.whl` env — the assertion passes and the wheel has 0 duplicate local headers.

Q00:
Good catch on the wheel/sdist asymmetry, @ouroboros-agent. Fixed in `daf172d6`:

- Moved `src/ouroboros/opencode/plugin/**` out of the **global** `[tool.hatch.build].exclude` and into **`[tool.hatch.build.targets.wheel].exclude`**. The exclude now scopes only to the wheel's `packages=` sweep, where it's needed to prevent duplicate local headers.
- The sdist keeps the files at their natural `src/ouroboros/opencode/plugin/` path, so a downstream rebuild-from-sdist reproduces the same wheel via this same config.

Verified in a clean worktree with `uv build`:

| artifact | opencode/plugin files | duplicate local headers |
|---|---|---|
| wheel | 6 | 0 |
| sdist | 6 (under `src/`) | n/a |

Also ran `uv run --with dist/*.tar.gz python -c "..."` — the sdist-driven wheel rebuild passes the `importlib.resources.files("ouroboros.opencode.plugin")` assertion for `ouroboros-bridge.ts`, `package.json`, and `tsconfig.json`.

PR #464, opened 2026-04-21
fix(mcp): cancel linked job runner tasks

Q00:
## Summary

- Fixes #341 by making `ouroboros_cancel_job` cancel the in-process job task even when the job is linked to an orchestration session.
- Preserves the existing session cancellation signal while ensuring `CancelledError` reaches runtime subprocess cleanup immediately.
- Adds an integration regression test with a real child process so cancelled jobs cannot leave CLI subprocesses alive.

## Changes

- `JobManager.cancel_job()` now always cancels tracked job/runner tasks after persisting `cancel_requested`.
- Linked session jobs still call `request_cancellation(session_id)` for runner/event-store visibility.
- New integration test starts a Python subprocess through a linked job and asserts it exits after `cancel_job()`.

## Tests

- `uv run pytest tests/integration/test_cancel_subprocess_termination.py`
- `uv run pytest tests/unit/mcp/test_job_manager.py::TestJobManager::test_cancel_job_cancels_non_session_task`

### DIFF
--- a/src/ouroboros/mcp/job_manager.py
+++ b/src/ouroboros/mcp/job_manager.py
@@ -11,7 +11,7 @@ from typing import Any
 from uuid import uuid4
 
 from ouroboros.events.base import BaseEvent
-from ouroboros.orchestrator.runner import request_cancellation
+from ouroboros.orchestrator.runner import clear_cancellation, request_cancellation
 from ouroboros.orchestrator.session import SessionRepository
 from ouroboros.persistence.event_store import EventStore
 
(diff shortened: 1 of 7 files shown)


Q00:
Addressed the latest review finding on the linked cancellation path.

Changes made:
- linked-session cancellation now only persists `orchestrator.session.cancelled` when the session could be reconstructed successfully
- before writing terminal cancellation, `cancel_job()` replays the latest session state again and fails instead of synthesizing a cancellation if that replay fails
- local task teardown still happens before the persisted terminal-state decision, so the background job is stopped without corrupting the event-sourced session state under partial reads
- added regression coverage for both initial reconstruct failure and latest reconstruct failure before persistence

Local checks:
- `uv run ruff format --check src/ouroboros/mcp/job_manager.py tests/unit/mcp/test_job_manager.py`
- `uv run ruff check src/ouroboros/mcp/job_manager.py tests/unit/mcp/test_job_manager.py`
- `uv run pytest tests/unit/mcp/test_job_manager.py tests/unit/orchestrator/test_inflight_cancellation.py tests/unit/orchestrator/test_runner.py`

CI is green on the latest push. Please re-review when you get a chance.

Issue #472, opened 2026-04-22
Introduce unified Directive vocabulary for the control plane


shaun0927:
## Context

Depends on the framing in #471 (control-plane RFC).

Today, "what happens next" is decided independently at each site:

- `src/ouroboros/evaluation/` — decides pass/fail and whether to trigger evolve
- `src/ouroboros/evolution/` — decides continue / converge
- `src/ouroboros/resilience/` — decides retry / unstuck persona
- `src/ouroboros/orchestrator/` — decides phase routing
- `src/ouroboros/observability/` — signals drift but does not emit a decision

Each site has its own enum / boolean / string literals. There is no shared type that captures "a control-plane decision."

## Problem

Adding a new workflow or a new decision source (e.g., the planned dynamic bridge server addition from #280) requires the author to:

1. Invent a new ad-hoc signal at their site.
2. Teach every downstream consumer to recognize it.
3. Reproduce drift/cancel/compact semantics manually.

This is O(decision_sites × workflows) coupling, and it's the root cause of the O(handlers) plumbing debt observed after #279 / #280.

## Proposal

Introduce a first-class `Directive` type in `src/ouroboros/core/`:

```python
class Directive(StrEnum):
    CONTINUE  = "continue"   # proceed with current plan
    EVALUATE  = "evaluate"   # hand off to evaluation pipeline
    EVOLVE    = "evolve"     # emit next-generation seed proposal
    UNSTUCK   = "unstuck"    # invoke lateral persona
    RETRY     = "retry"      # re-execute last unit
    COMPACT   = "compact"    # compress context before continuing
    WAIT      = "wait"       # block on external input
    CANCEL    = "cancel"     # terminate this execution
    CONVERGE  = "converge"   # terminal success (ontology similarity ≥ threshold)
```

(Final set to be agreed in review — this is a starting point, not final.)

Each directive has documented:

- **Precondition** — what state the system must be in to emit it
- **Effect** — what the consumer must do
- **Terminal?** — whether it ends the run (`CONVERGE`, `CANCEL` only)

Invariants to enforce in tests:

- At most one terminal directive per execution
- `WAIT` must be paired with an external-input contract
- `RETRY` must respect existing retry budget in `resilience/`
- `COMPACT` must not lose event lineage

## Migration strategy

Non-breaking, one site at a time:

1. Land the type + invariants with no callers.
2. Convert one decision site (suggest `evaluation/` terminal branches) as a reference.
[...]

Issue #471, opened 2026-04-22
[RFC] Control plane vs Capability plane: framing the post-#280 scaling path


shaun0927:
## Context

PR #280 (MCPBridge) shipped a lifecycle-managed server-to-server MCP layer and, together with #279, introduced `mcp_manager` plumbing into `ExecuteSeedHandler`. The bridge unlocks external MCP tools for the child agent during `execute_seed`.

However, `docs/guides/mcp-bridge.md` explicitly records two limitations:

> - Evolution loop (`evolve_step`) does not yet pass the bridge manager
> - No dynamic server addition after initial connection

These are not incidental — they point to a structural property of how decisions currently flow through the system. This RFC proposes a shared vocabulary for reasoning about that property before we open implementation issues.

## Two planes, not one

I find it useful to separate Ouroboros's extensibility surface into two planes:

| Plane | What it extends | Post-#280 status |
|---|---|---|
| **Capability plane** | The set of tools / actions a child agent can invoke | Well-factored. `MCPBridge → MCPClientManager` owns lifecycle, discovery, and tool merging. |
| **Control plane** | The set of decisions the workflow takes next (`evaluate`, `evolve`, `unstuck`, `ralph`, `retry`, `cancel`, `compact`, `converge`, …) | No central authority. Decisions are distributed across `orchestrator/`, `evaluation/`, `evolution/`, `resilience/`, `observability/`. |

#280 is a capability-plane advance. The Known Limitations above are control-plane symptoms: each decision site (`evolve_step`, `unstuck`, `ralph`, future workflows) needs its own explicit plumbing of `mcp_manager`, because there is no bus-like entity a decision site can subscribe to.

## Why this matters for future workflows

Consider introducing a workflow that does not match the interview → seed → Double-Diamond → 3-stage-eval → Wonder/Reflect shape — e.g., a propose/run/adjudicate experiment loop, or a generator/evaluator adversarial loop. Today that requires touching `core/seed`, `execution/`, `evaluation/`, `evolution/`, `resilience/`, and adding an MCP surface to each. The *capability* side is already clean (bridge just works); the *control* side scales as O(handlers × workflows).

[...]

Issue #476, opened 2026-04-22
[RFC] Agent OS runtime contract: capabilities, policy, directives, and replayable agent processes


Q00:
## Context

This follows the thread from #353, #471, and #474.

- #353 taught the runtime engine what tools mean before it schedules them.
- #471 framed the gap between the capability plane and the control plane.
- #474 proposes replacing per-handler `mcp_manager` plumbing with a shared bus-like abstraction.

The direction feels right, but I want to sharpen the architectural framing before we commit to names and topology.

The larger idea is that Ouroboros is starting to become less like one more agent harness, and more like the execution substrate underneath several harnesses, runtimes, workflows, and MCP surfaces.

That is the sense in which I would use the phrase **Agent OS**.

Not an operating system in the literal sense, and not a generic agent framework for every workflow. More specifically:

> An Agent OS is the layer that turns non-deterministic agent behavior into a replayable, observable, policy-bound execution contract.

Ouroboros should keep its product identity: spec-first execution, acceptance criteria, evaluation, and evolve-until-converge. But underneath that workflow, it can expose a small set of runtime primitives that other agent workflows can share.

## Core framing

The thin-harness vs fat-skill debate mostly answers a vertical question:

> How should one model, one harness, and one skill interact?

Ouroboros is beginning to face a horizontal question:

> What happens when multiple agents, multiple harnesses, multiple runtimes, and multiple MCP servers need to coexist inside the same execution environment?

At that point, the important layer is no longer just the harness. It is the shared runtime contract underneath the harnesses.

## Proposed Agent OS layers

```mermaid
flowchart TD
    A[External tools / MCP servers / provider-native tools / skills]
    B[Capability Layer<br/>What can this environment do?]
    C[Policy Layer<br/>Who may use which capability?]
    D[Directive Layer<br/>What should happen next?]
    E[Event Journal<br/>Why did the system move?]
    F[Agent Processes<br/>execute_seed / evolve_step / unstuck / ralph / future workflows]

    A --> B
    B --> C
    C --> D
    D --> E
    E --> F
    F --> D
```

### Capability Layer

The capability layer answers: **what can this agent environment do?**

This is where #353 matters. Tools should not be passed around as plain strings. They need engine-owned meaning:

- read-only vs mutating
- safe vs serialized vs isolated
- interruptible vs non-interruptible
- default vs elevated approval
- builtin vs attached MCP vs provider-native vs future runtime
- kernel vs sidecar vs attachment vs shell-only scope

[...]

PR #477, opened 2026-04-22
feat(core): introduce Directive StrEnum for control-plane decisions

shaun0927:
## Summary

- Introduce a shared `Directive` `StrEnum` for control-plane decisions (`continue`, `evaluate`, `evolve`, `unstuck`, `retry`, `compact`, `wait`, `cancel`, `converge`).
- Pure addition — no decision site is modified in this PR. Subsequent migrations will land as separate, independently reviewable PRs.

## Motivation

First step of #472. Post-#280, `docs/guides/mcp-bridge.md` records two Known Limitations: `evolve_step` not receiving the bridge manager, and no dynamic upstream-server addition. #471 frames these as **control-plane** symptoms — decision sites across `evaluation/`, `evolution/`, `resilience/`, `orchestrator/`, and `observability/` each emit ad-hoc signals (enums, booleans, literal strings) with no shared vocabulary, which is why new plumbing is required on every handler added post-#280.

#472 proposes introducing that shared vocabulary. This PR lands only the type and its invariants, with no caller modifications, matching the surgical-PR pattern established by recent merges (e.g., #470, #438, #444). Downstream migrations of individual decision sites become their own review units.

## Changes

- **`src/ouroboros/core/directive.py`** — new `Directive` `StrEnum`. Each member has a short docstring stating precondition and effect. Terminality is exposed through `is_terminal`, backed by a module-level `_TERMINAL_DIRECTIVES` `frozenset`. Exactly two terminal members: `CANCEL`, `CONVERGE`.
- **`src/ouroboros/core/__init__.py`** — lazy re-export of `Directive` alongside `Result` and the core type aliases, preserving the existing lazy-loader pattern used for every other core symbol.
- **`tests/unit/core/test_directive.py`** — 10 unit tests covering:
  - string-value stability (lowercased names, uniqueness, `StrEnum` substitutability)
  - terminal-set closure (exactly `CANCEL` and `CONVERGE`; everything else is non-terminal)
  - required vocabulary membership (guards accidental rename/removal)
  - `Directive(value)` round-trip
  - lazy re-export through `ouroboros.core`

## Not changed

No existing decision site is migrated in this PR. Migration is deferred so each call site becomes its own small, reviewable change:

- `evaluation/` terminal branches — follow-up PR
- `evolution/` convergence / `resilience/` stagnation handlers — follow-up PRs
- `orchestrator/` routing — follow-up PR

## Verification

```
uv run ruff check src/ouroboros/core/directive.py src/ouroboros/core/__init__.py tests/unit/core/test_directive.py
uv run ruff format --check src/ouroboros/core/directive.py src/ouroboros/core/__init__.py tests/unit/core/test_directive.py
uv run pytest tests/unit/core/ -q
```

Result: `296 passed` (10 new), ruff clean, format clean.

## References

- #471 (RFC — control plane vs capability plane framing)
- #472 (parent issue — unified Directive vocabulary)
- #280 (source of the Known Limitations this vocabulary is designed to address over follow-up PRs)

### DIFF
--- a/src/ouroboros/core/__init__.py
+++ b/src/ouroboros/core/__init__.py
@@ -16,6 +16,8 @@ _EXPORTS: dict[str, tuple[str, str]] = {
     "EventPayload": ("ouroboros.core.types", "EventPayload"),
     "CostUnits": ("ouroboros.core.types", "CostUnits"),
     "DriftScore": ("ouroboros.core.types", "DriftScore"),
+    # Control-plane directive vocabulary
+    "Directive": ("ouroboros.core.directive", "Directive"),
     # Errors
     "OuroborosError": ("ouroboros.core.errors", "OuroborosError"),
     "ProviderError": ("ouroboros.core.errors", "ProviderError"),
--- a/src/ouroboros/core/directive.py
+++ b/src/ouroboros/core/directive.py
@@ -1,20 +1,46 @@
-"""Directive vocabulary for control-plane decisions.
+"""Directive vocabulary for the Phase 2 control plane.
 
-This module defines the shared type used to describe "what should happen next"
-at any decision site across the Ouroboros workflow. Decision sites currently
-distributed across `evaluation/`, `evolution/`, `resilience/`, `orchestrator/`,
-and `observability/` will be migrated to this vocabulary in follow-up changes;
-this module adds only the type and its metadata, with no caller modifications.
+This module defines the shared value type that every decision site uses to
+describe "what should happen next." In the Agent OS framing introduced by
+the Phase-2 RFC, ``Directive`` members act as runtime-level syscalls: a
+small, stable alphabet through which decision sites (evaluator, evolver,
+resilience handlers, orchestrator, job manager) express control flow
+without each inventing its own signalling.
 
-Design notes:
-- The enum is additive. Each member has a deliberate precondition and effect.
-- Exactly two members are terminal: ``CANCEL`` and ``CONVERGE``. Every other
-  member implies the run continues.
-- The vocabulary is intentionally small. New directives are only added when an
-  existing one cannot cover the semantics without loss of meaning.
+Positioning within the Agent OS layers described by the RFC:
 
-See the control-plane RFC and the Directive introduction issue for rationale
-and the planned migration path.
+- Capability layer   — answers *what can this environment do?*
+- Policy layer       — answers *who may use which capability?*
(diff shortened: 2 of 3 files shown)


shaun0927:
Aligning with #476 (Phase 2 framing): this PR is in scope and unchanged. The `Directive` type as defined here matches the vocabulary enumerated in that RFC. Placement in `core/` is the intended Phase 2 home per my answer at #476#issuecomment-4294892689 (Q2) — relocatable via a one-line lazy re-export change if a dedicated `runtime/` module is introduced later.

No code change needed. `ouroboros-agent` APPROVED, all CI green.

shaun0927:
Correction to my earlier alignment note — `ouroboros-agent` APPROVED covers style and diff-scoped correctness, not conceptual alignment with #476. Re-reading the module docstring against the Phase 2 RFC surfaced two gaps:

1. **Framing**: `Directive` was described as "control-plane decisions" but not positioned inside the Agent OS layer stack. The RFC treats directives as runtime-level syscalls in the Control Layer; the docstring now says so, next to the other four layer names so future readers see where it sits.
2. **Migration posture**: the RFC explicitly calls for mapping existing enums (`StepAction` example) rather than replacing them. That stance was implicit in the PR body but missing from the module itself; the docstring now records the reference `StepAction → Directive` mapping at the type's site of truth.

New commit: `1606c6e docs(core): tie Directive module to Phase 2 control-plane framing` — docstring only, no behavior change, existing tests pass (10/10).

Q00:
@shaun0927 LGTM on the Agent OS direction.

Re-reading #477 against the Agent OS / Phase 2 framing, I think the shape is right:

- `Directive` belongs in `core/` for now as a small shared runtime value type, and can move later if/when a real `runtime/` layer earns its keep.
- The enum is correctly scoped to workflow control, not capability policy.
- The incremental migration posture is the right one: existing local enums like `StepAction` should map into `Directive` at boundaries rather than being replaced in one broad rewrite.
- This sets up #478 cleanly: directives become the shared vocabulary, while `control.directive.emitted` becomes the replayable decision-journal primitive.

One non-blocking cleanup before merge would be nice: the module docstring markdown for the `Directive layer` bullet has broken emphasis markup:

```text
- **Directive layer  — answers what should happen next?*   (this module)*
```

No runtime or lint impact, but it would be cleaner to fix before merging since this docstring is carrying the Agent OS framing.

Q00:
let's make a new PR to follow up the non blocking cleanup!

PR #485, opened 2026-04-22
feat(mcp): add streamable HTTP serve transport

andrew-adamson:
## Summary
- add streamable HTTP as a supported `ouroboros mcp serve` transport
- update the CLI and server adapter to wire the new transport through cleanly
- cover the transport behavior with focused docs and tests

## Testing
- `uv run pytest tests/unit/cli/test_mcp_validate_transport_stderr.py tests/unit/mcp/server/test_adapter.py`

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -627,7 +627,7 @@ ouroboros mcp serve [OPTIONS]
 |--------|-------------|
 | `-h, --host TEXT` | Host to bind to (default: localhost) |
 | `-p, --port INTEGER` | Port to bind to (default: 8080) |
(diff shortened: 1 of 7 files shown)


shaun0927:
Thanks for the clean slice here — the direction makes sense, and it lines up with the recent MCP transport expansion work.

Before merge, I want to tighten the serve-side contract a bit more:

1. Please make the user-facing contract explicit for `streamable-http`:
   - the exact endpoint path clients should use,
   - why `http` is still only a client-side alias and not a serve transport,
   - and where startup messages go for network transports.

2. Please add or strengthen one user-visible contract test around the advertised network address/path, not just transport validation.

3. Since transport work in this repo tends to become security/ops surface area very quickly, please make sure the docs call out any FastMCP-specific limitations/caveats clearly.

The scope is good as-is; I just want the docs + runtime contract to be a bit sharper before this lands.

Q00 (approved):
LGTM

PR #486, opened 2026-04-22
fix(brownfield): scan repos with any origin remote

andrew-adamson:
## Summary
- broaden brownfield repository scanning so repos with any configured `origin` remote are considered
- align setup and brownfield MCP handling with the updated scan behavior
- add focused coverage for the scan and handler paths

## Testing
- `uv run pytest tests/unit/bigbang/test_brownfield.py tests/unit/mcp/tools/test_brownfield_handler.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ Inside AI coding agent sessions, use `ooo <cmd>` skills. From the terminal, use 
 | `ooo pm`         | *(via MCP)*                                                       | PM-focused interview + PRD generation                        |
 | `ooo qa`         | *(via skill)*                                                     | General-purpose QA verdict for any artifact                  |
 | `ooo update`     | `ouroboros update`                                                | Check for updates + upgrade to latest                        |
-| `ooo brownfield` | *(via skill)*                                                     | Scan and manage brownfield repo defaults                     |
+| `ooo brownfield` | *(via skill)*                                                     | Scan and manage brownfield repo/worktree defaults            |
 | `ooo publish`    | *(skill/runtime surface; uses `gh` CLI)*                          | Publish a Seed as GitHub Epic/Task issues for team workflows |
 
 > Not all skills have direct CLI equivalents. Some (`evaluate`, `evolve`, `unstuck`, `ralph`, `publish`) are available through agent skills, runtime rules, or MCP tools rather than a direct `ouroboros <subcommand>` shell command.
--- a/docs/api/mcp.md
+++ b/docs/api/mcp.md
@@ -261,6 +261,25 @@ Ouroboros exposes workflow tools as ordinary MCP handlers. The MCP server does
 not parse `ooo` syntax and does not receive channel-specific identifiers for
 skill routing.
 
+### `ouroboros_brownfield` Scan Boundaries
+
+The brownfield MCP tool registers existing codebases for PM/interview context.
+For `{"action": "scan"}`, `scan_root` controls only the filesystem walk for
+seed repositories and worktrees. When `scan_root` is omitted, the walk starts at
+the current user's home directory. Dot-prefixed directories and known noisy
+directories such as `node_modules` are not walked as seed locations.
+
+Linked worktree expansion is asymmetric. When the walk finds a normal repo root
+with a `.git` directory, Ouroboros runs `git worktree list --porcelain` from
+that repo and may register Git-reported linked worktrees even when those paths
+are outside `scan_root`. When the walk finds a linked worktree root with a
+`.git` file, Ouroboros registers that worktree itself but does not use it to
+register the main worktree or sibling worktrees outside `scan_root`. This keeps
+narrow scans scoped when a user intentionally passes one worktree as AI context.
+Local repos, repos without remotes, and repos whose remotes are not named
+`origin` are all eligible. Stale linked worktree paths are skipped when the path
+no longer exists or is no longer a valid Git working tree.
+
 Runtime integrations that support `ooo <skill>` or `/ouroboros:<skill>` use the
 shared stateless router in `ouroboros.router` before invoking MCP. The router
 loads packaged `SKILL.md` frontmatter, validates `mcp_tool` and `mcp_args`,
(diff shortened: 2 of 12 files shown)


shaun0927:
I understand the motivation here. Broadening brownfield scan from "GitHub origin only" to "any repo with an `origin` remote" is a reasonable direction for local-first brownfield UX.

That said, I don't think this is merge-ready yet because the contract is still blurry in two places:

1. `_has_github_origin()` no longer matches its own name/meaning.
   If the new policy is "any configured origin remote," then please keep the legacy helper GitHub-specific and migrate callers to a new `_has_origin_remote()` predicate instead of changing the semantics under the old name.

2. The MCP/tool copy now says `scan` discovers repos/worktrees, but the actual scan path still centers on directories that literally contain a `.git` entry. If we want to claim worktree support here, the implementation/tests need to cover that explicitly; otherwise the contract text should stay narrower.

The direction is fine. I just want the naming, docs, and real behavior to line up cleanly.

Q00 (changes requested):
Thanks for tightening the brownfield scan behavior here. I like the direction of moving away from a GitHub-only assumption; for Agent OS, brownfield discovery should work with the repos people actually have locally, not only with GitHub-hosted origins.

I would hold this before merging though. Two things need to be clarified/fixed first:

1. The helper naming now hides an important semantic change. `_has_github_origin()` no longer means GitHub origin, it means any `origin` remote. I think we should keep the GitHub-specific helper honest and introduce/use a separate `_has_origin_remote()` for the new behavior.

2. The MCP-facing language says this scans repos/worktrees, but the scanner still only detects a `.git` directory. Git worktrees usually have a `.git` file, so they will be silently skipped. That is especially important for Agent OS workflows where local worktrees are a normal way to run multiple branches/agents side by side.

Also Ruff is currently failing on import ordering, so that needs a quick cleanup too.

Requesting changes mainly to keep the brownfield contract precise: either support `.git` file worktrees with tests, or narrow the wording so we do not promise worktree discovery yet.

andrew-adamson:
Updated this PR after the review discussion to make the brownfield scan boundary explicit and less surprising.

Scope changes:
- `scan_root` is now the documented boundary for filesystem walking. When omitted, it defaults to the current user's home directory.
- The scanner now accepts any valid Git repo/worktree under `scan_root`, including local repos with no remotes and repos whose remote is not named `origin`.
- Normal repo roots with a `.git` directory still expand through `git worktree list --porcelain`, so Git-reported linked worktrees can be registered even outside `scan_root`.
- Linked worktree roots with a `.git` file are registered self-only; they no longer pull in the main worktree or sibling worktrees outside `scan_root`.
- `ouroboros setup scan` now accepts an optional `[SCAN_ROOT]` and passes that root explicitly through the scan pipeline.
- Docs, packaged skill docs, MCP parameter descriptions, CLI help/docstrings, and schema comments now describe the same scan boundaries.

Why:
- Local-only repositories and non-`origin` remotes are valid brownfield context, so requiring an `origin` remote was too narrow.
- Git-linked worktrees may live outside the scanned directory, so normal repo roots should use Git metadata rather than relying only on directory walking.
- A linked worktree can be intentionally passed to AI as a narrow context boundary. Treating it as a portal to the main repo and sibling worktrees can unexpectedly include projects the user meant to keep out of scope.
- The docs now distinguish filesystem discovery from Git-reported worktree expansion so users can predict what a scan may register.

Validation run after rebasing on `upstream/main`:
- `uv run ruff check .`
- `git diff --check`
- `uv run pytest tests/unit/cli/test_setup.py::TestScanAndRegisterRepos tests/unit/cli/test_setup.py::TestScanRegisterPipeline tests/unit/cli/test_setup.py::TestScanCommand tests/unit/bigbang/test_brownfield.py tests/unit/mcp/tools/test_brownfield_handler.py`
- `uv run mypy src/ouroboros`

shaun0927:
Hi @andrew-adamson — to keep momentum on the latest reviews, I went ahead and prepared a small fixup branch addressing the two remaining items, in case it's useful. No expectation that you take it as-is; happy to defer entirely if you prefer to land it your way.

Branch: [`shaun0927:pr-486-fixups`](https://github.com/shaun0927/ouroboros/tree/pr-486-fixups) (two commits on top of `24bcede`)
Compare: https://github.com/Q00/ouroboros/compare/main...shaun0927:ouroboros:pr-486-fixups

**1. `scan_and_register` no longer leaks the full DB on a narrow/empty scan** ([codex review on `24bcede`](https://github.com/Q00/ouroboros/pull/486#pullrequestreview-563559433))
- Empty filesystem walk → returns `[]`
- Non-empty walk → returns only the repos this scan discovered and upserted (filtered by `scanned_paths`)
- Updated docstring directs callers that need the full registry to `BrownfieldStore.list` directly
- Updated `test_mocked_scan_empty_returns_store_list` → `..._returns_empty_list`, added `test_mocked_scan_returns_only_scanned_repos`

**2. MCP `_handle_scan` validates `scan_root`** ([codex follow-up on `f9e7669`](https://github.com/Q00/ouroboros/pull/486#pullrequestreview-560547594))
- `Path(scan_root_str).resolve()` then reject with `MCPToolError` when the resolved path doesn't exist or isn't a directory (CLI was already covered by Typer's `exists=True, dir_okay=True`)
- `.resolve()` also closes a small symlink/TOCTOU gap so the validation and the walk see the same target
- Added `test_scan_action_rejects_nonexistent_scan_root` and `test_scan_action_rejects_file_scan_root`

Local verification on the worktree: `uv run ruff check .` ✅ · `uv run mypy src/ouroboros` ✅ · `uv run pytest tests/unit/bigbang/test_brownfield.py tests/unit/mcp/tools/test_brownfield_handler.py tests/unit/cli/test_setup.py` → 205 passed.

Use whatever's easiest:
- `git fetch https://github.com/shaun0927/ouroboros.git pr-486-fixups && git cherry-pick 948627f 65a5631`
- Or copy the patch by hand
- Or tell me to open it as a separate PR if you'd rather review it that way

Either way, thanks for the careful iteration on this one — the worktree-vs-repo-root asymmetry came out really cleanly.

shaun0927:
Pushed two follow-ups via maintainer-edit on top of `24bcede`:

- `948627f` fix(brownfield): return only scanned repos from scan_and_register — addresses the empty-scan leak from the codex review
- `65a5631` fix(brownfield): validate MCP scan_root before walking — addresses the missing input validation

Local checks: ruff/mypy/pytest (3 affected files, 205 passed) clean. Detail and rationale in the [earlier comment](https://github.com/Q00/ouroboros/pull/486#issuecomment-4345120275).

@andrew-adamson if you'd rather take this differently, feel free to amend or force-push over these — just wanted to keep the review loop tight. Thanks again for the careful work on the worktree boundary.

shaun0927:
Merged as c29aebc. Thanks @andrew-adamson — really nice work clarifying the worktree boundary semantics, and appreciate your patience through the review iterations. The brownfield scan is now meaningfully more useful for local-first workflows.

PR #487, opened 2026-04-22
fix(mcp): share brownfield store for serve --db

andrew-adamson:
## Summary
- pass a shared `BrownfieldStore` through `mcp serve --db` instead of falling back to a separate default store
- keep brownfield handler state aligned with the server database selection
- add focused coverage for injected-store reuse and initialization

## Testing
- `uv run pytest tests/unit/mcp/server/test_adapter.py::TestCreateOuroborosServerBrownfieldStore::test_injected_store_is_shared_with_handler_and_owned_by_server tests/unit/mcp/tools/test_brownfield_handler.py::TestBrownfieldHandlerDispatch::test_injected_store_initializes_once`


shaun0927:
The overall direction is good. Sharing the brownfield store across `mcp serve --db` is more consistent with the rest of the server lifecycle and avoids per-request drift.

My concern is startup-failure semantics.

Right now it looks possible for `brownfield_store.initialize()` to fail partway through startup, the server to continue running, and the handler path to later treat the injected store as effectively ready based on partial internal state. That would leave brownfield access wedged until restart instead of giving us a clean retry/fail-fast boundary.

Please tighten that before merge:

1. Track "successfully initialized" explicitly instead of inferring readiness from `_engine`-style internals, or
2. guarantee a safe retry path after failed shared-store initialization.

Please also add a regression test for the failed shared-init path so we know the next brownfield request behaves intentionally rather than inheriting a half-initialized store.

So: good direction, but I want the failure mode made safe.

Q00 (dismissed):
Thanks for pushing this forward. Sharing the brownfield store with the `mcp serve --db` database is the right direction for Agent OS: MCP tools should see the same persistent project context as the rest of the server instead of drifting into a separate local store.

I would not merge this version yet because the readiness check can wedge the brownfield store after a failed initialization.

The current handler infers readiness from `_engine` being non-null, but `BrownfieldStore.initialize()` can assign `_engine` before the full initialization succeeds. If startup initialization fails after that point and the server continues, the request handler will treat the store as ready and skip retrying initialization. In practice, brownfield MCP tools can stay broken until the process restarts.

For the Agent OS path, this needs a more explicit success state. I think either of these would be fine:

- track `store_ready` only after `initialize()` returns successfully, and leave it false on failure so the next request can retry; or
- fail fast for `--db` if the shared brownfield store cannot initialize.

Please also add a regression test for the failed shared-init path so we know the server either retries safely or stops clearly instead of running with a half-initialized store.

shaun0927:
Pushed a follow-up commit (`12a7ad3`) on this branch addressing the readiness-contract concern from @ouroboros-agent and @Q00.

**What changed**
- `BrownfieldStore.initialize()` flips `_initialized` only after schema creation and migrations succeed end-to-end; locally-created engines are disposed on failure so the next caller retries against a clean state.
- `BrownfieldHandler._get_store()` drops the `_engine`-based inference and flips `_store_ready` only on a successful `initialize()` — failed shared-init recovers on the next request.
- `_run_mcp_server` no longer swallows store-init exceptions; persistent stores must come up before the server starts serving (orphan cleanup remains best-effort).

**Regression coverage**
- `test_failed_shared_init_recovers_on_next_request`
- `test_brownfield_store_init_failure_aborts_startup`
- `test_event_store_init_failure_aborts_startup` (renamed from the old swallowing-behavior test)

Full mcp + cli + persistence suite passes locally (1298/1298). @andrew-adamson — happy to drop this if you would prefer to land the fix yourself; just say the word.

shaun0927:
@Q00 readiness contract is now explicit (`_initialized` flips only on a clean `initialize()`) plus fail-fast at startup, with regression tests for both the recovery and the abort path — ready for another look when you have a minute.

shaun0927 (approved):
Reviewed end-to-end. Original CHANGES_REQUESTED concerns are resolved:

- 12a7ad3 — `BrownfieldStore._initialized` only flips True on a fully successful `initialize()`. Failed init disposes any locally-created engine and propagates so the next caller can retry safely.
- 12a7ad3 — `BrownfieldHandler._store_ready` replaces the `_engine`-inferred readiness check, so a half-initialized shared store no longer wedges request handling.
- 12a7ad3 — `_run_mcp_server()` no longer swallows persistent-store init exceptions; event/brownfield init failures abort startup cleanly instead of running with a partially initialized store.
- 2dde69f — Regression coverage: `initialize()` rollback test at the persistence layer, recovery test at the handler layer.

CI green across Lint/MyPy/py3.12-3.14. ouroboros-agent bot APPROVE x2 on `2dde69f` and `f91d981`.

Test isolation (BrownfieldStore stub) and lazy-path concurrent-init lock have been moved to follow-up PR #507 to keep this PR focused on the readiness contract.

Approving for self-merge under write permission.

andrew-adamson:
@shaun0927 Thanks for taking over on this one - I bit off a bit more than I could chew with my initial 5x PRs at the same time

Issue #184, opened 2026-03-23
feat(config): Add reasoning_effort level to ModelConfig for per-model reasoning control


Q00:
## Summary

Currently, Ouroboros delegates reasoning effort entirely to the underlying runtime's global default. For example, the documented Codex baseline is "GPT-5.4 with medium reasoning effort" (PR #179), but this is set at the Codex CLI level — Ouroboros has no way to control or vary reasoning effort per role, per phase, or per task complexity.

This feature request proposes adding **explicit reasoning effort configuration** as an advanced option in `~/.ouroboros/config.yaml`, enabling users to set a minimum reasoning floor per model and let the router auto-select the appropriate level based on task complexity.

## Motivation

- Different phases of the Ouroboros pipeline have vastly different reasoning demands:
  - **Routine coding / Stage 1 fixes** → `medium` is sufficient
  - **Semantic evaluation / ontology analysis** → `high` or `xhigh` needed
  - **Lateral thinking / consensus** → maximum reasoning desired
- Currently all LLM calls inherit the runtime's global reasoning level (e.g., Codex CLI's `medium`), which means:
  - Over-spending on simple tasks (unnecessary deep reasoning)
  - Under-reasoning on complex tasks (shallow analysis where depth is needed)
- PR #179 documented `gpt-5.4` with `medium` reasoning as the Codex baseline, but this is a **documentation convention only** — the code never passes reasoning effort to any adapter

### What exists
- `ModelConfig(provider, model)` — model ID only, no reasoning parameter
- `CompletionConfig(model, temperature, max_tokens, ...)` — no reasoning field
- Per-role model overrides in `config.yaml` (e.g., `llm.qa_model`, `evaluation.semantic_model`, `consensus.*`) — model string only
- `model_tier` parameter in `execute_seed` MCP tool (`small/medium/large`) — this is model **size** tier, not reasoning effort

### What's missing
- No `reasoning_effort` field in `ModelConfig` or `CompletionConfig`
- No adapter-level pass-through to provider APIs
- Ouroboros completely delegates reasoning level to the runtime's global setting

### 1. Extend `ModelConfig` with reasoning bounds

```python
class ModelConfig(BaseModel, frozen=True):
    provider: str
    model: str
    # Advanced: reasoning effort bounds (optional)
    min_reasoning: Literal["low", "medium", "high", "xhigh"] = "medium"
    max_reasoning: Literal["low", "medium", "high", "xhigh"] = "xhigh"
```

[...]

PR #488, opened 2026-04-22
fix(codex): ignore user config in subprocesses

andrew-adamson:
## Summary
- add `--ignore-user-config` to Codex subprocess invocations so runtime behavior does not inherit local user config unexpectedly
- update the runtime and adapter coverage to match the new invocation shape
- keep the tests portable across path semantics used in local environments

## Testing
- `uv run pytest tests/unit/orchestrator/test_codex_cli_runtime.py tests/unit/providers/test_codex_cli_adapter.py`

### DIFF
--- a/src/ouroboros/orchestrator/codex_cli_runtime.py
+++ b/src/ouroboros/orchestrator/codex_cli_runtime.py
@@ -552,7 +552,7 @@ class CodexCliRuntime:
         prompt: str | None = None,
     ) -> list[str]:
         """Build the CLI command args.  Prompt is fed via stdin separately."""
-        command = [self._cli_path, "exec"]
+        command = [self._cli_path, "exec", "--ignore-user-config"]
 
         command.extend(
             [
--- a/src/ouroboros/providers/codex_cli_adapter.py
+++ b/src/ouroboros/providers/codex_cli_adapter.py
@@ -358,6 +358,7 @@ class CodexCliLLMAdapter:
         command = [
             self._cli_path,
             "exec",
+            "--ignore-user-config",
             "--json",
             "--skip-git-repo-check",
             "-C",
--- a/tests/unit/orchestrator/test_codex_cli_runtime.py
+++ b/tests/unit/orchestrator/test_codex_cli_runtime.py
@@ -209,13 +209,15 @@ class TestCodexCliRuntime:
             output_last_message_path="/tmp/out.txt",
         )
 
-        assert command[:2] == ["/usr/local/bin/codex", "exec"]
+        assert Path(command[0]) == Path("/usr/local/bin/codex")
+        assert command[1] == "exec"
+        assert "--ignore-user-config" in command
         assert "--json" in command
         assert "--full-auto" in command
         assert "--model" in command
         assert "o3" in command
         assert "-C" in command
-        assert "/tmp/project" in command
+        assert Path(command[command.index("-C") + 1]) == Path("/tmp/project")
 
     def test_build_command_for_resume(self) -> None:
         """Builds an exec resume command when a session id is provided."""
@@ -233,7 +235,7 @@ class TestCodexCliRuntime:
         assert command.index("--skip-git-repo-check") < resume_index
         assert command.index("--output-last-message") < resume_index
         assert command.index("-C") < resume_index
-        assert command[command.index("-C") + 1] == "/tmp/project"
+        assert Path(command[command.index("-C") + 1]) == Path("/tmp/project")
 
     def test_resolve_cli_path_falls_back_from_wrapper(self, tmp_path: Path) -> None:
         """Runtime should bypass wrappers the same way provider adapters do."""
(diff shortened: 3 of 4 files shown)


shaun0927:
I don't think we should merge this in its current form.

The isolation goal is understandable, but forcing `codex exec --ignore-user-config` crosses a real contract boundary for this repo. We explicitly treat `~/.codex/config.toml` as the setup-managed Codex MCP/env hookup point. If we unconditionally ignore user config here, we are not just filtering unrelated local customization — we are also potentially disabling Ouroboros's own Codex-side integration path.

That makes this too broad.

If there is a specific inheritance problem we want to solve here, please narrow it first:
- parent session/thread inheritance,
- wrapper resolution drift,
- or a particular user-config side effect.

Then solve that targeted problem directly instead of globally opting out of the configured Codex environment.

At minimum, any version of this change needs coverage showing that a setup-managed Codex MCP/env hookup still works end-to-end after the change. Right now the regression risk is larger than the isolation benefit.

andrew-adamson:
I agree that this is a bit of a blunt instrument. The intent was to be able to have different settings for codex invocations that are run by ouroboros vs those I've initiated interactively from the desktop or cli. For instance, when I'm working on small tasks interactively without ouroboros, I want the best model, fast mode, and I want particular notify actions (sounds), that I don't want when ouroboros is using codex. I also don't necessarily want certain mcp configs enabled for ouroboros workers.

Digging deeper, I think the best way to do this might be to use codex profiles, which are still in the config.toml file, and then modify ouroboros to take a codex_profile config setting that gets passed to `codex exec --profile` through the runtime command builder. Does that sound sane, and if so, do you want it as a new PR?

andrew-adamson:
The profile selection approach might also be a good way to implement the reasoning tier levels needed for  https://github.com/Q00/ouroboros/issues/184

Q00 (changes requested):
I understand the intent here: Codex subprocesses should be isolated enough that a parent/user session does not accidentally leak behavior into Agent OS runs. That is a real problem to solve.

I do not think `--ignore-user-config` is the right default lever, though. In Ouroboros, setup-managed Codex integration lives in the user config path as well, including MCP/env wiring. If we force `codex exec --ignore-user-config`, we can also disable the very configuration that makes Codex work correctly inside Ouroboros.

For Agent OS, the safer contract is: default runs should preserve the setup-managed Codex configuration, and isolation should be explicit/narrow. A profile-based design sounds much better to me, e.g. an optional `codex_profile` config that passes `codex exec --profile ...` when set. That would let us avoid parent-session drift without globally discarding user config.

Requesting changes because this version is too broad and can break the default Codex/MCP path even though CI is green. I would be happy with a replacement PR that keeps the default config path intact and adds coverage showing setup-managed MCP/env config still works.

Q00:
I think the Codex profile idea is the right direction, but I would probably frame it as the Codex mapping of a more general Agent OS `runtime_profile` concept rather than landing a top-level `codex_profile` knob directly.

The contract I have in mind is:

```text
Agent OS asks for `runtime_profile = worker`
→ Codex maps that to `codex exec --profile ouroboros-worker`
→ OpenCode maps it to its config/profile/env overlay equivalent
→ Hermes maps it to its runtime config/env
→ Claude Code maps it to model / permission / env settings
→ LiteLLM/OpenAI maps it to `CompletionConfig` fields such as `reasoning_effort`
```

That keeps the default Codex config path intact, avoids the broad behavior of `--ignore-user-config`, and gives us a clean path toward #184 without making the first implementation solve every backend at once.

A good first slice could be:

- add `orchestrator.runtime_profile`
- add a small `runtime_profiles.<name>.backend_profiles.codex` mapping
- have the Codex runtime/adapter pass `--profile <mapped-name>` only when that mapping exists
- leave the default behavior unchanged when no runtime profile is configured

So yes, I like the profile direction. I would just prefer to land it under the Agent OS runtime-profile abstraction, with Codex profile as the first backend mapping, instead of making `codex_profile` the main public concept.

shaun0927:
@andrew-adamson — thanks for sticking with this through the redesign discussion. The Agent OS \`runtime_profile\` slice you and @Q00 sketched is now up at #505, following the four-step scope from his comment above:

1. \`orchestrator.runtime_profile\` config field (default \`None\` → no behavior change)
2. Module-local \`_RUNTIME_PROFILE_TO_CODEX_PROFILE\` mapping
3. Codex runtime + adapter pass \`--profile ouroboros-worker\` only when the mapping fires
4. \`ouroboros setup --runtime codex\` writes \`[profiles.ouroboros-worker]\` next to the existing \`[mcp_servers.ouroboros]\` block, so the MCP/env hookup keeps working

Your *interactive vs worker* use case is exactly what the worker profile is for — drop your per-worker overrides (model, notify, sandbox) into that section and they apply to every Ouroboros-spawned \`codex exec\` without touching your interactive defaults. The profile abstraction also gives us a clean home for #184 and the next backends (OpenCode/Hermes/Claude/LiteLLM) as separate slices.

@Q00 — happy to leave the close on this PR to you once #505 lands; or if you'd rather close now, fine by me.

andrew-adamson:
Yep, #505 sounds like the way to go, so I'm happy to close this.

PR #489, opened 2026-04-22
fix(parallel): honor configured worker caps

andrew-adamson:
## Summary
- honor configured max parallel worker limits when the environment override is not set
- thread the runtime lookup through the relevant CLI and execution paths
- add focused config and CLI coverage for the worker-cap resolution behavior

## Testing
- `uv run pytest tests/unit/cli/test_run_qa.py tests/unit/config/test_loader.py::TestRuntimeHelperLookups::test_get_max_parallel_workers_prefers_env tests/unit/config/test_loader.py::TestRuntimeHelperLookups::test_get_max_parallel_workers_falls_back_to_config`


shaun0927:
This is directionally right.

Recent work in this area has been about making runtime behavior more explicit and less surprising, and centralizing the worker-cap lookup fits that well. Threading the configured cap through both CLI and MCP execution paths is the right move.

I do want one follow-up before we treat this as fully done: please avoid silently swallowing invalid config/env values.

If `get_max_parallel_workers()` quietly falls back to `3` for malformed env/config input, users can believe their configured cap is in effect when the runner is actually using the default. In this codebase we've generally been moving toward explicit contracts rather than silent fallback.

So please tighten that behavior — either by surfacing the invalid setting more clearly or by locking the intended fallback semantics down with stronger tests.

Overall though, this PR is in the right shape.

Q00 (changes requested):
This is the right feature direction. Agent OS needs the worker cap to be part of the runtime contract, not just a CLI-local knob, and wiring it through MCP execution is important for predictable multi-agent behavior.

I would like to see one behavior tightened before merge: invalid worker-cap configuration should not silently collapse back to `3`.

Right now `get_max_parallel_workers()` catches invalid env/config cases and returns the default. That makes a bad setting look like a successful setting, which is risky for Agent OS orchestration. If someone sets `OUROBOROS_MAX_PARALLEL_WORKERS=0`, a typo, or an invalid config value, we should either surface a clear error or very deliberately warn and test that fallback policy. Since the config model already says `ge=1`, my preference is to fail clearly instead of swallowing it.

Ruff is also failing on import ordering, so that needs cleanup before merge.

Requesting changes because the feature is good, but the failure mode should be explicit. Worker caps are a control-plane decision; silent fallback makes the system harder to reason about when multiple agents are running.

andrew-adamson:
Damn this review agent is a slave driver. Even load_config() itself also only catches yaml.YAMLError, not OSError....
It's thorough though - what's it based on?

shaun0927:
@andrew-adamson — pushed `1660549` to your branch (maintainer-edits is on) to clear the bot's last finding. `get_max_parallel_workers()` now also normalises `OSError` (`IsADirectoryError`, `PermissionError`, ...) into `ConfigError`, so the CLI / MCP wrappers don't get bypassed.

Two regression tests added: directory at config path → `ConfigError(error_type=IsADirectoryError)`, and `PermissionError` on `open` → `ConfigError(error_type=PermissionError)`. Diff is small (+6 in `loader.py`, +40 tests). Local ruff / mypy / `tests/unit/config/test_loader.py::TestRuntimeHelperLookups` (30/30) and the CLI/MCP tests this PR touches (102/102) all green; CI is now green on `1660549` as well.

Looping in @Q00 for the final pass — happy to revert the push if you'd rather land it yourself. Thanks for the very patient iterations on this; splitting `_parse_max_parallel_workers` out from `load_config` was the right call.

shaun0927:
@Q00 PR #489 should be ready for the final pass. Author addressed the silent-fallback you flagged in earlier commits, and I just pushed `1660549` to normalise the remaining `OSError` edge case the bot caught on `e7f84e1`. CI green, ruff/mypy clean. Could you re-review and dismiss the prior `CHANGES_REQUESTED` if you're satisfied? I'll squash-merge once you give the nod.

andrew-adamson:
@shaun0927 what about that other review (might have appeared while you were fixing the first one) where "OpenCode plugin executions silently ignore orchestrator.max_parallel_workers". I'd rather leave that as an issue and move on, since I know nothing about opencode, but can probably fix if necessary.

And thanks for your patience - this is the first public code I've ever contributed to, so probably making a bunch of mistakes.

shaun0927:
@Q00 @andrew-adamson — pushed `dc428d7` to address the last bot finding on `1660549` (OpenCode plugin path silently ignoring `max_parallel_workers`).

**What changed:**
- `get_max_parallel_workers()` now resolves *before* `should_dispatch_via_plugin()` in both `ExecuteSeedHandler` and `StartExecuteSeedHandler`, so plugin / in-process / background paths all see the same configured cap.
- `build_execute_subagent()` accepts `max_parallel_workers` and surfaces it in both the child prompt and the payload context, so OpenCode plugin consumers can honor it.
- `ConfigError` on the worker cap now fails the plugin path too, with the same `"Execution handler config error"` prefix as the in-process path.

**Tests:** +4 regression tests (2 unit on `build_execute_subagent`, 2 wiring on `ExecuteSeedHandler` / `StartExecuteSeedHandler` plugin payloads). Local ruff / mypy clean, 1414/1414 pytest pass on `tests/unit/mcp tests/unit/cli tests/unit/config`.

I think this clears the last blocking finding — re-review when you have a moment, happy to squash-merge once both prior `CHANGES_REQUESTED` reviews are dismissed.

PR #495, opened 2026-04-26
feat(core): render seed ontology in execution contract

### LINKED ISSUE #493: Elevate Seed ontology into the execution contract
## Summary

Seed already carries an `ontology_schema`, and the evolutionary loop already treats ontology as a first-class artifact through Wonder, Reflect, lineage snapshots, and ontology drift. However, the execution prompt currently exposes the Seed goal, constraints, acceptance criteria, and evaluation principles while largely omitting the ontology.

This leaves the ontology as schema/evolution metadata instead of a runtime conceptual lens. Execution agents should interpret the crystallized Seed through an explicit execution contract, where ontology is part of the contract rather than a loose prompt fragment.

## Problem

Current state:

- `Seed` is the immutable specification created from interview crystallization.
- `ontology_schema` is required on Seed.
- Wonder/Reflect use ontology to identify tensions and evolve child Seeds.
- `build_system_prompt()` does not render ontology into the execution instructions.

This creates a semantic gap:

```text
Interview -> Seed ontology -> Evolution loop
                         \-> Execution prompt: mostly goal/constraints/AC only
```

The result is that execution can satisfy AC text while missing the conceptual model the Seed was meant to preserve.

## Design Direction

Introduce a small interpretation layer between Seed and prompts:

```text
Seed = immutable crystallized source
SeedContract = execution-ready semantic contract derived from Seed
Prompt = runtime-specific rendering of that contract
```

Suggested modules:

- `src/ouroboros/core/seed_contract.py`
  - Pure domain interpretation of `Seed`
  - Converts ontology into an `OntologyLens`
  - Keeps Seed immutable and unmodified

- `src/ouroboros/orchestrator/seed_prompt.py`
  - Renders `SeedContract` into execution prompt sections
  - Keeps string composition out of `runner.py`

`runner.build_system_prompt()` should continue exposing the same public API but delegate Seed contract rendering to the new layer.

## Scope

In scope:

- Add `SeedContract` / `OntologyLens` domain models or equivalent pure helpers.
- Render ontology as an execution-time conceptual lens.
- Include ontology in `build_system_prompt()` output.
- Keep `build_task_prompt()` focused on concrete AC execution for now.
- Add regression tests for ontology rendering and prompt integration.

Out of scope for this issue:

- Changing the Seed YAML/schema format.
- Adding relation/value-judgment fields to `OntologySchema`.
- Changing Wonder/Reflect mutation behavior.
- Changing drift calculation.
- Changing interview completion gates.

## Acceptance Criteria

- `build_system_prompt(seed)` includes a dedicated Seed contract / ontology lens section.
- The section includes ontology name, description, and fields when fields exist.
- The section clearly instructs execution agents to use ontology as a conceptual lens for ambiguous implementation decisions.
- Empty ontology fields render gracefully without noisy placeholders.
- Existing public APIs remain compatible.
- Unit tests cover the new contract renderer and runner prompt integration.

## Test Plan

- `uv run pytest tests/unit/orchestrator/test_runner.py -q`
- `uv run pytest tests/e2e/test_full_workflow.py -q`
- `uv run ruff check src/ouroboros/core/seed_contract.py src/ouroboros/orchestrator/seed_prompt.py src/ouroboros/orchestrator/runner.py tests/unit/orchestrator/test_runner.py tests/e2e/test_full_workflow.py`

Q00:
## Summary

- Closes #493 by deriving a SeedContract from immutable Seeds and rendering ontology as a runtime conceptual lens.
- Moves goal, ACs, constraints, ontology, evaluation principles, exit conditions, and brownfield context into one prompt contract.
- Keeps task prompts focused on concrete AC execution while system prompts carry the Seed-level conceptual contract.

## Changes

- Add `SeedContract`, `OntologyLens`, and prompt renderers for Seed contract sections.
- Reframe ontology docs from output schema to conceptual lens.
- Update `build_system_prompt()` to delegate Seed contract rendering instead of composing goal/constraints inline.
- Add prompt renderer, runner, and e2e coverage for ontology lens rendering and empty ontology fields.

## Notes

This PR intentionally keeps Wonder/Reflect behavior, drift calculation, and evaluation gating out of scope. Those are follow-up contract-enforcement changes.

Closes #493

### DIFF
--- a/src/ouroboros/core/__init__.py
+++ b/src/ouroboros/core/__init__.py
@@ -29,6 +29,9 @@ _EXPORTS: dict[str, tuple[str, str]] = {
     "OntologyField": ("ouroboros.core.seed", "OntologyField"),
     "EvaluationPrinciple": ("ouroboros.core.seed", "EvaluationPrinciple"),
     "ExitCondition": ("ouroboros.core.seed", "ExitCondition"),
+    "OntologyConcept": ("ouroboros.core.seed_contract", "OntologyConcept"),
+    "OntologyLens": ("ouroboros.core.seed_contract", "OntologyLens"),
+    "SeedContract": ("ouroboros.core.seed_contract", "SeedContract"),
     # Context management
     "WorkflowContext": ("ouroboros.core.context", "WorkflowContext"),
     "ContextMetrics": ("ouroboros.core.context", "ContextMetrics"),
--- a/src/ouroboros/core/seed.py
+++ b/src/ouroboros/core/seed.py
@@ -54,13 +54,13 @@ class EvaluationPrinciple(BaseModel, frozen=True):
 
 
 class OntologyField(BaseModel, frozen=True):
-    """A field in the ontology schema.
+    """A concept in the ontology lens.
 
     Attributes:
-        name: Field identifier.
-        field_type: Type of the field (string, number, boolean, array, object).
-        description: Purpose of this field.
-        required: Whether this field is required.
+        name: Concept identifier.
+        field_type: Concept kind (string, number, boolean, array, object).
+        description: Purpose of this concept.
+        required: Whether this concept is required.
     """
 
     model_config = {"populate_by_name": True}
@@ -72,15 +72,16 @@ class OntologyField(BaseModel, frozen=True):
 
 
 class OntologySchema(BaseModel, frozen=True):
-    """Schema defining the structure of workflow outputs.
+    """Conceptual lens defining workflow domain coherence.
 
-    The ontology schema defines what data structure the workflow should
-    produce and maintain throughout iterations.
+    The ontology schema names the domain concepts and boundaries the workflow
+    should preserve throughout iterations. It is not, by itself, a mandatory
+    output shape.
 
     Attributes:
         name: Name of the ontology.
-        description: Purpose and scope of this ontology.
-        fields: List of fields in the ontology.
+        description: Purpose, scope, and perspective of this ontology.
+        fields: Concepts in the ontology.
     """
 
     name: str = Field(..., min_length=1)
@@ -167,7 +168,7 @@ class Seed(BaseModel, frozen=True):
         goal: The primary objective of the workflow.
         constraints: Hard constraints that must be satisfied.
         acceptance_criteria: Specific criteria for success.
-        ontology_schema: Schema for workflow output structure.
+        ontology_schema: Conceptual lens for workflow coherence.
         evaluation_principles: Principles for evaluating outputs.
         exit_conditions: Conditions for terminating the workflow.
         metadata: Generation metadata (version, timestamp, etc.).
@@ -227,10 +228,10 @@ class Seed(BaseModel, frozen=True):
         description="Specific criteria for success evaluation",
     )
 
-    # Structure
+    # Conceptual lens
     ontology_schema: OntologySchema = Field(
         ...,
-        description="Schema defining the structure of workflow outputs",
+        description="Ontology defining the workflow's conceptual lens",
     )
 
     # Evaluation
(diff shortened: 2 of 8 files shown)


Sathvik-1007:
@Q00 , Hey, went through `ef7be45` in detail. Overall the `SeedContract` extraction is solid — clean separation, ontology-as-lens is the right call. Few things I noticed:

---

**1. ACs are now duplicated across system prompt and task prompt**

This is the main one. Before this PR, `build_system_prompt()` rendered goal + constraints + eval principles. ACs only lived in `build_task_prompt()`. Now `render_seed_contract_for_execution()` pulls in `render_acceptance_criteria_section()` at `seed_contract_prompt.py:109`, so system prompt has the full numbered AC list. But `build_task_prompt()` at `runner.py:288` still renders them independently too.

End result: every AC appears twice in the combined prompt sent to the agent. Goal also appears twice. For a 7-AC seed that's ~200-400 extra tokens per run for no benefit. No test catches this either — `test_does_not_duplicate_system_ontology_lens` checks ontology isn't in task prompt, but nothing checks AC duplication across the two prompts.

Probably easiest fix: drop ACs from the system prompt contract (they're more of a task-level concern anyway), or strip them from `build_task_prompt()` since the contract already has them.

---

**2. Eval principles format changed — weight is now visible to agents**

Old format in system prompt:
```
- completeness: All requirements are met
```

New format via `render_evaluation_principles_section()` at `seed_contract_prompt.py:95`:
```
- completeness (weight 1.00): All requirements are met
```

Weight was previously hidden from execution agents. Now they can see `weight 0.80` vs `weight 1.00` and might prioritize differently. Not saying this is wrong — might even be better — but it's a silent behavioral change from the prompt refactor. Worth being intentional about.

---

**3. Seed docstrings now say "concept" but class is still `OntologyField`**

`seed.py:57` — class is `OntologyField`, but PR changed the docstring to `"A concept in the ontology lens"`. Same for the attributes at `seed.py:62-65`: "Concept identifier", "Concept kind", etc. `OntologySchema` docstring at `seed.py:75` now says "Conceptual lens defining workflow domain coherence".

The thing is, `Seed` is the YAML-facing schema layer — it parses `ontology_schema.fields` from seed files. "Field" and "Schema" are the right names for that layer. The interpretation layer (`seed_contract.py`) already has `OntologyConcept` and `OntologyLens` with the right vocabulary. Having both layers use "concept" language muddies which one is the schema and which is the interpretation.

Suggest: revert the `seed.py` docstrings to original "field"/"schema" language. Let `SeedContract` own the "concept"/"lens" vocabulary exclusively.

---

**4. `decision_context` param is orphaned**

`render_ontology_lens_section()` at `seed_contract_prompt.py:51` takes `decision_context="execution decisions"`. This was designed for `render_seed_contract_for_evaluation()` which passed `"evaluation judgments"` — but that function was removed in this commit. So now only the default value is ever used.

If PR #496 re-adds the eval renderer, this is fine as forward-looking API. If not, it's dead flexibility in a public `__all__` export.

---

**5. E2E test doesn't check `field_type`**

`test_full_workflow.py:88-91` iterates over `seed.ontology_schema.fields` and asserts `field.name in prompt` and `field.description in prompt`. But doesn't check `field.field_type`. That's exactly what commit `6c961e0` fixed (the dropped field_type). Unit test at `test_runner.py:121` covers it via the full string match `"- tasks [array]: List of tasks (required concept)"`, but the e2e doesn't guard the regression directly.

Quick fix: add `assert field.field_type in prompt` to the e2e loop.

---

**6. Recovery is sequential-only (just noting)**

Recovery loop at `runner.py:1773-1805` only fires in sequential `execute_seed`. Parallel execution (`_execute_parallel`) and `resume_session` don't have it. Makes sense — parallel has per-AC retry semantics, and resume is already a recovery mechanism. Just noting the asymmetry since there's no comment in code explaining why it's sequential-only. A one-liner would help future readers.

---

That's what I found. no.1 (AC duplication) is the most actionable. Rest is polish. Nice work on the extraction overall.

Issue #516, opened 2026-04-29
Directive emission site — StepAction in evolution (slice 1 of #472)


shaun0927:
# Directive emission site — StepAction in evolution (slice 1 of #472)

## Context

`core/directive.py` (#477) declares a 9-member `Directive` vocabulary
(`CONTINUE`, `EVALUATE`, `EVOLVE`, `UNSTUCK`, `RETRY`, `COMPACT`, `WAIT`,
`CANCEL`, `CONVERGE`). The module docstring is explicit:

> This PR adds only the type and its invariants; no caller is modified
> here.

#472 asks for the migration of existing local enums onto `Directive`
across the codebase. Per the maintainer alignment in #476 (Q5), the **first
reference migration is `StepAction` in evolution**, not the evaluator.

Reasons (verbatim from #476 Q5):

> - The mapping is unambiguous — no semantics invented:
>   ```
>   StepAction.STAGNATED  →  Directive.UNSTUCK
>   StepAction.CONVERGED  →  Directive.CONVERGE
>   StepAction.FAILED     →  Directive.RETRY or Directive.CANCEL
>                            (budget-dependent)
>   ```
> - Evolution already has event coverage under `evolution.*`, so the
>   directive emission can be cross-validated against existing lineage in
>   the same PR.
> - It validates the "local enums don't need to disappear immediately"
>   stance — the adapter is a module-level mapping function, not a rewrite.

Evaluator terminal branches are the second migration (separate slice),
because they have more variants (mechanical / semantic / consensus) and
should inherit the pattern set by this one.

## Scope

1. **Mapping**
   - `StepAction.STAGNATED` → `Directive.UNSTUCK`
   - `StepAction.CONVERGED` → `Directive.CONVERGE`
   - `StepAction.FAILED` → `Directive.RETRY` or `Directive.CANCEL`
     (decision based on resilience layer's retry budget)
   - The decision rule for `FAILED → RETRY vs CANCEL` is documented inline
     in `evolution/` so future migrators have a reference matrix.

2. **Emission**
   - At each `StepAction`-producing site, build a
     `control.directive.emitted` event via the existing factory and append
     it to the EventStore.
   - Required fields per #476 Q-Q00: `target_type="lineage"`,
     `target_id=<lineage_id>`, `emitted_by="evolver"`,
     `directive=<mapped>`, `reason=<short text>`.
   - Optional correlation: `lineage_id`, `generation_number`, `phase`.

3. **Backward compatibility**
[...]

Issue #515, opened 2026-04-29
Introduce ControlBus on AgentRuntimeContext for control.directive.emitted


shaun0927:
# Introduce ControlBus on AgentRuntimeContext for control.directive.emitted

## Context

`events/control.py` (#492) ends with:

> Reactive consumption via a ControlBus subscription surface is a separate,
> later concern so the primitive stays stable while projections evolve.

After the EventStore registration / lineage projector for
`control.directive.emitted` lands (closes #473), projections can *see*
directive events but cannot *react* to them.

Per the maintainer alignment in #476 (Q1) and the renamed framing in #474,
the reactive surface lands as **`ControlBus` exposed through
`AgentRuntimeContext`**, not as a standalone `PolicyBus`. The split is
deliberate: `AgentRuntimeContext` describes *what the runtime offers* (the
context struct), and `ControlBus` describes *where directives flow* (the
subscription surface). Two different concerns, two different names.

This is intentionally **single-runtime in-process**. The cross-runtime
generalization is the Mesh (separate sub-RFC of #476). ControlBus exists so
#474 (`AgentRuntimeContext + ControlBus`) and #475 (`evolve_step` /
`unstuck` / `ralph` migration) can use a stable subscription contract
before the Mesh is built.

## Scope

1. **`ControlBus` class**
   - `subscribe(predicate, handler) → SubscriptionHandle`
   - `unsubscribe(handle)`
   - `publish(event)` — invokes matching handlers
   - Lives in `orchestrator/` (or `runtime/` if/when that layer is
     introduced per #476 Q2 reasoning).

2. **`AgentRuntimeContext` membership**
   - `ControlBus` is exposed on `AgentRuntimeContext` as `context.control`
   - `AgentRuntimeContext` keeps the narrow membership commitment from
     #476 Q1 — no service-locator drift; new fields require a per-PR
     justification line
   - First minimal shape per #476 Q1:
     ```python
     @dataclass(frozen=True, slots=True)
     class AgentRuntimeContext:
         event_store: EventStore
         runtime_backend: str | None
         llm_backend: str | None
         mcp_bridge: MCPBridge | None = None
         control: ControlBus | None = None  # this PR adds this slot
     ```

3. **Event source wiring**
   - When `control.directive.emitted` is appended to the EventStore, the
     ControlBus is notified (no polling)
   - Multiple subscribers receive the event independently; one handler's
     failure does not block others (per-handler task spawn)

4. **Failure semantics**
[...]

Issue #514, opened 2026-04-29
Wire control.directive.emitted into EventStore + lineage projector (closes #473)


shaun0927:
# Wire control.directive.emitted into EventStore + lineage projector (closes #473)

## Context

PR #492 added `events/control.py` with the
`create_control_directive_emitted_event` factory. The module docstring
explicitly states the PR is **observational-first**:

> This module is observational-first. It persists the event; no emission
> site is wired, and no reactive consumer is added.

Issue #473 asks for the next step: register the new category with the
EventStore and surface it through the lineage projector, so directive
emissions become visible alongside the existing five categories
(`decomposition`, `evaluation`, `interview`, `lineage`, `ontology`).

This is the smallest possible PR that turns the dormant Stage-2 primitive
into something the TUI / MCP tooling actually sees.

## Scope

1. **EventStore registration**
   - Register `control.directive.emitted` as a known category in
     `persistence/event_store.py` (and any schema or migration files that
     enumerate categories)
   - Ensure existing aggregate filters (`aggregate_type=lineage`,
     `aggregate_type=execution`, etc.) naturally see directive events for
     their target, per the target-oriented aggregation rule baked into
     `events/control.py`
   - Add a migration if the event_store schema enumerates categories
     anywhere (verify with #436 `event_version` work)

2. **Lineage projector**
   - The existing lineage projector (or its closest equivalent in
     `core/lineage.py` / `tui/`) reads directive events for the lineage
     aggregate and renders them inline with state events
   - Render produces a single-line entry per directive: `<emitted_by>
     <directive> → <reason>`

3. **Observability test**
   - Round-trip test: emit a directive event via the factory, persist it,
     query the EventStore by `aggregate_type=lineage` and confirm the
     directive shows up alongside lineage state events
   - Snapshot test: lineage projector output includes the directive line

## Out of scope

- ControlBus subscription surface (separate issue — required to react to
  directives, not just observe them)
- Any actual emission site (separate issue — a callsite that *produces* a
  directive event)
- New `target_type` values beyond what `events/control.py` already declares

## Acceptance

- [ ] `control.directive.emitted` registered in EventStore category
      enumeration
- [ ] Lineage aggregate query returns directive events without extra opt-in
- [ ] Lineage projector renders one line per directive
- [ ] Round-trip integration test passes
- [ ] Projector snapshot test added
- [ ] No emission site added (kept observational)

## Dependencies

- Closes: #473
- Builds on: #492
[...]

PR #523, opened 2026-04-30
feat(events): fold control.directive.emitted onto LineageProjector

### LINKED ISSUE #473: Add control_directive event category to EventStore
## Context

Depends on #471 (control-plane RFC) and the Directive vocabulary issue.

Ouroboros's strength is event sourcing: the SQLite EventStore reconstructs any prior state from immutable events. Current categories include `decomposition`, `evaluation`, `interview`, `ontology`, and `lineage`.

Conspicuously absent: the decisions themselves. Today, *why* a retry fired, *why* `unstuck` was invoked, or *why* `evolve` advanced a generation is not a first-class event — it is implied by downstream effects.

## Problem

This creates three concrete gaps:

1. **No audit trail for control decisions.** Post-hoc debugging of "why did the loop pick X?" requires reading logs, not the event store.
2. **Directive lineage is invisible.** The `lineage` tracker captures data causality but not decision causality.
3. **TUI `lineage` screen cannot render decision timelines.** Users see phase progress but not the decisions that produced it.

Without this, the new `Directive` type from the sibling issue is ephemeral — it describes decisions that happen but are not remembered.

## Proposal

Add a new event category `control` with a primary event type `ControlDirectiveEmitted`:

```python
@dataclass(frozen=True)
class ControlDirectiveEmitted:
    event_id: UUID
    session_id: UUID
    emitted_by: str                 # logical source, e.g. "evaluator", "evolver"
    directive: Directive            # from Directive-vocabulary issue
    reason: str                     # free-form rationale captured at emission
    context_snapshot_id: UUID | None   # optional pointer into an existing snapshot
    timestamp: datetime
```

Additional fields may join during review (e.g., `parent_directive_id` for causal chains).

### Integration

- SQLAlchemy model + migration under `src/ouroboros/persistence/migrations/`
- Event builder helpers under `src/ouroboros/events/control.py`
- Replay: `reconstruct_state` must accept `control` events without changing prior-state semantics (they are observational relative to existing state machines)

### Rendering

- Extend TUI `lineage` and `debug` screens to show a directive timeline per session
- `ouroboros status execution <id>` shows the last N directives

## Acceptance Criteria

- [ ] `events/control.py` defines `ControlDirectiveEmitted` with the schema above
- [ ] Alembic migration adds the new table and is reversible
- [ ] Round-trip test: emit → persist → reconstruct → assert equality
- [ ] One production emission site wired (e.g., `evaluation/` terminal decisions)
- [ ] TUI lineage screen renders at least one directive per execution in a manual demo
- [ ] Backward compatibility: existing databases migrate without data loss

## Out of Scope

- Wiring every decision site — single reference site only; remainder handled in the `PolicyBus` wiring issue
- Dispatching directives to consumers — that is the `PolicyBus` issue
- Replacing existing event categories

## Risks

- **Schema churn**: the event shape may evolve. Mitigation: store `reason` as free-form and add structured fields via additive migrations only.
- **Event volume**: high-frequency emissions could bloat the store. Mitigation: no per-token events; emit one event per decision, not per step.

shaun0927:
## Summary

Folds `control.directive.emitted` events (introduced as observational-first by #492) onto the projected `OntologyLineage` read model so downstream callers — TUI lineage view, reports, the evaluator follow-up work in #516/#517/#518 — can render the audit trail without a second event-store query.

The EventStore itself is schema-free for categories (`aggregate_type` is a free-form string), so no migration or registration step is needed. `replay_lineage()` already returns the directive rows because `events/control.py` aggregates them by `target_type`/`target_id`; this PR adds the projector branch and the read-model fields.

Closes #473. Partially advances #514 (TUI rendering of the new field is deferred to a follow-up so this PR stays single-responsibility).

## Changes

- `core/lineage.py` — new `ControlDirectiveEmission` view model and `directive_emissions: tuple[ControlDirectiveEmission, ...]` field on `OntologyLineage`, plus a `with_directive_emission()` helper that mirrors the existing `with_generation` / `with_status` mutation pattern.
- `evolution/projector.py` — `LineageProjector.project()` now folds `control.directive.emitted` events into a per-replay list and attaches them in event-order via `model_copy`. Malformed payloads (missing or empty `directive` field) are silently skipped so a single bad row cannot corrupt the projection.
- `tests/unit/test_projector_directives.py` — five focused cases.

## Verification

| Check | Result |
|---|---|
| `uv run ruff check src/ouroboros/core/lineage.py src/ouroboros/evolution/projector.py tests/unit/test_projector_directives.py` | clean |
| `uv run ruff format ...` | no diff |
| `uv run pytest tests/unit/test_projector_directives.py` | 5 passed |
| `uv run pytest tests/unit/test_projector_rewind.py tests/unit/core/test_lineage.py tests/unit/events/test_control_events.py` (regression) | 42 passed |

## Pre-merge checklist (from #514 verification protocol)

- [x] EventStore append accepts `control.directive.emitted` events (existing factory output already valid; no schema change needed)
- [x] `EventStore.replay("lineage", lineage_id)` returns directive events alongside `lineage.*` state events (existing aggregate-based replay)
- [x] `LineageProjector.project()` exposes one `ControlDirectiveEmission` per emitted event on `OntologyLineage.directive_emissions`
- [x] Round-trip test: emit → persist (would happen via `event_store.append`) → query by aggregate → directive present on the projection (`tests/unit/test_projector_directives.py`)
- [x] Existing `lineage.*` queries are unaffected (regression suite green)
- [ ] TUI lineage view shows the directive line in the timeline at the correct position — *deferred to a follow-up PR; this PR's responsibility ends at the projected read model*
- [x] No destructive ALTER, no data rewrite (no migration introduced — EventStore stores `aggregate_type` as a free-form string already)

## Post-merge checklist

- [ ] On a real install, run a fixture session that emits at least one directive event; query the event_store and confirm the row exists with the correct `aggregate_type` / `aggregate_id`
- [ ] `OntologyLineage.directive_emissions` populates on `replay_lineage(lineage_id)` of a fresh run
- [ ] Open the TUI follow-up issue once #515 / #516 land so the rendering layer can consume the new field

## Rollback

The change is additive: a new field on a frozen Pydantic model and one new branch in the projector. Rollback steps:

1. Revert this PR. Already-emitted directive events stay in the EventStore as inert rows.
2. Pre-existing `LineageProjector` consumers continue to work unchanged because the new field defaults to an empty tuple, and the new projector branch only fires for events that did not exist on the read model before.
3. No data migration needed for rollback.

Closes #473.

### DIFF
--- a/src/ouroboros/core/lineage.py
+++ b/src/ouroboros/core/lineage.py
@@ -369,6 +369,24 @@ class RewindRecord(BaseModel, frozen=True):
     discarded_generations: tuple[GenerationRecord, ...] = ()
 
 
+class ControlDirectiveEmission(BaseModel, frozen=True):
+    """One ``control.directive.emitted`` event folded onto a lineage.
+
+    Captures the audit-level "who/what/why/when" of a control-plane
+    decision so it appears in the lineage timeline alongside state events
+    (per RFC #476 M2 / sub-RFC #511). The full event row remains in the
+    EventStore; this is the projected view used by callers.
+    """
+
+    directive: str
+    reason: str
+    emitted_by: str
+    timestamp: datetime
+    generation_number: int | None = None
+    phase: str | None = None
+    is_terminal: bool = False
+
+
 class OntologyLineage(BaseModel, frozen=True):
     """Tracks O₁ → O₂ → O₃ evolution across generations.
 
@@ -382,6 +400,7 @@ class OntologyLineage(BaseModel, frozen=True):
     goal: str
     generations: tuple[GenerationRecord, ...] = Field(default_factory=tuple)
     rewind_history: tuple[RewindRecord, ...] = Field(default_factory=tuple)
+    directive_emissions: tuple[ControlDirectiveEmission, ...] = Field(default_factory=tuple)
     status: LineageStatus = LineageStatus.ACTIVE
     created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
 
@@ -409,6 +428,18 @@ class OntologyLineage(BaseModel, frozen=True):
     def with_status(self, status: LineageStatus) -> OntologyLineage:
         """Return new lineage with updated status."""
         return self.model_copy(update={"status": status})
+
+    def with_directive_emission(self, emission: ControlDirectiveEmission) -> OntologyLineage:
+        """Return new lineage with the directive emission appended.
+
+        Folds one ``control.directive.emitted`` event into the lineage's
+        timeline view. Existing emissions are preserved; ordering is
+        whatever the projector replayed (timestamp order from the
+        EventStore).
+        """
+        return self.model_copy(
+            update={"directive_emissions": self.directive_emissions + (emission,)}
+        )
 
     def rewind_to(self, generation_number: int) -> OntologyLineage:
         """Return lineage truncated to the given generation.
--- a/src/ouroboros/core/lineage.py
+++ b/src/ouroboros/core/lineage.py
@@ -464,9 +464,15 @@ class OntologyLineage(BaseModel, frozen=True):
             raise ValueError(f"Generation {generation_number} out of range [1, {max_gen}]")
 
         truncated = tuple(g for g in self.generations if g.generation_number <= generation_number)
+        retained_directives = tuple(
+            emission
+            for emission in self.directive_emissions
+            if emission.generation_number is None or emission.generation_number <= generation_number
+        )
         return self.model_copy(
             update={
                 "generations": truncated,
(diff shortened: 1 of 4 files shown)


shaun0927:
Autopilot follow-up: pushed review-addressing commits and confirmed the GitHub check suite is green on the latest head. Current merge state is clean; remaining CHANGES_REQUESTED status appears to be awaiting fresh bot/human re-review rather than failing CI.

Q00 (approved):
Reviewed the directive projector changes. Folding control.directive.emitted into the lineage projection is the right observational layer, and the implementation keeps legacy lineage state intact.

Issue #511, opened 2026-04-29
[RFC] MCP Mesh wire format & Coordinator handshake (sub-RFC of #476)


shaun0927:
# [RFC] MCP Mesh wire format & Coordinator handshake (sub-RFC of #476)

## Context

#476 frames the Phase 2 Agent OS as a horizontal **Contract layer** that lets
heterogeneous runtimes (Claude Code, Codex, OpenCode, Hermes, Gemini CLI, …)
attach to a single Coordinator through MCP-style tool schema, polling, and
ResultEvent. The Phase-2 RFC settles the *what*; this sub-RFC settles the
*how* of the **Mesh primitive itself** so that downstream PRs (#474, #475 and
the multi-runtime Coordinator work) have a fixed protocol surface to build
against.

`core/directive.py` (#477) and `events/control.py` (#492) already give the
control vocabulary and the journal entry. They are intentionally
*observational-first* — the Mesh is the missing piece that turns those types
into actual cross-runtime traffic.

## Scope

This RFC decides:

1. **Transport** — UDS, TCP loopback, stdio multiplex, SSE, or a tier
   (e.g., UDS preferred, TCP fallback). Constraint: must work on macOS,
   Linux, and Windows without elevated privileges.
2. **Envelope schema** — required and optional fields for both *request*
   (Coordinator → runtime) and *result* (runtime → Coordinator). Must include
   `contract_id`, `target_type`, `target_id`, `directive`, `correlation_id`,
   and a forward-compatible `extra` slot. Versioning rule (`schema_version`
   field, additive-only) must be specified.
3. **Polling vs push** — long-poll loop, server-push streaming, or hybrid.
   The slide framing ("polling, ResultEvent가 IPC처럼 작동") leans toward
   long-poll, but a concrete decision is required so polling intervals,
   backpressure, and reconnection semantics are unambiguous.
4. **Coordinator topology** — separate daemon process, embedded in the
   orchestrator session, or pluggable per deployment. Affects #474 because
   PolicyBus is the in-process precursor; the Mesh is the cross-process
   generalization.
5. **Runtime registration & capability advertisement** — handshake message
   shape, when registration happens (startup vs lazy), and how a runtime
   advertises which `Directive` members it can satisfy.
6. **Ordering guarantees** — FIFO per `contract_id`, at-least-once,
   exactly-once, or best-effort. Must be stated explicitly because the
   Disposable Memory RFC depends on it (replay correctness).
[...]

Issue #475, opened 2026-04-22
Wire evolve_step / unstuck / ralph through AgentRuntimeContext (closes #280 known limitations)


shaun0927:
## Context

Depends on #471 (control-plane RFC) and the `PolicyBus` issue. Closes the two Known Limitations recorded in `docs/guides/mcp-bridge.md` from PR #280:

> - Evolution loop (`evolve_step`) does not yet pass the bridge manager
> - No dynamic server addition after initial connection

The `PolicyBus` issue provides the single injection point. This issue consumes it across the remaining decision surfaces so that external MCP tools (via `MCPBridge`) are uniformly available beyond `execute_seed`.

## Problem

Currently, `MCPBridge` tool visibility is scoped to `ExecuteSeedHandler → OrchestratorRunner`. Decision surfaces outside this path — evolution, resilience (unstuck personas), and persistent loops (`ralph`) — cannot reach upstream MCP tools, defeating most of the bridge's value for long-running sessions.

## Proposal

Split the work into three independent PRs under this issue, so each is reviewable on its own:

### PR 1: `evolve_step`

- Accept `PolicyBus` in the evolution entry point
- Allow the evolutionary proposal step to query upstream MCP tools (e.g., filesystem browsing during retrospection)
- End-to-end test: a seed whose evolution requires an external MCP lookup succeeds

### PR 2: `unstuck`

- Each of the 5 lateral personas receives `PolicyBus`
- Personas may call upstream tools to inspect current state before producing a lateral suggestion
- End-to-end test: invoke `unstuck` with a bridge configured; verify tool usage is recorded in `control_directive` events

### PR 3: `ralph`

- Persistent loop uses the same bus across session boundaries
- Bridge lifecycle is re-established on session restart; health checks follow existing `health_check_interval`
- End-to-end test: interrupt and resume `ralph`; verify bridge re-connects and remaining generations succeed

### Dynamic server addition (bonus)

Given `PolicyBus` holds a live `MCPClientManager` reference, expose `bus.add_mcp_server(config)` and verify that all three surfaces pick up the new server without restart. This closes the second Known Limitation.

## Acceptance Criteria

- [ ] `evolve_step` end-to-end test passes with upstream MCP tool involvement
- [ ] `unstuck` personas can invoke upstream tools; directive events recorded
- [ ] `ralph` survives session restart with bridge re-connection
- [ ] Dynamic server addition works at runtime and is covered by tests
[...]

Issue #474, opened 2026-04-22
Replace per-handler mcp_manager plumbing with AgentRuntimeContext + ControlBus


shaun0927:
## Context

Depends on #471 (control-plane RFC), the Directive vocabulary issue, and the `control_directive` event-category issue.

PRs #279 and #280 wire `mcp_manager` through `ExecuteSeedHandler` via explicit constructor injection. Each additional handler that needs MCP capability — `evolve_step`, `unstuck`, `ralph`, and any future workflow — must repeat the same injection.

Evidence from #280:

- `src/ouroboros/cli/commands/mcp.py`: bridge is constructed and passed into `create_ouroboros_server(..., mcp_bridge=mcp_bridge)`
- `src/ouroboros/mcp/tools/definitions.py::execute_seed_handler` added `mcp_manager` and `mcp_tool_prefix` parameters
- `docs/guides/mcp-bridge.md` records as a Known Limitation that `evolve_step` does not yet receive the manager

## Problem

Per-handler injection is O(handlers). It has three failure modes:

1. A new handler silently omits the injection → runtime failure only.
2. Future plumbing (event store, LLM backend, directive emitter) compounds the parameter list.
3. Dynamic capability addition (also a stated limitation: "No dynamic server addition after initial connection") has no seam to plug into — each handler holds a static reference.

## Proposal

Introduce a single `PolicyBus` object (name open — `ControlBus` / `DirectiveBus` are alternatives) that owns the dependencies currently passed à la carte:

```python
@dataclass
class PolicyBus:
    event_store: EventStore
    mcp_manager: MCPClientManager | None   # from MCPBridge, optional
    directive_emitter: DirectiveEmitter    # emits ControlDirectiveEmitted
    llm_backend: str | None
    runtime_backend: str | None
    # ... single source of shared dependencies
```

Construction: once per server lifecycle, at `_run_mcp_server` time. Handlers receive `PolicyBus` by reference.

### Handler contract change

Before:

```python
ExecuteSeedHandler(
    event_store=event_store,
    mcp_manager=mcp_manager,
    mcp_tool_prefix=prefix,
    runtime_backend=rb,
    llm_backend=lb,
)
```

After:

```python
ExecuteSeedHandler(bus=bus)
```

### Dynamic capability support

Because `PolicyBus` holds a live reference to the manager, future dynamic `add_server` calls on the manager propagate to all handlers without re-injection — unlocking the second half of #280's Known Limitations.

## Migration strategy

[...]

PR #524, opened 2026-04-30
feat(orchestrator): introduce ControlBus on AgentRuntimeContext

### LINKED ISSUE #515: Introduce ControlBus on AgentRuntimeContext for control.directive.emitted

shaun0927:
## Summary

Adds the reactive surface paired with the observational event factory from #492. `ControlBus` is an in-process publish/subscribe primitive; `AgentRuntimeContext` is the narrow, frozen dataclass agreed in #476 Q1 that carries the bus alongside the existing per-handler dependencies (EventStore, MCP bridge, runtime/llm backend labels). The `control` slot on the context defaults to `None` so the upcoming #474 migration can adopt it incrementally without breaking existing callers.

Closes #515.

## Changes

- `src/ouroboros/orchestrator/control_bus.py` — new `ControlBus` class with `subscribe(predicate, handler) → SubscriptionHandle`, `unsubscribe(handle)`, and `publish(event) → tuple[Task, ...]`. Per-handler asyncio task spawn isolates a slow or failing handler from siblings. Predicate exceptions are treated as non-matches and logged; the predicate is *not* auto-unsubscribed.
- `src/ouroboros/orchestrator/agent_runtime_context.py` — new frozen, slotted dataclass with the minimal field set from #476 Q1: `event_store`, `runtime_backend`, `llm_backend`, `mcp_bridge`, `control`. A regression test pins the field set so reviewers see explicit diffs when the membership grows.
- `tests/unit/orchestrator/test_control_bus.py` (8 cases) and `tests/unit/orchestrator/test_agent_runtime_context.py` (3 cases).

## Scope choices baked in

- **No coupling to persistence.** EventStore stays untouched; callsites publish to the bus themselves after `EventStore.append`. This honours the "no service locator" guardrail from #476 Q1 (narrow membership for the runtime context) and keeps the bus a pure in-process primitive that future callers (#474, #475) can adopt without modifying EventStore.
- **No cross-runtime delivery.** That is the Mesh's responsibility (sub-RFC #511). The bus is single-runtime by design.
- **No forging resistance.** Cooperative trust per #476: any in-process actor can publish; the bus does not police identity.

## Verification

| Check | Result |
|---|---|
| `uv run ruff check src/ouroboros/orchestrator/control_bus.py src/ouroboros/orchestrator/agent_runtime_context.py tests/unit/orchestrator/test_control_bus.py tests/unit/orchestrator/test_agent_runtime_context.py` | clean (after one auto-fix for import order) |
| `uv run ruff format ...` | no diff |
| `uv run pytest tests/unit/orchestrator/test_control_bus.py tests/unit/orchestrator/test_agent_runtime_context.py` | 11 passed |

## Pre-merge checklist (from #515 verification protocol)

- [x] `ControlBus` class with `subscribe` / `unsubscribe` / `publish`
- [x] Slot added to `AgentRuntimeContext` (`control`) with narrow-membership rationale in this PR body and pinned by `test_context_membership_is_narrow`
- [x] Predicate matching: subscribers only see events for which their predicate returns `True`
- [x] Multi-subscriber: all matching subscribers receive the event for one publish call
- [x] Handler failure isolation: a subscriber raising an exception does not affect other subscribers' delivery for the same event
- [x] Slow subscriber test: a handler that sleeps does not delay another handler that completes immediately
- [x] No deadlock when a handler subscribes a new handler from inside `publish`
- [x] `unsubscribe` is idempotent and stops further delivery
- [x] Existing `AgentRuntimeContext` consumers that do not pass `control=` continue to work; `context.control is None` is a valid state — verified by the test where the dataclass is constructed without the slot
- [ ] **Deferred to follow-up:** EventStore append → bus auto-publish wiring. This PR keeps the bus standalone so it stays single-responsibility; the wiring lands either alongside the #474 migration (where the runtime context is actually threaded through handlers) or in a tiny dedicated PR. The current callsite contract is "publish to the bus immediately after `event_store.append` in the same code path" — the bus does not depend on EventStore for its API, and the verification scenario from #515 (subscriber fires on a real directive event) is achievable today by any callsite that holds both objects.

## Post-merge checklist

- [ ] On a real install, register a no-op subscriber via a small dev script; emit one directive; confirm the subscriber fires
- [ ] Confirm `control_bus.handler_raised` log entry appears when a deliberately broken subscriber is registered
- [ ] Open the EventStore-wiring follow-up issue (or fold the wiring into the first slice of #474)

## Rollback

The bus is dormant until callers register subscribers, and #474/#475 callsite migration is in a separate PR. Rollback steps:

1. Revert this PR.
2. No data to clean up — the bus does not persist anything.
3. The new `AgentRuntimeContext` slot disappears; any pre-emptive callers (none in this PR) would fail to construct, surfaced quickly through CI.

Closes #515.

### DIFF
--- a/src/ouroboros/orchestrator/agent_runtime_context.py
+++ b/src/ouroboros/orchestrator/agent_runtime_context.py
@@ -1,0 +1,45 @@
+"""``AgentRuntimeContext`` — the runtime-offering side of #476 Q1.
+
+The runtime context describes *what the orchestrator runtime offers* a
+handler — the EventStore it appends into, optional capability sources,
+and the :class:`ControlBus` that lets handlers react to directive events.
+The context is intentionally narrow: per the maintainer commitment in
+#476 Q1, every new field added later must include a one-line PR-body
+justification so the type does not drift into a service locator.
+
+This first step keeps the membership minimal:
+
+* ``event_store`` — the persistence boundary handlers already share.
+* ``runtime_backend`` / ``llm_backend`` — informational labels existing
+  callers (rate-limit, telemetry) already pass alongside the bridge.
+* ``mcp_bridge`` — the existing capability source from #280's bridge
+  work; optional so non-MCP code paths stay valid.
+* ``control`` — the :class:`ControlBus` introduced alongside this
+  context; ``None`` while #474's migration is in progress so existing
+  callers compile unchanged.
+
+Subsequent issues (#474, #475) wire concrete handlers to consume the
+context. This module deliberately does *not* import any handler-side
+type so that adopting the context becomes a one-import change.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ouroboros.mcp.bridge.bridge import MCPBridge
+    from ouroboros.orchestrator.control_bus import ControlBus
+    from ouroboros.persistence.event_store import EventStore
+
+
+@dataclass(frozen=True, slots=True)
+class AgentRuntimeContext:
+    """Minimal, narrow-membership runtime context handed to MCP handlers."""
+
+    event_store: EventStore
+    runtime_backend: str | None = None
+    llm_backend: str | None = None
+    mcp_bridge: MCPBridge | None = None
+    control: ControlBus | None = None
--- a/src/ouroboros/orchestrator/control_bus.py
+++ b/src/ouroboros/orchestrator/control_bus.py
@@ -37,7 +37,7 @@ import asyncio
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from ouroboros.events.base import BaseEvent
@@ -55,9 +55,12 @@ class SubscriptionHandle:
     """Opaque token returned by :meth:`ControlBus.subscribe`.
 
     Callers pass it back to :meth:`ControlBus.unsubscribe` to detach.
+    Handles are scoped to the bus that created them so identical local
+    integer IDs from different buses cannot detach each other.
     """
 
     _id: int
+    _owner: Any
 
 
 @dataclass(slots=True)
@@ -80,6 +83,8 @@ class ControlBus:
 
     _subscriptions: list[_Subscription] = field(default_factory=list)
     _next_id: int = 0
+    _owner: object = field(default_factory=object)
+    _tasks: set[asyncio.Task[None]] = field(default_factory=set)
     _spawn: Callable[[Awaitable[None]], asyncio.Task[None]] | None = None
 
     def subscribe(self, predicate: Predicate, handler: Handler) -> SubscriptionHandle:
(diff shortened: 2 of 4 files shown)


Q00 (approved):
Reviewed AgentRuntimeContext and ControlBus. The context stays narrow, and ControlBus is appropriately scoped as an in-process transport. Terminology-wise, I would still keep ControlBus under the broader ControlPlane/ControlContract concept, but the code shape here looks good.

PR #557, opened 2026-05-01
docs(agent-os): lock kernel terminology

Q00:
## Summary

- Lock the Agent OS kernel vocabulary around `AgentRuntimeContext`, `ControlPlane`, `ControlContract`, `Directive`, `ControlBus`, and `IOJournal`.
- Clarify that `ControlBus` is an in-process transport inside the broader `ControlPlane`, not the top-level control concept.
- Add review invariants for control contracts, durable control decisions, resume preservation, and I/O journaling.

## Changes

- Add `docs/contributing/agent-os-kernel-terminology.md` as the canonical terminology reference.
- Link the terminology reference from the docs index.
- Point architecture readers to the locked Agent OS vocabulary.

## Notes

This is a docs-only terminology lock intended to guide the Agent OS PR stack before the naming hardens further in code.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -48,6 +48,7 @@ replayable execution contract on your choice of runtime backend.
 
 - [Contributing Guide](../CONTRIBUTING.md) - How to set up, code, test, and submit PRs
 - [Architecture for Contributors](./contributing/architecture-overview.md) - How modules connect
+- [Agent OS Kernel Terminology](./contributing/agent-os-kernel-terminology.md) - Locked vocabulary for `AgentRuntimeContext`, `ControlPlane`, `ControlContract`, `Directive`, `ControlBus`, and `IOJournal`
 - [Testing Guide](./contributing/testing-guide.md) - Writing and running tests
 - [Key Patterns](./contributing/key-patterns.md) - Result type, immutability, event sourcing, protocols
 - [Findings Registry](./contributing/findings-registry.md) - Documentation audit findings registry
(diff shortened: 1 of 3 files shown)


Q00:
@shaun0927 I wanted to leave the framing behind this small terminology PR, since it touches the Agent OS direction you have been building toward.

The main point is: `AgentRuntimeContext` feels like the right direction. It gives each agent/process a stable execution envelope instead of passing runtime dependencies and capabilities through scattered handler-specific arguments.

The part I think is worth locking early is the control vocabulary. `ControlBus` is a good implementation name for in-process pub/sub, but it feels too transport-shaped to carry the kernel-level concept by itself. I would prefer we reserve the higher-level language like this:

- `ControlPlane`: the architecture layer for workflow control.
- `ControlContract`: the stable schema/invariants for control behavior, including terminal semantics, retry budgets, timeout behavior, resume preservation, and required journaling.
- `Directive`: the small command vocabulary emitted by decision sites.
- `ControlBus`: the local in-process delivery mechanism inside the plane.
- `IOJournal`: the durable black box for LLM/tool I/O.
- `AgentRuntimeContext`: the execution envelope that exposes the relevant runtime capabilities and control surfaces.

That distinction matters because upcoming work like `ooo auto`, AgentProcess, mesh, and disposable memory will fail or succeed on OS-like contracts: resume must preserve runtime/cwd/MCP capability/safety options, terminal directives must really be terminal, and retries must be explainable from persisted control + I/O events rather than local process state.

So this PR is intentionally docs-only: it does not rename implementation yet, but it tries to prevent `ControlBus` from becoming the umbrella term before the architecture has a chance to settle around `ControlPlane` / `ControlContract`.

Issue #455, opened 2026-04-18
[Bug] ouroboros_interview fails with exit code 1 / empty stderr while ouroboros_qa works (v0.28.5)


jcfernandez-890825:
## Summary

`ouroboros_interview` fails on first call with `Claude Agent SDK request failed: Command failed with exit code 1`, empty stderr, despite the MCP server having `CLAUDECODE=1` in its environment. `ouroboros_qa` invoked in the same session with the same adapter pattern (`max_turns=1 + create_llm_adapter + Agent SDK subprocess`) works correctly and returns a full structured verdict. This strongly suggests the bug is InterviewHandler-specific, not environmental or shared-adapter-related.

## Environment

- **Ouroboros version:** v0.28.5 (installed via `uv tool install ouroboros-ai[mcp,claude]`)
- **OS:** Linux Mint 22 (kernel 6.8.0-107-generic)
- **Python:** 3.14 (uv tool venv at `~/.local/share/uv/tools/ouroboros-ai/`)
- **claude-agent-sdk:** 0.1.58
- **Claude Code:** 2.1.114
- **claude CLI:** `/home/batman/.local/bin/claude` v2.1.114
- **Config:** `~/.ouroboros/config.yaml` has `permission_mode: bypassPermissions`, `cli_path: /home/batman/.local/bin/claude`

## Steps to reproduce

1. Fresh Claude Code session.
2. Invoke `ouroboros_interview` MCP tool with any `initial_context` and `cwd`.
3. Observe failure with the error output below.
4. In the same session, invoke `ouroboros_qa` with any artifact + quality_bar.
5. Observe success (full QA verdict returned).

## Actual error

```
Error executing tool ouroboros_interview: Claude Agent SDK request failed:
Command failed with exit code 1 (exit code: 1)
Error output: Check stderr output for details
details: {
  'error_type': 'Exception',
  'session_id': 'ecc50fab-041b-4e45-b819-3180fe10b7b4',
  'stderr': '',
  'claudecode_present': True,
  'claude_code_entrypoint': 'cli'
}
```

## Diagnostic evidence

- Issue reproduces even after Claude Code full restart.
- MCP server process env has `CLAUDECODE=1` (verified via `/proc/<pid>/environ`).
- `ouroboros_qa` works in the same session, returning score 0.62 REVISE verdict on a dummy `hello_world()` artifact.
- `ouroboros_lateral_think` (templated, no Agent SDK subprocess) also works.
[...]

PR #563, opened 2026-05-02
fix(evolution): rebuild Reflect adapter on llm.backend config drift (#562)

### LINKED ISSUE #562: [Bug] EvolveStep Reflect uses stale llm_adapter when llm.backend config changes mid-session (v0.31.1)
## Summary

`mcp/server/adapter.py` wires the LLM adapter ONCE at MCP server creation:

```python
llm_adapter = create_llm_adapter(backend=llm_backend, max_turns=1, ...)
...
reflect_engine = ReflectEngine(llm_adapter=llm_adapter, ...)
evolutionary_loop = EvolutionaryLoop(..., reflect_engine=reflect_engine, ...)
```

When a user changes `~/.ouroboros/config.yaml` after the server started (e.g. switching `llm.backend: codex` → `claude_code` to work around an unrelated proxy issue), every subsequent `evolve_step` Reflect call still routes through the stale adapter.

`ouroboros_evaluate` does NOT exhibit this bug because `EvaluateHandler` constructs its own adapter per call inside `handle()` from `self.llm_backend` (which is `None` at init, so `create_llm_adapter(backend=None)` falls back to `get_llm_backend()` reading current config). Reflect should follow the same pattern.

## Environment

- Ouroboros: v0.31.1
- Python: 3.10.11 (also reproduces on 3.12)
- macOS Darwin 24.6.0

## Reproducer

1. Start Claude Code with `~/.ouroboros/config.yaml` containing `llm: { backend: claude_code }`.
2. Run any tool to materialize the cached adapter (interview/evolve_step).
3. While the session is still alive, edit the config: `llm: { backend: codex }`.
4. Call `ouroboros_evolve_step` → fails with the *claude_code* adapter's error (`Claude Agent SDK request failed: claudecode_present: False`) even though config now says codex.
5. Call `ouroboros_evaluate` in the same session → succeeds (uses current codex backend correctly).

Same external symptom appears whenever the captured backend stops working mid-session for any reason (proxy outage, model removal, env-var change picked up by `get_llm_backend()`).

## Relevant code

- `src/ouroboros/mcp/server/adapter.py:851-897` constructs the shared `llm_adapter` and the `reflect_engine` once at server creation.
- `src/ouroboros/evolution/reflect.py:77` stores the adapter as a dataclass field; `:119` uses `self.llm_adapter.complete(...)` directly.
- `src/ouroboros/mcp/tools/evaluation_handlers.py:499-503` shows the working pattern: `create_llm_adapter(backend=self.llm_backend, ...)` per call.

## Proposed fix

Make `ReflectEngine` (and similarly `WonderEngine`) honor live config without a server restart. PR #563 implements this for Reflect:

- Snapshot the configured backend in `__post_init__`.
- Optional `adapter_factory` callable for callers that want full per-call freshness.
- `_resolve_adapter()` rebuilds via `create_llm_adapter()` once when `get_llm_backend()` drifts from the captured snapshot, then updates the cache.

Behavior is unchanged when config does not drift, so direct `ReflectEngine` consumers and existing tests do not regress.

## Related

- #455 closed-but-unfixed; same external "stale-cached-adapter-after-config-change" pattern in InterviewHandler.
- #560 / #561 (codex adapter swallows stdout error events) — different bug, same family of "diagnostic clarity" issues.
- #519 (Stage-Runtime mapping) — once stage-specific runtime selection lands, this freshness logic still applies per stage.

## Checklist

- [x] Reproduced on v0.31.1
- [x] Verified `ouroboros_evaluate` already works correctly (different code path)
- [x] PR with tested fix prepared

minsing-jin:
## Summary

Fixes #562.

`ReflectEngine` captures `llm_adapter` at MCP server startup. When a user changes `~/.ouroboros/config.yaml` (e.g. switching `llm.backend` mid-session) without restarting Claude Code, every subsequent `evolve_step` Reflect call still uses the stale adapter — even though `ouroboros_evaluate` works correctly because `EvaluateHandler` constructs its own adapter per call.

This change makes Reflect honor live config without a server restart.

## What changed

`src/ouroboros/evolution/reflect.py`:

- Added `from ouroboros.config import get_llm_backend` for live backend reads.
- Snapshot the configured backend in `__post_init__` (via `_captured_backend`).
- New optional `adapter_factory` field — zero-arg callable for callers that want full per-call freshness (no behavior change when omitted).
- New `_resolve_adapter()` method:
  1. If `adapter_factory` is supplied, invoke it.
  2. Otherwise compare `get_llm_backend()` to the captured snapshot; if drifted, rebuild via `create_llm_adapter()` once, update both the cache and the snapshot.
  3. Otherwise return the originally injected `llm_adapter`.
- The `complete()` call site uses `self._resolve_adapter()` instead of `self.llm_adapter` directly.

Defensive: failures in factory/config-read paths log and fall back to the captured adapter, so the engine never crashes harder than before.

## Test plan

```bash
python -c \"import ast; ast.parse(open('src/ouroboros/evolution/reflect.py').read())\"
python -c \"from ouroboros.evolution.reflect import ReflectEngine\"
```

Both pass. Behavior is unchanged for callers that don't touch config mid-session (captured == current ⇒ identity returns the injected adapter), so existing engines/tests don't regress.

For full verification of the rebuild path, an integration test would need to mock `get_llm_backend()` returning a different value between calls and assert the adapter type changes — happy to add that if reviewers want it inline (kept the diff minimal here).

## Why this matters

Without this fix, any user who switches `llm.backend` without restarting Claude Code sees `evolve_step` keep failing on the previous backend's adapter, even when other tools (`ouroboros_evaluate`) work correctly with the new backend. The asymmetry is confusing and the workaround (full session restart) is heavyweight.

This also covers the broader \"cached-stale-adapter\" pattern that surfaced as #455 (closed-but-unfixed) and parts of #475.

## Follow-up

`WonderEngine` has the same caching pattern and should get the same treatment — happy to do that in a follow-up PR (omitted here to keep this PR focused on the Reflect path that actively blocks `evolve_step`).

### DIFF
--- a/src/ouroboros/evolution/reflect.py
+++ b/src/ouroboros/evolution/reflect.py
@@ -89,17 +89,18 @@ class ReflectEngine:
     llm_adapter: LLMAdapter
     model: str = field(default_factory=get_reflect_model)
     adapter_factory: Callable[[], LLMAdapter | None] | None = field(default=None)
+    adapter_backend: str | None = None
     _captured_backend: str | None = field(default=None, init=False, repr=False)
 
     def __post_init__(self) -> None:
         try:
-            self._captured_backend = get_llm_backend()
+            self._captured_backend = self.adapter_backend or get_llm_backend()
         except Exception:  # noqa: BLE001 — never fail engine init on config read
             self._captured_backend = None
 
     def _resolve_adapter(self) -> LLMAdapter:
         """Return the adapter the next ``complete()`` call should use."""
-        current_backend = self._current_backend()
+        current_backend = self._selected_backend()
         backend_drifted = (
             self._captured_backend is not None
             and current_backend
@@ -143,7 +144,9 @@ class ReflectEngine:
 
         return self.llm_adapter
 
-    def _current_backend(self) -> str | None:
+    def _selected_backend(self) -> str | None:
+        if self.adapter_backend is not None:
+            return self.adapter_backend
         try:
             return get_llm_backend()
         except Exception:  # noqa: BLE001
--- a/src/ouroboros/evolution/reflect.py
+++ b/src/ouroboros/evolution/reflect.py
@@ -12,6 +12,7 @@ Reflect handles all subsequent generations autonomously.
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass, field
 import json
 import logging
@@ -87,51 +88,48 @@ class ReflectEngine:
 
     llm_adapter: LLMAdapter
     model: str = field(default_factory=get_reflect_model)
-    adapter_factory: object = field(default=None)
+    adapter_factory: Callable[[], LLMAdapter | None] | None = field(default=None)
     _captured_backend: str | None = field(default=None, init=False, repr=False)
 
     def __post_init__(self) -> None:
-        # Snapshot the backend at construction time so we can detect drift.
         try:
             self._captured_backend = get_llm_backend()
         except Exception:  # noqa: BLE001 — never fail engine init on config read
             self._captured_backend = None
 
     def _resolve_adapter(self) -> LLMAdapter:
-        """Return the adapter the next ``complete()`` call should use.
+        """Return the adapter the next ``complete()`` call should use."""
+        current_backend = self._current_backend()
+        backend_drifted = (
+            self._captured_backend is not None
+            and current_backend
+            and current_backend != self._captured_backend
+        )
 
-        Priority:
-            1. ``adapter_factory()`` if supplied (always per-call fresh).
-            2. A freshly constructed adapter when the live ``llm.backend``
-               differs from the one captured at engine init.
-            3. The originally injected ``llm_adapter`` (today's behavior).
-        """
         if self.adapter_factory is not None:
             try:
                 fresh = self.adapter_factory()
                 if fresh is not None:
+                    if backend_drifted:
+                        self._captured_backend = current_backend
+                        self.model = get_reflect_model(current_backend)
                     return fresh
             except Exception:  # noqa: BLE001 — fall through to captured adapter
                 logger.exception("ReflectEngine adapter_factory raised; using captured adapter")
                 return self.llm_adapter
 
-        # Detect post-startup config drift and rebuild the adapter inline.
-        try:
-            current_backend = get_llm_backend()
-        except Exception:  # noqa: BLE001
-            return self.llm_adapter
-
-        if (
-            self._captured_backend is not None
-            and current_backend
-            and current_backend != self._captured_backend
-        ):
+        if backend_drifted:
             try:
                 from ouroboros.providers.factory import create_llm_adapter
 
-                rebuilt = create_llm_adapter(backend=current_backend, max_turns=1)
+                rebuilt = create_llm_adapter(
+                    backend=current_backend,
+                    cwd=_adapter_cwd(self.llm_adapter),
+                    max_turns=_adapter_max_turns(self.llm_adapter),
+                )
                 self.llm_adapter = rebuilt
                 self._captured_backend = current_backend
+                self.model = get_reflect_model(current_backend)
                 logger.info(
                     "reflect.adapter_rebuilt_for_backend_drift",
                     extra={"new_backend": current_backend},
@@ -145,6 +143,12 @@ class ReflectEngine:
                 return self.llm_adapter
 
         return self.llm_adapter
+
+    def _current_backend(self) -> str | None:
+        try:
+            return get_llm_backend()
+        except Exception:  # noqa: BLE001
+            return None
 
     async def reflect(
         self,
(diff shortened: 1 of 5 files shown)


shaun0927:
Autopilot follow-up: addressed the latest bot blocker on factory fallback coherence. Commit `1c9bb0f4` now treats every non-null Reflect/Wonder `adapter_factory` result as the latest known-good `llm_adapter`, so a later transient factory failure falls back to the fresh adapter instead of the stale startup adapter after backend/model state has moved. Added regression coverage for both ReflectEngine and WonderEngine. Verification: `uv run pytest tests/unit/evolution/test_backend_drift.py -q`; `uv run pytest tests/unit/evolution tests/unit/mcp/server/test_adapter.py tests/integration/mcp/test_server_adapter.py -q`; `uv run mypy src/ouroboros/evolution/reflect.py src/ouroboros/evolution/wonder.py src/ouroboros/mcp/server/adapter.py`; `uv run ruff format --check src/ tests/`; `uv run ruff check src/ tests/`; `git diff --check`; all GitHub checks are green on latest head.

shaun0927:
Autopilot follow-up: addressed the latest bot design-note blocker with commit `f02cb4d2`. Reflect/Wonder now refresh model selection on every successful per-call adapter factory result, including same-backend `resilience.reflect_model` / `resilience.wonder_model` or env-var changes, and construct `CompletionConfig` only after adapter resolution so the refreshed model is the one sent to the provider. Added regression coverage proving the refreshed model reaches the actual completion config for both engines, and adjusted explicit-backend tests to keep model resolution pinned to the selected backend.\n\nVerification: `uv run pytest tests/unit/evolution/test_backend_drift.py -q`; `uv run pytest tests/unit/evolution tests/unit/mcp/server/test_adapter.py tests/integration/mcp/test_server_adapter.py -q`; `uv run mypy src/ouroboros/evolution/reflect.py src/ouroboros/evolution/wonder.py src/ouroboros/mcp/server/adapter.py`; `uv run ruff format --check src/ tests/`; `uv run ruff check src/ tests/`; `git diff --check`. GitHub checks are green on latest head; stale bot change-request reviews on superseded commits were dismissed with the same rationale.

shaun0927:
@Q00 Could you please re-review this PR?

PR #563 fixes the stale LLM adapter issue in MCP `evolve_step` reported in #562. The original problem was that ReflectEngine/WonderEngine kept using the `llm_adapter` and backend/model settings captured at MCP server startup. As a result, even after a user changed `llm.backend` or the evolution model settings in `~/.ouroboros/config.yaml`, Reflect/Wonder could keep calling the previous provider/model until the server was restarted.

This PR now improves that boundary to the following level:

- Reflect/Wonder can receive a fresh adapter through a per-call `adapter_factory`.
- MCP wiring creates fresh adapters while preserving `cwd=effective_cwd`, `max_turns`, and explicit `llm_backend` overrides.
- On backend drift, it refreshes the backend-specific model along with the adapter rather than swapping only the adapter.
- Same-backend changes to `resilience.reflect_model` / `resilience.wonder_model` or the corresponding env vars are also reflected.
- `CompletionConfig` is constructed only after adapter/model resolution, so the refreshed model is the one actually sent to the provider.
- If a transient factory failure occurs, the engine falls back to the latest successful adapter instead of the stale startup adapter.
- I also addressed the latest bot design-note follow-up warning: direct non-MCP consumers that pass `ReflectEngine(..., model=...)` / `WonderEngine(..., model=...)` keep that explicit model override as pinned caller intent, while the MCP path omits explicit model arguments and continues to use config-derived live model refresh.

Validation evidence:

- Latest head: `9d96ffde5c567d4ded569ed25b2b14c26dcb54b2`
- GitHub: `MERGEABLE`, review decision `APPROVED`
- GitHub checks are green: Ruff Lint, MyPy Type Check, Bridge TypeScript, Python 3.12/3.13/3.14 tests
- Local verification:
  - `uv run pytest tests/unit/evolution/test_backend_drift.py -q` → 13 passed
  - `uv run pytest tests/unit/evolution tests/unit/mcp/server/test_adapter.py tests/integration/mcp/test_server_adapter.py -q` → 93 passed
  - `uv run mypy src/ouroboros/evolution/reflect.py src/ouroboros/evolution/wonder.py src/ouroboros/mcp/server/adapter.py` → success
  - `uv run ruff format --check src/ tests/` → pass
  - `uv run ruff check src/ tests/` → pass
  - `git diff --check` → pass

My assessment is that all bot blocking findings have been resolved, and the final design-note warning was addressed in follow-up commit `9d96ffde` without regressing the explicit model override contract. This PR looks merge-ready to me.

Q00 (approved):
Reviewed fetched head 9d96ffd. Verification passed: backend-drift/evolution/MCP adapter suites (93 passed) and ruff check on the changed Reflect/Wonder/MCP files. Explicit model/backend override behavior is preserved; no blocking findings.

PR #583, opened 2026-05-02
fix(events): fail closed for invalid IO journal preview mode

### LINKED ISSUE #580: Fail closed for invalid OUROBOROS_IO_JOURNAL_PREVIEWS values
## Summary

Post-v0.31.1 audit found that `OUROBOROS_IO_JOURNAL_PREVIEWS` currently fails open on unrecognized values. If an operator explicitly configures a typo such as `of`, `OFF `, or another invalid value, `get_privacy_mode()` falls back to `PrivacyMode.ON`, which preserves I/O previews instead of disabling or redacting them.

This is not a remote exploit by itself, but it is a privacy-control footgun for the new I/O Journal surface.

## Evidence

- `src/ouroboros/events/io.py`: `get_privacy_mode()` catches invalid enum values and returns `PrivacyMode.ON`.
- `src/ouroboros/events/io.py`: preview shaping includes content previews when mode is `ON`.
- `src/ouroboros/events/io_recorder.py`: recorder paths pass raw prompts, completions, tool args, and tool results into the event factories.
- `tests/unit/events/test_io_events.py`: current tests lock the invalid-env fallback to `ON`.

Read-only reproduction during audit: with `OUROBOROS_IO_JOURNAL_PREVIEWS=of`, `get_privacy_mode()` returned `on` and emitted a prompt preview containing the test secret string.

## Impact

A misconfigured privacy environment variable can silently preserve journal previews even though the operator attempted to disable/redact them. For an Agent OS-style runtime, control-plane and journal privacy settings should fail closed when explicitly configured incorrectly.

## Expected behavior

- Unset env should keep the current default if desired.
- Explicit but unrecognized env values should fail closed to `off` or `redacted`, not `on`.
- Values should be normalized with at least `strip().lower()`.
- A warning should be emitted when falling back because of an invalid value.

## Acceptance criteria

- Invalid explicit `OUROBOROS_IO_JOURNAL_PREVIEWS` values do not preserve raw previews.
- Whitespace/case variants of valid values are handled deterministically.
- Tests cover unset, valid `on`, valid `redacted`, valid `off`, and invalid explicit values.
- The behavior is documented in or near the env parser so future privacy modes do not reintroduce fail-open behavior.

## Minimal fix scope

Start in `src/ouroboros/events/io.py` and the focused event tests. This should be a small, isolated bugfix PR.

Q00:
## Summary

- normalize `OUROBOROS_IO_JOURNAL_PREVIEWS` with `strip().lower()`
- keep unset env defaulting to `on`
- fail closed to `off` for explicit invalid values and log a warning
- add coverage that invalid values do not preserve raw previews

Closes #580.

## Verification

- `uv run pytest tests/unit/events/test_io_events.py -q`
- `uv run pytest tests/unit/events/test_io_recorder.py -q`
- `uv run ruff check src/ouroboros/events/io.py tests/unit/events/test_io_events.py`

### DIFF
--- a/src/ouroboros/events/io.py
+++ b/src/ouroboros/events/io.py
@@ -46,6 +46,7 @@ from typing import Any, Final
 from ouroboros.events.base import BaseEvent
 
 logger = logging.getLogger(__name__)
+_WARNED_INVALID_PRIVACY_VALUES: set[str] = set()
 
 # ---------------------------------------------------------------------------
 # Constants
@@ -111,12 +112,14 @@ def get_privacy_mode() -> PrivacyMode:
     try:
         return PrivacyMode(raw)
     except ValueError:
-        logger.warning(
-            "Invalid %s=%r; falling back to %s to avoid preserving I/O previews",
-            PRIVACY_ENV_VAR,
-            configured,
-            PrivacyMode.OFF.value,
-        )
+        if raw not in _WARNED_INVALID_PRIVACY_VALUES:
+            _WARNED_INVALID_PRIVACY_VALUES.add(raw)
+            logger.warning(
+                "Invalid %s=%r; falling back to %s to avoid preserving I/O previews",
+                PRIVACY_ENV_VAR,
+                configured,
+                PrivacyMode.OFF.value,
+            )
         return PrivacyMode.OFF
 
 
--- a/src/ouroboros/events/io.py
+++ b/src/ouroboros/events/io.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 from enum import StrEnum
 import hashlib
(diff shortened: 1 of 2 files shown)


Q00:
Addressed the non-blocking review note in 1cc66a31. Invalid `OUROBOROS_IO_JOURNAL_PREVIEWS` values now warn once per normalized invalid value instead of logging on every preview-shaping call.

Focused verification:
- `uv run pytest tests/unit/events/test_io_events.py tests/unit/events/test_io_recorder.py -q`
- `uv run ruff check src/ouroboros/events/io.py tests/unit/events/test_io_events.py`

PR #584, opened 2026-05-02
fix(orchestrator): own ControlBus shutdown lifecycle

### LINKED ISSUE #582: ControlBus should be an owned lifecycle resource and drain subscriber tasks on shutdown
## Summary

Post-v0.31.1 audit found that `ControlBus.publish()` fans out subscriber callbacks into background tasks, but the bus has no lifecycle API and is not registered as an MCP server owned resource. As a result, subscriber tasks can survive server shutdown and continue running after resources they depend on are closed.

For an Agent OS control-plane primitive, this leaves the transport below the ownership/lifecycle standard expected from the runtime context.

## Evidence

- `src/ouroboros/orchestrator/control_bus.py`: `publish()` creates one task per matching subscription.
- `src/ouroboros/orchestrator/control_bus.py`: `_spawn_handler()` uses `asyncio.ensure_future()` and tracks tasks in `_tasks`, but there is no `close()`, `drain()`, bounded queue, semaphore, or shutdown timeout.
- `src/ouroboros/mcp/server/adapter.py`: `shutdown()` closes only `_owned_resources`.
- `src/ouroboros/mcp/server/adapter.py`: `event_store` and `brownfield_store` are registered as owned resources.
- `src/ouroboros/mcp/server/adapter.py`: `AgentRuntimeContext(control=ControlBus())` creates the bus, but the bus is not registered as an owned resource.

Runtime probe during audit: after publishing a blocking subscriber and calling `server.shutdown()`, the subscriber task remained unfinished and tracked by the bus.

## Impact

A control subscriber can outlive the MCP server shutdown path. That creates possible resource-use-after-close behavior, background task leaks, and nondeterministic behavior in tests or long-running Agent OS sessions. It also makes ControlBus feel like a utility object rather than a runtime-owned control-plane component.

## Expected behavior

- The runtime should own the `ControlBus` lifecycle.
- Shutdown should stop new publishes, clear subscriptions, and drain or cancel subscriber tasks deterministically.
- The shutdown order should account for resources subscriber tasks may still use, especially `event_store` and bridge resources.

## Acceptance criteria

- Add an idempotent `ControlBus.close()` and/or `ControlBus.drain()` lifecycle API.
- Close/drain marks the bus closed and prevents new publish fan-out after shutdown starts.
- Pending subscriber tasks are awaited with a bounded timeout and stragglers are cancelled.
- `create_ouroboros_server()` registers the runtime context bus as an owned resource, or otherwise guarantees it is closed on server shutdown.
- Tests cover normal drain, timeout cancellation, publish-after-close behavior, and server shutdown draining the runtime context bus.

## Related context

Adjacent but not duplicate: #515 / #524 introduced the initial `ControlBus`; #574 and #575 cover higher-level control contract/journal semantics. This issue is specifically about the in-process transport lifecycle and shutdown ownership.

Q00:
## Summary

- add idempotent `ControlBus.close()` / `drain()` lifecycle cleanup
- reject new subscriptions after close and make publish-after-close a no-op
- await subscriber tasks with a bounded timeout, then cancel stragglers
- register the runtime `ControlBus` as a server-owned resource before stores/bridges it may be used with

Closes #582.

## Verification

- `uv run pytest tests/unit/orchestrator/test_control_bus.py -q`
- `uv run pytest tests/unit/mcp/server/test_adapter.py -q`
- `uv run pytest tests/unit/mcp/bridge/test_handler_wiring.py tests/unit/mcp/tools/test_mcp_manager_wiring.py tests/unit/orchestrator/test_agent_runtime_context.py -q`
- `uv run ruff check src/ouroboros/orchestrator/control_bus.py src/ouroboros/mcp/server/adapter.py tests/unit/orchestrator/test_control_bus.py tests/unit/mcp/server/test_adapter.py`

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -36,7 +36,7 @@ from ouroboros.mcp.types import (
     ToolInputType,
 )
 from ouroboros.orchestrator.agent_runtime_context import AgentRuntimeContext
-from ouroboros.orchestrator.control_bus import ControlBus
+from ouroboros.orchestrator.control_bus import ControlBus, ControlBusDrainError
 
 log = structlog.get_logger(__name__)
 
@@ -721,6 +721,12 @@ class MCPServerAdapter:
             if callable(close_fn):
                 try:
                     await close_fn()
+                except ControlBusDrainError:
+                    log.error(
+                        "mcp.server.control_bus_close_failed",
+                        resource=type(resource).__name__,
+                    )
+                    raise
                 except Exception as exc:
                     log.warning(
                         "mcp.server.resource_close_failed",
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -1467,24 +1467,27 @@ def create_ouroboros_server(
         rate_limit_config=rate_limit_config,
     )
 
-    # The server owns the shared event store lifecycle
-    server.register_owned_resource(event_store)
-    if brownfield_store is not None:
-        server.register_owned_resource(brownfield_store)
-
     # Build the AgentRuntimeContext that #474 funnels through every
     # handler. For now the context only exposes the EventStore, the
     # backend labels, the optional MCP bridge, and a fresh ControlBus
     # for #515. Subsequent migration slices move handler internals to
     # consume context.mcp_bridge directly instead of self.mcp_manager.
+    control_bus = ControlBus()
     agent_runtime_context = AgentRuntimeContext(
         event_store=event_store,
         runtime_backend=resolved_runtime_backend,
         llm_backend=llm_backend,
         mcp_bridge=mcp_bridge,
-        control=ControlBus(),
+        control=control_bus,
     )
     server.set_runtime_context(agent_runtime_context)
+
+    # Close the reactive control surface before stores/bridges it may
+    # reference from subscriber tasks.
+    server.register_owned_resource(control_bus)
+    server.register_owned_resource(event_store)
+    if brownfield_store is not None:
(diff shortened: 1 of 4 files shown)


Q00:
Addressed the blocking drain issue in d6947512. `ControlBus.drain()` now has a second bounded cancel grace timeout after `task.cancel()`, logs stubborn tasks instead of awaiting them indefinitely, and includes a regression test for subscribers that catch `CancelledError` and keep running.

Focused verification:
- `uv run pytest tests/unit/orchestrator/test_control_bus.py tests/unit/mcp/server/test_adapter.py -q`
- `uv run ruff check src/ouroboros/orchestrator/control_bus.py tests/unit/orchestrator/test_control_bus.py tests/unit/mcp/server/test_adapter.py`

Q00:
Adjusted the shutdown policy again in 72f461fc based on the second review.

New behavior:
- `ControlBus.drain()` still has bounded wait + bounded cancel grace.
- If a subscriber ignores cancellation past the grace window, `ControlBusDrainError` is raised instead of returning while the task is still live.
- `MCPServerAdapter.shutdown()` does not swallow `ControlBusDrainError`, so it stops before closing dependent resources such as the event store / bridge.
- Added regression coverage proving shutdown does not proceed to close a dependent resource while a control subscriber is still running.

Focused verification:
- `uv run pytest tests/unit/orchestrator/test_control_bus.py tests/unit/mcp/server/test_adapter.py -q`
- `uv run ruff check src/ouroboros/orchestrator/control_bus.py src/ouroboros/mcp/server/adapter.py tests/unit/orchestrator/test_control_bus.py tests/unit/mcp/server/test_adapter.py`

PR #590, opened 2026-05-03
fix(persistence): include sessions without started events

### LINKED ISSUE #508: Orphan detection ignores sessions without an explicit `orchestrator.session.started` event
## Background

While reviewing #311 / v0.30.0's snapshot-based orphan detection, the bot review surfaced a Low-severity sidecase that still applies on main:

- `EventStore.get_session_activity_snapshots()` (`src/ouroboros/persistence/event_store.py:445+`) anchors its candidate set on \`event_type == \"orchestrator.session.started\"\`.
- Any aggregate whose stream **does not contain a \`started\` event** but does carry later lifecycle events (e.g. imported / migrated / partial-history streams) is silently invisible to orphan cleanup.
- This is structurally the same Low-#5 finding called out on PR #311; #326 inherited the same anchor.

## Impact

Low — affects only non-canonical / imported event streams. Canonical sessions always emit \`started\` first. But it means:

1. Cleanup can leave stale RUNNING/PAUSED markers if a stream is missing \`started\`.
2. Audit/reporting that derives from the snapshot will under-count such sessions.

## Proposed fix (one of)

- **A. Anchor on \`aggregate_type == \"session\"\`** instead of the \`started\` event type, taking the earliest event per aggregate as the synthetic start row (with \`session_id\` from \`aggregate_id\`, fallback \`start_time\` from the row's timestamp).
- **B. Keep the \`started\` anchor for the typed identity fields, but \`UNION\` in any aggregate that has session events without a \`started\` row, marking those as origin-unknown.

A is simpler and matches the pre-#326 enumerator semantics more closely.

## Test coverage to add

- Snapshot row produced for an aggregate that has only \`session.completed\` / \`session.failed\` events (no \`started\`).
- Snapshot row produced for an aggregate that has \`session.heartbeat\` / progress events but no \`started\`.
- Existing canonical (started → … → terminal) path still passes.

## References

- PR #311 bot review #2, finding 5 — [link](https://github.com/Q00/ouroboros/pull/311#pullrequestreview-)
- PR #326 / v0.30.0 — main implementation
- Issue #310 — original startup-timeout report

shaun0927:
## Summary

- Enumerate session snapshots from all `aggregate_type == "session"` streams instead of anchoring only on `orchestrator.session.started`
- Preserve started-event metadata when present, with timestamp fallback for imported/partial streams
- Add regression coverage for terminal-only and progress-only session aggregates

## Verification

- `uv run ruff check src/ouroboros/persistence/event_store.py tests/unit/persistence/test_event_store.py`
- `uv run mypy src/ouroboros/persistence/event_store.py`
- `uv run pytest tests/unit/persistence/test_event_store.py::TestSessionActivitySnapshots -q`

Closes #508.

### DIFF
--- a/src/ouroboros/persistence/event_store.py
+++ b/src/ouroboros/persistence/event_store.py
@@ -535,12 +535,15 @@ class EventStore:
             func.json_extract(events_table.c.payload, "$.runtime_status"),
         )
 
-        started_ranked = (
+        first_session_event_ranked = (
             select(
                 events_table.c.aggregate_id.label("session_id"),
                 func.json_extract(events_table.c.payload, "$.execution_id").label("execution_id"),
                 func.json_extract(events_table.c.payload, "$.seed_id").label("seed_id"),
-                func.json_extract(events_table.c.payload, "$.start_time").label("start_time"),
+                func.coalesce(
+                    func.json_extract(events_table.c.payload, "$.start_time"),
+                    events_table.c.timestamp,
+                ).label("start_time"),
                 func.row_number()
                 .over(
                     partition_by=events_table.c.aggregate_id,
@@ -549,7 +552,6 @@ class EventStore:
                 .label("rn"),
             )
             .where(events_table.c.aggregate_type == "session")
-            .where(events_table.c.event_type == "orchestrator.session.started")
             .subquery()
         )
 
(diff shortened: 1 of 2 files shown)


Q00 (approved):
Reviewed fetched head aa2d9a2. Targeted verification passed: pytest tests/unit/persistence/test_event_store.py::TestSessionActivitySnapshots (3 passed) and ruff check on event_store/test_event_store. No blocking findings.

Issue #598, opened 2026-05-04
Plugin hooks broken on Windows (v0.31.x): wrong script path + UnicodeEncodeError on non-UTF-8 locales (regression from #86)


Flangom:
## Summary

The Python-based plugin hooks shipped in v0.31.x fail on every prompt submission / Write|Edit tool use on Windows. There are three independent issues stacked on top of each other, the worst of which is a regression of the fix for #86.

## Environment

- OS: Windows 11 (Korean locale, system codepage cp949)
- Python: 3.14 at `C:\Python314\python.exe` (also `python3` Windows App alias)
- Ouroboros plugin: 0.31.1 (installed via `claude plugin install ouroboros@ouroboros`)
- Claude Code: latest

## Symptoms

Every user prompt and every Write/Edit tool call shows:

```
UserPromptSubmit hook error
Failed with non-blocking status code: Traceback (most recent call last):
```

Non-blocking, but noisy on every turn.

### 1. Hook command uses `CLAUDE_PROJECT_DIR` + relative script path (regression of #86)

`.claude/settings.json` in the plugin ships:

```json
"command": "cd \"${CLAUDE_PROJECT_DIR:-.}\" && python3 scripts/keyword-detector.py"
```

`CLAUDE_PROJECT_DIR` resolves to the *user's* project (e.g. `C:\my\unrelated\repo`), where `scripts/keyword-detector.py` does not exist. The actual script lives in the plugin cache (`~/.claude/plugins/cache/ouroboros/ouroboros/0.31.1/scripts/`).

Issue #86 was closed by switching to `${CLAUDE_PLUGIN_ROOT}/scripts/keyword-detector.mjs`. The Node→Python migration in 0.31.x re-introduced the bug — `${CLAUDE_PLUGIN_ROOT}` was replaced with `cd "${CLAUDE_PROJECT_DIR:-.}"`.

### 2. UnicodeEncodeError on non-UTF-8 Windows locales

Even if you fix the path, `keyword-detector.py` prints emoji (`🎯`, U+1F3AF) to stdout. On Windows with cp949 (Korean), cp936 (Chinese), shift-jis (Japanese), etc., Python's default stdout encoding follows the system codepage and crashes:

```
UnicodeEncodeError: 'cp949' codec can't encode character '\U0001f3af' in position 32: illegal multibyte sequence
```

This is a real Python traceback on every prompt — the script never reaches stdin pass-through.

### 3. `python3` on Windows often points to a Microsoft Store stub

[...]

PR #603, opened 2026-05-04
fix(claude): fall back to python for hook executable

Q00:
## Summary
- keep `python3` as the preferred hook interpreter on Unix-like systems
- fall back to unversioned `python` when `python3` fails or is unavailable (covers the Windows Store alias case from #598)
- extends hook settings regression coverage for the interpreter fallback

References #598

Now based directly on `main` after #601 and #602 landed.

## Test Plan
- `uv run ruff format --check tests/unit/scripts/test_claude_hook_settings.py`
- `uv run ruff check tests/unit/scripts/test_claude_hook_settings.py`
- `uv run pytest tests/unit/scripts/test_claude_hook_settings.py -q`
- manual fallback simulation with a fake failing `python3` ahead of PATH for both hook commands

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "python3 \"${CLAUDE_PLUGIN_ROOT}/scripts/keyword-detector.py\"",
+            "command": "python3 \"${CLAUDE_PLUGIN_ROOT}/scripts/keyword-detector.py\" || python \"${CLAUDE_PLUGIN_ROOT}/scripts/keyword-detector.py\"",
             "timeout": 5
           }
         ]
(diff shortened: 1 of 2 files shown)


Q00:
Addressed the requested change in `849df4e`.

Instead of replacing `python3` with hard-coded `python`, the hook command now keeps the existing Unix-friendly `python3` path and falls back to `python` only if `python3` fails/is unavailable:

```sh
python3 "${CLAUDE_PLUGIN_ROOT}/scripts/..." || python "${CLAUDE_PLUGIN_ROOT}/scripts/..."
```

Also narrowed the test wording/assertions to match that contract and verified the Windows Store-alias style failure with a fake failing `python3` in PATH.

Q00:
Rebased this PR onto current `main` after #601/#602 landed and retargeted the PR base from `fix/598-plugin-hook-root` to `main`.

New head: `7bf23ff`

Verified the PR diff is now only the interpreter fallback change against `main`, with no stacked-base conflict:

```text
.claude/settings.json                           | 4 ++--
tests/unit/scripts/test_claude_hook_settings.py | 8 ++++++++
```

Local checks passed again, including fake failing `python3` fallback simulation for both hook commands.

Issue #528, opened 2026-04-30
`ooo ralph` is documented as a persistent MCP loop, but does not actually loop reliably


andrew-adamson:
## Problem

I tried to use Ralph in the intended high-level way:

1. Had the agent reason that Ralph was the right Ouroboros workflow.
2. Restarted the chat/session after installing Ouroboros so the `ooo` commands were available.
3. Ran `ooo ralph ...`.

Expected: Ralph starts a persistent loop and continues running generations until pass/convergence/limit.

Actual: it did not loop as documented.

This is not just an onboarding/docs issue. I found the intended advanced workflow, restarted so the skills/routing were available, and used the documented command. The command still did not behave like a persistent loop.

## Current Contract

The docs currently describe Ralph as a persistent loop:

- `ooo ralph` runs the evolutionary loop persistently until convergence.
- `ooo evolve` is a single evolution step.
- `ooo ralph` is listed as available “via MCP” / “MCP only”.

But the implementation appears to rely on client-side skill behavior repeatedly calling one-generation MCP tools. The MCP server does not appear to expose a first-class `ouroboros_ralph` tool.

That makes the user contract fragile: if the client skill is stale, not loaded, not followed exactly, or routes incorrectly, Ralph does not actually loop.

## Workaround

My workaround was to use the CLI Ralph script instead of `ooo ralph`.

However, I am on Windows and working natively in PowerShell (no WSL). The existing script path is shell/Python-oriented, so I had to make a hacky local PowerShell port of `ralph.sh` / `ralph.py` to `ralph.ps1`.

That workaround is not release-ready, and needing it undermines the documented “MCP-only” Ralph workflow.

## Related Issues

This relates to #362 because users need examples and guidance for using Ouroboros beyond basic coding tasks. But this case is stronger than documentation/discovery: I discovered the intended workflow and the advertised command still did not work as expected.

This also relates to #475 and the AgentRuntimeContext / ControlBus / AgentProcess direction. That may be the correct long-term architecture for Ralph, but the immediate user-facing requirement is simpler:

`ooo ralph` just needs to work.

## Suggested Fix

Add a first-class MCP-owned Ralph loop, exposed as a public MCP tool:

- `ouroboros_ralph`: starts a job-managed Ralph loop and returns a `job_id`.
- Internally, implement the loop as a reusable `RalphLoopRunner` so stop conditions and multi-generation behavior can be tested directly.
- Route `ooo ralph` to `ouroboros_ralph`.

[...]

Issue #518, opened 2026-04-29
M6 AgentProcess lifecycle (lite) — spawn/pause/resume/cancel/replay


shaun0927:
# M6 AgentProcess lifecycle (lite) — spawn/pause/resume/cancel/replay

## Context

RFC #476 promotes process lifecycle to **Tier-1 Must (M6)**:

> **M6. Process Lifecycle (lite).** Workflow = `AgentProcess` with
> `spawn / pause / resume / cancel / replay`. `ralph`, `evolve_step`,
> `execute_seed` share the abstraction.
>
> *Pass:* pause `ralph`, inspect, resume — no context loss.

Today these three workflows are separate code paths with separate
cancellation, no shared pause/resume primitive, and no shared replay
contract. Adding a new workflow means re-inventing lifecycle wiring each
time — exactly the problem #474 is solving for capability access, applied
to lifecycle.

This issue introduces the **`AgentProcess`** abstraction in lite form: an
opaque handle whose 5 verbs every long-running workflow shares. It is
**not** a fork/branch primitive (that is Tier-3 C1, gated by evidence).

## Scope

1. **`AgentProcess` interface**
   - `spawn(intent, context: AgentRuntimeContext) → AgentProcessHandle`
   - `handle.pause()` — durable; survives a process restart if the
     orchestrator persists the pause directive (see #514)
   - `handle.resume()` — picks up at the same step
   - `handle.cancel()` — terminal; emits
     `control.directive.emitted CANCEL` and stops the work loop
   - `handle.replay()` — replays from the EventStore; deterministic up to
     the I/O journal preview fidelity (see M3 issue)
   - `handle.status() → AgentProcessStatus` — `running / paused /
     cancelled / completed / failed`

2. **Directive integration**
   - Every lifecycle transition emits a `control.directive.emitted` event
     with `target_type="agent_process"` and `target_id=<process_id>`.
     This is the first non-forward-compatible use of the
     `agent_process` target_type declared in `events/control.py`.
   - Existing `Directive` members map directly:
     `WAIT` (paused), `CONTINUE` (resumed), `CANCEL` (cancelled),
     `CONVERGE` (completed), `RETRY` / `EVOLVE` (loop iterations,
     emitted by callers, not by lifecycle).

3. **Three reference migrations**
   - `ralph` → `AgentProcess` (loop until verified)
   - `evolve_step` → `AgentProcess` (single generation)
   - `execute_seed` → `AgentProcess` (single execution)
[...]

PR #617, opened 2026-05-05
feat(mcp): add a first-class Ralph loop job

shaun0927:
## Summary

Adds a first-class MCP-owned Ralph loop surface:

- `RalphLoopRunner` runs repeated `evolve_step` generations until QA passes, convergence occurs, a terminal evolution action occurs, or `max_generations` is reached.
- `RalphHandler` exposes `ouroboros_ralph` as a background JobManager job and returns a `job_id` immediately.
- The default MCP tool registry and server adapter register the new tool.
- The existing watchdog timeout test now uses file-backed SQLite so multi-version CI does not hit per-connection `:memory:` schema loss during cancellation cleanup.

## Why

#528 reports that `ooo ralph` was documented as a persistent MCP loop but the server only exposed one-generation evolve tools. This PR adds the missing runtime-owned loop primitive so clients no longer have to own the multi-generation loop in prompt pseudo-code.

## Scope

This is the runtime/tool slice only. It does not migrate Ralph onto `AgentProcess` yet; durable pause/resume/replay remains follow-up work under #518.

## Verification

- `uv run ruff check --fix src/ouroboros/ralph_loop.py src/ouroboros/mcp/tools/ralph_handlers.py src/ouroboros/mcp/tools/definitions.py src/ouroboros/mcp/tools/__init__.py src/ouroboros/mcp/server/adapter.py tests/unit/mcp/tools/test_ralph_handler.py tests/unit/mcp/tools/test_definitions.py`
- `uv run pytest tests/unit/evolution/test_generation_watchdog.py tests/unit/mcp/tools/test_ralph_handler.py tests/unit/mcp/tools/test_definitions.py tests/unit/mcp/tools/test_round5_fixes.py tests/integration/mcp/test_server_adapter.py -q` — 186 passed

## Follow-up

- #618 updates `skills/ralph/SKILL.md` to call `ouroboros_ralph` directly and poll `ouroboros_job_wait` / `ouroboros_job_result`.
- Wire Ralph through `AgentProcess` after #518 durability/replay slices land.

Addresses #528.


shaun0927:
@ouroboros-agent please re-review the latest head (9cb0c706). The stale blocking review was for 6b480a9; this head adds the outer plugin-mode Ralph dispatch boundary, a generation cap, nested-delegation guard, seed prompt-data hardening, and regression coverage. All GitHub checks are now green.

shaun0927:
@ouroboros-agent please re-review latest head 1106923. It specifically addresses your 9cb0c70 blockers: max_generations now completes normally, public docs reference ouroboros_cancel_job and plugin job_id=None, and QA-pass termination is covered by regression tests. All GitHub checks are green.

shaun0927:
@ouroboros-agent please re-review latest head dc9c851. It addresses the 1106923 blocker by enforcing Ralph recursion in the OpenCode bridge itself: Ralph child session IDs are tracked, and any child attempt to dispatch ouroboros_ralph is rejected before a nested child can be created. Bridge TS checks and tests are green, and all GitHub checks are green.

shaun0927:
@ouroboros-agent please re-review latest head f4f6a76. It addresses the dc9c851 ancestry gap by propagating Ralph-owned session state to every descendant created from a Ralph-owned session, so any later ouroboros_ralph call from a child/grandchild path is rejected before dispatch. Bridge TS checks/tests and all GitHub checks are green.

shaun0927:

This PR adds `ouroboros_ralph` as a first-class MCP tool/job so Ralph loops are owned by the runtime instead of client-side skill pseudo-code repeatedly calling `evolve_step`.

Since the initial bot review, the PR has been hardened in the following ways:

- Preserved the existing OpenCode plugin/subprocess boundary: in plugin mode `RalphHandler` returns a single `_subagent` delegation with `job_id=None` instead of enqueueing a local background job whose inner `evolve_step` would only return delegation envelopes.
- Added a Ralph-specific plugin payload and regression coverage for plugin delegation, factory wiring, and payload shape.
- Capped `max_generations` at 1-10, encoded seed content as JSON data with backtick escaping, and added explicit non-nested Ralph context.
- Clarified completion semantics: reaching `max_generations` is now a normal completed stop condition, while true failed/interrupted/stagnated/exhausted actions still fail the loop.
- Fixed the public contract text to point users to `ouroboros_cancel_job` and to describe the plugin-mode `job_id=None` behavior.
- Enforced the nested-Ralph safety boundary in the OpenCode bridge itself: Ralph-owned child sessions propagate ancestry to descendants, and any `ouroboros_ralph` dispatch from that ancestry is rejected before a new child session can be created.

Validation evidence:

- GitHub checks are all green: Ruff Lint, MyPy Type Check, Bridge TypeScript, and Tests on Python 3.12 / 3.13 / 3.14.
- Local full suite: `uv run pytest -q` → 6326 passed, 2 skipped.
- Targeted MCP suite: `251 passed`.
- Bridge checks: `bunx tsc --noEmit` and `bun test` → 96 passed.
- `ouroboros-agent[bot]` re-reviewed latest head `f4f6a76` and returned **APPROVE**.

Given the runtime-boundary fix, bounded execution safeguards, updated public contract, regression tests, and green CI, I think this PR is now merge-ready.

PR #627, opened 2026-05-05
test(ralph): lock job cancellation contract

shaun0927:
## Summary

Adds focused regression coverage for the MCP-owned Ralph loop introduced for #528: a running `ouroboros_ralph` background job can be cancelled through the existing `JobManager` / job-tool contract, and the terminal job snapshot reports `cancelled` with the Ralph lineage link preserved.

## Why

#528 should close with Ralph converging on the existing MCP job surface rather than adding a Ralph-specific control path. This PR verifies that the new server-owned Ralph loop can be stopped through the same cancellation/status mechanism used by other long-running MCP jobs.

## Scope

- Adds a blocking fake evolve handler for deterministic cancellation timing.
- Starts `RalphHandler`, captures its `job_id`, cancels through `JobManager.cancel_job()`, and asserts the job reaches `JobStatus.CANCELLED`.
- Confirms the linked `lineage_id` remains visible for later status/result inspection.

## Risks

- This is unit-level coverage; it does not prove a live client can interrupt a real long-running model/tool call mid-flight.
- It intentionally does not add Ralph-specific pause/resume/cancel APIs; broader lifecycle controls remain under #518 / `AgentProcess`.
- The assertion allows the immediate cancel response to be either `cancel_requested` or already `cancelled`, because the in-process runner may terminate before the caller observes the intermediate state.

## Verification

- `uv run ruff check tests/unit/mcp/tools/test_ralph_handler.py`
- `uv run ruff format --check tests/unit/mcp/tools/test_ralph_handler.py`
- `uv run mypy src/ouroboros/mcp/tools/ralph_handlers.py src/ouroboros/ralph_loop.py`
- `uv run pytest tests/unit/mcp/tools/test_ralph_handler.py tests/unit/mcp/test_job_manager.py -q` → 36 passed

Refs #528.

### DIFF
--- a/tests/unit/mcp/tools/test_ralph_handler.py
+++ b/tests/unit/mcp/tools/test_ralph_handler.py
@@ -51,6 +51,30 @@ class _FakeEvolveHandler:
         )
 
 
+@dataclass
+class _BlockingEvolveHandler:
+    """Fake evolve handler that blocks until the Ralph job is cancelled."""
+
+    started: asyncio.Event = field(default_factory=asyncio.Event)
+    calls: int = 0
+
+    async def handle(self, arguments: dict[str, Any]):  # noqa: ARG002 - protocol fixture
+        self.calls += 1
+        self.started.set()
+        await asyncio.sleep(60)
+        return Result.ok(
+            MCPToolResult(
+                content=(MCPContentItem(type=ContentType.TEXT, text="late"),),
+                is_error=False,
+                meta={
+                    "lineage_id": arguments["lineage_id"],
+                    "generation": self.calls,
+                    "action": "continue",
+                },
+            )
+        )
+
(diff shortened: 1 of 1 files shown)


shaun0927:

This is a deliberately small #528 regression-test PR: it locks that `RalphHandler` jobs can be cancelled through the shared `JobManager` while a generation is active/blocked, and preserves the lineage link in the resulting job state.

Risk profile: low; tests only, no production behavior changes. It complements the MCP-owned Ralph loop by pinning the cancellation contract that users need once Ralph is job-backed.

Verification: ouroboros-agent APPROVE; mergeable clean; Ruff, MyPy, Bridge TypeScript, and Python 3.12/3.13/3.14 CI are green.

PR #628, opened 2026-05-05
feat(agent-process): project lifecycle state from events

shaun0927:
## Summary

Adds the first #518 durable lifecycle state-model slice: `AgentProcessSnapshot` plus `project_agent_process_snapshot()`, a small read-model projection that reconstructs one agent-process lifecycle from existing `control.directive.emitted` rows.

## Why

#518 should move toward durable pause/resume/replay through additive EventStore contracts before widening live runtime behavior. The current `AgentProcessHandle` is in-memory; this PR creates the replayable read-model foundation that later CheckpointStore and replay work can depend on without inspecting live handles.

## Scope

- Adds `AgentProcessSnapshot` with process id, status, intent, directive count, last reason, and terminal helper.
- Projects only existing `target_type="agent_process"` / `aggregate_type="agent_process"` directive events.
- Skips malformed lifecycle rows instead of corrupting state.
- Adds tests for normal lifecycle projection, process filtering, and malformed-row tolerance.

## Risks

- This is intentionally not durable pause/resume yet; it only defines the read model that those slices should build on.
- It projects from the current `extra.lifecycle_status` / `extra.intent` fields, so future event-schema changes must preserve or migrate those fields.
- It does not query EventStore directly; callers still supply the replayed event rows.

## Verification

- `uv run ruff check src/ouroboros/orchestrator/agent_process.py tests/unit/orchestrator/test_agent_process.py`
- `uv run ruff format --check src/ouroboros/orchestrator/agent_process.py tests/unit/orchestrator/test_agent_process.py`
- `uv run pytest tests/unit/orchestrator/test_agent_process.py -q` → 18 passed
- `uv run mypy src/ouroboros/orchestrator/agent_process.py`

Refs #518.

### DIFF
--- a/src/ouroboros/orchestrator/agent_process.py
+++ b/src/ouroboros/orchestrator/agent_process.py
@@ -183,17 +183,15 @@ def project_agent_process_snapshot(
             continue
 
         raw_intent = extra.get("intent")
-        if not isinstance(raw_intent, str) or not raw_intent:
-            continue
+        intent = raw_intent if isinstance(raw_intent, str) and raw_intent else None
         raw_reason = data.get("reason")
-        if not isinstance(raw_reason, str) or not raw_reason:
-            continue
+        reason = raw_reason if isinstance(raw_reason, str) and raw_reason else None
         snapshot = AgentProcessSnapshot(
             process_id=event_process_id,
             status=status,
-            intent=raw_intent,
+            intent=intent,
             directive_count=(snapshot.directive_count if snapshot is not None else 0) + 1,
-            last_reason=raw_reason,
+            last_reason=reason,
         )
 
     return snapshot
--- a/src/ouroboros/orchestrator/agent_process.py
+++ b/src/ouroboros/orchestrator/agent_process.py
@@ -52,13 +52,14 @@ reference migrations in slices 4–6 of #518.
 from __future__ import annotations
 
 import asyncio
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Iterable
 from dataclasses import dataclass, field
 from enum import StrEnum
 import logging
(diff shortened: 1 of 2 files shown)


shaun0927:
@ouroboros-agent please re-review the latest head. The stale cf661c9 blocker has been addressed: project_agent_process_snapshot now follows EventStore's canonical (timestamp, id) ordering rather than lifecycle-rank ordering, preserves optional metadata across minimal rows, and the timestamp-tie tests now pin deterministic IDs. Current CI is green across Ruff, MyPy, Bridge TypeScript, and Python 3.12/3.13/3.14 tests.

shaun0927:
@ouroboros-agent please re-review latest head 4515b819. The unordered-row blocker from 5fdf147 is now addressed by validating timestamp and id before appending rows to the sorted projection set, plus a regression test for an event-like row without ordering metadata. Current CI is green across Ruff, MyPy, Bridge TypeScript, and Python 3.12/3.13/3.14.

shaun0927:

This PR is a small #518 AgentProcess follow-up: it adds a replayable lifecycle snapshot projector over existing `control.directive.emitted` agent_process rows. It intentionally stays additive: no handler migration, checkpointing, or replay execution semantics are included here.

Risk profile: narrow but correctness-sensitive because this is a durable read model. The bot review drove the final contract: lifecycle status is the mandatory field; intent/reason are sticky optional metadata; projection follows EventStore's canonical `(timestamp, id)` ordering; malformed rows, including rows without comparable ordering metadata, are skipped instead of crashing reconstruction.

Verification: ouroboros-agent APPROVE; mergeable clean; Ruff, MyPy, Bridge TypeScript, and Python 3.12/3.13/3.14 CI are green.

PR #633, opened 2026-05-05
refactor(mcp): tighten JobManager runner cleanup boundary

Q00:
## Summary

`_run_job` carried two unreachable branches — a coroutine→Task conversion in the try block and a `runner.close()` call in the finally block — both gated on `runner_task is None and inspect.iscoroutine(runner)`. Since `start_job` already converted coroutines to Tasks before invoking `_run_job`, that condition could never hold during cleanup.

## Changes

| File | Change |
|------|--------|
| `src/ouroboros/mcp/job_manager.py` | `start_job` now normalises every input — pre-built Task, coroutine, or bare awaitable — into a Task at the boundary. `_runner_tasks` registration consolidated to a single statement. `_run_job` signature narrowed to `asyncio.Task[Any]`; the `runner_task` ceremony is removed in favour of `runner.done()` / `runner.cancel()` directly. |
| `tests/unit/mcp/test_job_manager.py` | New coverage verifies externally-created Task tracking/cleanup, bare Future wrapping, and cancellation reaching externally-created runner Tasks. |

## Why

- Removes dead code that obscured the actual cleanup contract.
- Eliminates a Future-vs-Task gap in the cancellation handler (Future inputs would have skipped runner cancel under the old typing).
- Centralises the "always a Task" invariant so future contributors do not reintroduce coroutine cleanup paths.

## Test plan

- [x] `uv run pytest tests/unit/mcp/test_job_manager.py -q` — 30 passed
- [x] Wider regression `tests/unit/mcp/` — 781 passed
- [x] `ruff format` / `ruff check` clean
- [x] `ouroboros-agent` review — APPROVED after edge-case test hardening

## Out of scope

`_append_event` failure before runner conversion can still strand a coroutine for GC. Separate follow-up.

### DIFF
--- a/src/ouroboros/mcp/job_manager.py
+++ b/src/ouroboros/mcp/job_manager.py
@@ -125,31 +125,49 @@ class JobManager:
         )
 
         self._known_job_ids.add(job_id)
-        if inspect.iscoroutine(runner):
-            runner = asyncio.create_task(runner)
-            self._runner_tasks[job_id] = runner
-        task = asyncio.create_task(self._run_job(job_id, job_type, runner))
+        # Normalise ``runner`` to a Task so ``_run_job`` can rely on Task
+        # semantics (cancellation, ``done()``). Coroutines are wrapped via
+        # ``create_task``; pre-built Tasks are reused; bare awaitables (e.g.
+        # ``Future``) are wrapped through an inner coroutine so cancellation
+        # and GC remain consistent.
+        if isinstance(runner, asyncio.Task):
+            runner_task = runner
+        elif inspect.iscoroutine(runner):
+            runner_task = asyncio.create_task(runner)
+        else:
+
+            async def _await_runner(_awaitable: Any = runner) -> Any:
+                return await _awaitable
+
+            runner_task = asyncio.create_task(_await_runner())
+        self._runner_tasks[job_id] = runner_task
+        task = asyncio.create_task(self._run_job(job_id, job_type, runner_task))
         self._tasks[job_id] = task
         self._monitors[job_id] = asyncio.create_task(self._monitor_job(job_id))
 
         return await self.get_snapshot(job_id)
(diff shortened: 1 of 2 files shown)


shaun0927:
@ouroboros-agent please re-review latest head ecf1eba. I addressed your non-blocking design note by adding coverage for both remaining runner-normalization edges: a bare `asyncio.Future` is wrapped into a Task and completes the job, and cancellation reaches an externally-created Task registered as the runner. Local verification: `uv run pytest tests/unit/mcp/test_job_manager.py -q` (30 passed), `uv run pytest tests/unit/mcp/ -q` (781 passed), and ruff check/format on the touched files.

shaun0927:

This PR tightens the `JobManager` runner lifecycle boundary by making `start_job()` normalize every runner input into a single internal `asyncio.Task` invariant before `_run_job()` starts. That removes the unreachable coroutine cleanup branches in `_run_job()` and makes cancellation/cleanup routing simpler and more consistent.

I reviewed the bot design notes and added the two missing edge-case tests it called out:

- bare `asyncio.Future` runners are wrapped into a Task and still complete/persist the job result;
- externally-created runner Tasks are cancelled through `cancel_job()` and observed by the caller-owned Task.

Current validation:

- `ouroboros-agent` approved latest head `ecf1eba` with no blocking or non-blocking findings.
- GitHub checks are all green: Bridge TypeScript, MyPy, Ruff, Python 3.12, Python 3.13, and Python 3.14.
- Local targeted validation passed:
  - `uv run pytest tests/unit/mcp/test_job_manager.py -q` — 30 passed
  - `uv run pytest tests/unit/mcp/ -q` — 781 passed
  - ruff check/format on the touched files passed

Given the narrow scope, added regression coverage for the remaining review risks, bot approval, and green CI, I think this PR is now safe to merge.

PR #654, opened 2026-05-06
feat(install): preserve user backend across upgrades

Q00:
## Summary
Re-running `install.sh` (or `ouroboros setup --non-interactive`) silently rewrote `orchestrator.runtime_backend` to whatever auto-detection landed on — typically `claude` when multiple CLIs are present. Users who deliberately switched to codex/hermes/gemini lost that choice on every upgrade, with no warning.

This PR makes upgrades **preserve the user's existing backend by default** and gives explicit hooks for scripted installs to pick one.

### `scripts/install.sh`
- Read `~/.ouroboros/config.yaml` first; if a prior runtime is set, preserve it (logged as `preserved from <path>`) and skip the auto-detection branch.
- New flags / env vars:
  - `--runtime <name>` / `OUROBOROS_INSTALL_RUNTIME=<name>` — pin explicitly (claude / codex / hermes / all).
  - `--reconfigure` / `OUROBOROS_INSTALL_RECONFIGURE=1` — force re-detection / re-prompt.
- First install with no CLI on PATH and a TTY now also offers an explicit `[0] None` option so users can install the base package and pick a backend later.
- Final hint after success points users to:
  - `ouroboros config backend <claude|codex|hermes|gemini>` for runtime swap
  - `ouroboros setup --runtime <name>` for kiro/copilot/opencode (which `config backend` doesn't yet support).

### `src/ouroboros/cli/commands/setup.py`
- `--non-interactive` auto-select: when a `current_backend` is configured and present in available runtimes, use it instead of falling through to a hardcoded `claude` default. Direct callers (`ouroboros setup --non-interactive` from CI/scripts) now match install.sh's preserving behaviour without needing a flag.

### Tests
`TestNonInteractiveAutoSelect` covers:
- Persisted codex → preserved (does not flip to claude)
- No persisted backend → claude default kept (first-install path)
- No persisted backend, no claude on PATH → first-available picked deterministically

## Why this matters
Reported by user: every `install.sh` re-run reset their backend. Combined with auto-update flows on dev machines, the backend choice was effectively non-persistent.

## Test plan
- [x] `uv run pytest tests/unit/cli` — 667 passed
- [x] `uv run ruff check / format --check / mypy` — clean
- [x] `bash -n scripts/install.sh` — syntax OK
- [x] Manual: yaml-parse logic returns `codex` for a sandbox config with `runtime_backend: codex`

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,11 +1,41 @@
 #!/bin/bash
 # Ouroboros installer — auto-detects runtime and installs accordingly.
 # Usage: curl -fsSL https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.sh | bash
+#
+# Runtime selection (first match wins):
+#   1. OUROBOROS_INSTALL_RUNTIME env var (claude|codex|opencode|hermes|all)
+#   2. Existing ~/.ouroboros/config.yaml runtime — preserved on upgrade
+#      unless OUROBOROS_INSTALL_RECONFIGURE=1 (or --reconfigure flag) is set.
+#   3. Interactive prompt when stdin is a TTY.
+#   4. Auto-detect single CLI on PATH; default to claude in pipe mode.
 set -euo pipefail
 
 PACKAGE_NAME="ouroboros-ai"
 MIN_PYTHON="3.12"
 IS_LOCAL=false
+RECONFIGURE="${OUROBOROS_INSTALL_RECONFIGURE:-}"
+EXPLICIT_RUNTIME="${OUROBOROS_INSTALL_RUNTIME:-}"
+
+# Parse simple flags: --reconfigure, --runtime <name>
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --reconfigure)
+      RECONFIGURE="1"
+      shift
+      ;;
+    --runtime)
+      EXPLICIT_RUNTIME="${2:-}"
+      shift 2
+      ;;
+    --runtime=*)
+      EXPLICIT_RUNTIME="${1#--runtime=}"
+      shift
+      ;;
+    *)
+      shift
+      ;;
+  esac
+done
 
 # Override PACKAGE_NAME if running inside the repository clone
 if [ -f "pyproject.toml" ] && grep -q "name = \"ouroboros-ai\"" pyproject.toml; then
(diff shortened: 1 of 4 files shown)


shaun0927:
@ouroboros-agent please re-review the latest PR head.

Bot re-review request for 170496f3. The latest commit addresses the previous blocking design notes by aligning install.sh runtime selection with setup.py's backend matrix and restoring Claude's MCP+Claude extras contract.

shaun0927:
@Q00 could you re-review this PR when you have a chance?

Merge-readiness summary for 170496f3:

This PR fixes the upgrade/install path so re-running `install.sh` or `ouroboros setup --non-interactive` preserves an already configured `orchestrator.runtime_backend` instead of silently falling back to auto-detection (usually Claude). That is the right default for upgrades because the configured backend is a deliberate user choice.

I pushed `fix(install): preserve runtime selection consistently` to address the previous bot design notes:

- `install.sh` now uses the same backend matrix as `setup.py` for explicit and preserved runtimes: `claude`, `codex`, `opencode`, `hermes`, `gemini`, `kiro`, and `copilot` (plus `all`).
- Existing configs such as `runtime_backend: opencode` are now preserved instead of being ignored and re-detected.
- Claude selection now consistently installs/registers the MCP + Claude dependency contract (`[mcp,claude]`), matching the MCP entry used by setup.
- Installer guidance now points users to the runtime-aware `ouroboros setup --runtime ...` path for all supported backends.
- Added regression coverage for installer syntax, preserving an existing OpenCode backend, and explicit Claude installing MCP+Claude extras.

Verification:

- Local: `bash -n scripts/install.sh`
- Local: `uv run pytest tests/unit/scripts/test_install_runtime_selection.py tests/unit/cli/test_setup.py -q` (131 passed)
- Local: `uv run ruff check tests/unit/scripts/test_install_runtime_selection.py tests/unit/cli/test_setup.py src/ouroboros/cli/commands/setup.py`
- GitHub checks on the latest head are green: Python 3.12/3.13/3.14 tests, Ruff, MyPy, and Bridge TypeScript.
- `ouroboros-agent[bot]` re-reviewed the latest head and approved it with no in-scope blocking findings.

Given the clean merge state, green CI, and fresh bot approval on the current head, I think this is now merge-ready.

Issue #578, opened 2026-05-02
Agent OS: define RuntimeControls watchdog contract


Q00:
# Agent OS: RuntimeControls watchdog v1 (minimum) + v2 expansion path

> **2026-05-22 status update.** This issue is lifted from `needs-design`
> to **implementation-ready v1 RFC** as part of the #1157 *minimal-substrate*
> redesign. The v1 contract below is intentionally narrow; richer
> designs originally proposed in this issue's earlier draft are
> preserved as a *v2 expansion path* and become slices only when
> evidence demands them.
>
> Implementation roadmap tracked in #1172. Wave 1: `ooo auto` SSOT (#1157).

---

### One timer

```yaml
runtime_controls:
  session_wall_clock_seconds: 14400   # 4h default; configurable per goal
```

Per-session wall-clock budget only. No `idle_timeout`, no
`no_progress_timeout`, no `safety_timeout` — until evidence demands them.

### One event family

```python
class RuntimeWatchdogCancelEvent:
    """Recorded when a session's wall-clock budget is exceeded."""

    aggregate_type: str = "runtime_control"
    aggregate_id: str   # the auto_session_id
    event_kind: str = "runtime.watchdog.cancel"
    reason: str = "wall_clock_exceeded"   # only reason for v1
    session_started_at: datetime
    fired_at: datetime
    elapsed_seconds: int
    configured_budget_seconds: int
```

The `runtime_control` aggregate type is **confirmed additive** to the
projection v1 vocabulary by reading `src/ouroboros/persistence/schema.py`
(`aggregate_type` is a free-form `VARCHAR(100)`, no enum constraint, no
Python allowlist). Informational confirmation posted on #946.

### One new stop_reason_code

Adds `watchdog_wall_clock_exceeded` to the existing 8-code taxonomy
(#1167 added the previous 8th), bringing it to 9 codes. The auto
pipeline catches the event, transitions to BLOCKED with this code, and
returns a resumable `auto_session_id` per existing L4 semantics.

### Resume semantics

Timer state is *implicit* in `AutoPipelineState.session_started_at`.
On every watchdog tick: read `session_started_at` from the auto
session state, compare against `now`. No separate serialization.
On resume, `session_started_at` is preserved by the existing state
restore path — the watchdog naturally sees the *real* elapsed time
since the original start.

### Cancellation behavior

When the watchdog fires:

1. Append `runtime.watchdog.cancel` event to EventStore.
[...]

PR #664, opened 2026-05-06
test(watchdog): make parent cancellation assertion deterministic

shaun0927:
## Summary
- Preserve the watchdog invariant that the EventStore baseline is initialized before the watched generation starts.
- Replace the sleep-based parent-cancellation test with an explicit child-start synchronization point so cancellation is asserted only after the generation task exists.

## Review response
- Addresses the bot review on commit edbaf7b: the production change that created the watched task before baseline initialization has been reverted, avoiding the startup cursor race that could hide first progress events.

## Validation
- uv run pytest tests/unit/evolution/test_generation_watchdog.py -q
- uv run pytest tests/unit/evolution/test_generation_watchdog.py tests/unit/scripts/test_keyword_detector.py -q
- uv run ruff check src/ouroboros/evolution/watchdog.py tests/unit/evolution/test_generation_watchdog.py scripts/keyword-detector.py tests/unit/scripts/test_keyword_detector.py

Refs #578

### DIFF
--- a/src/ouroboros/evolution/watchdog.py
+++ b/src/ouroboros/evolution/watchdog.py
@@ -105,9 +105,9 @@ class GenerationProgressWatchdog:
 
     async def watch[T](self, awaitable: Awaitable[T]) -> T:
         """Run *awaitable* until it finishes or watchdog policy cancels it."""
-        await self.initialize_baseline()
         task: asyncio.Task[T] = asyncio.create_task(awaitable)
         try:
+            await self.initialize_baseline()
             while True:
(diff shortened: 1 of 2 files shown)


shaun0927:
Follow-up pushed in 89368b78 to address the blocking bot review: restored baseline initialization before starting the watched task, and changed the cancellation test to wait until the child generation has actually started before cancelling the parent. Local validation: watchdog unit file 10 passed; combined watchdog+keyword focused tests 24 passed; ruff focused check passed.

Issue #689, opened 2026-05-07
Add direct operational-task path for concrete PR and merge goals in ooo auto


shaun0927:
## Background
This goal was already a concrete operational task: analyze open PRs, iterate them up to a mergeable state, and merge them once the merge conditions are satisfied. Today's auto pipeline routes every goal through `Interview → Seed generation → Review/Repair → Run handoff`, so a goal like this can be blocked at the Socratic interview before it ever reaches the PR work step.

## Reproduction / observation evidence
- Observed when the Discord/Hermes deterministic gateway rewrite executed:
  - `ooo auto --runtime codex 'Carefully review the open PRs at https://github.com/shaun0927/opensafari/pulls and iteratively improve them until they are mergeable. Once they are mergeable, go ahead and merge.'`
- Auto state example: `auto_78c98678de5d`
- State path example: `/opt/data/home/.ouroboros/data/auto_78c98678de5d.json`
- Final phase: `blocked`
- `last_tool_name`: `interview.start`
- `last_error`: `interview.start timed out after 60s for auto_78c98678de5d`

## Goal
Allow operational goals that already carry sufficient execution context — PR/issue URLs, explicit merge / review / fix asks — to bypass or minimize the unnecessary interview-first path.

## Scope
- Add a goal classifier that flags `github.com/.../pulls`, `github.com/.../pull/<n>`, "mergeable", `PR review/fix/merge` style asks as operational tasks.
- Operational tasks get an option to enter a bounded execution plan or run handoff directly instead of seed/interview.
- Make the dangerous side effect (merge) gated explicitly by a policy / permission / CI guardrail before execution.

## Affected files / paths
- `src/ouroboros/auto/pipeline.py`
- `src/ouroboros/auto/state.py`
- `src/ouroboros/cli/commands/auto.py`
- Runtime / handoff modules
- GitHub / PR task classifier tests

## Implementation steps
1. Define an auto-goal classification result schema: `interview_required`, `direct_run_allowed`, `side_effect_risk`, `requires_confirmation`.
2. Add a deterministic classifier that detects PR / issue URL goals.
3. Design phase transitions so the direct path still preserves ledger, state, and replayability.
4. Define the prerequisite matrix for performing a merge: target branch, CI success, review/mergeability, repository permission.
5. Separate tests for the existing interview-first path and the new direct operational path.

## Success criteria
- A concrete PR-URL merge/fix goal is no longer blocked entirely by a first-question timeout.
[...]

PR #693, opened 2026-05-07
feat(auto): expose ledger provenance in auto pipeline result

shaun0927:
## Summary
Surface which Seed sections came from repo facts vs assumptions through the auto pipeline result and MCP response.

- `SeedDraftLedger.summary()` now returns:
  - `provenance`: a dict mapping each `LedgerSource` value to the list of sections that have at least one *active* entry of that source.
  - `evidence_backed_sections`: sections containing any active entry from `USER_GOAL`, `REPO_FACT`, `EXISTING_CONVENTION`, or `INFERENCE`.
  - `assumption_only_sections`: sections that have entries but none from those evidence-backed sources.
  - WEAK/CONFLICTING/BLOCKED entries are excluded so superseded or risky entries do not falsely claim evidence.
- `AutoPipelineResult` carries the same three fields so CLI and MCP consumers see them without having to deserialize the full ledger.
- `ouroboros_auto` MCP handler:
  - Adds an `Evidence:` block to the human-readable text body (`evidence-backed:` / `assumption-only:` lines).
  - Exposes `provenance`, `evidence_backed_sections`, and `assumption_only_sections` in `MCPToolResult.meta`.

## Why
This implements the second acceptance criterion from #640:

> The final Seed or ledger shows which constraints are evidence-backed versus assumed.

The provenance taxonomy already lives on each `LedgerEntry` (added by #646/#666), but it was previously invisible above the raw ledger dict. Callers had no machine-readable answer to "which parts of this Seed came from repo inspection vs which were filled with conservative defaults?". This PR keeps the change bounded to the auto/MCP surface — it does not modify the immutable `Seed` schema or `bigbang/seed_generator.py`, since those touch interview/eval/run paths and have a much wider blast radius.

## Out of scope
- Modifying the immutable `Seed` model in `core/seed.py` (frozen schema; would need coordinated migration).
- Modifying `bigbang/seed_generator.py` YAML output.
- Adding new repo crawling or repo-fact discovery (covered separately by #676/#681).
- Risky-fallback gating (a follow-up addresses that as a separate slice).

## Tests
- New `test_ledger_summary_groups_active_sections_by_provenance_source` and `test_ledger_summary_excludes_inactive_entries_from_provenance` in `tests/unit/auto/test_ledger_grading_answerer.py`.
[...]

PR #683, opened 2026-05-07
feat(auto): classify selected-driver answer risk

shaun0927:
## Summary
- Add post-response selected-driver answer risk classification for risky generated answer text.
- With `brake=on`, block high-risk actual driver responses before they are sent back to the interview backend.
- With `brake=off`, continue autopilot while recording combined risk metadata and ledger provenance.

## Discussion / implementation notes
- This PR is stacked on #682 conceptually and completes the #675 behavior slice.
- It preserves the existing pre-response/question risk classifier and adds a second classifier over the actual selected-driver answer text.
- The classifier is deterministic and local; it does not call external services.

## Validation
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/auto/test_driver_answerer.py -q` → 10 passed
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/auto/driver_answerer.py tests/unit/auto/test_driver_answerer.py` → passed
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check src/ouroboros/auto/driver_answerer.py tests/unit/auto/test_driver_answerer.py` → passed
- `git diff --check` → passed

Closes #675
Depends on #682
Depends on #672

PR #682, opened 2026-05-07
feat(auto): add selected-driver answer metadata

shaun0927:
## Summary
- Add `AutoAnswerMetadata` as structured metadata on auto answers.
- Populate selected-driver answers and selected-driver blockers with risk, confidence, and provenance metadata.
- Preserve existing selected-driver behavior; this slice only adds metadata plumbing for downstream classifier/enforcement work.

## Discussion / implementation notes
- This is the first #675 slice and intentionally uses `Refs #675` instead of closing it.
- The follow-up slice should consume this metadata for post-response classification and brake-specific enforcement.
- Existing text tags and ledger provenance are preserved for compatibility.

## Validation
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/auto/test_driver_answerer.py -q` → 7 passed
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/auto/answerer.py src/ouroboros/auto/driver_answerer.py src/ouroboros/auto/__init__.py tests/unit/auto/test_driver_answerer.py` → passed
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check src/ouroboros/auto/answerer.py src/ouroboros/auto/driver_answerer.py src/ouroboros/auto/__init__.py tests/unit/auto/test_driver_answerer.py` → passed
- `git diff --check` → passed

Refs #675
Depends on #672

PR #695, opened 2026-05-07
feat(auto): block risky-fallback answers for regulated topics

shaun0927:
## Summary
Add a narrow risky-fallback gate to the deterministic auto answerer so high-risk topics surface for human review instead of being silently filled with a generic default.

When the deterministic answerer would otherwise return `CONSERVATIVE_DEFAULT` or `ASSUMPTION` for a question whose topic has no defensible generic default, it now returns a `BLOCKER` with a concrete reason.

Targeted topics (intentionally narrow):
- Regulated personal data: `PII`, `personally identifiable information`, `GDPR`, `HIPAA`, `SOX`, `PCI-DSS`.
- Destructive bulk schema/table operations: `truncate`/`purge` of `table(s)` or `schema(s)`.

Excluded by design (already covered or would over-trigger):
- Production deployment authority — already covered by `_blocker_for` for the explicit verb+target pairs (`deploy/release/publish to/against/on production/prod/live/external`).
- Real-money authority for credit cards / billing accounts — already covered by `_blocker_for`.
- Generic 'destructive' keywords like `drop ... database` — already covered by `_blocker_for`.

## Why
This implements the fourth piece of #640's acceptance criteria:

> block or ask the user when a required answer would otherwise be a risky fallback

The provenance taxonomy is in place after #646/#666, but the auto answerer was still willing to fabricate a generic answer for topics where no generic answer is safe. PR-B keeps the change deliberately narrow — only the topics where a fallback is unambiguously wrong — so the existing safe-allowlists for product-feature questions about credentials/branches keep working.

## Behavior
- `_blocker_for` runs first — explicit sensitive-authority questions still block at their original reason and message.
- Deterministic routing then picks a category answerer (verification / acceptance / actor-IO / runtime / product behavior / default).
- After routing, if the answer source is `CONSERVATIVE_DEFAULT` or `ASSUMPTION` *and* the question matches a risky-fallback pattern *and* the safe-product allowlists do not exempt it, the answer is replaced with a `BLOCKER` carrying the matched reason.
- A `REPO_FACT` / `USER_GOAL` / `EXISTING_CONVENTION` / `INFERENCE` answer is never gated (callers can still ground the answer with bounded repo facts via `AutoAnswerContext`).

## Out of scope
- Selected-driver post-response risk classification (handled by #675 / #682 / #683).
- Expanding the regulated-topic vocabulary beyond PII/GDPR/HIPAA/SOX/PCI-DSS (can grow incrementally).
- Adding a manual-approval flow as the default path (kept off the autopilot critical path).

## Tests
- New \`test_auto_answerer_blocks_regulated_data_questions_instead_of_falling_back\` — PII / HIPAA / 'purge tables' all block with their matched reason.
- New \`test_auto_answerer_does_not_block_regulated_topic_when_repo_fact_supplied\` — a HIPAA-adjacent runtime question with supplied repo facts still returns \`REPO_FACT\`.
- New \`test_auto_answerer_skips_risky_fallback_for_safe_product_credential_questions\` — feature questions about credentials remain unblocked through the existing safe-product allowlists.
- Existing \`test_auto_answerer_allows_safe_production_and_project_feature_questions\` and the production-credential authority test continue to pass.

## Validation
- \`UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/auto/ tests/unit/mcp/ -q\` → 998 passed
- \`UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/auto/answerer.py tests/unit/auto/test_ledger_grading_answerer.py\` → passed
- \`UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check\` (same files) → passed

Refs #640
Independent of #693 (provenance surface). Either PR can land first.


Q00 (changes requested):
Reviewed. The regulated-topic risky fallback gate is a policy boundary, so it needs fully green tests. Checks are still running, so I cannot approve yet.

Q00 (approved):
Reviewed across OS/UserLevel/Program boundaries, auto scope, and UX complexity. Approving: the risky-fallback gate narrows automation for regulated/destructive topics, which reduces scope risk rather than expanding it.

shaun0927:
Cross-linking from #689 (closed as out-of-scope for core).

Surfacing a boundary concern, not asking for an immediate decision: this PR places the regulated-topic vocabulary (`PII / GDPR / HIPAA / SOX / PCI-DSS`) and destructive bulk-schema verbs (`truncate / purge` of `tables / schemas`) **directly inside `src/ouroboros/auto/answerer.py`**.

The *surface* — block instead of silently fabricating a `CONSERVATIVE_DEFAULT` for risky topics — fits core (safety boundary). What I am less sure about is the **vocabulary** living in core:

- The set is team- and jurisdiction-specific (e.g. some teams care about KYC/AML; others care about FERPA; others care about export-control). A core PR that hard-codes one regional/regulatory subset will keep accreting siblings.
- It is exactly the same pattern @Q00 closed #689 PRs for: a typed classifier with a hard-coded vocabulary in core, instead of "stable primitive + policy lookup."

Possible reframings (any of these would address the concern without losing the safety property):

1. Keep the *gate* in core, move the *vocabulary* to a config/policy file the gate consults. Default config can ship the current PII/GDPR/HIPAA/SOX/PCI-DSS list so behavior is unchanged; teams can extend without patching core.
2. Or: keep core's job as "any `CONSERVATIVE_DEFAULT` / `ASSUMPTION` for an answer marked sensitive becomes a `BLOCKER`" and let an external classifier decide what is sensitive. Core then has zero domain words.
3. Or: explicitly accept this as a narrow temporary primitive (with a comment + a tracker for plugin-extraction) and freeze the vocabulary at exactly the current 5 + bulk-schema verbs.

Happy to take any of those directions — flagging now so the boundary call is explicit before merge rather than litigated later.

shaun0927:
Following up on #725 ("Design a UserLevel plugin manager for operational workflows"): recommend **HOLD on this PR until #725 v0 contract lands.**

Reasons:

1. The current diff places regulated-topic vocabulary (`PII / GDPR / HIPAA / SOX / PCI-DSS`) and destructive bulk-schema verbs directly inside `src/ouroboros/auto/answerer.py`. Per the #725 boundary, vocabulary belongs in a UserLevel skill, not in core auto.
2. Merging now would create the same anti-pattern that got #689 closed: core slowly accumulates domain words. After #725 v0 (`RiskAssessor` protocol + skill-owned config), this PR's intent can land as **(a)** a one-liner core change introducing `RiskAssessor`, plus **(b)** a default-shipping skill that owns the current vocabulary and can be replaced/extended without patching core.

The *gate* (block instead of fabricate when the answerer would otherwise return `CONSERVATIVE_DEFAULT` for a topic with no defensible default) is correct and should land in core. Only the **vocabulary list and matching logic** moves out.

Concrete suggested redesign (after #725 v0):
- core: `RiskAssessor` protocol with one method `assess(question, candidate_answer) -> RiskVerdict`.
- core: existing `_blocker_for` keeps its current behavior; new gate runs `RiskAssessor` after the deterministic answerer returns and converts a non-empty verdict to a `BLOCKER`.
- skill: `regulated-topics-default` ships with the same PII/GDPR/HIPAA/SOX/PCI-DSS list as today — behavior unchanged for users who install nothing extra.
- users: can install a different `RiskAssessor`-providing skill (e.g. `regulated-topics-fintech`, `regulated-topics-healthcare`) without forking core.

Happy to split this PR into the core gate + the default skill once the v0 manifest lands. Until then, holding so the contract isn't pre-committed to.

shaun0927:
Closing this PR per the boundary established in #689 / #725 v0 discussion.

The intent is correct (block instead of fabricate for risky topics) and will return after #725 v0 lands as **two cleaner pieces**:
1. core `RiskAssessor` protocol + the gate semantics (small, generic),
2. a default-shipping `regulated-topics-default` skill carrying the current PII / GDPR / HIPAA / SOX / PCI-DSS vocabulary.

That split keeps the safety property (block instead of silently fabricate) without committing core to a specific regional/regulatory vocabulary.

Tests in this PR (`test_auto_answerer_blocks_regulated_data_questions_instead_of_falling_back`, `..._does_not_block_regulated_topic_when_repo_fact_supplied`, `..._skips_risky_fallback_for_safe_product_credential_questions`) will be ported to the post-v0 implementation.

Refs #725.

PR #701, opened 2026-05-07
feat(auto): allowlisted gateway provenance on AutoPipelineState (#691) [1/3]

shaun0927:
## Summary

PR 1/3 of the gateway-provenance stack for #691. Adds an optional, allowlisted ``provenance`` dict on ``AutoPipelineState`` so the durable auto-session JSON can record why a particular ``ooo auto`` run exists when an external gateway (Discord/Hermes deterministic command rewrite) translated a natural-language request into a shell invocation.

- New ``PROVENANCE_ALLOWED_KEYS`` + ``redact_provenance()`` helper in ``src/ouroboros/auto/state.py``
- New ``provenance: dict | None`` field on ``AutoPipelineState`` (default ``None`` so direct CLI runs are untouched)
- ``_validate_loaded`` refuses unredacted payloads on save
- ``from_dict`` backfills ``None`` so legacy auto-session JSON files load unchanged
- 7 new tests in ``tests/unit/auto/test_state_provenance.py``

Refs #691, epic #692. Stack: **PR-A (this)** → PR-B (CLI/env wiring) → PR-C (reporting). Subsequent PRs are stacked on this branch and link back here for context.

## Test plan

- [x] ``pytest tests/unit/auto/test_state_provenance.py tests/unit/auto/test_state.py``
- [x] ``ruff check`` clean on touched files
- [x] Legacy state JSON without ``provenance`` field still loads via ``AutoStore.load``
- [ ] Reviewer confirms allowlist keys cover the gateway-integration shape called out in #691

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #692, opened 2026-05-07
Track ooo auto Discord gateway rewrite incident follow-up work


shaun0927:
## Background
Tracker for the follow-up work on the incident where the Discord/Hermes deterministic gateway rewrite translated a natural-language PR-operations request into the shell command `ooo auto --runtime codex ...` and the Ouroboros auto pipeline became blocked at the `interview.start` step.

The observed auto session was `auto_78c98678de5d`, with the final state `phase=blocked`, `last_tool_name=interview.start`, `last_error=interview.start timed out after 60s for auto_78c98678de5d`. The session never reached the open-PR analysis / repair / merge step.

## Goal
Track the timeout, resume, interview-first UX, runtime semantics, and gateway-provenance problems exposed by this incident as independent implementation units.

## Scope
- Create leaf issues and maintain a recommended order.
- Maintain success / verification criteria so each issue can be implemented as a small PR.
- Distinguish external-gateway problems from Ouroboros-internal improvements.

## Affected files / paths
- `src/ouroboros/auto/state.py`
- `src/ouroboros/auto/interview_driver.py`
- `src/ouroboros/auto/pipeline.py`
- `src/ouroboros/cli/commands/auto.py`
- `src/ouroboros/mcp/tools/authoring_handlers.py`
- `src/ouroboros/mcp/tools/subagent.py`

## Implementation breakdown
1. #686 — ooo auto interview timeout config is ignored and start can block before work begins
2. #687 — Persist interview session id before first question generation in ooo auto
3. #688 — Make ooo auto resume semantics truthful for interview.start timeouts
4. #689 — Add direct operational-task path for concrete PR and merge goals in ooo auto
5. #690 — Clarify and test --runtime codex authoring versus run handoff semantics in ooo auto
6. #691 — Record external gateway provenance when rewritten commands invoke ooo auto

## Success criteria
- Each leaf issue can be implemented and verified as its own PR.
- Timeout and resume problems are guarded by reproducible regression tests.
- A path is defined where operational tasks (PR URL / merge goals) are not entirely blocked by a single interview-start timeout.
- Gateway-rewrite provenance carries enough information to track external integration problems while never persisting sensitive data.

## Verification
- Each leaf-issue PR runs the tests listed in that issue's `Verification` section.
- The incident follow-up is considered complete when every leaf issue is closed or explicitly deferred.
[...]

Issue #691, opened 2026-05-07
Record external gateway provenance when rewritten commands invoke ooo auto


shaun0927:
## Background
This run was not a direct user invocation of `ooo auto` — it was the Discord/Hermes deterministic gateway rewrite translating a natural-language request into a shell command. The rewrite trigger itself is not in the Ouroboros repo, and `ooo auto` state currently records no external gateway / rewrite provenance, which makes post-incident root-cause analysis difficult.

## Reproduction / observation evidence
- Observed when the Discord/Hermes deterministic gateway rewrite executed:
  - `ooo auto --runtime codex 'Carefully review the open PRs at https://github.com/shaun0927/opensafari/pulls and iteratively improve them until they are mergeable. Once they are mergeable, go ahead and merge.'`
- Auto state example: `auto_78c98678de5d`
- State path example: `/opt/data/home/.ouroboros/data/auto_78c98678de5d.json`
- Final phase: `blocked`
- `last_tool_name`: `interview.start`
- `last_error`: `interview.start timed out after 60s for auto_78c98678de5d`

## Goal
Add an opt-in metadata contract that lets external gateways safely record the original user utterance, rewrite indication, source platform/channel, and command provenance when invoking `ooo auto`.

## Scope
- Design a CLI- or environment/flag-based input mechanism for provenance metadata.
- Persist provenance in auto state / ledger.
- Minimize the scope of stored sensitive data and channel identifiers.
- Distinguish rewrite-driven invocations from direct CLI ones in blocker reports.

## Affected files / paths
- `src/ouroboros/auto/state.py`
- `src/ouroboros/cli/commands/auto.py`
- `src/ouroboros/auto/pipeline.py`
- gateway integration docs or CLI docs
- state serialization tests

## Implementation steps
1. Add an opt-in `provenance` dict/schema field to `AutoPipelineState`.
2. Define a CLI flag or env-var contract — e.g. `--provenance-file` or `OUROBOROS_AUTO_PROVENANCE_JSON`.
3. Apply an allowlist schema and redaction rules before persistence.
4. Surface a low-risk piece of metadata such as `invoked_by: direct|gateway|unknown` in the blocker / summary output.
5. Keep the existing CLI behavior unchanged when the gateway provides no metadata.

## Success criteria
- Auto sessions launched via rewrite carry provenance in state / ledger.
- The persisted metadata contains no token, full credential, or private message content.
- Direct CLI runs without provenance remain backwards-compatible.

## Verification
[...]

PR #702, opened 2026-05-07
feat(auto): wire --provenance-file and env var into ooo auto (#691) [2/3]

shaun0927:
## Summary

PR 2/3 of the gateway-provenance stack for #691. Adds the input contract a Discord/Hermes-style gateway uses to attach provenance to an ``ooo auto`` run.

- New ``src/ouroboros/auto/provenance.py`` (``resolve_provenance`` + ``PROVENANCE_ENV_VAR``)
- New ``--provenance-file PATH`` typer option in ``src/ouroboros/cli/commands/auto.py``; takes precedence over ``OUROBOROS_AUTO_PROVENANCE_JSON`` env var on conflict
- Both inputs are JSON-decoded then narrowed by ``redact_provenance`` from PR-A (#701) before reaching state, so unknown keys (tokens, raw utterances, plaintext channel ids) are silently dropped
- Direct CLI runs leave both inputs unset and observe the prior behaviour: ``state.provenance`` stays ``None`` and downstream code paths are unchanged
- Resume rejects payloads that conflict with persisted provenance to keep gateway integrations honest
- 10 new tests in ``tests/unit/cli/test_auto_provenance.py``

> ⚠️ **Stacked on #701** — review and merge PR-A (#701) first. Diff against ``main`` here will include PR-A commits until #701 lands.

Refs #691, epic #692. Stack: PR-A (#701) → **PR-B (this)** → PR-C.

## Test plan

- [x] ``pytest tests/unit/cli/test_auto_provenance.py tests/unit/cli/test_auto_command.py``
- [x] CLI flag overrides env var on conflict; either input alone works
- [x] Empty/whitespace env, malformed JSON, non-object payloads all raise actionable ``ValueError``
- [x] Direct invocation (no flag, no env) keeps ``state.provenance == None``
- [ ] Manual: gateway integration smoke against a sample payload file in a test repo

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -18,6 +18,20 @@ class HandlerError(RuntimeError):
     """Raised when an MCP handler returns an error result."""
 
 
+class PartialInterviewStartError(HandlerError):
+    """Raised when interview start failed but a session_id was persisted server-side.
+
+    Carries the persisted ``session_id`` so callers (e.g. the auto interview
+    driver) can record it on durable state and resume the same interview
+    after a transient first-question failure such as an LLM timeout.
+    See Q00/ouroboros#687.
+    """
+
+    def __init__(self, message: str, *, session_id: str) -> None:
+        super().__init__(message)
+        self.session_id = session_id
+
+
 def _unwrap(result, *, tool_name: str) -> MCPToolResult:
     if result.is_err:
         error: MCPServerError = result.error
@@ -36,11 +50,33 @@ class HandlerInterviewBackend(InterviewBackend):
         self.handler = handler
         self.cwd = cwd
 
-    async def start(self, goal: str, *, cwd: str) -> InterviewTurn:
-        result = _unwrap(
-            await self.handler.handle({"initial_context": goal, "cwd": cwd or self.cwd}),
-            tool_name="ouroboros_interview",
-        )
+    async def start(self, goal: str, *, cwd: str, interview_id: str | None = None) -> InterviewTurn:
+        arguments: dict[str, str] = {"initial_context": goal, "cwd": cwd or self.cwd}
+        if interview_id:
+            arguments["interview_id"] = interview_id
+        outcome = await self.handler.handle(arguments)
+        # Recoverable error path: handler persisted state but failed to
+        # produce the first question.  ONLY trust an explicit
+        # ``meta.session_id`` from the handler — never fall back to the
+        # caller-supplied ``interview_id``, otherwise auto state would
+        # record persistence evidence that the handler never produced
+        # (Q00/ouroboros#723 review).
+        if not outcome.is_err:
+            value = outcome.value
+            if value.is_error:
+                meta = value.meta or {}
+                session_id = _optional_str(meta.get("session_id"))
+                if session_id:
+                    text = (
+                        value.content[0].text
+                        if value.content
+                        else "ouroboros_interview returned error"
+                    )
+                    raise PartialInterviewStartError(
+                        f"ouroboros_interview failed: {text}",
+                        session_id=session_id,
+                    )
+        result = _unwrap(outcome, tool_name="ouroboros_interview")
         return _turn_from_result(result)
 
     async def answer(self, session_id: str, answer: str) -> InterviewTurn:
@@ -56,6 +92,22 @@ class HandlerInterviewBackend(InterviewBackend):
             tool_name="ouroboros_interview",
         )
         return _turn_from_result(result, fallback_session_id=session_id)
+
+    def is_session_persisted(self, session_id: str) -> bool:
+        """Return True when ``interview_<session_id>.json`` exists on disk.
+
+        Used by ``AutoInterviewDriver`` to decide whether a pre-allocated
+        id may be retained on auto state after a driver-level
+        ``asyncio.wait_for`` cancel — without this probe the driver cannot
+        distinguish "handler crashed before persisting" from "handler
+        persisted then got cancelled".  Routes through
+        ``InterviewHandler.resolved_state_dir`` so the probe always
+        targets the directory the engine actually writes to (Q00/ouroboros#723).
+        """
+        if not session_id:
+            return False
+        state_dir = self.handler.resolved_state_dir()
+        return (state_dir / f"interview_{session_id}.json").exists()
 
 
 class HandlerSeedGenerator:
--- a/src/ouroboros/auto/interview_driver.py
+++ b/src/ouroboros/auto/interview_driver.py
@@ -5,8 +5,12 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
+import inspect
 import re
 from typing import Protocol
+from uuid import uuid4
+
+import structlog
 
 from ouroboros.auto.answerer import (
     AutoAnswer,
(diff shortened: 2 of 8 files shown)


shaun0927:
**Validator round (architect / security / code-reviewer)** — all approved-with-suggestions. Force-pushed updates:

- `provenance.py`: `env` parameter now defaults to `os.environ` instead of `{}` so a future caller that forgets the kwarg still picks up `OUROBOROS_AUTO_PROVENANCE_JSON` (architect + code-reviewer MEDIUM, both flagged the same latent bug).
- `provenance.py`: 4 KiB cap on both `--provenance-file` reads and the env var, returns an actionable `ValueError` past the cap. Bounds memory if a misconfigured caller points the flag at `/dev/urandom` or a multi-GB file (security MEDIUM).
- `auto.py`: resume now refuses to attach provenance to a session originally invoked without it (`cannot attach provenance on resume...`). Re-attribution would mislead audit (code-reviewer MEDIUM, was an unnoticed gap in the conflict guard).
- 4 new tests cover the env default, both size caps, and the resume reject-attach path. 14/14 in the file pass; 904 globally green; ruff clean.

shaun0927:
**Bot blocking finding addressed (round 2):** ``stat().st_size`` is unreliable for FIFOs, character devices (``/dev/urandom``), and procfs paths — they report ``0`` yet stream unbounded data. Replaced the size pre-check with a proper guard:

1. ``path.exists()`` → keeps the actionable "not found" error path.
2. ``path.is_file()`` rejects FIFOs, devices, directories, and symlinks-to-pipes (it follows symlinks, so the ``S_ISREG`` check applies to the resolved target).
3. **Bounded read** of ``MAX + 1`` bytes regardless of stat — defends against a stat that lies (kernel-fakable for special files), making this a defense-in-depth pair instead of a single trust point.
4. UTF-8 decode is now explicit so binary garbage produces an actionable error rather than a UnicodeDecodeError that bubbles past the boundary.

New regressions in ``tests/unit/cli/test_auto_provenance.py``:

- ``test_resolve_provenance_rejects_directory``
- ``test_resolve_provenance_rejects_fifo`` (skips on platforms without ``os.mkfifo``)
- ``test_resolve_provenance_bounded_read_caps_oversize_file``
- ``test_resolve_provenance_rejects_invalid_utf8``

25/25 in the file pass; ruff format + check clean.

Q00 (changes requested):
Reviewed. The CLI provenance flag and environment-variable wiring are user-facing behavior. CI is still running and prior requested changes remain, so I cannot approve yet.

shaun0927:
@Q00 — re-review ping. CI is now fully green (Ruff Lint + Tests on 3.12/3.13/3.14 + MyPy + Bridge TS), and ouroboros-agent[bot] returned **APPROVE** on the latest commit (`15aecf5`, review record ``e487bc00-0228-4c2e-91b3-cbc225cd7d0f``).

**Scope of this PR (2/3 of the gateway-provenance stack):** add the input contract that lets a Discord/Hermes-style deterministic command-rewrite gateway tag an ``ooo auto`` invocation with hash-only audit metadata. No change to control flow for direct CLI users — ``state.provenance`` remains ``None`` and every existing code path observes the prior behaviour.

**What landed across the two review rounds**

| Concern | Source | Resolution |
|---|---|---|
| ``env=None`` silently skips the env var | architect + code-reviewer (MEDIUM) | ``env`` now defaults to ``os.environ`` so a future caller that forgets the kwarg still picks up ``OUROBOROS_AUTO_PROVENANCE_JSON``. Tests pass an explicit ``env={}`` to suppress lookup. |
| Unbounded read on ``--provenance-file`` | security review (MEDIUM) | 4 KiB cap; the cap is enforced via a **bounded read** (``read(MAX + 1)`` with a length check) so it cannot be bypassed by a stat that reports ``0``. |
| ``stat().st_size`` lies for FIFOs/devices/procfs | ouroboros-agent[bot] (BLOCKING, round 2) | ``path.exists()`` then ``path.is_file()`` rejects FIFOs, character devices, directories, and symlinks-to-pipes upfront. Bounded read is still the second line of defense. UTF-8 decode is now explicit so binary garbage produces an actionable ``ValueError`` at the boundary instead of a leaked ``UnicodeDecodeError``. |
| Re-attribution on resume | code-reviewer (MEDIUM) | Resume now refuses to attach provenance to a session that was originally invoked without it (``cannot attach provenance on resume...``). Conflict-equality guard is preserved. |

**Why this is merge-ready**

1. **Strict-allowlist redaction at the trust boundary.** ``redact_provenance`` (PR #701) drops every key not on a 7-entry allowlist, so a misconfigured gateway cannot smuggle tokens, raw user utterances, or plaintext channel/user ids through this path. Hex-only enforcement on ``*_hash`` keys prevents hash-named smuggling.
2. **Defense-in-depth on file ingest.** Three independent guards: regular-file check, bounded read, explicit UTF-8 decode. Any single one being bypassable does not make the system unsafe.
3. **Backward compatibility is mechanically guaranteed.** ``resolve_provenance`` returns ``None`` when no input is supplied; the CLI does not touch ``state.provenance`` in that case; ``AutoPipelineState.from_dict`` backfills ``provenance=None`` for legacy session JSON (covered in PR #701's ``test_legacy_state_without_provenance_field_loads``).
4. **Test coverage is contract-shaped, not line-shaped.** 25 tests in ``tests/unit/cli/test_auto_provenance.py`` assert: env default, both size caps, directory/FIFO/non-UTF-8 rejection, resume reject-attach, file-wins-over-env precedence, redaction-drops-unknown-keys, and direct-CLI-keeps-provenance-None. They prove the *contract* — not just the lines.
5. **No change to direct-CLI hot paths.** Every existing ``ooo auto`` test still passes; the only branch that activates is a new no-op when neither flag nor env var is supplied.

**Stack ordering reminder:** PR #701 (already approved by you) must merge first. Once #701 lands, this PR's diff against ``main`` shrinks to the +366 LOC that are actually unique to it.

Q00 (changes requested):
Reviewed across OS/UserLevel/Program boundaries, auto scope, and UX complexity. Requesting changes: --provenance-file plus env input adds another UserLevel entry point for gateway metadata. The Program need is valid, but the UX should stay gateway/internal unless there is a clear user story for exposing a CLI flag.

shaun0927:
@Q00 — design pivot applied based on your feedback ("--provenance-file plus env input adds another UserLevel entry point ... the UX should stay gateway/internal unless there is a clear user story for exposing a CLI flag").

**What changed (force-pushed):**

- **Removed** the ``--provenance-file`` typer option entirely. There is now no user-facing CLI flag for provenance.
- **Kept only** ``OUROBOROS_AUTO_PROVENANCE_JSON`` as the gateway-internal contract: the gateway sets the env var in the child environment before exec'ing ``ooo auto`` (the deterministic command-rewrite use case from #691). Direct CLI users never see this.
- ``resolve_provenance(env=...)`` now accepts only the env mapping; ``file_path`` parameter removed. The signature shrinks to one kwarg with ``os.environ`` as the implicit default.
- Dropped all file-ingest plumbing: ``is_file()`` regular-file check, bounded read, UTF-8 decode, oversized-file rejection, FIFO/directory/non-UTF-8 regression tests. The env-var size cap (4 KiB) is preserved since the env path is still untrusted gateway input.
- Documentation in ``provenance.py`` rewritten to reflect the gateway-internal scope explicitly: ``"There is no user-facing CLI flag — this metadata is not part of the direct CLI UX and we do not want to surface it as such."``

**Test impact:**

- ``tests/unit/cli/test_auto_provenance.py`` rewritten to cover only the env path (10 tests; was 25 before — the removed 15 covered file-mode behaviours that no longer exist).
- ``tests/unit/auto/test_pipeline_provenance.py`` (PR #704) updated to use ``monkeypatch.setenv`` instead of ``--provenance-file`` for the gateway-tagged scenarios.

**Rationale**: this matches your boundary call. A gateway integration is a Program-level concern; the env-var contract is sufficient for it (the gateway controls the child environment), and the absence of a CLI flag means there is no UserLevel surface to mis-document, mis-discover, or mis-use. Direct CLI runs remain mechanically untouched (``state.provenance`` stays ``None``, no new flags appear in ``--help``).

CI is re-running on the new commit (43128b93). Bot review should land shortly.

PR #705, opened 2026-05-07
feat(auto): add AutoProgressEvent callback contract

shaun0927:
## Summary

Lays down the **observer contract** for live `ooo auto` progress reporting. This is the foundation that the follow-up CLI / MCP / TUI surface PRs in the #639 series will plug into. No user-visible behavior changes by itself.

## Changes

- New module `src/ouroboros/auto/progress.py`:
  - `AutoProgressEvent` — frozen, slotted dataclass with `auto_session_id`, `phase`, `kind` (`phase` / `grade` / `repair`), `message`, optional `round` / `grade`, and an ISO timestamp.
  - `AutoProgressCallback` type alias for the subscriber signature.
- `src/ouroboros/auto/pipeline.py`:
  - `AutoPipeline.progress_callback` (optional, defaults to `None`).
  - Internal `_maybe_emit_phase` / `_maybe_emit_grade` / `_maybe_emit_repair` helpers that dedupe so observers don't see noisy duplicate events for unchanged state.
  - `_save` now opportunistically emits a `phase` event when the persisted phase has actually changed since the last emission, so we don't have to instrument every `state.transition` site.
  - Grade/repair emissions live alongside the existing `repairer.converge(...)` block.
  - Callback exceptions are swallowed — observers can never break the pipeline.
- `tests/unit/auto/test_progress.py`:
  - Dataclass immutability + default factory.
  - No-op behavior when no callback is registered.
  - Happy-path emit ordering through `SEED_GENERATION` → `REVIEW` → `COMPLETE` with `phase` and `grade` events captured (uses stub reviewer/repairer to avoid coupling to grading internals).
  - Exploding callback does not break a successful run.
  - Terminal `BLOCKED` path emits a final `phase` event.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/auto/progress.py src/ouroboros/auto/pipeline.py tests/unit/auto/test_progress.py` → passes.
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/auto -q` → 204 passed.

## Risks / Notes

- The contract is intentionally narrow (3 event kinds). Future PRs may extend it (e.g. `round`-kind events emitted from `AutoInterviewDriver`), but this PR keeps every emission inside `AutoPipeline` to limit churn.
- Phase emission uses a per-instance dedup tracker. Concurrent reuse of the same `AutoPipeline` instance across overlapping `run()` calls would race; the codebase does not do this today.

[...]

PR #698, opened 2026-05-07
feat(auto): tag seed_origin in persisted auto state

shaun0927:
## Summary

First narrow slice of the remaining #639 surface gaps. Adds a `SeedOrigin` enum (`none` / `auto_pipeline` / `external_authoring`) to `AutoPipelineState` so callers can tell a Seed produced by the auto pipeline apart from one persisted via a side-channel authoring call (e.g. a manual `ouroboros_generate_seed`).

This was explicitly requested in the #639 follow-up comment that pointed out an `ooo auto` run could appear to succeed because a Seed had been produced, even though the auto pipeline itself was `phase=blocked` with `seed_path=null`.

## Changes

- `src/ouroboros/auto/state.py`
  - New `SeedOrigin` `StrEnum`.
  - `AutoPipelineState.seed_origin` field, default `SeedOrigin.NONE`.
  - `to_dict` serializes the enum value; `from_dict` hydrates it (legacy sessions backfill to `none`).
  - Validates against the enum and raises an actionable error on unknown values.
- `src/ouroboros/auto/pipeline.py`
  - Sets `state.seed_origin = SeedOrigin.AUTO_PIPELINE` whenever the configured `seed_generator` returns a fresh Seed.
  - Adds `seed_origin` to `AutoPipelineResult` and `_result()`.
- `src/ouroboros/mcp/tools/auto_handler.py`
  - `_result_meta` exposes `seed_origin` so non-CLI clients (Codex/OMX bridges) can render it.
- `src/ouroboros/cli/commands/auto.py`
  - `_print_status` and `_print_result` render `Seed origin: <value>`.
- `tests/unit/auto/test_surface.py`
  - Default value, round-trip, legacy hydration, invalid value rejection.
  - `AutoPipelineResult` default.
  - CLI renders `Seed origin: auto_pipeline` after pipeline-set value.
  - Pipeline integration test that drives `SEED_GENERATION` through a stubbed `seed_generator` and asserts both `state.seed_origin` and `result.seed_origin`.
  - Updated existing strict-equality MCP meta test to include `seed_origin`.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/auto/state.py src/ouroboros/auto/pipeline.py src/ouroboros/cli/commands/auto.py src/ouroboros/mcp/tools/auto_handler.py tests/unit/auto/test_surface.py` → passes.
[...]

Issue #639, opened 2026-05-05
Expose live phase and interview progress for ooo auto sessions


shaun0927:
## Problem

Even when the real `ouroboros_auto` pipeline runs, the current implementation appears to return a final formatted result rather than providing a live user-visible phase trace through the Codex/OMX UI. The pipeline persists state under `~/.ouroboros/data/auto_*.json`, but that state is separate from the visible OMX/Codex runtime status surfaces.

For a long-running `ooo auto` command, users need to know what is happening now:

- Is the command in interview, seed generation, review/repair, or run phase?
- Which interview round is active?
- Which auto answers were selected?
- Did the run block, complete, or start a background execution?
- What command resumes the exact session?

## Evidence

The code already tracks useful state:

- `AutoPipelineState.phase`
- `interview_session_id`
- `current_round`
- `pending_question`
- `last_progress`
- `seed_path`
- `review_grade`
- run handles and resume command

However, the visible runtime behavior observed from Codex/OMX did not show an auto session id, phase transitions, interview rounds, or selected answers. State persistence alone is not enough if the user cannot see it during the run.

## Proposed fix

Add a first-class progress/status surface for `ooo auto`:

- Emit structured progress events at each phase transition.
- Surface `auto_session_id`, phase, current round, and last progress in the user-visible output/HUD.
- Provide a simple status command/path for active auto sessions.
- Record selected auto interview answers in a compact ledger view.

## Acceptance criteria

- A running `ooo auto` session visibly reports phase transitions.
- The first response includes `Auto session: auto_...` as soon as the session is created.
- Interview progress shows round count and source-tagged auto answer summaries.
- Seed review/repair progress shows current grade and repair round.
- The final output includes resume/status instructions even on blocked or failed runs.

---

## Narrowed first PR scope

The first implementation PR should not attempt full live streaming or OMX HUD integration.

Start with a read-only status surface for persisted auto sessions, for example one of:

```text
ouroboros auto status auto_<id>
```

or:

```text
ooo auto --resume auto_<id> --status
```

The command should read the persisted `~/.ouroboros/data/auto_<id>.json` state and print a compact status summary.

## First PR output fields

The initial status command should include:

- `auto_session_id`
- `phase`
- `last_progress`
- `interview_session_id`
- `current_round`
- `pending_question` summary, if present
- `seed_path`
- `last_grade`
- `job_id` / `execution_id` / `run_session_id`, if present
- `last_error` or blocker, if present
- resume command

[...]

PR #713, opened 2026-05-07
feat(auto): stream live phase trace in ooo auto CLI

shaun0927:
## Summary

Wires the `AutoProgressEvent` callback into the `ooo auto` CLI so users see live phase / grade / repair lines as the auto pipeline advances, instead of only the final summary. A `--quiet` flag opts out.

This is the third slice of the #639 surface follow-ups, after the persisted-state `--status` view (#645) and the MCP `_result_meta` (#667 / #698).

## Stacking / dependency

Stacked on **#705** (\`feat/639-B-progress-contract\`). This branch contains those commits at the bottom of its history; please review **#705 first**, then the single new commit on top here. Once #705 merges I will rebase this on `main` and the diff will collapse to the CLI-only changes.

## Changes

- `src/ouroboros/cli/commands/auto.py`
  - New `--quiet` typer option on `auto_command`.
  - `_run_auto` now accepts `progress_callback` and threads it into `AutoPipeline(...)`.
  - New `_make_progress_renderer(*, quiet)` factory:
    - quiet → returns `None` (no callback wired, no streaming).
    - otherwise → returns a callback that prints \`\\[auto] {label} — {message}\` via `print_info`. The leading `[` is escaped so Rich treats \`[auto]\` as literal text rather than a markup style name.
- `tests/unit/auto/test_cli_progress.py`
  - Quiet flag returns `None`.
  - Renderer formats phase, grade, and repair events via captured stdout.
  - `--quiet` is documented in `--help`.
  - `_run_auto(progress_callback=cb)` actually forwards `cb` to `AutoPipeline(progress_callback=...)`.
  - End-to-end CliRunner check: without `--quiet`, the streamed phase line appears in the final output; with `--quiet`, no `[auto]` lines appear and the callback is `None` at the pipeline boundary.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/cli/commands/auto.py tests/unit/auto/test_cli_progress.py` → passes.
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/auto -q` → 209 passed.

## Risks / Notes

- The renderer prints once per distinct phase change (dedup happens inside `AutoPipeline`); long-running phases will not flood stdout.
- Existing CLI snapshot tests / scripted users that grep for the final-summary lines are unchanged. The new lines all carry an unmistakable `\\[auto]` prefix so downstream pipes can filter on it.

Refs #639

### DIFF
--- a/src/ouroboros/auto/interview_driver.py
+++ b/src/ouroboros/auto/interview_driver.py
@@ -17,6 +17,7 @@ from ouroboros.auto.answerer import (
 )
 from ouroboros.auto.gap_detector import Gap, GapDetector
 from ouroboros.auto.ledger import LedgerStatus, SeedDraftLedger
+from ouroboros.auto.progress import AutoProgressCallback, AutoProgressEvent
 from ouroboros.auto.repo_context import repo_auto_answer_context
 from ouroboros.auto.state import AutoPhase, AutoPipelineState, AutoStore
 
@@ -70,9 +71,36 @@ class AutoInterviewDriver:
     store: AutoStore | None = None
     timeout_seconds: float = 60.0
     max_rounds: int = 12
+    progress_callback: AutoProgressCallback | None = None
+    _last_emitted_message: str | None = field(default=None, init=False, repr=False)
+
+    def _emit(self, state: AutoPipelineState) -> None:
+        """Emit a progress snapshot for the current state via the callback.
+
+        Deduped on ``last_progress_message`` so consumers do not see a
+        torrent of identical events for unchanged state. Callback errors
+        are swallowed so an observer can never break the interview loop.
+        """
+        if self.progress_callback is None:
+            return
+        message = state.last_progress_message
+        if message == self._last_emitted_message:
+            return
+        self._last_emitted_message = message
+        event = AutoProgressEvent(
+            auto_session_id=state.auto_session_id,
+            phase=state.phase.value,
+            kind="phase",
+            message=message,
+        )
+        try:
+            self.progress_callback(event)
+        except Exception:
+            pass
 
     async def run(self, state: AutoPipelineState, ledger: SeedDraftLedger) -> AutoInterviewResult:
         """Run bounded auto interview until Seed-ready or blocked."""
+        self._last_emitted_message = None
         self._ensure_interview_phase(state)
         answer_context = self.context_provider(state.cwd)
         interview_tool_name = "interview.start"
@@ -267,6 +295,10 @@ class AutoInterviewDriver:
     def _save(self, state: AutoPipelineState) -> None:
         if self.store is not None:
             self.store.save(state)
+        # Per-round / per-error progress lives in ``state.last_progress_message``;
+        # emit it here so observers see every interview-loop save without each
+        # call site needing to remember to fire the callback.
+        self._emit(state)
 
 
 class FunctionInterviewBackend:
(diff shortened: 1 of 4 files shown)


Q00 (changes requested):
Reviewed. The CLI live trace direction looks useful, but checks are currently running again. I will hold approval until all checks complete successfully.

Q00 (approved):
Reviewed across OS/UserLevel/Program boundaries, auto scope, and UX complexity. Approving: live CLI progress is a UserLevel improvement, but it is opt-out with --quiet, based on an internal observer contract, and does not add a new decision branch to ooo auto.

shaun0927:
@Q00 re-review ping.

This PR wires the **AutoProgressEvent** callback (introduced in #705) into the \`ooo auto\` CLI so users see live phase / grade / repair lines as the auto pipeline advances, instead of only the final summary. \`--quiet\` opts out.

### What changed since the first round of review

The deep review on \`cda3f4d\` raised two non-blocking notes; the deep review on \`0c2be3e\` then raised a more structural concern about the contract not covering interview rounds. Both lines of feedback are now addressed:

1. **\"\`print_info()\` renders a full Rich panel for every event\"** → fixed: \`_make_progress_renderer()\` now emits a single dim line per event via \`console.print(rf\"[dim]\\[auto][/] {label} — {message}\")\`. The leading \`[\` is escaped so consumers can grep on the literal \`[auto]\` prefix without parsing panel chrome.
2. **\"\`RUN\`-resume path updates \`state.last_grade\` but never calls \`_maybe_emit_grade()\`\"** → fixed (the same fix is also in #705 so PR-B is internally consistent).
3. **\"The progress-stream contract is incomplete because the callback is only attached to \`AutoPipeline\`, while \`AutoInterviewDriver\` persists most interview progress through its own \`_save()\` path\"** → fixed by threading the callback down one layer:
   - \`AutoInterviewDriver\` gained a \`progress_callback\` field and a small \`_emit()\` helper that dedupes on \`state.last_progress_message\` and swallows observer exceptions so the loop is never broken by a buggy consumer.
   - \`AutoInterviewDriver._save()\` now fires \`_emit()\` after every persist call, so each centralized save site (round entry, gap-steered answer, blocker, completion) automatically streams a snapshot.
   - \`AutoPipeline.run()\` pushes its own \`progress_callback\` into the driver at the start of every run, so callers wire the observer in one place and the longest-running phase emits through the same contract.

The latest deep review (commit \`4652c75\`) **APPROVED** the change after policy filtering, with no remaining blocking findings. There is one non-blocking *follow-up* note about the \`phase\` discriminator now also carrying intra-phase round progress; that is best addressed by extending \`AutoProgressKind\` (e.g. adding \`\"round\"\`) in a separate slice — it does not change the merge-readiness of this PR but is tracked for the next iteration.

### Why this is safe to merge

- **Strict additive contract**: \`progress_callback\` defaults to \`None\` everywhere; with no callback the CLI behaves exactly as before. The \`--quiet\` flag is also additive.
- **Observer isolation**: callback exceptions are caught and swallowed at both the pipeline and driver layers — no observer can break the auto-pipeline control flow.
- **CI green** across Ruff lint + format, MyPy, pytest on Python 3.12 / 3.13 / 3.14, and Bridge TypeScript.
- **Stacked on #705** — please review #705 first; once it merges, this branch will rebase to drop the duplicate commits and the diff will collapse to the CLI + driver wiring on top.

Refs #639. Ready for human review.

PR #716, opened 2026-05-07
feat(auto): expose auto progress event history through MCP meta

shaun0927:
## Summary

Records every `AutoProgressEvent` emitted during a single `ouroboros_auto` invocation and surfaces the accumulated list as `meta[\"progress_events\"]` so non-CLI bridges (Codex / OMX) can render the same phase / grade / repair trace the CLI streams.

This is the MCP-side companion to the CLI live trace (#713) and depends on the same `AutoProgressEvent` callback contract introduced in #705.

## Stacking / dependency

Stacked on **#705** (\`feat/639-B-progress-contract\`). This branch contains those commits at the bottom of its history; please review **#705 first**, then the single new commit on top here. Once #705 merges I will rebase this on `main` and the diff will collapse to the MCP-only changes.

This PR is independent of #713 (CLI trace) and can land in either order.

## Changes

- `src/ouroboros/mcp/tools/auto_handler.py`
  - `handle()` now allocates a `progress_events: list[AutoProgressEvent]` and threads `progress_events.append` as the pipeline's `progress_callback`.
  - `_run()` accepts an optional `progress_events` list (keyword-only) and wires the callback into `AutoPipeline(progress_callback=...)` only when the caller supplied a list.
  - `_result_meta()` accepts the optional list and adds `meta[\"progress_events\"] = [...]` only when at least one event was captured.
  - New helper `_progress_event_to_dict()` returns a JSON-friendly dict with the keys `{phase, kind, message, round, grade, timestamp}`. The redundant per-event `auto_session_id` is intentionally elided.
- `tests/unit/auto/test_mcp_progress_history.py` (new):
  - Empty event list omits `progress_events` from meta.
  - Three serialized events appear in emit order with the expected keys.
  - End-to-end `AutoHandler.handle({...})` pipes events emitted by a fake pipeline into `meta[\"progress_events\"]`.
  - Pipeline that emits no events leaves the meta key absent.
- `tests/unit/auto/test_surface.py`:
  - Two pre-existing `_run` monkeypatches grew a `**_kwargs` parameter to absorb the new `progress_events` keyword. No functional changes.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ouroboros/mcp/tools/auto_handler.py tests/unit/auto/test_mcp_progress_history.py` → passes.
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/auto -q` → 208 passed.

## Risks / Notes

- Long-running auto sessions accumulate one entry per phase change + one per grade + one per repair round. With the existing dedup in `AutoPipeline._maybe_emit_phase`, the worst case for a 12-round interview + 5 repair rounds is ~25 entries — comfortably below any reasonable MCP payload threshold.
- The `meta[\"progress_events\"]` key is additive and absent when no events were emitted, so existing clients that do not consume it are unaffected.

Refs #639

### DIFF
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -4,12 +4,13 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Awaitable, Callable
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 from typing import Any
 
 from ouroboros.auto.grading import GradeGate
 from ouroboros.auto.interview_driver import AutoInterviewDriver
 from ouroboros.auto.ledger import SeedDraftLedger
+from ouroboros.auto.progress import AutoProgressCallback, AutoProgressEvent
 from ouroboros.auto.seed_repairer import SeedRepairer
 from ouroboros.auto.seed_reviewer import SeedReview, SeedReviewer
 from ouroboros.auto.state import AutoPhase, AutoPipelineState, AutoStore, utc_now_iso
@@ -75,9 +76,16 @@ class AutoPipeline:
     reconcile_source: str | None = None
     seed_timeout_seconds: float = 120.0
     run_start_timeout_seconds: float = 60.0
+    progress_callback: AutoProgressCallback | None = None
+    _last_emitted_phase: str | None = field(default=None, init=False, repr=False)
+    _last_emitted_grade: str | None = field(default=None, init=False, repr=False)
+    _last_emitted_repair: int | None = field(default=None, init=False, repr=False)
 
     async def run(self, state: AutoPipelineState) -> AutoPipelineResult:
         """Run a bounded auto pipeline using injected side-effecting dependencies."""
+        self._last_emitted_phase = None
+        self._last_emitted_grade = None
+        self._last_emitted_repair = None
         ledger = (
             SeedDraftLedger.from_dict(state.ledger)
             if state.ledger
@@ -251,6 +259,8 @@ class AutoPipeline:
             state.last_grade = review.grade_result.grade.value
             state.findings = [asdict(finding) for finding in review.findings]
             state.ledger = ledger.to_dict()
+            self._maybe_emit_repair(state)
+            self._maybe_emit_grade(state)
             if self.seed_saver is not None:
                 try:
                     state.seed_path = self.seed_saver(seed)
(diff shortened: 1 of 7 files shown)


Q00 (changes requested):
Reviewed. Exposing MCP progress history is a public contract change, so the existing requested changes need to be resolved first. I am not approving this yet.

Q00 (changes requested):
Reviewed across OS/UserLevel/Program boundaries, auto scope, and UX complexity. Requesting changes: exposing raw progress_events through MCP meta creates a Program-level contract and can grow into a second event API. Please define a stable schema/version and bounded retention/summary behavior before making bridge clients depend on this payload.

shaun0927:
@Q00 re-review ping.

This PR is the MCP-side companion to the CLI live trace (#713) and depends on the same \`AutoProgressEvent\` callback contract introduced in #705. It records every \`AutoProgressEvent\` emitted during a \`ouroboros_auto\` invocation onto the **persisted** \`AutoPipelineState\` and surfaces the accumulated history as \`meta[\"progress_events\"]\` so non-CLI bridges (Codex / OMX) can render the same phase / grade / repair trace the CLI streams — and, crucially, can keep that history coherent across resume boundaries.

### Bot verdict: APPROVE (commit \`9aceeef\`)

> The change is directionally sound: progress history is now persisted on the session state and re-hydrated by the MCP handler from a single durable source of truth, which fixes the \"one save behind\" problem and makes resumes observable across invocations.

### What changed across the review cycles

The first deep review on \`fa26294\` flagged a fundamental contract gap; the bot raised three additional concrete issues on the fix commits, all of which are now closed:

1. **\"\`progress_events\` is initialized as a fresh in-memory list for each \`handle()\` call ... clients will receive a truncated history after any multi-call or resumed session\"** → fixed: progress history was moved onto the persisted \`AutoPipelineState.progress_events\` field. \`AutoHandler.handle()\` now reloads the state after \`_run\` completes and surfaces the durable log as \`meta[\"progress_events\"]\`.
2. **\"\`_save()\` persists state *before* \`_maybe_emit_phase()\` appends the new phase event ... terminal complete/blocked/failed events are never persisted at all\"** → fixed: \`_save()\` now emits *before* persisting, so the matching event lands in \`state.progress_events\` and gets durably written by the same store call. New regression test runs the pipeline end-to-end through a passing reviewer/repairer and asserts the persisted state ends in COMPLETE with the matching final \`phase\` event.
3. **\"\`run()\` resets \`_last_emitted_phase\` and then immediately calls \`_save(state)\` before any transition, so every resume/no-op invocation appends a fresh \`phase\` event for the current phase even when nothing changed\"** → fixed: dedup trackers are now seeded from the *current* persisted state on \`run()\` entry, so a no-op resume — including a resume of an already-terminal session — appends no spurious entries. New regression test resumes a complete session and asserts the persisted history length is unchanged.
4. **\"The new persisted ring-buffer behavior is untested at the 100-entry boundary\"** *(non-blocking note in the approval review)* → addressed by an additional unit test that fires \`AutoPipeline._emit\` past \`_PROGRESS_EVENT_LIMIT + 25\` and asserts the cap, FIFO eviction, and newest-entry survival.

### Why this is safe to merge

- **Strictly additive on the persisted schema**: \`progress_events\` defaults to \`[]\`, hydrates legacy persisted sessions to \`[]\`, and is validated as a list of dicts on load.
- **Bounded persistent log**: ring buffer caps at \`_PROGRESS_EVENT_LIMIT = 100\` (≈ 25 KB worst case), so resume state files stay compact even on long-running sessions; eviction is FIFO and now boundary-tested.
- **Single source of truth**: every event consumer (callback, MCP meta, future CLI/HUD) reads from the same \`state.progress_events\` log. No more in-memory drift between \`AutoPipeline\` and \`AutoHandler\`.
- **Defensive against degraded storage**: an unloadable \`AutoStore\` after \`_run\` is caught — the meta payload is still produced, just without the history block. Covered by a regression test.
- **CI green** across Ruff lint + format, MyPy, pytest on Python 3.12 / 3.13 / 3.14, and Bridge TypeScript.
- **Stacked on #705** — please review #705 first; once it merges, this branch will rebase to drop the duplicate commits and the diff will collapse to the persistence + handler-reload changes on top.

Refs #639. Ready for human review.

shaun0927:
Closing per Q00's review feedback combined with rebase complexity.

Q00's review:
> "Exposing raw progress_events through MCP meta creates a Program-level contract and can grow into a second event API. Please define a stable schema/version and bounded retention/summary behavior before making bridge clients depend on this payload."

The PR has 9 commits with internally superseding revisions (e.g., one commit seeds dedup from live state, a later commit re-implements the same dedup from `progress_events` log). The cleanest path is to rebuild as a focused PR that:

1. Defines a versioned schema (`progress_events_v1`) before exposing it
2. Bounded retention (e.g., last N events with a documented cap) up front
3. Single, final dedup design (not the multi-revision history this branch carries)
4. Drops the `progress_events` MCP exposure if the schema/retention design isn't ready in v0 — UX cost low (no client depends yet), schema cost high (locks in a Program contract)

Rebase tried — `state.py` `for field_name in (...)` block, `pipeline.py` dedup logic, multiple test additions all conflict with current main. Conflict count makes manual resolution risky vs a clean fresh PR.

Refs #639, #725.

PR #746, opened 2026-05-07
feat(plugin): lockfile + per-user trust store

shaun0927:
Closes #732. Sub-issue of #725 (Phase 2). **Stacked on #745 (CR-3 manifest loader)** — review after that lands; the diff against `main` will be larger until then.

### `src/ouroboros/plugin/lockfile.py`

`LockEntry` dataclass + `Lockfile` manager. Persists at `~/.ouroboros/plugins.lock` (TOML).

- Atomic writes (temp file + `os.replace`) — partial-write crash leaves the original intact.
- POSIX `fcntl.flock` for concurrent-write safety.
- Deterministic name-sorted entry order so diffs are reviewable.
- Schema versioned (`schema_version = "0.1"`); unsupported versions raise on read.
- Hand-rolled minimal TOML serialization (the shape is small and fixed); reading uses stdlib `tomllib`.

### `src/ouroboros/plugin/trust_store.py`

`GrantedScope`, `TrustRecord` dataclasses + `TrustStore` manager. Implements all four locked answers from Q00/ouroboros-plugins#9 that affect persistence:

| Q | Behavior |
|---|---|
| Q3 | Exact-scope only. `TrustRecord.has_scope("github:pull_request")` does **not** match `github:pull_request:write`. |
| Q4 | Version-bump invalidation. `grant(plugin=X, version=0.2.0, ...)` after `0.1.0` was trusted drops the prior grants. Explicit `reset_for_version_bump()` available too. |
| Q5 | Per-user storage at `~/.ouroboros/plugins/<name>/trust.json`. Each plugin gets its own file. |
| Q6 | `granted_by`, `granted_scope`, `granted_at` recorded. Firewall (#729) and CLI (#731) source `plugin.trusted` audit events from these fields. |

**No tokens stored.** The store has no token API; `test_no_raw_token_in_persisted_file` enforces this.

## Tests (16/16 pass)

```
$ PYTHONPATH=src python3 -m pytest \
    tests/unit/plugin/test_lockfile.py \
    tests/unit/plugin/test_trust_store.py -v
============================== 16 passed in 0.78s ==============================
```

### Lockfile (7)

1. install → read round-trip
2. remove drops entry; idempotent on already-absent
3. entries written in sorted name order
4. `schema_version` header present
5. unsupported `schema_version` rejected on read
6. atomic write: simulated `os.replace` crash leaves the original file intact, no leftover temp file
7. concurrent multiprocess writes (10 entries from 2 procs) do not corrupt the file or lose entries

[...]

PR #745, opened 2026-05-07
feat(plugin): manifest loader + vendored 0.1 schemas

shaun0927:
Closes #728. Sub-issue of #725 (Phase 1).

Implements the locked spec for the UserLevel plugin manifest loader.

## Module: `src/ouroboros/plugin/manifest.py`

- `PluginManifest` — frozen dataclass with the locked **8 required + 2 optional** field shape (per Q00/ouroboros-plugins#6).
- `load_manifest(path)` — returns a frozen, validated manifest.
- `PluginManifestError` exposes structured fields: `path`, `json_pointer`, `expected`, `got`. Tests assert on `json_pointer` rather than parsing message text — schema changes that move the failing field cannot silently slip through.
- `source.type=first_party` is a real branch (per Q00/ouroboros-plugins#8); `source.path` / `source.repository` optional for it.
- `schema_version` routes to `schemas/<major>/plugin.schema.json` (per Q00/ouroboros-plugins#11). Out-of-window versions reject with a clear message naming the supported window.

## Vendored schemas

`src/ouroboros/plugin/schemas/0.1/{plugin,audit-event}.schema.json` mirror upstream `Q00/ouroboros-plugins/schemas/0.1/` with all locked decisions applied:

- 8 required + 2 optional (#6)
- 3-value risk enum (#10)
- per-major archive layout (#11)

`schemas/0.1/_source.json` records the upstream repo + intent. A `scripts/sync-plugin-schemas.sh` follow-up will sync these once Q00/ouroboros-plugins#13 / #16 / #19 / #20 land on upstream main.

## Dependency

`pyproject.toml` adds `jsonschema>=4.21.0,<5.0.0` as a runtime dep.

## Tests — `tests/unit/plugin/test_manifest.py` (13/13 pass)

| # | Test | Asserts |
|---|---|---|
| 1 | Reference manifest loads cleanly | `name == "github-pr-ops"`, command count, source.type |
| 2 | Missing required field | `json_pointer == ""`; message names "name" |
| 3 | Pattern violation on `name` | `json_pointer == "/name"` |
| 4 | Unknown capability | `json_pointer == "/capabilities/0/name"` |
| 5 | Unknown source.type | `json_pointer == "/source/type"` |
| 6 | additionalProperties:false | unknown key reported by name |
| 7 | Unsupported `schema_version` | `json_pointer == "/schema_version"`; message names support window |
| 8 | Returned manifest is frozen | `FrozenInstanceError` |
[...]

Issue #732, opened 2026-05-07
Phase 2: Implement trust store and ~/.ouroboros/plugins.lock


shaun0927:
> ## ✅ Decisions LOCKED — implementation-ready
>
> Upstream contract decisions resolved 2026-05-07 by @shaun0927:
> - **Trust UX (Q00/ouroboros-plugins#9)**:
>   - Q4 lock: trust **invalidated on any version bump**.
>   - Q5 lock: trust storage is **per-user** at `~/.ouroboros/plugins/<name>/trust.json`.
>   - Q6 lock: `plugin.trusted` event has 4 mandatory provenance keys: `granted_by`, `granted_scope`.

## Sub-issue of #725 (Phase 2)

Persist plugin trust state and installation provenance.

### 1. Lockfile at `~/.ouroboros/plugins.lock` (TOML)

```toml
schema_version = "0.1"

[[plugin]]
name = "github-pr-ops"
version = "0.1.0"
source_kind = "git"
repository = "https://github.com/Q00/ouroboros-plugins"
git_sha = "b3a91f2..."
manifest_checksum = "sha256:..."
installed_at = "2026-05-08T03:14:00Z"
plugin_home = "~/.ouroboros/plugins/github-pr-ops"
```

- Atomic writes (temp file + rename)
- File-locked for concurrent safety (`portalocker` or stdlib `fcntl`)
- Deterministic ordering (sorted by name) so diffs are reviewable

### 2. Trust store at `~/.ouroboros/plugins/<name>/trust.json` (per Q00/ouroboros-plugins#9 Q5 lock)

```json
{
  "schema_version": "0.1",
  "plugin": "github-pr-ops",
  "version": "0.1.0",
  "granted_scopes": [
    {
      "scope": "github:read",
      "granted_at": "2026-05-08T03:20:00Z",
      "granted_by": "user:shaun0927"
    }
  ]
}
```

- Per-user, per-plugin (one file per installed plugin)
- Trust invalidated on version bump (Q4 lock): when `ooo plugin update` changes the version, the trust file is **emptied of granted_scopes** and a `plugin.trusted` event with `result.status="success", message="trust reset due to version bump"` is emitted (or no event if no scopes were granted).
- Never stores raw tokens

## Required behavior

- `ooo plugin install` (#731): writes lockfile entry, never trust
[...]

Issue #731, opened 2026-05-07
Phase 2: Add ooo plugin {add,discover,inspect,install,trust,disable,remove,list} CLI


shaun0927:
## Summary

Sub-issue of #725 (Phase 2).

Add the user-facing `ooo plugin` CLI: install, inspect, trust, disable, remove, list. The UX must mirror the lifecycle documented in [Q00/ouroboros-plugins/docs/lifecycle.md](https://github.com/Q00/ouroboros-plugins/blob/main/docs/lifecycle.md) and the repo-URL flow agreed in #725.

## Required commands

```
ooo plugin add <repo-url>           # multi-select prompt, see #725 last comment
ooo plugin add <repo-url> --plugin <name>   # non-interactive
ooo plugin add <local-path>         # local-path source

ooo plugin discover <path>          # inspect manifest without registering
ooo plugin install <name>           # actually register (no trust grant)
ooo plugin inspect <name>            # show manifest, capabilities, permissions, trust state

ooo plugin trust <name> --scope <permission-scope>
ooo plugin disable <name>
ooo plugin remove <name>

ooo plugin list                      # all installed plugins + trust state
```

## Required behavior

- **Install string sanitization**: reject the subdirectory-leaking pattern documented as anti-pattern in the lifecycle doc:

  ```
  git+https://github.com/Q00/ouroboros-plugins.git#plugins/github-pr-ops
  ```

  Error message: `error: subdirectory-form install strings (#plugins/...) are not supported. Use 'ooo plugin add <repo-url> --plugin <name>' instead.`

- **Catalog discovery**: when `add` receives a repo URL, perform a shallow clone (`git clone --depth 1`) into a temp directory under `~/.ouroboros/cache/`, read each `plugins/<name>/ouroboros.plugin.json`, present the multi-select. After install, the temp clone is moved to `~/.ouroboros/plugins/<name>/` per selected plugin (Q00/ouroboros#732 owns the destination convention).

- **Trust grants are scoped**: `trust <name> --scope github:read` does **not** imply `github:pull_request:write`.

- **Audit emission**: every command that mutates plugin state emits the appropriate `plugin.*` event from the audit schema (via the firewall in #729). `discover` and `inspect` are read-only and emit only `plugin.discovered` (no trust granted).

## Required interactive UX

When `add <repo-url>` finds a multi-plugin catalog, the prompt:

```
Repository: Q00/ouroboros-plugins (b3a91f2)

Select plugins to install:

[...]

PR #747, opened 2026-05-07
feat(plugin): UserLevel program registry

### LINKED ISSUE #730: Phase 1: Extend src/ouroboros/plugin/skills/registry.py to register UserLevel programs
## Summary

Sub-issue of #725 (Phase 1).

Extend the existing `src/ouroboros/plugin/skills/registry.py` to also register **UserLevel programs** alongside bundled skills, so both flow through the same lookup table.

The decision agreed in #725 discussion: **reuse, do not introduce a new system**.

## Why this matters

If UserLevel plugins get their own parallel registry:

- Two places to look up "does this command exist"
- Risk of namespace collisions across the two registries
- Doubled CLI dispatch logic

Sharing the registry keeps a single source of truth.

## Required changes

1. **Define a `RegisteredEntry` interface** that both bundled skills and UserLevel programs implement (`name`, `commands`, `invoke(...)`, `entry_kind`).
2. **Add `register_userlevel_program(manifest: PluginManifest, ...)`** alongside the existing skill registration entry points.
3. **Namespace collision detection**: a UserLevel program declaring `namespace=foo` must collide-check against existing skill names and against other UserLevel namespaces.
4. **Lookup**: existing `lookup(name)` continues to work; it now also returns UserLevel matches.
5. **Distinguishability**: the API can ask "is this a UserLevel entry?" without breaking existing skill code paths.

## Tests

- `tests/plugin/skills/test_registry.py` — extend
- Test: register `github-pr-ops`; lookup by namespace returns it
- Test: registering a UserLevel program whose namespace collides with an existing skill is rejected
- Test: existing skill registration tests still pass unchanged
- Test: lookup-by-command-name routes to the right entry kind

## Acceptance criteria

- [ ] Single registry handles both kinds of entries
- [ ] Namespace collisions rejected with a clear error
- [ ] No existing skill test broken
- [ ] CLI dispatch (Phase 2 — separate sub-issue O6) can route via this single registry

## Related

- Parent: #725
- Phase 1 siblings: O3 (loader), O4 (firewall)
- Consumed by: O6 (CLI routing)

shaun0927:
Closes #730. Sub-issue of #725 (Phase 1). **Stacked on #745 (CR-3 manifest loader)**.

## Module: `src/ouroboros/plugin/userlevel_registry.py`

In-memory registry of installed UserLevel programs, keyed by plugin name and namespace.

- `register(manifest, replace=False)` — namespace ownership enforced; first registration wins, subsequent for the same namespace raise `RegistryError`. Duplicate name requires `replace=True`.
- `unregister(name)` — releases both name AND namespace.
- `get(name)`, `get_by_namespace(namespace)`, `all_programs()`.
- Process-wide singleton via `get_userlevel_registry()`; `reset_userlevel_registry()` for tests.

### Cross-registry lookup

`lookup_command(name)` — top-level helper that satisfies the locked spec's "single lookup":

```
lookup_command(name)
  → UserLevel registry by namespace
  → UserLevel registry by name
  → bundled SkillRegistry by name
  → LookupResult(kind="none")
```

Returns `LookupResult` with `kind ∈ {"userlevel", "skill", "none"}`.

## Design choice — alongside, not merged into SkillRegistry

The locked spec asked to "extend `src/ouroboros/plugin/skills/registry.py`". After reading the existing module (700 LOC, watchdog hot-reload, YAML frontmatter parsing) I concluded that **merging JSON-manifest plugins into SkillRegistry's class would more than double its size and lifecycle complexity**. The two artifact shapes have fundamentally different semantics:

- Skills: `.claude-plugin/skills/<name>/SKILL.md`, hot-reloaded via watchdog, YAML frontmatter.
- UserLevel programs: `~/.ouroboros/plugins/<name>/ouroboros.plugin.json`, install/trust lifecycle, JSON manifest with audit contract.

Instead, the new registry sits alongside. The cross-registry helper is the "one lookup table" the spec implies — and existing skill tests are **unchanged** (regression checked: 46 passed, 1 skipped).

If the maintainer prefers the literal "extend the same class" interpretation, that's a follow-up refactor; this PR's interface (`lookup_command`) lets the firewall (#729) and CLI (#731) consume both registries without caring about the underlying class structure.

## Tests — 10/10 pass; no regression on skills/test_registry.py

```
$ PYTHONPATH=src python3 -m pytest tests/unit/plugin/test_userlevel_registry.py -v
============================== 10 passed in 0.24s ==============================

$ PYTHONPATH=src python3 -m pytest tests/unit/plugin/skills/test_registry.py
======================== 46 passed, 1 skipped in 0.29s =========================
```

| # | Test | Asserts |
|---|---|---|
| 1 | register/get/get_by_namespace happy path | round-trip, unknown returns None |
| 2 | namespace collision rejected | `RegistryError("already owned")` |
| 3 | duplicate name requires `replace=True` | first error, then succeeds with `replace=True` |
| 4 | unregister releases name AND namespace | another plugin can claim the namespace after |
| 5 | `all_programs()` returns every entry | set-equality check |
| 6 | `lookup_command` via namespace | `kind="userlevel"` |
| 7 | `lookup_command` via name | `kind="userlevel"` |
| 8 | unknown name returns `kind="none"` | |
| 9 | global singleton identity | `get_userlevel_registry()` is `get_userlevel_registry()` |
| 10 | skill registry unaffected | regression guard |

## What's next

#729 (firewall) and #731 (CLI) consume `lookup_command` to dispatch. The firewall instantiates an in-memory `UserLevelProgramRegistry` from the lockfile (#732 — already in #746) on startup; the CLI registers/unregisters as plugins are installed/removed.

### DIFF
--- a/src/ouroboros/plugin/userlevel_registry.py
+++ b/src/ouroboros/plugin/userlevel_registry.py
@@ -62,11 +62,26 @@ class RegisteredProgram:
 
 
 class UserLevelProgramRegistry:
-    """In-memory registry of installed UserLevel programs."""
+    """In-memory registry of installed UserLevel programs.
+
+    Maintains three indexes that must stay in lockstep with `_by_name`:
+
+    - `_namespace_owner` — `namespace → plugin name` (one namespace per
+      plugin).
+    - `_command_owner` — `command name → plugin name` (command names are
+      globally unique because `lookup_command(name)` resolves bare command
+      names; without uniqueness the same name could dispatch to different
+      plugins depending on registration order).
+
+    Every mutation (`register`, `unregister`) is performed under `_lock`
+    and updates all relevant indexes atomically; tests that manipulate
+    these indexes from the outside are expected to use the public API.
+    """
 
     def __init__(self) -> None:
         self._by_name: dict[str, RegisteredProgram] = {}
         self._namespace_owner: dict[str, str] = {}  # namespace -> plugin name
+        self._command_owner: dict[str, str] = {}  # command name -> plugin name
         self._lock = RLock()
 
     def register(self, manifest: PluginManifest, *, replace: bool = False) -> RegisteredProgram:
@@ -81,7 +96,8 @@ class UserLevelProgramRegistry:
             The registered program.
 
         Raises:
-            RegistryError: on namespace collision or duplicate name without
+            RegistryError: on namespace collision, command-name collision
+                across plugins, or duplicate plugin name without
                 `replace=True`.
         """
         if not manifest.commands:
@@ -96,6 +112,24 @@ class UserLevelProgramRegistry:
             )
         namespace = namespaces.pop()
 
+        # Within-manifest command-name uniqueness. The vendored 0.1 schema
+        # does not enforce uniqueness on `commands[*].name`, but the
+        # registry treats command names as a primary lookup key — without
+        # this guard a manifest with two commands named `review` would
+        # accept silently and `find_command()` / `get_by_command()` would
+        # always return the first array entry, hiding the second spec.
+        seen: set[str] = set()
+        duplicates: list[str] = []
+        for c in manifest.commands:
+            if c.name in seen and c.name not in duplicates:
+                duplicates.append(c.name)
+            seen.add(c.name)
+        if duplicates:
+            raise RegistryError(
+                f"{manifest.name}: duplicate command name(s) within manifest: {sorted(duplicates)}"
+            )
+        new_command_names = seen
+
         with self._lock:
             existing = self._by_name.get(manifest.name)
             if existing is not None and not replace:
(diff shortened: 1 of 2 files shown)


shaun0927:
@Q00 — re-review ping. The bot just APPROVED `3dcddda6` and the PR is `MERGEABLE` / `CLEAN`. Quick recap of where this landed and why I think it can ship.

## What this PR is

Per #730 / Phase 1 of #725, this is the **in-memory UserLevel program registry** that the firewall (#729) and CLI (#731) consume. Stacked on #745 (manifest loader). The PR also folds in tightening for the parent loader's contract surface (the bot kept finding gaps as soon as the registry exposed new lookup paths into it, so I fixed them here rather than splitting yet another stack).

The deliverables are unchanged from the original description:

- `UserLevelProgramRegistry` with `register / unregister / get / get_by_namespace / get_by_command / all_programs`
- `lookup_command(name)` cross-registry resolver: namespace → plugin name → command name → bundled SkillRegistry
- Process-wide singleton + `reset_userlevel_registry()` for tests
- 0 changes to the existing skill `SkillRegistry`; `tests/unit/plugin/skills/test_registry.py` runs green untouched

## What changed during the review loop

The bot ran 7 review rounds; each surfaced a real correctness gap that the previous patch had not yet covered. They are all addressed in atomic commits — happy to land them squashed if you prefer.

| # | Finding | Fix |
|---|---|---|
| 1 | `register(replace=True)` left the old namespace bound to the plugin in `_namespace_owner` if the new manifest declared a different namespace — `get_by_namespace(old_ns)` returned the new program and the old slot was permanently leased to a name that no longer existed | Release the stale namespace inside the lock, transactional with the new install (a897f61a) |
| 2 | The 0.1 schema only required `source.type`, so `{"source": {"type": "local_path"}}` validated and built `SourceSpec(path=None)`, pushing failure into install/run code instead of the loader's contract boundary | Post-schema check requiring `source.path` for `local_path` / `plugin_home`; `first_party` exempt per Q00/ouroboros-plugins#8 (2d8c5e13) |
| 3 | `lookup_command()` advertised resolution by command name but only checked namespace + plugin name; the skills branch silently returned `kind="none"` in a fresh process because `discover_all()` had never run | Added command-name resolution; documented the skill discovery precondition explicitly (the helper deliberately does not trigger `discover_all` to avoid async-from-sync); regression test exercises the post-discovery success path with a tmp skill_dir (e97feba1) |
| 4 | Two plugins could legally register different namespaces while both exposing a command like `review`; `lookup_command("review")` would dispatch by registration order | Third index `_command_owner`, global command-name uniqueness in `register`, transactional cleanup on `replace`, exposed `get_by_command` for O(1) (890664d0) |
| 5 | `load_manifest` only translated `JSONDecodeError`; `OSError` from a `PermissionError`, broken mount, etc. leaked past the loader contract | `OSError` arm wrapping with the same structured shape as the rest (84631d88) |
| 6 | A manifest with invalid UTF-8 raised `UnicodeDecodeError` (not an `OSError`) and bypassed the I/O wrapper | Dedicated `UnicodeDecodeError` arm reporting encoding + byte offset (79d23488) |
| 7 | A single manifest declaring the same command name twice was silently accepted (set-deduped); the second spec became unreachable through `find_command()` | Within-manifest duplicate-name check before deduping (d31e23a9) |
| 8 | Per-axis uniqueness was not enough: a name reserved on a higher-priority axis by one plugin could shadow a different plugin's lower-axis match (`name="review"` vs another plugin's command `review`, or namespace vs command name) | Single cross-axis sweep treating `{plugin_name} ∪ {namespace} ∪ commands` as one identifier space; rejects shadowing by other plugins; self-owned identifiers exempt so `replace=True` still works (3dcddda6) |

Each fix landed with a regression test. Total test count: **179 passed, 1 skipped** in the plugin suite (`tests/unit/plugin/`), with the 46 existing skill-registry tests untouched.

## Why this is merge-ready

- **Contract boundary holds**: `load_manifest()` now translates every observable failure mode (schema, JSON decode, UTF-8 decode, OS-level read errors, missing-path semantics) into structured `PluginManifestError` with a JSON Pointer.
- **Registry invariants are mechanical**: three indexes (`_by_name`, `_namespace_owner`, `_command_owner`) updated atomically under `RLock`; `register / unregister / replace` paths each have a regression test covering both happy-path and stale-cleanup.
- **`lookup_command` dispatch is unambiguous**: cross-axis collisions can no longer register, so any single string resolves to exactly one program (or none).
- **CI is fully green**: Ruff Lint, Ruff Format, MyPy, Bridge TypeScript, Tests on Python 3.12 / 3.13 / 3.14 — all SUCCESS on `3dcddda6`.
- **Bot APPROVED** at the latest commit with no remaining findings.

## What is intentionally out of scope

- Subsuming `SkillRegistry` into a unified class. The original spec language said "extend `src/ouroboros/plugin/skills/registry.py`"; I took the pragmatic interpretation that a shared `lookup_command` cross-registry helper is the "single lookup table" the spec implies, and left the two registries as siblings. The skill registry module is unmodified and its 46 existing tests still pass. Happy to do the merge as a follow-up if you'd rather see literal extension.

Holding off on `Squash and merge` for your call. Thanks for the review!

PR #748, opened 2026-05-07
feat(plugin): invocation firewall + audit-event emitter

shaun0927:
Closes #729. Sub-issue of #725 (Phase 1).

**Stacked on** #745 (CR-3 manifest loader), #746 (CR-7 lockfile + trust store), #747 (CR-5 UserLevel registry). Branches merged into this one for cross-module integration; review after the three foundation PRs land.

## Module: `src/ouroboros/plugin/firewall.py`

`invoke_plugin(program, command_name, argv, *, trust_record, event_sink, correlation_id, confirm, subprocess_runner)`.

### Behavior, in locked order

1. **Pre-invocation trust check** (Q00/ouroboros-plugins#9 Q1):
   - Every `required: true` permission must be in the trust record (exact-scope match per Q3).
   - On any missing scope: emit **only** `plugin.failed` (status=blocked) with a clean error message; **NO `plugin.invoked` is emitted**. Return `InvocationResult(status="blocked")`.
   - First-party programs (`source.type=first_party`) skip the check (per Q00/ouroboros-plugins#8 lock).

2. **Confirmation gate** (Q00/ouroboros-plugins#9 Q2):
   - If `command.requires_confirmation=true`, call `confirm()` exactly once.
   - On decline: `plugin.failed` (blocked, "user declined confirmation"); subprocess never launched.

3. **Emit `plugin.invoked`** before launching subprocess.

4. **Emit one `plugin.permission_used` per `required: true` permission** (locked Option (a) coarse rule). Optional permissions are NOT emitted in v0.

5. **Run entrypoint** (subprocess.run by default; injectable via `subprocess_runner` for tests).

6. **Emit `plugin.completed`** (success) or `plugin.failed` (failed) with exit code in `result.message`.

### Bounded payloads

- `argv` stored as-is (already user-facing).
- Raw stdout/stderr NOT copied to events; only sha256 hashes in provenance (`stdout_sha256`, `stderr_sha256`).
- Tokens, channel IDs, free-form messages forbidden by contract.

All events conform to `schemas/0.1/audit-event.schema.json` (vendored in #745).

`EventSink` is a callable — the core ledger writer in production, `list.append` in tests. The firewall does not own the audit log.

## Tests — 9/9 firewall, 193 plugin total

```
$ PYTHONPATH=src python3 -m pytest tests/unit/plugin/ -v
======================== 193 passed, 1 skipped in 0.64s ========================
```

| # | Test | Asserts |
|---|---|---|
[...]

Issue #733, opened 2026-05-07
Phase 3: End-to-end contract proof with github-pr-ops (read-only path only)


shaun0927:
> ## ✅ Decisions LOCKED — implementation-ready
>
> Upstream decisions resolved 2026-05-07 by @shaun0927:
> - **Reference plugin scope (Q00/ouroboros-plugins#7)**: `merge` removed from v0. v0 reference is **read-only `review` only**.
> - **Trust UX (Q00/ouroboros-plugins#9)**: Q1 (refuse + clear error format) consumed by Path 2 expected behavior.

## Sub-issue of #725 (Phase 3)

End-to-end contract proof: install [Q00/ouroboros-plugins/plugins/github-pr-ops](https://github.com/Q00/ouroboros-plugins/tree/main/plugins/github-pr-ops), invoke `ooo github-pr review <pr-url>`, and verify the audit trail.

### Path 1 — read-only success

```bash
ooo plugin add https://github.com/Q00/ouroboros-plugins
# → multi-select, choose github-pr-ops
ooo plugin trust github-pr-ops --scope github:read
ooo github-pr review https://github.com/Q00/ouroboros/pull/725
```

Expected ledger trail (in this exact order):

```
1. plugin.discovered
2. plugin.installed
3. plugin.trusted        # provenance contains granted_by, granted_scope (per #9 Q6)
4. plugin.invoked
5. plugin.permission_used  # scope=github:read (per #729 firewall step 4)
6. plugin.completed
```

### Path 2 — trust violation

```bash
# After install, skip trust step
ooo github-pr review https://github.com/Q00/ouroboros/pull/725
```

Expected per locked #9 Q1: error message format `"plugin requires \`github:read\` (read_only), which is not yet trusted. Run: ooo plugin trust github-pr-ops --scope github:read"`.

Ledger trail:

```
1. plugin.discovered    # from earlier add
2. plugin.installed     # from earlier install
3. plugin.failed        # status=blocked, message string-matches the format above
                        # NO plugin.invoked (per #729 firewall early-return rule)
```

### Path 3 — subprocess failure

Force the plugin entrypoint to exit non-zero (e.g., malformed PR URL after trust grant; entrypoint validates URL after firewall pre-check passes).

Ledger trail:

```
1. plugin.invoked
2. plugin.permission_used  # github:read still emitted at start
3. plugin.failed           # status=failed, exit code in result.message
```

## What this exercises

[...]

Issue #725, opened 2026-05-07
Design a UserLevel plugin manager for operational workflows


Q00:
## Vision

Ouroboros should be understood as an OS-like core plus UserLevel programs built on top of it.

The core should provide stable primitives. UserLevel programs should compose those primitives into useful workflows. Some UserLevel programs can be first-party and ship with Ouroboros, such as `ooo auto`. Others should be installable plugins maintained by teams or the ecosystem.

This gives us a cleaner boundary than putting every useful workflow into core, or into `ooo auto`.

## Layer Model

```text
+-------------------------------------------------------------------+
|                Installable UserLevel Programs                      |
|                                                                   |
|  github-pr-ops   merge-assistant   jira-sync   linear-triage       |
|  slack-incident  release-coordinator  customer-debugger  ...       |
+-------------------------------+-----------------------------------+
                                |
                                | plugin contract / declared scopes
                                v
+-------------------------------------------------------------------+
|                First-party UserLevel Programs                      |
|                                                                   |
|  ooo auto     ooo run     ooo pm     ooo review?     ...           |
|                                                                   |
|  These are product-level workflows maintained with Ouroboros,       |
|  but they are still programs above the core, not the core itself.   |
+-------------------------------+-----------------------------------+
                                |
                                | uses stable OS primitives
                                v
+-------------------------------------------------------------------+
|                         Ouroboros Core / OS                         |
|                                                                   |
|  Seed      Ledger      State      Runtime      MCP                 |
|  Provenance  Safety Boundaries  Progress/Status  Handoff           |
+-------------------------------+-----------------------------------+
                                |
                                | bounded adapters / external calls
                                v
+-------------------------------------------------------------------+
|                    External Systems / Runtimes                      |
|                                                                   |
|  GitHub   Jira   Linear   Slack   CI   Local repo   Agent CLIs      |
+-------------------------------------------------------------------+
```

## Core Boundary

Core should provide primitives that can support many workflows:

- Seed generation and validation
- Ledger and evidence/provenance tracking
- Durable workflow state
- Runtime/provider abstraction
- MCP tool surfaces
- Safety boundaries and permission checks
- Progress, recovery, and status reporting
- Execution handoff and result attachment

Core should not directly encode every domain workflow. It should not need built-in knowledge of GitHub PR merge policy, Jira triage semantics, Slack incident response flows, release checklist conventions, or customer support debugging playbooks.

Those belong at the UserLevel.

## First-party UserLevel Programs

[...]

Issue #729, opened 2026-05-07
Phase 1: Implement plugin invocation firewall + audit-event emitter


shaun0927:
> ## ✅ Decisions LOCKED — implementation-ready
>
> Upstream contract decisions resolved 2026-05-07 by @shaun0927:
> - **Trust UX (Q00/ouroboros-plugins#9)**: 6 specific answers locked. Most relevant for the firewall: Q1 (refuse + clear error, no interactive prompt), Q2 (single confirmation per invocation, command-level), Q3 (exact-scope only).
> - **Manifest shape (Q00/ouroboros-plugins#6)**: 8 required + 2 optional fields.
> - **Risk enum (Q00/ouroboros-plugins#10)**: single 3-value `read_only | write | destructive`.

## Sub-issue of #725 (Phase 1)

Implement the **invocation firewall** — a single wrapper that every UserLevel plugin command must traverse.

### Module: `src/ouroboros/plugin/firewall.py`

Public API (unchanged from earlier revision):

```python
def invoke_plugin(
    plugin: LoadedPlugin,
    command_name: str,
    argv: list[str],
    *,
    ledger: Ledger,
    trust_state: TrustState,
    correlation_id: str,
) -> InvocationResult:
    ...
```

### Required behavior, in order

1. **Pre-invocation trust check** (per Q00/ouroboros-plugins#9 Q1 lock):
   - For each `permission` in manifest where `required: true`, the trust store must contain that exact scope (per Q00/ouroboros-plugins#9 Q3 lock — exact-scope only, no parent implication).
   - If any required permission is not trusted → emit **only** `plugin.failed` with `result.status="blocked"`, message format: `"plugin requires \`<scope>\` (<risk>), which is not yet trusted. Run: ooo plugin trust <plugin-name> --scope <scope>"`.
   - Do NOT emit `plugin.invoked`. Return early with `InvocationResult(status="blocked")`.
2. **Confirmation gate** (per Q00/ouroboros-plugins#9 Q2 lock):
   - If the resolved command has `requires_confirmation: true`, show a single confirmation prompt: `"This command is destructive and requires confirmation.\nPlugin: <name> <version>\nAction: <command-name> <argv>\nContinue? [y/N]"`.
   - On `n` or empty → emit `plugin.failed (status=blocked)`, return early.
   - **No second prompt** for permission-level destructive flag. Permission risk has been handled at trust grant time.
3. **Emit `plugin.invoked`** before launching the entrypoint subprocess.
[...]

PR #749, opened 2026-05-07
feat(ledger): plugin audit-event adapter for core EventStore

### LINKED ISSUE #737: RFC: audit-event.schema.json compatibility with core ledger writer
## Summary

Cross-repo coordination issue.

The plugin audit event format is defined in [Q00/ouroboros-plugins/schemas/audit-event.schema.json](https://github.com/Q00/ouroboros-plugins/blob/main/schemas/audit-event.schema.json). The Ouroboros core ledger writer must emit events conforming to this schema (or a superset of it). This issue tracks **whether the existing ledger schema collides with the audit-event schema** and what to reconcile.

## Why this matters

The firewall (O4) emits `plugin.invoked`, `plugin.permission_used`, `plugin.completed`, `plugin.failed`. If the core ledger writer expects a different shape, every plugin invocation produces malformed ledger entries, and the audit-event schema becomes aspirational rather than load-bearing.

## Specific compatibility checks needed

1. **Top-level event envelope**: does the ledger wrap audit events with its own envelope (`event_id`, `seq`, `actor`, …)? If so, document the wrapping rule. The audit-event schema's `additionalProperties: false` means we cannot add fields **inside** the schema, so wrapping must happen at a layer above it.

2. **`event_type` value space**: does the ledger already use names like `plugin.invoked`? If yes (or close), confirm no conflict. If not, add the new event types to the ledger's accepted enum.

3. **`occurred_at` format**: schema requires RFC3339 `date-time`. Ledger today may use a different format (epoch ns? ISO8601 with timezone?). Reconcile.

4. **`plugin.source_type` values**: schema requires `local_path | plugin_home | first_party`. If ledger has its own actor-source taxonomy, document the mapping.

5. **`provenance` map**: schema allows `additionalProperties: true` for string-only values. Ledger's existing provenance may have nested objects or non-string values; reconcile or reject.

6. **`result.status` enum**: schema requires `success | blocked | failed`. Ledger may have other statuses.

## Required output

- A compatibility memo in `docs/rfc/userlevel-plugins.md` (or a referenced sub-doc)
- Either: the existing ledger writer accepts the audit-event schema as-is (preferred), OR a small adapter is written and the cases where adaptation is lossy are documented
- A test fixture: every audit-event from `audit-event.schema.json` round-trips through the ledger writer

## Acceptance criteria

- [ ] Compatibility analysis documented
- [ ] Ledger writer accepts plugin audit events
- [ ] Round-trip test passes for all 7 event types
- [ ] No silent data loss (e.g. truncated `provenance` keys)
- [ ] No silent expansion (e.g. firewall emits a field that the schema's `additionalProperties: false` forbids — must error instead)

## Related

- Parent: #725
- Phase 1: O4 (firewall is the consumer of this compatibility)
- Cross-repo: Q00/ouroboros-plugins schemas

shaun0927:
Closes #737. Sub-issue of #725 (Phase 1). **Stacked on #748 (CR-4 firewall)**.

## Investigation finding

The core `EventStore` (`src/ouroboros/persistence/event_store.py`) stores rows shaped as `(id, aggregate_type, aggregate_id, event_type, payload, timestamp)` — a row envelope around an opaque JSON payload.

The plugin audit-event schema (`schemas/0.1/audit-event.schema.json`) declares `additionalProperties: false`, so wrapping fields the core ledger needs **must** live above the audit-event boundary. The shapes are compatible without modifying the audit-event schema.

### `wrap_plugin_event(audit_event, *, correlation_id, aggregate_id=None, envelope_id=None) -> dict`

Wraps a firewall-emitted audit event in a core-ledger row envelope:

| Envelope field | Source |
|---|---|
| `id` | `uuid4()` (override-able for deterministic tests) |
| `aggregate_type` | `"plugin"` (constant) |
| `aggregate_id` | `correlation_id` (override-able) |
| `event_type` | `audit_event["event_type"]` (mirrored so `events_table.event_type` is queryable without parsing JSON) |
| `payload` | the full audit event (shallow copy — caller cannot mutate stored form) |
| `timestamp` | `audit_event["occurred_at"]` |

The audit event itself is **not** mutated; the payload remains schema-valid for downstream consumers that read it back.

### `unwrap_plugin_event(envelope) -> dict`

Recovers the original audit event from a stored row. Rejects envelopes whose `aggregate_type != "plugin"` so other event categories cannot accidentally pass through plugin-only consumers.

### `make_event_sink(append_fn, *, correlation_id, ...) -> EventSink`

Factory producing a firewall-compatible `EventSink` callable. Production wires `append_fn` to a real `EventStore.append` adapter; tests pass `list.append`.

## Why this module does NOT import `event_store`

The async-SQLAlchemy machinery in `event_store.py` is a heavy dependency. Keeping the adapter pure-Python lets:

- The firewall stay synchronous (no async leakage from a logging concern).
- Tests stay fast (no async event loop or SQLite setup).
- The CLI (#731) be the integration point that takes a real async session and bridges to `append_fn`.

This matches the `Invocation Contract` framing in `docs/rfc/userlevel-plugins.md` (#743).

## Tests — 14/14 pass; full plugin suite 207/207

```
$ PYTHONPATH=src python3 -m pytest tests/unit/plugin/test_ledger_adapter.py -v
============================== 14 passed in 0.28s ==============================

$ PYTHONPATH=src python3 -m pytest tests/unit/plugin/ --tb=no -q
207 passed, 1 skipped in 0.71s
```

| # | Test | Asserts |
|---|---|---|
| 1 | basic wrap envelope shape | aggregate_type, aggregate_id, event_type, timestamp, id |
| 2 | wrap does not mutate input | input dict unchanged after wrap |
| 3 | payload stays schema-valid | `additionalProperties:false` honored |
| 4 | unwrap returns audit event | round-trip equality |
| 5 | **round-trip for all 7 event types** | the locked spec's main AC |
| 6 | unwrap rejects non-plugin envelope | safety guard |
| 7 | wrap requires event_type | fails fast |
| 8 | wrap requires occurred_at | fails fast |
| 9 | aggregate_id override | overrides correlation_id default |
| 10 | make_event_sink integration | rows appended with envelope shape |
| 11 | id factory for deterministic tests | id sequence preserved |
| 12 | envelope.event_type == payload.event_type | mirrored consistently |
| 13 | wrap rejects non-dict input | TypeError |
| 14 | no token-shaped keys in envelope | sanity guard |

## Cross-repo

- The locked Q00/ouroboros-plugins#9 Q6 trust event shape (with `granted_by`/`granted_scope` in provenance) is exercised by the round-trip test.
- `docs/rfc/userlevel-plugins.md` Section 6 (added in #743) documents this compatibility contract; this PR is the implementation.

## What's next

- #731 (CLI) — bridges `append_fn` to a real `EventStore.append` async session.
- #733 (E2E) — exercises the full firewall → adapter → ledger flow against the `github-pr-ops` plugin.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ requires-python = ">=3.12"
 dependencies = [
     "aiosqlite>=0.20.0,<1.0.0",
     "anyio>=4.0.0,<5.0.0",
+    "jsonschema>=4.21.0,<5.0.0",
     "pydantic>=2.0.0,<3.0.0",
     "prompt-toolkit>=3.0.0,<4.0.0",
     "pyyaml>=6.0.0,<7.0.0",
--- a/src/ouroboros/plugin/firewall.py
+++ b/src/ouroboros/plugin/firewall.py
@@ -203,6 +203,19 @@ def invoke_plugin(
             message=f"unknown command {command_name!r} in namespace {namespace!r}",
         )
 
+    # Defense-in-depth: a TrustRecord must positively identify the plugin
+    # and version it was granted for. A record from a previous version (the
+    # locked Q4 says version-bumps invalidate trust) or a record loaded for
+    # a different plugin that happens to grant the same scope strings must
+    # NOT authorize execution. Mismatched records are dropped here so the
+    # downstream check sees no trust and fails the call closed. Per the
+    # documented "single chokepoint" contract, the firewall — not callers —
+    # owns this invariant.
+    if trust_record is not None and (
+        trust_record.plugin != manifest.name or trust_record.version != manifest.version
+    ):
+        trust_record = None
+
     trust_state = _trust_state_label(manifest, trust_record)
     risks = _scope_risk_index(manifest)
     emitted: list[dict] = []
--- a/src/ouroboros/plugin/firewall.py
+++ b/src/ouroboros/plugin/firewall.py
@@ -30,18 +30,17 @@ ledger writer (#737). Tests pass a list-appender for inspection.
 
 from __future__ import annotations
 
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass, field
+from datetime import UTC, datetime
 import hashlib
 import shlex
 import subprocess
-from collections.abc import Callable
-from dataclasses import dataclass, field
-from datetime import datetime, timezone
-from typing import Iterable, Literal
+from typing import Literal
 
 from ouroboros.plugin.manifest import PluginManifest
 from ouroboros.plugin.trust_store import TrustRecord
 from ouroboros.plugin.userlevel_registry import RegisteredProgram
-
 
 SCHEMA_VERSION = "0.1"
 
@@ -60,7 +59,7 @@ class InvocationResult:
 
 
 def _utc_now_iso() -> str:
-    return datetime.now(tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+    return datetime.now(tz=UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
 
 
 def _source_type_for_event(manifest: PluginManifest) -> str:
@@ -292,8 +291,22 @@ def invoke_plugin(
             text=True,
             check=False,
         )
-    except FileNotFoundError as exc:
-        message = f"entrypoint not found: {cmd_argv[0]!r} ({exc})"
+    except OSError as exc:
+        # Cover the full launch-failure surface (FileNotFoundError,
+        # PermissionError, IsADirectoryError, ENOEXEC, etc.). The firewall is
+        # the documented single chokepoint, so every launch failure must emit
+        # a terminal `plugin.failed` event rather than escaping as an
+        # uncaught exception. Exit code uses the conventional shell mapping
+        # (127 = not found, 126 = found-but-not-executable, otherwise 1).
+        if isinstance(exc, FileNotFoundError):
+            message = f"entrypoint not found: {cmd_argv[0]!r} ({exc})"
+            launch_exit_code = 127
+        elif isinstance(exc, PermissionError):
+            message = f"entrypoint not executable: {cmd_argv[0]!r} ({exc})"
+            launch_exit_code = 126
+        else:
+            message = f"entrypoint launch failed: {cmd_argv[0]!r} ({exc})"
+            launch_exit_code = 1
         _emit(
             _event_envelope(
                 event_type="plugin.failed",
@@ -308,7 +321,7 @@ def invoke_plugin(
         )
         return InvocationResult(
             status="failed",
-            exit_code=127,
+            exit_code=launch_exit_code,
             message=message,
             events=tuple(emitted),
         )
--- a/src/ouroboros/plugin/firewall.py
+++ b/src/ouroboros/plugin/firewall.py
@@ -160,6 +160,31 @@ def _scope_risk_index(manifest: PluginManifest) -> dict[str, str]:
     return {p.scope: p.risk for p in manifest.permissions}
 
 
+def _resolve_plugin_cwd(manifest: PluginManifest) -> str | None:
+    """Compute the working directory the entrypoint subprocess should run in.
+
+    For `local_path` and `plugin_home` sources, the manifest is now
+    required to declare `source.path` (per the schema's conditional
+    required block) — entrypoints typically reference their own
+    installation tree (`./run.sh`, `python -m foo`), so the firewall
+    runs the subprocess with cwd set to the resolved plugin directory.
+    `~`-style expansion is handled here so a manifest like
+    `{"path": "~/.ouroboros/plugins/x"}` works even when the calling
+    process didn't pre-expand it. First-party plugins keep cwd=None
+    (current process cwd) because they are launched from the parent
+    ouroboros runtime and have no installation tree of their own.
+    """
+    if manifest.source.type == "first_party":
+        return None
+    raw = manifest.source.path
+    if not raw:  # pragma: no cover — schema now rejects this at load time
+        return None
+    from os.path import expanduser
+    from pathlib import Path
+
+    return str(Path(expanduser(raw)).resolve())
+
+
 def _deny_confirmation(_msg: str) -> bool:
     """Fail-closed default for `invoke_plugin(confirm=...)`.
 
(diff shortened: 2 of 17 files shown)


shaun0927:
@Q00 — re-review ping. This PR is now bot-approved on commit `55ddfc4` and the merge gate is `CLEAN`. Below is a structured summary of what changed and why I believe it is ready to land.

## Original PR scope

`feat(ledger): plugin audit-event adapter for core EventStore` — adds `src/ouroboros/plugin/ledger_adapter.py`, the dict-envelope wrapper between the firewall (#729) and the core `EventStore` (#737). The adapter deliberately stays import-light (no `event_store` async machinery) because the PR description frames the CLI (#731) as the integration boundary. Sub-issue of #725 Phase 1, stacked on #748.

## Improvements applied in 9 review rounds with `ouroboros-agent[bot]`

The bot ran a deep review on every push and called out contract-level gaps that pre-existed in the stacked branches. Each round was addressed with targeted commits + regression tests, never by suppressing the finding. The fixes split cleanly across the four boundaries the PR touches:

**Manifest contract (`plugin/manifest.py` + schema)**
- Reject duplicate `commands[].name` (was silently shadowed by `find_command`).
- Reject duplicate `permissions[].scope` and `capabilities[].name` (uniqueness now natural-key, not whole-object).
- Convert `permissions` / `capabilities` from `frozenset` to ordered `tuple` so the firewall's `plugin.permission_used` event sequence and "first missing scope" message are deterministic.
- Conditionally require `source.path` for `local_path` / `plugin_home` (only `first_party` is exempt) via a JSON Schema `if/then` block.
- Reject blank/whitespace-only `entrypoint.command` (would otherwise let the firewall try to exec the literal subcommand name).
- Track `manifest_path` on `PluginManifest` so downstream can resolve relative `source.path` correctly.

**Registry (`plugin/userlevel_registry.py`)**
- Fix `register(replace=True)` namespace-owner stale-mapping bug: when a replace changes namespace, the old namespace alias is now released so other plugins can claim it (and `get_by_namespace(old)` no longer resolves to the new program).

**Trust store (`plugin/trust_store.py`)**
- Per-plugin POSIX `fcntl.flock` around all mutating ops (`grant`, `reset_for_version_bump`, `remove`) — closes the silent-scope-loss race under concurrent CLI/upgrade traffic.
- Sticky sidecar `trust.json.lock`: never unlinked, so future writers always lock the same inode that any concurrent process still holds open. Caller-side directory cleanup is now `ooo plugin remove`'s job, as documented.

**Firewall (`plugin/firewall.py`)**
- `_trust_state_label` now reports `installed` for partial trust instead of contradicting itself (`trust_state="trusted"` while `result.status="blocked"`).
- Defense-in-depth: invocation drops a `TrustRecord` whose `(plugin, version)` doesn't match the program — stale or foreign records can no longer authorize execution even if the caller mishands them.
- `confirm` parameter now defaults to **fail-closed** (`_deny_confirmation`); a caller that forgets to wire a real prompt cannot silently execute a `requires_confirmation: true` command.
- Subprocess launch uses `cwd=` resolved from `manifest.source.path`, with `~` expansion. Relative paths resolve against the *manifest's* directory (via `manifest_path`), not the caller's cwd, so an installed plugin works from any working directory. First-party plugins keep `cwd=None`.
- `OSError` is the launch-failure surface (was only `FileNotFoundError`); `PermissionError` now emits a terminal `plugin.failed` (exit 126), generic `OSError` exit 1 — preserving the documented "single chokepoint" contract.
- `_emit` isolates `event_sink` exceptions: a ledger outage on a terminal emit no longer turns into an uncaught exception that loses the `InvocationResult` after the subprocess has already run. Failures are logged structurally (`plugin.firewall.event_sink_failed`).

**Ledger adapter (`plugin/ledger_adapter.py`)**
- Add `envelope_to_base_event(envelope) -> BaseEvent` so the integration layer (#731) can translate dict envelopes into the typed `BaseEvent` that `EventStore.append` actually accepts. `BaseEvent` is imported lazily so wrap/unwrap consumers don't pay the pydantic cost.
- `make_event_sink` docstring shows the exact production wiring snippet (asyncio bridge + `envelope_to_base_event`).

**Lint / format / test plumbing**
- Round 1 also resolved the original CI Lint failure: `ruff check --fix` and `ruff format` over the touched files (I001 / UP035 / UP017 / F401), plus pre-existing format drift in `plugin/lockfile.py` and `tests/unit/plugin/test_trust_store.py`.

## Verification

- Bot final verdict: **APPROVE** on commit `55ddfc4`.
- CI on the PR HEAD: `Ruff Lint`, `MyPy Type Check`, `Bridge TypeScript`, `Test Python 3.12`, `Test Python 3.13`, `Test Python 3.14` — all green.
- Local: `uv run ruff check src/ tests/` clean, `uv run ruff format --check` clean, `uv run mypy src/ouroboros/plugin/` clean (18 files), `PYTHONPATH=src uv run pytest tests/unit/plugin/ -q` → **232 passed, 1 skipped** (up from 207 at the original PR open). Every fix is paired with a regression test that asserts the specific failure mode the bot reproduced.

## Why I think this is mergeable

1. The PR's core intent — a sync, async-free dict-envelope wrapper between the firewall and the core ledger — is unchanged and now has a concrete, tested bridge to the real `EventStore.append` contract via `envelope_to_base_event`.
2. The contract-level invariants the bot kept surfacing (audit coherence, trust identity, fail-closed confirmation, deterministic permission ordering, manifest-relative cwd, atomic trust mutations) are all defended *at the chokepoint* now, not in callers.
3. There are no remaining BLOCKING / warning items from the bot, no inline reviewer comments, and `mergeStateStatus = CLEAN`.
4. The 3.13 flake on commit `e0112d3` (`tests/unit/mcp/tools/test_ralph_handler.py::test_ralph_handler_returns_job_id_and_completes_loop`) was unrelated (3.12 / 3.14 passed for the same code) and disappeared on the next push.

If you'd like a different bisection (e.g. the trust-store and firewall hardening peeled into a follow-up PR), happy to split — but everything here is incremental and behind the firewall / trust-store boundary, so I'd lean toward landing it as a unit.

---

shaun0927:
Closing as **superseded by main**.

`src/ouroboros/plugin/ledger_adapter.py` is already on `upstream/main` (added via the #750 series) and is a superset of this branch:

- All four functions this PR introduces (`wrap_plugin_event`, `unwrap_plugin_event`, `make_event_sink`, `envelope_to_base_event`) are present on main with the same contract.
- main has additional helpers this branch lacks: `_parse_audit_timestamp`, `_default_plugin_event_class`.
- Test coverage on main has been extended beyond what this branch ships.

Rebase against current main would produce add/add conflicts that all resolve to *take the main version*. Merging would regress.

If a specific contract from this PR is not present on main, I'll open a focused follow-up rather than rebase this snapshot.

(Stacked PR #748 — invocation firewall — closed for the same reason.)

PR #751, opened 2026-05-07
feat(cli): ooo plugin {add,install,trust,disable,remove} (mutating)

shaun0927:
Closes #731 (combined with CR-6a #750). Sub-issue of #725 (Phase 2). **Stacked on #750 (CR-6a read-only CLI)**.

### `ooo plugin add <target> [--plugin <name> ...]`

- **Anti-pattern rejection**: any install string containing `#plugins/` is rejected with the locked error message: *"subdirectory-form install strings (#plugins/...) are not supported. Use `ooo plugin add <repo-url> --plugin <name>` instead."*
- URL form: shallow `git clone --depth 1` into `~/.ouroboros/cache/<sanitized>`; resolved git SHA written to the lockfile.
- Local path form: read directly.
- Catalog discovery: every `<repo>/plugins/<name>/ouroboros.plugin.json` parsed.
- Selection: `--plugin <name>` (repeatable) for non-interactive; `questionary` multi-select otherwise. Missing-questionary fallback surfaces a clean error pointing the user to `--plugin`.
- Per-selected-plugin: copy into `~/.ouroboros/plugins/<name>/` + lockfile entry. **No trust granted at install** (locked principle).

### `ooo plugin install <plugin-dir>`

Single-plugin form, no catalog convention required. Validates the manifest; surfaces `PluginManifestError`'s JSON Pointer on schema violation.

### `ooo plugin trust <name> --scope ... --granted-by <id>`

- Per Q00/ouroboros-plugins#9 **Q3 lock**: scopes are exact strings.
- Per **Q5 lock**: stored at `~/.ouroboros/plugins/<name>/trust.json`.
- Per **Q6 lock**: emits `plugin.trusted` event with `provenance.granted_by` + `provenance.granted_scope`. Wrapped through `ledger_adapter.wrap_plugin_event` and appended to `--audit-log` (JSON Lines) if provided.

### `ooo plugin disable <name>`

Wipes the trust file (lockfile entry preserved → user can re-enable via `trust ...`).

### `ooo plugin remove <name>`

Atomic teardown across lockfile entry, trust file, and plugin home directory.

## Test override paths

Every command accepts `--lockfile`, `--trust-root`, `--plugin-home-root`, `--cache-root`, `--audit-log`. Tests use these; production defaults to `~/.ouroboros/...`.

## Dependency

`pyproject.toml`: `questionary>=2.0,<3.0` added as a runtime dependency.

## Tests — 10/10 mutating + 8/8 read-only = 18/18 pass

```
[...]

Issue #735, opened 2026-05-07
Phase 4: CI lint guard preventing domain keywords from leaking into ooo auto


shaun0927:
## Summary

Sub-issue of #725 (Phase 4).

After `ooo auto` is cleaned (O9), add a CI lint rule that prevents future PRs from re-introducing domain-specific branching into `ooo auto`'s code path.

Without this guard, the boundary will silently re-erode the next time someone needs a "small" PR-specific tweak.

## Why this matters

The lesson from #689: a single "small" addition that crosses the boundary is hard to reject during review when there is no machine-checked rule. A 3-line grep guard makes the boundary self-enforcing.

## Required guard

A simple CI step (e.g. in `pyproject.toml` ruff custom rule, or a tiny `scripts/check-auto-boundary.py` invoked by CI):

```
Forbidden in src/ouroboros/cli/commands/auto.py and
src/ouroboros/auto/pipeline.py:

  - "github" (case-insensitive) outside docstrings/comments
  - "jira"
  - "linear" (when not part of the word "linearly")
  - "slack"
  - "pull_request"
  - URL patterns matching domain APIs

Domain-specific keywords belong in plugins, not core auto.
```

False-positive handling: a small allowlist comment marker (`# domain-keyword-allowed: <reason>`) for the rare necessary mention. Each allowlist usage is reviewed.

### Option A: ruff custom rule

Cleanest. Runs as part of existing lint.

### Option B: standalone script in `scripts/check-auto-boundary.py`

Simpler to write, runs as a separate CI step.

### Option C: pre-commit hook + CI

Catches drift before push.

**Recommended**: B (standalone) for v1; migrate to A if ruff custom rules are already used elsewhere.

## Tests

- The guard itself is tested: a synthetic `auto.py` containing `"github"` fails the check; a clean one passes.

## Acceptance criteria

- [ ] Guard implemented in CI
- [ ] Documented in `CONTRIBUTING.md` (or equivalent) so PR authors know to expect it
- [ ] Test for the guard itself (must catch the offending pattern, must not false-positive on docstrings)
- [ ] Allowlist mechanism present and used at most 0–2 times in the codebase

## Related

- Parent: #725
- Depends on: O9 (excision must land first; guarding a still-dirty file is pointless)

shaun0927:
Closing — implementation complete.

**Delivered:**
- `.github/workflows/auto-boundary.yml` runs on PR + push to main
[...]

Issue #734, opened 2026-05-07
Phase 4: Excise GitHub-PR domain branching from ooo auto


shaun0927:
## Summary

Sub-issue of #725 (Phase 4).

**Verified state at issue creation** (`grep -nE 'github|pull_request|/pulls?/' src/ouroboros/cli/commands/auto.py src/ouroboros/auto/pipeline.py` returns **empty**):

- No GitHub-PR branching exists in `ooo auto` or its pipeline today.
- The 6 PRs implementing #689 — `#697`, `#707`, `#712`, `#715`, `#721`, and the stacked branch — are all **CLOSED, never merged**.
- The boundary that #725 protects is **intact**.

This sub-issue's purpose is therefore **preventative and documentary**, not corrective: ensure the boundary stays intact, document the verified state in the RFC, and confirm that any future #689-class work lands in a plugin (`github-pr-ops` per Q00/ouroboros-plugins) rather than re-entering `ooo auto`.

The mechanical enforcement layer is #735 (CI lint guard). This issue captures the policy and the as-of-today verification.

## Why this matters

> "@Q00 in #725: #689 crossed both boundaries — it was not an OS primitive, and it was not part of the `ooo auto` product boundary either. It should be a separate UserLevel program if we want it at all."

The closed status of the #689 PR stack is the natural alignment point for that statement. This issue makes the alignment **explicit and verifiable**, rather than leaving "boundary holds" as a hope.

### 1. Re-verify the boundary at the time of merge

Re-run the grep at PR-merge time and capture the empty output as evidence:

```bash
grep -nE '\bgithub\b|\bpull_request\b|/pulls?/|api\.github\.com' \
  src/ouroboros/cli/commands/auto.py \
  src/ouroboros/auto/pipeline.py \
  src/ouroboros/auto/interview_driver.py \
  src/ouroboros/auto/state.py \
  src/ouroboros/auto/adapters.py \
  src/ouroboros/auto/grading.py \
  src/ouroboros/auto/seed_repairer.py \
  src/ouroboros/auto/seed_reviewer.py
# Expected: no matches
```

If the grep produces matches, this issue's premise has changed and the issue must be re-scoped (back to the original "excise" framing).

### 2. Document the policy in the RFC (#726)

Add to `docs/rfc/userlevel-plugins.md` Section 9 (`ooo auto Boundary`):

```markdown
The `ooo auto` product boundary is permanently:

  goal → clarification/interview → Seed → validation → execution handoff

[...]

Issue #11, opened 2026-01-14
[Story 4.3] Persona Rotation Strategy


Q00:
**Parent Epic:** #5

# Story 4.3: Persona Rotation Strategy

Status: ready-for-dev

## Story

As a developer,
I want intelligent persona selection,
so that the most appropriate lateral thinking approach is tried first.

## Acceptance Criteria

1. Selection considers stagnation pattern type
2. Previously failed personas deprioritized
3. Rotation continues until progress
4. All personas exhausted triggers human intervention
5. Selection history tracked

## Tasks / Subtasks

- [ ] Task 1: Implement selection logic (AC: 1)
  - [ ] Match stagnation pattern to best persona
  - [ ] Spinning → Contrarian
  - [ ] Oscillation → Architect
  - [ ] No Drift → Researcher
  - [ ] Diminishing → Simplifier
- [ ] Task 2: Track history (AC: 2, 5)
  - [ ] Track failed personas per stagnation
  - [ ] Deprioritize failed ones
- [ ] Task 3: Handle exhaustion (AC: 3, 4)
  - [ ] Rotate through remaining
  - [ ] Signal human intervention if all fail

## Dev Notes

- Smart selection reduces attempts
- Never infinite retry
- Human intervention as last resort

### References

- [Source: requirement/2_FULL_SPECIFICATION.md#Persona-Selection]

PR #752, opened 2026-05-07
test(plugin): end-to-end contract proof with github-pr-ops fixture

### LINKED ISSUE #733: Phase 3: End-to-end contract proof with github-pr-ops (read-only path only)

shaun0927:
Closes #733. Sub-issue of #725 (Phase 3). **Stacked on #751 (CR-6b mutating CLI)**.

## Fixture: `tests/fixtures/plugins-fixture/`

Local mirror of the post-Sprint-1 `Q00/ouroboros-plugins` layout. Manifest reflects all locked decisions:

- 8 required + 2 optional fields (#6)
- `merge` absent (#7 — verified by `test_fixture_does_not_declare_merge`)
- Single 3-value `risk` enum (#10)
- `schema_version: "0.1"` (#11)

The fixture entrypoint (`github_pr_ops/__main__.py`) is deterministic and **never contacts GitHub**: returns 0 with a JSON receipt by default; returns exit 2 + stderr `"synthetic failure"` if the URL contains `"fail"` (drives Path 3).

## E2E test: `tests/integration/plugin/test_e2e.py`

Full chain exercised: CLI `install` → CLI `trust` → `firewall.invoke_plugin` (real subprocess to fixture entrypoint) → `ledger_adapter.wrap_plugin_event`.

### Path 1 — read-only success

`invoke_plugin` returns `success`. Envelopes unwrap to:
`[plugin.invoked, plugin.permission_used, plugin.completed]`. sha256 hash in `completed.provenance`. Raw plugin stdout (`"status": "ok"`) **does not** appear in any envelope.

### Path 2 — trust violation

No trust granted. `invoke_plugin` returns `blocked` with `exit_code=None`. Envelopes unwrap to **only `[plugin.failed]`** — explicit absence of `plugin.invoked` asserted (locked Q00/ouroboros-plugins#9 Q1). Locked Q1 message format checked **verbatim**:

> plugin requires \`github:read\` (read_only), which is not yet trusted. Run: ooo plugin trust github-pr-ops --scope github:read

### Path 3 — subprocess failure after trust

URL contains `"fail"`. `invoke_plugin` returns `failed (exit_code=2)`. Envelopes unwrap to: `[plugin.invoked, plugin.permission_used, plugin.failed]` (in contrast with Path 2 — the subprocess actually launched). `failed.result.message` includes `"code 2"`.

## Cross-cutting checks

| Test | What it guards |
|---|---|
| `test_fixture_does_not_declare_merge` | upstream drift if `merge` re-appears (Q00/ouroboros-plugins#7) |
| `test_fixture_manifest_validates` | fixture validates against the vendored 0.1 schema |
| `test_no_token_shaped_strings_in_any_envelope` | `ghp_`, `"bearer "`, `"x-api-key"`, `token=`, `'"token":'` all absent |
| `test_envelopes_carry_all_emitted_event_types` | `AUDIT_EVENT_TYPES` tuple == the schema's 7 enum values |

## Tests — 7/7 E2E pass; full plugin suite 232/232 (1 skipped pre-existing)

```
$ PYTHONPATH=src python3 -m pytest tests/integration/plugin/test_e2e.py -v
============================== 7 passed in 0.63s ===============================
```

## What's next

CR-9 (#734 + #735) — RFC boundary verification text + `ooo auto` CI lint guard. Phase 4 closeout.


shaun0927:
Pushed `802dcb25` (on top of `7dc5753`) addressing both BLOCKING findings from the bot's last review:

**1. `git+http(s)://` clone failure** — `_normalize_git_clone_url()` strips the pip-style `git+` prefix before invoking `git clone`. The user-facing URL surface advertised by `_looks_like_url` is unchanged; git just gets a URL it accepts. Plain http(s) and ssh forms pass through untouched. Parametrized regression test added (`test_normalize_git_clone_url`).

**2. CLI trust UX vs. firewall version-bump invalidation** — `inspect` and `list` now mirror the firewall rule from `_trust_record_matches_manifest`: a trust record whose `version` does not match the installed manifest's version contributes zero scopes, so `trust_state` becomes `installed` (not `trusted`) until the user re-trusts the new version. `inspect` also surfaces a "trust file is stale: recorded for version X but installed manifest is Y; re-trust required" diagnostic so the user knows what to do. New tests `test_inspect_reports_installed_after_version_bump` and `test_list_reports_installed_after_version_bump` exercise the full post-upgrade view.

Also pushed a follow-up `chore(plugin)` commit applying `ruff format` to the seven files CI flagged on `7dc5753`, so Ruff Lint goes green.

Local: all 284 plugin-related tests pass (1 skipped pre-existing); `ruff format --check`, `ruff check`, and `mypy` clean.

shaun0927:
Pushed `323236b0` addressing both BLOCKING findings on `802dcb25`:

**1. Trust grants silently revoked on every reinstall** — `DEFAULT_TRUST_ROOT` and the `add`/`install` default `--plugin-home-root` both pointed at `~/.ouroboros/plugins/`, so the atomic-replace install of `<plugin_home_root>/<name>/` was moving the colocated `<trust_root>/<name>/trust.json` aside on every reinstall (including same-version reinstalls). Moved `DEFAULT_TRUST_ROOT` to `~/.ouroboros/trust` and documented the invariant. Two new regressions in `test_plugin_command_mutating.py`:
- `test_default_trust_root_is_disjoint_from_default_plugin_home_root` — fails loudly if a future refactor points one root inside the other.
- `test_reinstall_preserves_trust_grants` — walks install → trust → reinstall and asserts `trust.json` survives byte-identical and the granted scope is still readable from `TrustStore.read()`.

**2. Plugin invocation has no `cwd` / `PYTHONPATH` wiring** — `invoke_plugin()` ran the entrypoint subprocess with the parent's `cwd` and `env`, so manifests like `python -m github_pr_ops` only resolved when the package was globally installed; the integration test was carrying a PYTHONPATH workaround that flagged the gap. Plumbed `plugin_home` through:
- `RegisteredProgram` gains optional `plugin_home: Path | None`.
- `UserLevelProgramRegistry.register(..., plugin_home=...)` records it.
- `invoke_plugin()` now sets `cwd=program.plugin_home` and prepends `plugin_home` to `PYTHONPATH` (defensively, for non-`-m` entrypoints). When `plugin_home` is None (first-party programs, in-process unit tests), the parent's cwd / env are untouched.
- The e2e test no longer needs its PYTHONPATH workaround — `_python_exec_runner` is now the trivial `bare-python → sys.executable` swap.
- Two new firewall unit tests pin both the inject-cwd-and-pythonpath path and the no-clobber path.

Local: 288 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean. The `tests/unit/orchestrator/test_codex_cli_runtime.py::TestCodexCliRuntime::test_build_command_omits_profile_flag_when_runtime_profile_unset` failure on full sweep reproduces on `7dc5753c` (= pre-existing on the PR branch, unrelated to this work).

shaun0927:
Pushed `5da8dd8e` addressing both BLOCKING findings on `323236b0`, plus the trailing help-text drift:

**1. `remove` is now transactional.** Reordered to lockfile-first so a downstream cleanup failure can no longer strand the user with `plugin_home` deleted but the lockfile still claiming the plugin is installed. The trust-file / plugin-home cleanup is now best-effort with surfaced warnings, since the lockfile is the source-of-truth pointer for "is this installed?". New `test_remove_is_transactional_when_lockfile_write_fails` monkeypatches `Lockfile.remove` to raise and pins both invariants (no destructive cleanup, lockfile entry intact).

**2. `add` / `install` normalize `--plugin-home-root` to an absolute path** before any lockfile persistence. The lockfile stored the string verbatim, so a relative root made later commands and runtime invocations re-interpret the path against whatever cwd the user happened to have at invocation time — the install would silently break after `cd`. New `test_install_normalizes_relative_plugin_home_root` runs `install` from inside `tmp_path` with `--plugin-home-root homes` (relative) and asserts the stored `plugin_home` is absolute and resolves to the expected location.

**3. Help-text drift fix** (non-blocking from the bot's review): `inspect`, `list`, and `disable` still advertised the pre-fix `--trust-root` default of `~/.ouroboros/plugins`; updated to the actual `~/.ouroboros/trust`.

Local: 290 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `74a16cd0` addressing the BLOCKING and the follow-up warning on `5da8dd8e`:

**1. `add` / `install` now commit atomically across filesystem + lockfile.** Refactored `_atomic_install_dir` from a function into a context manager that extends atomicity to a follow-up commit step (the lockfile write). The directory swap happens on `__enter__`; if the body raises, the swap is rolled back (new dir removed, backup restored), so a lockfile write failure during an upgrade no longer leaves the user with the new tree on disk but the lockfile pointing at the old version. New `test_install_rolls_back_directory_swap_when_lockfile_write_fails` pre-installs v0.1.0 with a marker file, attempts a v0.2.0 upgrade with `Lockfile.add` monkeypatched to raise, and asserts the marker survives, the on-disk manifest is still v0.1.0, the lockfile is byte-identical, and no staging/backup paths are stranded.

**2. Firewall now normalizes all launch-time OSErrors as `plugin.failed`** (the bot's follow-up warning). `invoke_plugin` only caught `FileNotFoundError`, so a `PermissionError` (entrypoint not executable), a bad `cwd`, or `Exec format error` bubbled out as an unhandled CLI exception with no audit trail. Broadened the handler to `OSError` with two exit codes: 127 for `FileNotFoundError` (POSIX "not found"), 126 for the rest (POSIX "not executable"). New `test_invoke_handles_permission_error_as_failed_event` asserts the full `[invoked, permission_used, failed]` sequence with `exit_code=126` and the error class name in the message.

Local: 292 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `1c552cf4` addressing the BLOCKING finding on `74a16cd0` — `TrustStore.grant/reset/remove` now hold an exclusive POSIX flock on a per-plugin sibling `trust.json.lock`, mirroring the lockfile's discipline. Two concurrent `ooo plugin trust ...` invocations can no longer both read the same old `trust.json`, append different scopes, and have the second `os.replace` silently drop one of the grants. `remove()` also drops the sibling `.lock` so the empty per-plugin directory still gets pruned. New `test_grant_is_concurrent_safe` spawns 20 threads released simultaneously via a `Barrier` and asserts all 20 distinct scopes are present in the final record (without the flock the race silently drops grants).

Local: 293 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `2a04ca2f` addressing the BLOCKING finding on `1c552cf4` — `_enumerate_catalog` no longer eagerly validates every `plugins/*/ouroboros.plugin.json`. It now returns a list of `_CatalogEntry(name, manifest_path)` records and `_select_plugins` loads manifests lazily for only the entries the user actually selected, normalizing `PluginManifestError` into the same friendly "manifest invalid …" panel that `discover`/`install` emit. The interactive (questionary) path treats per-entry validation failures as warnings and hides bad siblings from the picker. Two regressions:

- `test_add_skips_malformed_sibling_plugin` — plants a valid `github-pr-ops` plugin next to a malformed `broken-plugin` and asserts `add ... --plugin github-pr-ops` exits 0 with the lockfile entry written.
- `test_add_emits_friendly_error_for_selected_invalid_plugin` — selecting the bad plugin via `--plugin` exits 1 with the normalized panel and the JSON Pointer (`/name`) for the offending field, never a raw exception.

Local: 295 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `cdab1a94` addressing both BLOCKING findings on `2a04ca2f`:

**1. `trust` now grants against the installed manifest's version.** It also fails fast when the lockfile entry's version diverges from the on-disk manifest (the situation that produced the bug — CLI reported "Granted" while the firewall silently rejected every invocation). New regressions: `test_trust_fails_fast_when_lockfile_and_manifest_versions_diverge` exercises the divergence path (no trust file written, "version mismatch" surfaced, exit 1), and `test_trust_grants_against_installed_manifest_version` pins the happy-path contract that the persisted record stores the installed manifest's version exactly.

**2. `disable` no longer prints success on a no-op.** When `TrustStore.remove` returns False (typically: wrong `--trust-root` so the real trust file is elsewhere), `disable` now exits 1 with a message naming the trust root in use and suggesting `--trust-root <path>`. The existing `test_disable_wipes_trust_grants` was passing precisely because of the silent no-op; updated to pass the matching `--trust-root` and to assert the trust file is actually gone afterwards. New `test_disable_errors_when_no_trust_file_under_root` grants under one trust root, attempts `disable` against a different (empty) one, and asserts exit 1 plus the original trust file remains intact.

Local: 298 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `efeeee22` addressing the BLOCKING finding on `cdab1a94`. `add` no longer reconstructs the install source path from `manifest.name` — `_select_plugins` now returns `(manifest, source_dir)` pairs threaded from the actual `_CatalogEntry.manifest_path.parent`, and `add_command` passes `source_dir` straight to `_atomic_install_dir`. The interactive (questionary) flow keys on the catalog dirname so picks round-trip back to the exact source directory; the title surfaces `[dirname]` for the human. New `test_add_uses_catalog_directory_when_dirname_differs_from_manifest_name` plants a catalog where the directory is `github-pr-ops-fork` but the manifest declares `name=github-pr-ops`, drops a `FORK_MARKER` file, and asserts the install copied that exact directory (marker present) into the plugin home.

Local: 299 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `8eda071e` addressing both BLOCKING findings on `efeeee22`:

**1. `list` now mirrors the firewall's view exactly.** It reads the installed manifest from `entry.plugin_home` and routes the trust check through `_granted_scopes_for_invocation()`. The reported `version` is the installed manifest's, not the lockfile's, so JSON consumers see the value the firewall actually evaluates. If the manifest is unreadable from the recorded path, the row falls back to `trust_state=installed` rather than fabricating `trusted` from the lockfile.

**2. Trust-store lock file moved outside the data directory.** Previously `<root>/<plugin>/trust.json.lock` was unlinked by `remove()` so the per-plugin dir could be pruned, but unlinking under a held flock is unsound — a pending waiter on the old inode could race a new caller that recreated the lock at a fresh inode. The lock now lives at `<root>/.<plugin>.lock` (outside the data dir), the path is stable across the data file's lifecycle, and `remove()` no longer touches it. Pruning the empty data dir still works.

Two regressions:
- `test_list_reflects_installed_manifest_version_not_lockfile` — installs v0.1.0, grants trust, rewrites the on-disk manifest to v0.2.0, and asserts `list --json` reports `version=0.2.0` + `trust_state=installed`, not `0.1.0` + `trusted` from the lockfile.
- `test_grant_remove_grant_does_not_lose_grants_across_lock_recreation` — grants once, removes, then races 20 concurrent grants. Pins all 20 distinct scopes in the final record (the path that previously could silently drop grants).

Local: 301 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean.

shaun0927:
Pushed `cde0799a` (trust_state correctness fix) and merged the latest `upstream/main` (`09ca1682`) into the branch — the PR was DIRTY because PR #750 (read-only commands) merged while this PR was iterating, and #751 (mutating commands) is still open downstream of the merge base. Used `-X ours` to keep the more-recent fixes from this PR's iteration history while pulling in everything else from main.

**`trust_state` is now strictly invokability** (BLOCKING from `3bf15e8f`): `_trust_state_label()` in both `cli/commands/plugin.py` and `plugin/firewall.py` now returns `trusted` only when *every required scope is granted for the installed version*, not just "any scope exists". Without this, granting an optional scope alongside a missing required one made `inspect`/`list` say `trusted` and the firewall's blocked-event audit envelope carry `trust_state="trusted"`, even though every invocation would be blocked. Two new regressions:

- `test_inspect_does_not_report_trusted_when_required_scope_missing` (CLI) — installs a manifest with one required + one optional scope, grants only the optional, asserts `inspect` shows `trust_state: installed` + the missing-scopes diagnostic and `list --json` reports `installed`.
- `test_invoke_trust_state_is_installed_when_required_scope_missing` (firewall) — exercises the same partial-grant case at the firewall and pins the audit envelope's `trust_state="installed"`.

Local: 312 plugin tests pass (1 skipped pre-existing); ruff format/check + mypy clean. CI on the merge commit incoming.

shaun0927:
@Q00 ready for re-review.

## What this PR is

`test(plugin): end-to-end contract proof with github-pr-ops fixture` — closes #733, sub-issue of #725 (Phase 3), stacked on #751. It bundles a deterministic post-Sprint-1 fixture mirror of `Q00/ouroboros-plugins` and an integration test that walks the full chain end-to-end: CLI `install` → CLI `trust` → `firewall.invoke_plugin` (real subprocess) → `ledger_adapter.wrap_plugin_event`. Three explicit invocation paths are pinned:

1. **Read-only success** — envelopes unwrap to `[plugin.invoked, plugin.permission_used, plugin.completed]`; sha256 hash in `completed.provenance`; raw plugin stdout never appears in any envelope.
2. **Trust violation** — no trust granted; `invoke_plugin` returns `blocked` with `exit_code=None`; envelopes unwrap to **only** `[plugin.failed]` (locked Q1 #9 verbatim message format checked).
3. **Subprocess failure after trust** — URL contains `"fail"`; `invoke_plugin` returns `failed (exit_code=2)`; envelopes are `[plugin.invoked, plugin.permission_used, plugin.failed]`.

The fixture itself never contacts GitHub: deterministic JSON receipt by default, deterministic synthetic stderr on the `"fail"` path. Manifest reflects every locked decision (8 required + 2 optional fields, no `merge` field, single 3-value `risk` enum, `schema_version: "0.1"`).

## Bot iteration — 7 rounds, all BLOCKING findings closed

The bot's last review on `4dfedd6` is **APPROVE** with only two non-blocking follow-up warnings (detailed below). Each iteration added a regression test that fails on the previous broken state:

| # | Commit | Fix |
|---|--------|-----|
| 1 | `00bff08c` | Friendly errors on corrupt `plugins.lock` / `trust.json` (`_read_lockfile_or_exit` / `_read_trust_or_exit`); `--plugin <name>` accepts manifest name as well as catalog dirname. |
| 2 | `235ed80a` | `_granted_scopes_for_invocation` short-circuits first-party and routes through the friendly error helper; `_trust_state_label` reports `trusted` for plugins with zero required scopes (matches firewall's `_missing_required==[]` invokability rule), in both firewall and CLI. |
| 3 | `f23887ef` | `plugin.trusted` audit envelope's `trust_state` is computed via `_trust_state_label` after the grant lands, not hard-coded `"trusted"` — granting an optional scope while a required scope is still missing now correctly tags the envelope `installed`. |
| 4 | `52de00fd` | `inspect` skips `trust.json` read for first-party manifests; corrupt/stale trust files for first-party plugins no longer break `inspect`. |
| 5 | `1a684eb6` | `shlex.split(ValueError)` and empty-result entrypoint commands now emit terminal `plugin.failed` events with `exit_code=2` instead of crashing the firewall caller; `UserLevelProgramRegistry.register(replace=True)` releases the previous namespace owner when the new manifest moves the plugin to a different namespace. |
| 6 | `0216ff97` | `--plugin <manifest.name>` rejects ambiguity when two catalog directories declare the same manifest name (instead of silent first-match-wins); `trust_command` opens the audit log fail-fast pre-loop and surfaces partial-state on mid-loop write failures. |
| 7 | `4dfedd69` | `inspect` and `list` skip `trust.json` reads for plugins with zero required scopes (mirrors firewall: those are unconditionally invokable so corrupt trust files are irrelevant). |

## Why merge-ready

- **Bot verdict on the head commit (`4dfedd6`): APPROVE** — "no in-scope blocking findings remained after policy filtering"; the design notes call out that "the trust-state logic is much closer to the firewall's real behavior, and the new install/remove rollback work is thoughtful."
- **CI green** on all 7 required checks (Ruff Lint, MyPy, Tests Python 3.12/3.13/3.14, Bridge TypeScript, enforce-boundary).
- **310 plugin tests pass** locally (1 skipped pre-existing); every BLOCKING finding has a paired regression test.
- **No new public API surface** beyond the documented manager contract — every fix sits inside `cli/commands/plugin.py`, `plugin/firewall.py`, `plugin/userlevel_registry.py`, or the trust store, behind helpers (`_read_lockfile_or_exit`, `_read_trust_or_exit`, `_granted_scopes_for_invocation`, `_trust_state_label`).
- **Single observable contract: invokability.** Across this iteration, `inspect`, `list`, the `plugin.trusted` audit envelope, and the firewall's `_trust_state_label` all converge on the same rule — `trusted` iff the firewall would let the plugin run *now*. CLI diagnostics no longer disagree with runtime.

## Bot's remaining follow-ups (non-blocking, declared out-of-scope by policy filter)

1. `_select_plugins()` slug-vs-different-plugin-name edge case: if `plugins/foo/` exists AND another directory declares `manifest.name = "foo"`, the slug match wins silently. The intra-name ambiguity check from `0216ff97` does not cover the cross-namespace case.
2. `remove_command()` lockfile-first ordering opens a window where a concurrent `install` could re-add an entry that the still-running `remove` then `rmtree`s.
3. `discover` help text on first-party scopes ("must be trusted before invocation") drifts from the runtime first-party bypass.

These are real but narrow; happy to file follow-up issues and address them in a stacked PR if you'd prefer to merge the locked Phase 3 contract proof first, or fold them in here if you'd rather keep the lifecycle work in one PR. Either is fine — flagging here so the merge decision is informed.

shaun0927:
@ouroboros-agent please re-review latest head `ea9faa3a`.

Scope note: this update re-stacks PR #752 on the latest #751 head and narrows the effective #752 diff to the deterministic `github-pr-ops` fixture plus `tests/integration/plugin/test_e2e.py` only. Local validation on the current head:
- `PYTHONPATH=src /opt/data/projects/ouroboros/.venv/bin/python -m pytest tests/integration/plugin/test_e2e.py tests/unit/cli/test_plugin_command_mutating.py tests/unit/plugin/test_firewall.py tests/unit/plugin/test_trust_store.py tests/unit/plugin/test_userlevel_registry.py -q` → 144 passed
- `ruff check` / `ruff format --check` on the new fixture + integration test → passed
- `git diff --check origin/feat/plugin-cli-mutating..HEAD` → passed

shaun0927:
@Q00 ready for re-review.

## What this PR is

`test(plugin): end-to-end contract proof with github-pr-ops fixture` — closes #733, sub-issue of #725 (Phase 3), stacked on #751. It bundles a deterministic local mirror of `Q00/ouroboros-plugins` and an integration test that walks the full UserLevel chain end-to-end: CLI `install` → CLI `trust` → `firewall.invoke_plugin` (real subprocess to a fixture entrypoint) → `ledger_adapter.wrap_plugin_event`. The fixture entrypoint never contacts GitHub, so CI is fully offline.

## Iteration summary (HEAD = `c882e3a5`)

The bot's approval on `c882e3a5` reads:

> **Verdict: APPROVE**. No in-scope blocking findings remained after policy filtering. The deterministic local fixture plus the E2E test is a good shape for this contract: it avoids network dependence, exercises the install/trust/invoke flow, and the digest-drift refusal path usefully covers a real security boundary. I did not find any in-scope blocking issues in the current snapshot.

What changed since the last `e321f5a` REQUEST_CHANGES round, in the order the bot raised them:

### 1. Audit `plugin.trusted` `trust_state` mirrors firewall invokability (`16fdb59e`)
`trust_command()` previously hardcoded `trust_state="trusted"` on every emitted event. That misstated the permission boundary whenever the cumulative grant set was still firewall-blocked (e.g. one required + one optional scope, user grants only the optional one — `inspect`/`list`/firewall correctly reported `"installed"`, audit log lied). Now the field is computed from `record.granted_scopes` against the manifest's required scopes, so the audit stream agrees with the firewall's gate. Coverage: `test_trust_partial_grant_records_installed_in_audit_event` exercises the bot's exact regression case; the inverse pin (`test_trust_full_required_grant_records_trusted_in_audit_event`) keeps the happy-path label honest.

### 2. Malformed `entrypoint.command` produces a terminal `plugin.failed` (`16fdb59e`)
The 0.1 schema enforces only `minLength: 1` on `entrypoint.command`, so a manifest with an unmatched quote (`shlex.split` → `ValueError`) or a whitespace-only command (`shlex.split` → `[]`) was installable but crashed the firewall pre-launch — skipping the required terminal `plugin.failed` event. Both shapes now produce a controlled `plugin.failed` with `exit_code=126` and `exception_type=ValueError` in provenance. Coverage: `test_entrypoint_unmatched_quote_emits_failed_not_crash`, `test_entrypoint_whitespace_only_command_emits_failed`.

### 3. Merge with `upstream/main` (`5586cfb3`)
The PR was DIRTY against main (PR #750/#796/#798/#805 had merged in parallel). Resolved without losing either side's contributions:
- `firewall.py` — kept the PR-side argv-redaction infrastructure AND main's `_argv_summary`; `_event_envelope` runs redaction first, stores the redacted argv in `cmd["argv"]`, computes `argv_summary` over the redacted form (so secret-shaped values do not contribute to a hash an audit consumer might reverse-correlate).
- `list_command` — merged HEAD's first-party / manifest-unreadable branches with main's degraded-row pattern (catch corrupt `trust.json` instead of crashing the listing) and main's new JSON schema fields (`missing_required_scopes`, `trust_version_stale`).
- `discover_command` — reworded the first-party hint to match main's `"advisory; first-party plugins bypass the trust gate"`.
- `inspect_command` — when a trust record is bound to a different version than the installed manifest, surface a labelled `trust_version: version bump invalidated trust …` line with concrete remediation.
- `_load_with_friendly_error` — surfaces `"manifest is unreadable"` as a distinct top-level header for permission-denied / broken-symlink failures so `discover` agrees with `inspect`/`list`'s diagnostic guarantees.
- `tests/unit/cli/test_plugin_command.py` — accepted main's 700+ lines of read-only command tests verbatim.

### 4. E2E test now exercises the production critical path (`80c9f9d1`)
The `5586cfb3` round had been APPROVED with a follow-up warning: the e2e test claimed "end-to-end contract proof" but shimmed PYTHONPATH inside `_python_exec_runner` (masking the cwd-from-`plugin_home` plumbing the production dispatcher relies on) and never passed `plugin_home` / `expected_artifact_digest` to `invoke_plugin(...)`. Both are now plumbed identically to how `plugin_dispatch.py` calls `invoke_plugin` in production; the shim is reduced to a single `python -> sys.executable` substitution for deterministic interpreter resolution. New `test_path_4_digest_drift_refuses_launch` installs the fixture, grants trust, mutates the plugin home post-grant, and asserts the firewall recomputes the digest, sees drift, refuses to launch (`status="trust_subject_changed"`, `current_artifact_digest` in provenance, terminal `plugin.failed` only — no `plugin.invoked`/`plugin.permission_used`). A regression in digest plumbing or short-circuited comparison fails this test red.

### 5. Closed three lifecycle gaps the bot flagged on `80c9f9d1` (`c882e3a5`)
- **Catalog selection silently dedup'd duplicate manifest names.** `_select_plugins` and `_install_named_from_url` now scan for collisions before any selection runs and refuse with both conflicting paths surfaced. Test: `test_add_duplicate_manifest_names_in_catalog_refused`.
- **Runtime dispatcher was hard-wired to global default state paths.** Manager subcommands accept `--lockfile` / `--trust-root`, but `plugin_dispatch.py` always read `DEFAULT_LOCKFILE_PATH` / `DEFAULT_TRUST_ROOT` — installs at non-default paths were write-only. Added env-var overrides (`OUROBOROS_PLUGIN_LOCKFILE`, `OUROBOROS_PLUGIN_TRUST_ROOT`); Click commands built lazily by the typer fallback can't accept their own flag (collides with the plugin's argv namespace), so env vars are the right surface. Test: `test_dispatch_honors_env_overrides_for_lockfile_and_trust_root`.
- **`remove` was lockfile-atomic but not operation-atomic.** After `lock.remove(name)` returned, the file lock was released and a concurrent `install` could re-add the entry plus recreate the plugin home before `shutil.rmtree(plugin_home)` ran — the still-running `remove` would then delete the freshly-installed bytes. Added `Lockfile.transaction()` (a public re-entrant context manager) and wrapped the whole `remove` body — lockfile mutation through trust wipe through rmtree — in the transaction; nested `add`/`remove` skip re-acquiring the flock so they don't deadlock against the outer fd. Tests: `test_transaction_allows_nested_add_and_remove`, `test_transaction_serializes_with_external_flock`.

## Why this is safe to merge

1. **Bot APPROVED on `c882e3a5`** with explicit "no in-scope blocking issues in the current snapshot." The lone non-blocking suggestion is a cosmetic note about the fixture manifest's `usage` example string (the dispatcher resolves both `ooo github-pr review …` via namespace and `ooo github-pr-ops review …` via name; not a behavior issue).
2. **GitHub merge state**: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, `reviewDecision=APPROVED`.
3. **CI**: all green on `c882e3a5` — Bridge TypeScript, MyPy, Ruff Lint, Test Python 3.12 / 3.13 / 3.14, enforce-boundary, enforce-envelope.
4. **Local validation**: 411 passed, 1 skipped (pre-existing) across the full plugin + CLI plugin test surface.
5. **Coverage of the contract this PR claims**: the four E2E paths (read-only success, trust violation, subprocess failure after trust, post-grant digest drift) plus the cross-cutting token-leak guard now exercise exactly the firewall arguments production dispatch supplies — `plugin_home`, `expected_artifact_digest`, no PYTHONPATH shim. A regression in any of those plumbing edges goes red.
6. **No silent behavior shifts**: every fix lands with a regression-catch test that pins the new behavior; the merge with main preserved both sides' contributions and the upstream tests for new fields (`missing_required_scopes`, `trust_version_stale`) all pass against the merged implementation.

PR #754, opened 2026-05-07
fix(providers,interview): isolate subprocess from host plugin env

ASak1104:
# fix(providers,interview): pass strict_mcp_config when allowed_tools is pinned

> Updated after [ouroboros-agent[bot] review](#) on c4d3bb7f.
> Replaced the broad approach (CLAUDE_PLUGIN_DATA wipe + wiring change)
> with a single SDK option flip. See d2b2ff62.

## Summary

Calling `mcp__plugin_ouroboros_ouroboros__ouroboros_interview` (and likely
other `ClaudeCodeAdapter`-backed tools) from inside Claude Code consistently
fails with `Command failed with exit code 1` and an empty stderr. The
spawned `claude` subprocess auto-loads plugin-provided MCP servers
(including ouroboros itself), registers `mcp__plugin_ouroboros_*` tools,
and recurses on `ouroboros_interview` until it exits on the
`--max-turns 1` boundary.

This PR fixes the issue by forwarding `strict_mcp_config=True` to the SDK
whenever the caller has pinned an explicit read-only tool envelope. No
changes to plugin env, hook configuration, or composition-root wiring.

## Reproduction

The failure is **specific to Claude Code MCP server context**: the
adapter only fails when its subprocess is spawned from inside an MCP
server that Claude Code itself launched (the path used by
`mcp__plugin_*` tools). Direct CLI invocations of `claude` with the
same flags succeed.

Independently reproduced on two environments running `ouroboros-ai`
v0.35.0:

| Host                                 | Claude Code | Python      | API surface           |
|--------------------------------------|-------------|-------------|-----------------------|
| macOS 15.7.3 (Darwin 24.6, x86_64)   | 2.1.132     | 3.14 (uvx)  | api.anthropic.com     |
| macOS (Darwin 25.2, arm64, internal) | 2.1.119     | 3.12 (pipx) | enterprise gateway    |

Reproducible regardless of `initial_context` length. Symptom:

    Claude Agent SDK request failed: Command failed with exit code 1
    Error output: Check stderr output for details

SDK metadata at the failing call:
`claudecode_present=False, claude_code_entrypoint=cli,
env_override_keys=['CLAUDE_AGENT_SDK_SKIP_VERSION_CHECK']`. The
subprocess emits zero stderr lines.

## Root cause

When `ClaudeCodeAdapter` spawns its SDK subprocess from inside an MCP
server context, the child reads the host's MCP configuration sources
(project `.mcp.json`, user/global settings, **plugin-provided servers**)
and registers every tool they expose. Plugin-provided servers include
ouroboros's own MCP server, which exposes `ouroboros_interview` to the
child as `mcp__plugin_ouroboros_ouroboros__ouroboros_interview`. The
child then invokes that tool, hits `--max-turns 1` immediately, and
exits with code 1. The CLI treats turn-limit termination as non-error,
so the SDK surfaces the exit code with no stderr.

The CLI's `--allowedTools` whitelist is not enough to block the
recursion: plugin-loaded MCP tools are not gated by it.

## Fix

`providers/claude_code_adapter.py` — when the caller has pinned an
explicit tool envelope (`allowed_tools is not None`, e.g. the
interview path's `_interview_allowed_tools()` whitelist), also forward
`strict_mcp_config=True` on the SDK call. This maps to the CLI's
`--strict-mcp-config` flag and, per `ClaudeAgentOptions.strict_mcp_config`
docstring, restricts the subprocess to MCP servers we pass explicitly
while ignoring plugin-provided servers. Plugin hooks and global plugin
env stay intact.

The change is gated on `self._allowed_tools is not None` so existing
permissive callers keep their current behaviour.

## What changed since c4d3bb7f

The original commit broadly cleared `CLAUDECODE` /
`CLAUDE_PLUGIN_DATA` / `CLAUDE_PLUGIN_ROOT` in `env_overrides`, removed
the `interview_engine` injection from `mcp/server/adapter.py`, and
forced a fresh adapter rebuild in `InterviewHandler.handle()`.
ouroboros-agent[bot] correctly flagged the first two as overreach:
clearing the plugin env disables every checked-in hook that references
`${CLAUDE_PLUGIN_ROOT}/...`, and dropping the injection silently
switches `ouroboros_interview` to the default state directory,
splitting persisted sessions in deployments that override `state_dir`.

`d2b2ff62` reverts all three changes and replaces them with a single
SDK option flip in `providers/claude_code_adapter.py`. The previously
modified `mcp/server/adapter.py` and `mcp/tools/authoring_handlers.py`
are now unchanged from `main`. Verified by ablation that the SDK flip
alone closes the recursion path.

## Tests / quality gates

- `uv run ruff check src/` — clean.
- `uv run ruff format --check src/` — clean.
- `uv run mypy src/ouroboros/providers/claude_code_adapter.py` — clean.
- `uv run pytest tests/unit/providers/test_claude_code_adapter.py
  tests/unit/mcp/tools/{test_handler_subagent_wiring,
  test_interview_persists_session_on_failure,
  test_interview_done_streak,test_interview_reopen_seed_ready}.py
  tests/unit/mcp/server/` — 169 passed. No tests modified; existing
  injection contract preserved.

End-to-end smoke: `mcp__plugin_ouroboros_ouroboros__ouroboros_interview`
returns the first round question after `/mcp` reconnect on macOS
15.7.3 / Claude Code 2.1.132. Without the SDK flip, the same call
returns `exit code 1` deterministically.

## Notes

- No new dependencies, no public API changes, no plugin/hook regressions.
- Diff is one hunk (5 lines) in a single file.
- `_interview_allowed_tools()` already expressed the intended read-only
  policy; this PR only ensures the corresponding MCP-server isolation
  is applied alongside it.

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -1710,7 +1710,6 @@ def create_ouroboros_server(
             event_store=event_store,
         ),
         InterviewHandler(
-            interview_engine=interview_engine,
             event_store=event_store,
             llm_adapter=llm_adapter,
             llm_backend=llm_backend,
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -1710,6 +1710,7 @@ def create_ouroboros_server(
             event_store=event_store,
         ),
         InterviewHandler(
+            interview_engine=interview_engine,
             event_store=event_store,
             llm_adapter=llm_adapter,
             llm_backend=llm_backend,
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -1249,11 +1249,21 @@ class InterviewHandler:
         # Use injected or create interview engine
         # max_turns=1: MCP is a pure question generator. No tool use needed.
         # Main session handles codebase exploration and answering.
+        #
+        # ``strict_mcp_config=True`` is opt-in here — and ONLY here — so the
+        # subprocess spawned for question generation cannot rediscover the
+        # plugin-provided ouroboros MCP server when this handler runs as a
+        # child of Claude Code's MCP host (where ``mcp__plugin_ouroboros_*``
+        # tools are auto-registered).  Without this, the subprocess
+        # recurses on ``ouroboros_interview`` and exits at ``--max-turns 1``.
+        # CLI interview entrypoints (``ooo init`` / ``ooo pm``) do NOT pass
+        # this flag, so they keep plugin/project ``.mcp.json`` servers.
         llm_adapter = self.llm_adapter or create_llm_adapter(
             backend=self.llm_backend,
             max_turns=1,
             use_case="interview",
             allowed_tools=_interview_allowed_tools(self.llm_backend),
+            strict_mcp_config=True,
         )
         engine = self.interview_engine or InterviewEngine(
             llm_adapter=llm_adapter,
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -1249,7 +1249,7 @@ class InterviewHandler:
         # Use injected or create interview engine
         # max_turns=1: MCP is a pure question generator. No tool use needed.
         # Main session handles codebase exploration and answering.
-        llm_adapter = self.llm_adapter or create_llm_adapter(
+        llm_adapter = create_llm_adapter(
             backend=self.llm_backend,
             max_turns=1,
             use_case="interview",
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -1249,7 +1249,7 @@ class InterviewHandler:
         # Use injected or create interview engine
         # max_turns=1: MCP is a pure question generator. No tool use needed.
         # Main session handles codebase exploration and answering.
-        llm_adapter = create_llm_adapter(
+        llm_adapter = self.llm_adapter or create_llm_adapter(
             backend=self.llm_backend,
             max_turns=1,
             use_case="interview",
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -601,9 +601,10 @@ class ClaudeCodeAdapter:
             "CLAUDE_AGENT_SDK_SKIP_VERSION_CHECK": os.environ.get(
                 "OUROBOROS_SKIP_VERSION_CHECK", "1"
             ),
+            "CLAUDECODE": "",
+            "CLAUDE_PLUGIN_DATA": "",
+            "CLAUDE_PLUGIN_ROOT": "",
         }
-        if claudecode_present:
-            env_overrides["CLAUDECODE"] = ""
 
         stderr_lines: list[str] = []
 
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -30,6 +30,8 @@ import asyncio
 from collections.abc import Callable
 from copy import deepcopy
 from dataclasses import replace
+from functools import lru_cache
+import inspect
 import json
 import os
 from pathlib import Path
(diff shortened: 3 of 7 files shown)


shaun0927:
@Q00 — re-review request 🙏

I did a fresh independent review of this PR (head `889090d5`) in an isolated worktree, including a re-read of every iteration the bot pushed back on. Posting the rationale for why I think this is in merge-ready shape.

### What the PR fixes

A deterministic, reproduced-on-two-hosts failure: invoking `mcp__plugin_ouroboros_ouroboros__ouroboros_interview` from inside Claude Code's MCP host context fails with `Command failed with exit code 1` and an empty stderr. Root cause is nested-MCP recursion — the spawned `claude` subprocess auto-loads plugin-provided MCP servers (including ouroboros itself), registers `mcp__plugin_ouroboros_*`, re-invokes `ouroboros_interview`, and trips the `--max-turns 1` guard. The CLI's `--allowedTools` does not gate plugin-loaded MCP tools, so this is the only thing that closes the recursion path.

### Final shape (after 8 review iterations)

The whole PR is now **6 files / +485 / −0**, with the production change scoped to three call sites:

| File | Change |
|------|--------|
| `providers/claude_code_adapter.py` | New `strict_mcp_config: bool = False` constructor flag. Forwarding uses `lru_cache`d SDK introspection: typed field if present → `extra_args` CLI passthrough on current SDKs → **fail-fast `ProviderError`** with an actionable upgrade message if neither surface exists. |
| `providers/factory.py` | Adds the kwarg as a pure passthrough. **Does NOT** auto-derive from `use_case="interview"`, so `ooo init` / `ooo pm` keep plugin and project `.mcp.json` servers reachable for brownfield repos. |
| `mcp/tools/authoring_handlers.py` | The single opt-in site: `InterviewHandler.handle()` passes `strict_mcp_config=True`. |

### Why each prior bot finding is now resolved

The bot ran through `c4d3bb7 → d2b2ff6 → a0168cb → d228ad7 → 5a203d0 → f0a13c3 → bda1a98 → 889090d`. Each `REQUEST_CHANGES` was addressed with progressively narrower scope:

1. **Composition-root regressions** (c4d3bb7): clearing `CLAUDE_PLUGIN_DATA`/`CLAUDE_PLUGIN_ROOT` and dropping `interview_engine` injection — fully reverted; `mcp/server/adapter.py` and `state_dir` injection contract are unchanged from `main`.
2. **Envelope over-reach** (d2b2ff6 / a0168cb): isolation no longer keys on `allowed_tools is not None` or on `use_case == "interview"`. It is opt-in per call site. The unit test at `tests/unit/providers/test_claude_code_adapter.py:687` now explicitly asserts that an envelope containing `mcp__ouroboros__qa` keeps MCP access.
3. **Silent SDK fallback** (d228ad7): removed. The adapter now raises `ProviderError` with `error_type=ConfigurationError` and `mock_options_cls.assert_not_called()` is locked in tests so options can never be constructed when isolation cannot be honored.
4. **Dependency-floor concern** (f0a13c3): the comment block in `claude_code_adapter.py:660-688` documents the empirical PyPI-history invariant (`extra_args` has been a field on `ClaudeAgentOptions` since the earliest published release `0.0.23`). The fail-fast branch is defense-in-depth against vendored / monkey-patched SDKs, not a normal runtime path.
5. **Test brittleness** (bda1a98): `test_live_claude_agent_sdk_supports_extra_args` is now wrapped in `pytest.importorskip("claude_agent_sdk")` and clears the `lru_cache` on both sides, so it neither requires the optional `[claude]` extra in CI nor leaks state into other tests.

### Verification on `889090d5` (independent run, fresh worktree)

```
uv run ruff check src/ouroboros/providers/claude_code_adapter.py \
                  src/ouroboros/providers/factory.py \
                  src/ouroboros/mcp/tools/authoring_handlers.py
→ All checks passed!

uv run mypy src/ouroboros/providers/claude_code_adapter.py
→ Success: no issues found in 1 source file

uv run pytest tests/unit/providers/test_claude_code_adapter.py \
              tests/unit/providers/test_factory.py \
              tests/unit/mcp/tools/test_interview_strict_mcp_config_wiring.py
→ 85 passed in 2.17s
```

GitHub CI is also green across Ruff Lint, MyPy Type Check, Bridge TypeScript, and Tests on Python 3.12 / 3.13 / 3.14.

### Why I think this is safe to merge

- **Surgical blast radius.** One opt-in call site (the nested MCP entrypoint). Generic explicit envelopes, default callers, and CLI interview flows (`ooo init` / `ooo pm`) are all asserted to keep their pre-PR behavior by dedicated tests.
- **No public API churn.** No new dependencies, no plugin/hook regressions, no env mutation, `mcp/server/adapter.py` untouched, no behavior change for any caller that does not explicitly opt in.
- **Defense-in-depth, not silent degradation.** If a future SDK swap drops both forwarding surfaces, the system fails loudly with an actionable upgrade message rather than re-opening the recursion path it is closing.
- **Lock tests for the failure mode.** `test_strict_mcp_config_fails_fast_when_sdk_lacks_surface` asserts `mock_options_cls.assert_not_called()`, so any future regression that makes the flag silently no-op fails CI before reaching production.
- **Reproduced + fixed on two distinct environments** (macOS 15.7.3 / Claude Code 2.1.132 and macOS Darwin 25.2 / Claude Code 2.1.119), with bot APPROVE on the final commit.

Ready for your final pass when you have a minute.

PR #761, opened 2026-05-07
feat(auto): steer interviews toward open ledger gaps

### LINKED ISSUE #757: Drive auto-interview answers from open ledger gaps instead of repeated generic fallbacks
## Background

In blocked auto-interview sessions, the loop can keep emitting the same generic conservative answer while required ledger gaps remain unchanged. That means the auto-interview is making progress in round count but not in seed readiness.

The auto mode should optimize for producing a seed-ready ledger. If an answer does not reduce open required gaps, the next step should be driven by those gaps rather than by another generic fallback.

## Objective

Drive auto-interview answers from open ledger gaps and prevent repeated generic answers from stalling convergence.

## Scope

- Track open ledger gaps before and after each auto answer.
- Detect when an answer fails to reduce required gaps.
- Detect repeated identical or semantically equivalent generic fallback answers.
- Route the next answer attempt toward the highest-priority missing required section.
- Make gap steering independent of narrow question patterns such as "anything else?".

## Target files and paths

- `src/ouroboros/auto/interview_driver.py`
- `src/ouroboros/auto/answerer.py`
- `src/ouroboros/auto/ledger.py`
- Existing or new tests under `tests/` for auto-interview convergence behavior

## Implementation steps

1. Capture open required gaps before applying an auto answer.
2. Apply the answer and recompute open required gaps.
3. If no required gap was reduced, invoke a gap-directed answer path.
4. Add repeated-fallback detection for identical/default answers across rounds.
5. Ensure the loop records why gap steering was invoked for debugging.
6. Add regression tests for broad benign goals where gaps must decrease over rounds.

## Acceptance criteria

- The same generic fallback answer cannot be emitted repeatedly while required gaps remain unchanged.
- If `actors`, `inputs`, or `outputs` are missing, a subsequent auto answer attempts to fill those sections explicitly.
- If `acceptance_criteria` or `verification_plan` are missing, a subsequent auto answer produces observable criteria or checks instead of another generic assumption.
- Open required gap count decreases across rounds for a broad benign goal fixture.
- The loop still blocks when a required gap cannot be safely inferred.
- Debug/state output identifies when gap-directed steering was used.

## Verification

- Run targeted unit tests for `interview_driver`, `answerer`, and `ledger` behavior.
- Add a regression test that reproduces the previous repeated-generic-answer failure mode.
- Confirm tests cover both successful gap reduction and safe blocking.

## Dependencies / recommended order

Recommended after the convergence contract issue, because the contract should define what "progress" means. It can be implemented before or alongside language-agnostic classification.

## Risks and cautions

- Do not force-fill unsafe or irreversible decisions just to reduce gap count.
- Avoid making gap steering so aggressive that it overwrites user-provided facts.
- Preserve auditability: generated assumptions should remain traceable.

## Non-goals

- No domain-specific defaults for games, frontends, or any other application category.
- No increase to `max_interview_rounds` as the primary fix.
- No change to seed execution backends.

## Ambiguity reducers

- "Progress" means reducing required open ledger gaps or explicitly classifying a gap as unsafe/blocking with a precise reason.
- "Repeated generic fallback" means a fallback answer that does not add new section-specific information and leaves the same required gaps open.
- Gap steering should prefer missing required ledger sections over continuing broad clarification loops.

shaun0927:
## Summary
- Adds gap-progress tracking in the auto-interview loop.
- Falls back to gap-directed answer generation when an auto answer does not reduce required ledger gaps.
- Adds repeated/default-answer detection so broad interviews cannot keep consuming rounds without ledger progress.
- Adds focused regression coverage for stalled generic fallback behavior.

## Discussion / implementation notes
- This PR keeps the fix general: it steers toward missing required ledger sections rather than adding domain-specific defaults.
- Unsafe/unresolved gaps are still allowed to block when they cannot be safely inferred.

## Validation
- [x] `git diff --check`
- [x] `PYTHONPATH=src /opt/hermes/.venv/bin/python -m pytest tests/unit/auto/test_ledger_grading_answerer.py tests/unit/auto/test_interview_pipeline.py -q` — 162 passed

Closes #757

### DIFF
--- a/src/ouroboros/auto/answerer.py
+++ b/src/ouroboros/auto/answerer.py
@@ -152,6 +152,57 @@ class AutoAnswerer:
                 )
         return answer
 
+    def answer_gap(
+        self,
+        section: str,
+        ledger: SeedDraftLedger,
+        context: AutoAnswerContext | None = None,
+    ) -> AutoAnswer:
+        """Return an answer targeted at an unresolved required ledger section.
+
+        This path is used when the backend keeps asking broad prompts and the
+        normal question-shaped answer would repeat or fail to close any open
+        required gap.  Keep it domain-neutral: it only fills Seed contract
+        sections with conservative, reversible defaults.
+        """
+        context = context or AutoAnswerContext()
+        if section == "goal":
+            blocker = AutoBlocker(reason="goal is unresolved", question="Clarify the primary goal.")
+            return AutoAnswer(
+                text="Cannot safely decide automatically: goal is unresolved",
+                source=AutoAnswerSource.BLOCKER,
+                confidence=1.0,
+                blocker=blocker,
+            )
+        if section in {"actors", "inputs", "outputs"}:
+            return self._io_actor_answer("Who are the actors, inputs, and outputs for this task?")
+        if section in {"constraints", "failure_modes"}:
+            return self._default_answer(
+                "What conservative constraints and failure modes should bound this MVP?", ledger
+            )
+        if section == "non_goals":
+            return self._non_goal_answer(
+                "What non-goals should explicitly remain out of scope?", ledger
+            )
+        if section in {"acceptance_criteria", "verification_plan"}:
+            return self._verification_answer(
+                "Which command output verifies the acceptance criteria?"
+            )
+        if section == "runtime_context":
+            return self._runtime_answer(
+                "Which runtime stack, repo, and project patterns should be used?", context
+            )
+        blocker = AutoBlocker(
+            reason=f"unsupported ledger gap section: {section}",
+            question=f"Clarify {section}.",
+        )
+        return AutoAnswer(
+            text=f"Cannot safely decide automatically: unsupported ledger gap section: {section}",
+            source=AutoAnswerSource.BLOCKER,
+            confidence=1.0,
+            blocker=blocker,
+        )
+
     def apply(self, answer: AutoAnswer, ledger: SeedDraftLedger, *, question: str) -> None:
         """Apply answer updates to ``ledger``."""
         ledger.record_qa(question, answer.prefixed_text)
--- a/src/ouroboros/auto/answerer.py
+++ b/src/ouroboros/auto/answerer.py
@@ -75,6 +75,13 @@ class AutoAnswer:
     assumptions: list[str] = field(default_factory=list)
     non_goals: list[str] = field(default_factory=list)
     blocker: AutoBlocker | None = None
+    # True only when the answer came from the catch-all generic-default route
+    # (``_default_answer``). Feature-specific helpers
+    # (e.g. ``_feature_acceptance_answer``, ``_runtime_answer``) leave this
+    # ``False`` even when their ``source`` is ``CONSERVATIVE_DEFAULT``, so the
+    # interview driver can preserve repeated specific follow-ups instead of
+    # treating them as repeated generic fallbacks.
+    generic_default: bool = False
 
     @property
     def prefixed_text(self) -> str:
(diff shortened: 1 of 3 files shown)


shaun0927:
Addressed review: gap steering now only replaces answers when the answer is a repeated generic fallback or the prompt is broad enough to benefit from open-gap steering. Specific non-reducing answers are preserved. Re-ran focused auto tests; CI is green.

shaun0927:
@Q00 ready for re-review.

## What this PR does
Closes #757. Adds gap-progress tracking to the auto-interview loop so that:
- the driver detects when the auto answer would not actually close any required Seed-ledger gap, and
- in that case steers the next reply toward the open ledger section instead of letting broad "anything else?"-style prompts repeat the same generic default forever.

The fix is intentionally domain-neutral: it only nudges toward missing required Seed sections, never adds task-specific defaults. Unsafe / unresolved gaps are still allowed to block.

## Iteration on top of the original commit
Two follow-up commits address every concern raised by `ouroboros-agent[bot]` across its three reviews:

1. **`b30be46d` — preserve specific answers and same-turn ledger repair**
   - Restores broad-prompt detection (`_can_steer_with_gap_prompt`) as the gating signal so backend-specific answers (e.g. follow-up acceptance / verification questions) are not silently replaced with gap-targeted fallbacks when they don't shrink `open_gaps()` in this turn.
   - Runs the same-turn repair check before raising a hard blocker on the first detected gap, so a `CONFLICTING` / `BLOCKED` first gap can be fixed by the current answer instead of permanently stalling resumed/persisted ledgers.
   - `goal` remains an unconditional blocker — it can never be auto-filled and must come from the user.

2. **`b250a161` — distinguish generic-default route from feature-specific answers**
   - Adds `AutoAnswer.generic_default` (default `False`), set `True` only on the catch-all `_default_answer` route. `_is_repeated_default_answer` now keys off this flag instead of `AutoAnswerSource`. Feature-specific helpers (`_feature_acceptance_answer`, `_runtime_answer`, `_io_actor_answer`, `_verification_answer`, …) also emit `CONSERVATIVE_DEFAULT` as their source, but they are no longer treated as repeated generic fallbacks — so the same specific question asked twice keeps getting a specific answer rather than being swapped for an unrelated gap-fill.

## Why it is safe to merge
- **Bot verdict**: `ouroboros-agent[bot]` reviewed `b250a161` and **APPROVED** it (see the latest review). Its design notes explicitly call the change "cohesive: it separates true catch-all defaults from feature-specific conservative answers, then uses that distinction to steer only when the backend is looping on broad/generic prompts."
- **Regression coverage** for every blocker the bot flagged is now in the test suite (`tests/unit/auto/test_interview_pipeline.py`):
  - `test_interview_driver_uses_gap_answers_when_generic_defaults_repeat` — broad prompt + repeated catch-all default → driver pivots to gap-targeted answer.
  - `test_interview_driver_blocks_blank_goal_before_gap_defaults` — `goal` weak → blocker, no answer sent to backend.
  - `test_interview_driver_preserves_specific_acceptance_answers_with_open_gaps` — two distinct feature/acceptance prompts with other gaps still open → both replies stay feature-specific (bot review #1's exact non-blocking suggestion).
  - `test_interview_driver_preserves_repeated_specific_acceptance_answer` — same feature/acceptance prompt asked twice → both replies stay feature-specific (bot review #3's exact regression scenario).
- **CI**: `Ruff Lint`, `MyPy Type Check`, `Bridge TypeScript`, and `Test Python 3.12 / 3.13 / 3.14` all green on `b250a161`. Two intermittent timing-flakes (`test_session_activity_resets_idle_timeout`, `test_ralph_handler_returns_job_id_and_completes_loop`) hit on the first runs but cleared on re-run; both are in unrelated files (`tests/unit/evolution/`, `tests/unit/mcp/tools/`) and were not touched by this PR.
- **Blast radius**: only `src/ouroboros/auto/interview_driver.py` (steering reorder + broad-prompt helper), `src/ouroboros/auto/answerer.py` (one new optional dataclass field with default `False`), and the new test cases. No public API removed; existing callers of `AutoAnswer(...)` keep working because `generic_default` defaults to `False`.

`mergeStateStatus` is `CLEAN`, `mergeable` is `MERGEABLE`, review decision is `APPROVED`. Please take another look when you have a minute. 🙏

PR #767, opened 2026-05-08
fix(claude_adapter): Windows MCP spawn + strict-mode + parse-loop hardening (4 fixes)

whshea9792:
## Summary

Four scoped fixes to `src/ouroboros/providers/claude_code_adapter.py` discovered during a Windows 10 (subscription auth) bring-up of `ouroboros-ai 0.36.0` + `claude_agent_sdk 0.1.76` + Claude Code CLI 2.1.132. Each fix is a separate commit so reviewers can adopt subsets.

| # | Fix | Commit | Severity |
|---|---|---|---|
| 1 | Add `AskUserQuestion` + 14 other UI/orchestration tools to `dangerous_tools` | `d5d8aed7` | **Blocker** (silent spawn failure) |
| 2 | Union `dangerous_tools` into strict-mode `disallowed` set | `ade7a794` | **Blocker** (strict callers re-introduce bug 1) |
| 3 | Skip `os.access(X_OK)` check on Windows (PATHEXT shims) | `93bec32f` | Major (silent CLI override rejection) |
| 4 | Bound `MessageParseError` loop with consecutive-error limit | `36257cd9` | Minor (potential worker-pinning hang) |

### Fix 1 — UI tools missing from `dangerous_tools`
On Windows / subscription auth, `mcp__plugin_ouroboros_ouroboros__ouroboros_auto` failed every run with `Command failed exit 1`, `stderr_lines=0`, `claudecode_present=False`. Initial assumption was env-passthrough or stdio contention, but the real cause is upstream model behaviour: in non-interactive contexts the model invokes `AskUserQuestion` to clarify ambiguous prompts. Without UI to receive the response, the request hangs at `max_turns` and the SDK never emits an `AssistantMessage TextBlock`. The adapter wraps this no-output condition as exit-1 — misleadingly identical to a real spawn failure.

After patching `dangerous_tools` to include `AskUserQuestion` (and related: `EnterPlanMode`, `ExitPlanMode`, `EnterWorktree`, `ExitWorktree`, `ScheduleWakeup`, `PushNotification`, `RemoteTrigger`, `Cron*`, `Skill`, `ToolSearch`, `Monitor`, `ListMcpResourcesTool`, `ReadMcpResourceTool`), `ouroboros_auto` and `ouroboros_interview` produce real LLM-generated content end-to-end. Verified across multiple full pipeline runs (interview → seed → execute → evaluate, 2348 messages, 614 tool calls, ~1h 19m wall, 15/15 ACs landed).

### Fix 2 — Strict-mode `all_tools` enumeration
The strict-mode branch (`allowed_tools is not None`) at lines 599–615 computes `disallowed` only from `all_tools`, which enumerates 13 classic builtins (Read/Write/Edit/Bash/WebFetch/WebSearch/Glob/Grep/Task/NotebookEdit/TodoRead/TodoWrite/LS). UI tools are absent from that list and therefore never appear in `disallowed` even when caller specifies a strict allow list. Fix 1 alone is incomplete because any caller passing `allowed_tools=["Read","Glob","Grep"]` (typical "read-only minimal" pattern) would still expose `AskUserQuestion`, re-introducing the bug. This patch builds `disallowed` as the union of `all_tools` and `dangerous_tools`, filtered to exclude `allowed_tools`.

### Fix 3 — `os.access(X_OK)` on Windows
`os.access(p, os.X_OK)` returns False on Windows for `.cmd` / `.bat` / `.ps1` shims that Windows treats as executable via `PATHEXT`. Custom CLI paths like `OUROBOROS_CLI_PATH=C:\Users\<u>\AppData\Roaming\npm\claude.cmd` (typical npm-global install on Windows) are silently rejected with `cli_not_executable` warning, then the adapter falls back to the SDK bundled CLI. Patch gates the X_OK check on `sys.platform != "win32"`. The upstream `.exists()` / `.is_file()` checks are sufficient on Windows; PATHEXT-based launchability is not exposed via `os.access`.

### Fix 4 — Unbounded `MessageParseError` swallow
`_safe_query()` catches `MessageParseError` and continues with no max-skip counter and no application-level deadline (default `self._timeout=None`). If the SDK enters a state where every message is unparseable (e.g. SDK / CLI version skew introducing a new message type the installed SDK doesn't model), the loop runs forever pinning a worker with no signal to the caller. Patch tracks consecutive parse errors; after 50 in a row with no other message type, raises `ProviderError` so the caller sees an actionable failure mode. Counter resets on any successful yield. Original "skip unknown, keep streaming known messages" semantics preserved for transient unknowns via the `else` clause.

## Out of scope (separate follow-up PRs planned)

The audit also surfaced bugs in other files; intentionally not bundled into this PR to keep the diff scoped to one file:

- **B1**: `os.kill(pid, 0)` raises bare `OSError` on Windows — `core/worktree.py:300`, `orchestrator/heartbeat.py:176`. Lift the existing Windows-aware `_pid_is_alive` shim from `cli/commands/mcp_doctor.py:327-360` into `core/process.py` and call from both sites.
- **M2**: `find_real_cli` looks for bare filename without `.exe` / `.cmd` / `.bat` — `codex/cli_policy.py:96-109`, `copilot/cli_policy.py:118-131`. Same root cause as Fix 3 (PATHEXT).
- **M3**: `_ensure_shell_env()` hardcodes `python3 -c …` literal — `cli/commands/mcp.py:130`, `setup.py:73`. Should use `sys.executable` or `shutil.which` fallback.
- **Mi1**: Silent persist failure on interview ambiguity — `authoring_handlers.py:1804-1809` logs warning and continues; sister branches at lines 1614 and 1550 correctly raise.

Happy to follow up with separate PRs for any of these if useful.

## Test plan

- [x] AST parse check passes (`python -c "import ast; ast.parse(open(path).read())"`)
- [x] `mcp__plugin_ouroboros_ouroboros__ouroboros_auto` produces real LLM content (was: silent exit 1)
- [x] `mcp__plugin_ouroboros_ouroboros__ouroboros_interview` produces real LLM content
- [x] Full evolutionary pipeline run end-to-end: 2348 messages, 614 tool calls, 15/15 acceptance criteria landed (~1h 19m wall, real downstream Viber install build)
- [ ] Reviewer to confirm: no impact on macOS / Linux smoke
- [ ] Reviewer to confirm: parse-error limit value of 50 is reasonable (configurable not added to keep diff small)

## Reproducer (Windows 10, before fix 1)

```
mcp__plugin_ouroboros_ouroboros__ouroboros_auto(goal="hello world test")
```

→ blocks immediately with `Command failed with exit code 1`, `stderr_lines=0`, `claudecode_present=False`.

After fix 1, returns real interview output and proceeds through the pipeline.

### DIFF
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -592,7 +592,38 @@ class ClaudeCodeAdapter:
         #
         # Strategy: Block dangerous tools explicitly, allow everything else
         # This enables MCP tools (mcp__*) and read-only built-in tools
-        dangerous_tools = ["Write", "Edit", "Bash", "Task", "NotebookEdit"]
+        #
+        # The list MUST include UI / orchestration tools that block headless
+        # requests waiting for user input. If a model invokes one of these in
+        # a non-interactive context (no UI to receive the response), the
+        # request hangs at max_turns and the parent SDK call returns no
+        # usable text — surfacing as a misleading "Command failed exit 1"
+        # because no AssistantMessage TextBlock is ever emitted.
+        dangerous_tools = [
+            # File-mutating builtins
+            "Write",
+            "Edit",
+            "Bash",
+            "Task",
+            "NotebookEdit",
+            # UI / orchestration tools — block in headless contexts
+            "AskUserQuestion",
+            "ExitPlanMode",
+            "EnterPlanMode",
+            "EnterWorktree",
+            "ExitWorktree",
+            "ScheduleWakeup",
+            "PushNotification",
+            "RemoteTrigger",
+            "CronCreate",
+            "CronDelete",
+            "CronList",
+            "Skill",
+            "ToolSearch",
+            "Monitor",
+            "ListMcpResourcesTool",
+            "ReadMcpResourceTool",
+        ]
 
         # If allowed_tools is explicitly set (even empty []), compute disallowed
         # from it (strict mode). None = permissive (only block dangerous).
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -820,14 +820,43 @@ class ClaudeCodeAdapter:
             from claude_agent_sdk._errors import MessageParseError as _MPE
 
             gen = query(prompt=prompt, options=options).__aiter__()
+            consecutive_parse_errors = 0
+            # Bound the parse-error skip loop. If every message coming back
+            # is unparseable (e.g. SDK / CLI version skew introducing a new
+            # message type the SDK doesn't model yet), the original loop
+            # would spin forever pinning a worker with no application-level
+            # signal. 50 consecutive parse errors with no other message
+            # type indicates a stuck stream, not transient unknown messages.
+            PARSE_ERROR_LIMIT = 50
+
             while True:
                 try:
-                    yield await gen.__anext__()
+                    msg = await gen.__anext__()
                 except _MPE as parse_err:
-                    log.debug("claude_code_adapter.skipping_unknown_message", error=str(parse_err))
+                    consecutive_parse_errors += 1
+                    if consecutive_parse_errors >= PARSE_ERROR_LIMIT:
+                        log.warning(
+                            "claude_code_adapter.parse_error_limit_exceeded",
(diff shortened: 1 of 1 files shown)


whshea9792:
Closing per reporter request — patches will stay local only. Apologies for the noise; should not have opened upstream without confirming reporting policy first.

PR #787, opened 2026-05-08
fix(auto): bound retry on run_handoff_status="unknown" with idempotency-key

### LINKED ISSUE #774: fix(auto): bound retry on run_handoff_status="unknown" with idempotency-key
Part of #772.

## Problem

`src/ouroboros/auto/pipeline.py` has three "unknown handoff" code paths centered on `_mark_unknown_run_handoff`:

- `pipeline.py:355-356` — duplicate-prevention guard: `if state.run_start_attempted: _mark_unknown_run_handoff(state)` blocks the pipeline rather than retry.
- `pipeline.py:410-413` — `asyncio.TimeoutError` from the run starter calls `_mark_unknown_run_handoff(state, status="unknown_timeout")`.
- `pipeline.py:430` — handler returned but no `job_id`/`execution_id`/`run_session_id` was present.

In the "single input → completed product" model, the user must currently re-invoke `ooo auto --resume <session_id>` after any of these. That breaks the one-shot promise of #772.

The reason a duplicate-execution race exists at all is that the run starter has no idempotency contract: a transient timeout on the first call may have already enqueued an execution server-side, and a naive retry would create a second one.

## Why now

The auto pipeline cannot fulfill its end-to-end promise in #772 if a single transient run-starter timeout requires manual user resume.

## Acceptance criteria

- [ ] `RunStarter.__call__` (and `HandlerRunStarter`) accept an `idempotency_key: str` parameter, populated from `state.auto_session_id` by `pipeline.py`.
- [ ] `StartExecuteSeedHandler` honors the key:
  - Maintains an in-memory `dict[str, ExecutionMeta]` mapping `idempotency_key → previously-returned metadata`.
  - On a second call with the same key, returns the *same* execution metadata it returned originally; never enqueues a new execution.
  - The map TTL is the process lifetime; on server restart the key is lost. **Documented non-goal**: persistent idempotency across server restarts. After restart, a duplicate request *can* enqueue a new execution; this is acceptable given that auto state recovery on the same restart will surface the duplicate.
  - Contract documented in `StartExecuteSeedHandler` docstring with the explicit non-goal called out.
- [ ] `pipeline.py` retries the run starter exactly once when `run_handoff_status` lands in `"unknown_timeout"` or in the no-handle branch *and* `state.run_start_attempted=True`. The duplicate-prevention guard at `pipeline.py:355-356` is replaced by this retry path.
- [ ] The retry uses the same `run_start_timeout_seconds` as the first attempt (currently 60s); retry is *not* an opportunity to extend the timeout.
- [ ] If the retry also fails to surface any handle, the pipeline transitions to `BLOCKED` with `tool_name="run_starter"` and a guidance message that contains the literal phrase `"retried once with idempotency key"`.
- [ ] Regression tests, all in `tests/unit/auto/test_run_starter_idempotency.py`:
  - First call raises `asyncio.TimeoutError`, second call (with same idempotency key) returns the original execution_id; pipeline reaches `COMPLETE`, not `BLOCKED`.
  - First call returns `{}` (no handle), second call returns valid handle; pipeline reaches `COMPLETE`.
  - Both calls fail; pipeline reaches `BLOCKED` and `state.last_error` contains the documented phrase.
- [ ] Server-side test in `tests/unit/mcp/tools/test_start_execute_seed_idempotency.py`:
  - Same idempotency_key → same returned `execution_id`, exactly one execution enqueued (verify by inspecting EventStore).
  - Different idempotency_keys → independent executions.
  - Process restart simulation (new handler instance) → key map is empty, behavior is "first time" again.

## Scope / Non-goals

- **In**: auto-side retry policy, idempotency-key plumbing, server-side handler contract for `StartExecuteSeedHandler`, in-memory key map.
- **Out**: persistent idempotency across server restarts; generalizing idempotency across other handlers; server-side TTL eviction (the in-memory map is bounded by process lifetime, no eviction policy in this issue).

## Files & touch points

- `src/ouroboros/auto/pipeline.py:355-356,410-413,430` and `_mark_unknown_run_handoff` helper.
- `src/ouroboros/auto/adapters.py:HandlerRunStarter`
- `src/ouroboros/mcp/tools/execution_handlers.py:StartExecuteSeedHandler`
- `tests/unit/auto/test_run_starter_idempotency.py` (new)
- `tests/unit/mcp/tools/test_start_execute_seed_idempotency.py` (new)

## Depends on / Blocks

- Independent. Can ship before or in parallel with #773.

## Test plan

- [ ] `pytest tests/unit/auto/test_run_starter_idempotency.py -v`
- [ ] `pytest tests/unit/mcp/tools/test_start_execute_seed_idempotency.py -v`

shaun0927:
Closes #774.

## Summary
- `RunStarter` is now a `Protocol` whose `__call__` accepts an `idempotency_key` keyword. `HandlerRunStarter` forwards it as a tool argument so server-side handlers can de-duplicate retries.
- `AutoPipeline.run()` retries the run starter exactly once on `unknown_timeout` / no-handle outcomes using `state.auto_session_id` as the idempotency key. The duplicate-prevention guard at `pipeline.py:355-356` is replaced by this bounded retry path; both attempts share the same `run_start_timeout_seconds` budget.
- On retry exhaustion the pipeline transitions to `BLOCKED` with `tool_name="run_starter"` and the literal phrase `retried once with idempotency key` in both `state.last_error` and `state.run_handoff_guidance`, so a later resume cannot escalate to a third call.
- `StartExecuteSeedHandler` maintains a process-local `dict[str, ExecutionMeta]` keyed by `idempotency_key`. Repeated calls with the same key replay the original metadata without enqueuing a new execution; persistence across server restarts is documented as a non-goal.
- Existing tests that asserted the old "block on duplicate" behavior were updated to reflect the new retry contract.

## Test plan
- [x] `pytest tests/unit/auto/test_run_starter_idempotency.py -v` (3 scenarios: timeout-then-handle, no-handle-then-handle, both-fail-blocks-with-phrase)
- [x] `pytest tests/unit/mcp/tools/test_start_execute_seed_idempotency.py -v` (same key replays / different keys diverge / restart resets / no key keeps legacy behavior)
- [x] Local regression: `pytest tests/unit/auto/ tests/unit/mcp/tools/ -x` — 975 passed
- [x] `ruff format --check src/ tests/` and `ruff check src/ tests/` — clean

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -136,12 +136,15 @@ class HandlerRunStarter:
         self.handler = handler
         self.cwd = cwd
 
-    async def __call__(self, seed: Seed) -> dict[str, object]:
+    async def __call__(self, seed: Seed, *, idempotency_key: str = "") -> dict[str, object]:
         seed_yaml = yaml.dump(
             seed.to_dict(), default_flow_style=False, allow_unicode=True, sort_keys=False
         )
+        arguments: dict[str, object] = {"seed_content": seed_yaml, "cwd": self.cwd}
+        if idempotency_key:
+            arguments["idempotency_key"] = idempotency_key
         result = _unwrap(
-            await self.handler.handle({"seed_content": seed_yaml, "cwd": self.cwd}),
+            await self.handler.handle(arguments),
             tool_name="ouroboros_start_execute_seed",
         )
         meta = result.meta or {}
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -7,7 +7,7 @@ from collections.abc import Awaitable, Callable
 from dataclasses import asdict, dataclass, field
 import inspect
 import threading
-from typing import Any
+from typing import Any, Protocol
 
 from ouroboros.auto.blocker_attribution import record_authoring_backend
 from ouroboros.auto.grading import GradeGate
@@ -27,9 +27,25 @@ from ouroboros.auto.state import (
 from ouroboros.core.seed import Seed
 
 SeedGenerator = Callable[[str], Awaitable[Seed]]
-RunStarter = Callable[[Seed], Awaitable[dict[str, Any]]]
+
+
+class RunStarter(Protocol):
+    """Protocol for run-starter callables.
+
+    Implementations accept an optional ``idempotency_key`` so the auto
+    pipeline can safely retry a single run-start attempt without enqueuing
+    a duplicate execution server-side. The key is populated from
+    ``state.auto_session_id`` by ``AutoPipeline.run``.
+    """
+
+    async def __call__(self, seed: Seed, *, idempotency_key: str = "") -> dict[str, Any]: ...
+
+
 SeedSaver = Callable[[Seed], str]
 SeedLoader = Callable[[str], Seed]
+
+
+_RETRY_GUIDANCE_PHRASE = "retried once with idempotency key"
 
 
 @dataclass(frozen=True, slots=True)
@@ -381,15 +397,6 @@ class AutoPipeline:
                 )
                 self._save(state)
                 return self._result(state, ledger, review=review)
-            if state.run_start_attempted:
-                _mark_unknown_run_handoff(state)
-                state.mark_blocked(
-                    state.run_handoff_guidance
-                    or "Run start status is unknown; refusing to start a duplicate execution",
-                    tool_name="run_starter",
-                )
-                self._save(state)
-                return self._result(state, ledger, review=review, blocker=state.last_error)
             if not _grade_meets_required(state.last_grade, state.required_grade):
                 state.mark_blocked(
                     f"Cannot start execution without a persisted grade meeting {state.required_grade}",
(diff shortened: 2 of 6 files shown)


shaun0927:
@Q00 — re-review ping. PR is at `4135e50b`, CI fully green, ouroboros-agent[bot] verdict **APPROVE** (commit `49ef295`), no in-scope blocking findings remaining, GitHub state `MERGEABLE` / `CLEAN`.

## What this PR does

Closes #774 by replacing the pre-existing "block on duplicate" guard at `pipeline.py:355-356` with a **bounded once-only retry** flow for `unknown_timeout` / no-handle outcomes from `run_starter`, paired with a **server-side idempotency replay** so the retry cannot duplicate-enqueue:

- `RunStarter` is now a `Protocol` whose `__call__` accepts `idempotency_key`; `HandlerRunStarter` forwards it as a tool argument.
- `AutoPipeline.run()` retries exactly once per session lifetime using `state.auto_session_id` as the idempotency key, sharing the same `run_start_timeout_seconds` budget. On exhaustion the pipeline transitions to `BLOCKED` with `tool_name="run_starter"` and the literal phrase `retried once with idempotency key` carried in both `state.last_error` and `state.run_handoff_guidance`.
- `StartExecuteSeedHandler` keeps a process-local `dict[str, ExecutionMeta]` keyed by `idempotency_key`. Repeated calls with the same key replay the original metadata for both the background-job path and the plugin-dispatch path; persistence across server restarts is documented as a non-goal.

## How it converged across three review cycles

The bot ran two REQUEST_CHANGES cycles before the APPROVE; the iteration is what gives me confidence the contract is now actually robust rather than just plausibly so.

**Cycle 1 (commit `e7f0337` → `4ac36a0`):** bot flagged two BLOCKING gaps in the original implementation:

1. Plugin-dispatch branch never wrote to the idempotency cache, so retries re-emitted `subagent.dispatched` events. Fixed by mirroring the cache shape (`_idempotency_plugin_payload`) and replaying the cached `_subagent` envelope + `response_shape` on the second call.
2. `pipeline.run()` reset `run_start_attempted` when the **retry attempt itself** raised a non-timeout exception, so a later resume could call `run_starter` a third time. Fixed by persisting `run_handoff_status="unknown_retry_failed"` + the documented retry phrase, leaving `run_start_attempted=True`, and short-circuiting at a symmetric guard at the top of the RUN branch.

**Cycle 2 (commit `4ac36a0` → `49ef295`):** bot flagged two more BLOCKING gaps that only surfaced after the first round of fixes narrowed the contract:

1. **Legacy session compat:** pre-#787 sessions persisted with `run_start_attempted=True` and no `run_handoff_status` field (`from_dict` defaults it to `None` on load) would fall through the new retry-only check and call `run_starter` again — exactly the duplicate-execution case the original guard prevented. Fixed by extending `prior_retry_exhausted` to block any `run_start_attempted=True` session whose status is **not** in the retryable set `{"unknown_no_handle", "unknown_timeout"}`. This also covers a mid-call crash where the loop saved `run_start_attempted=True` before `_mark_unknown_run_handoff` ran.
2. **Concurrent-retry race:** cache writes for `_idempotency_meta` and `_idempotency_plugin_payload` happen *after* dispatch, so two concurrent `handle()` calls with the same key both miss the cache and each enqueue/dispatch independently. Fixed by adding a per-key `asyncio.Lock` (`dict.setdefault` is atomic in single-threaded asyncio, so lock-creation is race-free) so the check-cache / dispatch / write-cache sequence is atomic per key — the second caller waits for the first to populate the cache and then takes the cache hit.

**Cycle 3 (commit `49ef295` → `4135e50b`):** bot approved with one non-blocking warning that `_mark_unknown_run_handoff` guidance text still claimed "Resume will not start another run automatically", contradicting the new auto-retry behavior. Updated both timeout and no-handle guidance strings to describe the actual recovery contract (one automatic retry on resume, then block).

## Why I think this is merge-ready

- **Bot blocking findings:** zero remaining. Bot's design-notes line: *"the bounded retry plus server-side idempotency design is sound, and the new handler-level per-key lock closes the main duplicate-enqueue race."*
- **Contract invariants are exhaustively covered by regression tests:**
  - `tests/unit/auto/test_run_starter_idempotency.py` — timeout-then-handle, no-handle-then-handle, both-fail-blocks-with-phrase, retry-raises-non-timeout (asserts a follow-up `pipeline.run()` does **not** call `run_starter` a third time), and legacy-session-resume-does-not-dispatch.
  - `tests/unit/mcp/tools/test_start_execute_seed_idempotency.py` — same-key replay (background-job + plugin paths), different-keys diverge, restart resets, no-key keeps legacy behavior, **concurrent same-key dedupe** for both background-job and plugin paths (the second test holds the first `start_job` open via an `asyncio.Event` so the in-flight race is actually exercised, not just stipulated).
- **CI:** Ruff Lint, MyPy Type Check, Tests (Python 3.12 / 3.13 / 3.14), Bridge TypeScript, enforce-boundary — all SUCCESS on `4135e50b`. Local sweep `pytest tests/unit/auto/ tests/unit/mcp/tools/` = 980 passed.
- **Behavioral compatibility:** legacy sessions persisted before this PR retain the conservative pre-#787 behavior (block on `run_start_attempted=True` without a retryable status); only sessions that explicitly recorded an `unknown_*` handoff opt into the new retry path. No status-quo session resumed after deploy can dispatch a duplicate enqueue.
- **Non-goal explicitly documented:** persistence across server restarts. The map TTL is process lifetime; on restart auto-side state recovery surfaces the duplicate. This is the intentional scope bound from #774.

The two non-blocking follow-ups the bot listed in its APPROVE (an adapter-boundary integration test, plus the guidance-text alignment which I addressed in `4135e50b`) are explicitly classified as non-blocking and sit comfortably in the "could be a follow-up PR" bucket — they don't compromise the contract being shipped here.

Ready for your re-review and merge whenever you're ready.

shaun0927:
@Q00 re-review ping — I re-checked PR #787 at latest head `4135e50b` and consider it merge-ready.

Summary of what changed:
- The PR closes #774 by replacing the previous conservative duplicate-run block with a bounded, exactly-once retry for `unknown_timeout` / no-handle run-start outcomes.
- `AutoPipeline` now reuses `state.auto_session_id` as the `idempotency_key` for the initial run-start attempt and the single retry, then durably blocks with the explicit `retried once with idempotency key` marker if the retry budget is exhausted.
- `HandlerRunStarter` forwards that key into `ouroboros_start_execute_seed`.
- `StartExecuteSeedHandler` now dedupes both normal background execution and plugin-dispatch mode, including concurrent same-key calls via a per-key lock, while keeping restart persistence explicitly out of scope.

Why I think it is safe to merge:
- The earlier bot blocking findings were addressed: plugin dispatch now stores/replays the idempotency payload, retry-side exceptions consume the retry budget, legacy pre-#787 `run_start_attempted=True` sessions conservatively block, and overlapping same-key handler calls are serialized.
- The latest ouroboros-agent[bot] review on `4135e50` is **APPROVE** with no in-scope blocking findings or non-blocking suggestions remaining.
- GitHub reports the PR as `MERGEABLE`, and all required checks are green: Ruff Lint, MyPy Type Check, Bridge TypeScript, ooo-auto-boundary, and Python 3.12/3.13/3.14 tests.
- I also verified the branch in an isolated worktree and ran:
  - `PYTHONPATH=/tmp/ouroboros-pr787-review/src /opt/data/projects/ouroboros/.venv/bin/pytest tests/unit/auto/test_run_starter_idempotency.py tests/unit/mcp/tools/test_start_execute_seed_idempotency.py -q` → `12 passed`
  - `PYTHONPATH=/tmp/ouroboros-pr787-review/src /opt/data/projects/ouroboros/.venv/bin/pytest tests/unit/auto/test_interview_pipeline.py tests/unit/auto/test_run_starter_idempotency.py -q` → `132 passed`
  - `PYTHONPATH=/tmp/ouroboros-pr787-review/src /opt/data/projects/ouroboros/.venv/bin/ruff check ...` on the changed implementation/tests → `All checks passed!`

I did not push any additional commits because the latest implementation already matches the intended bounded-retry/idempotency contract and the remaining bot design note is not a merge blocker for this PR.

Q00 (approved):
LGTM. Verified 12 idempotency-specific tests + 450 broader auto tests pass locally. Using `state.auto_session_id` as the `idempotency_key` is the right choice — it's stable across mid-call crashes, `unknown_timeout`, and `unknown_no_handle` resume paths, so the bounded retry can't accidentally double-enqueue. Splitting `unknown_*` into distinct status values gives the recovery branches enough signal to make the right call, and the `_RETRY_GUIDANCE_PHRASE` surface is helpful for downstream log analysis.

PR #795, opened 2026-05-08
fix(plugin): escape all C0/DEL chars in lockfile TOML basic strings

Q00:
## Summary
- ``_toml_str`` previously only handled ``\``, ``\"``, ``\n``, ``\t`` and ``\r``. Any other ``ord(ch) < 0x20`` byte (e.g. ``\x0b``, ``\x0c``, ``\x05``) was emitted verbatim, but TOML 1.0 §2.1 forbids bare control characters in basic strings.
- The very next ``Lockfile.read()`` after such an ``add()`` then crashed with ``tomllib.TOMLDecodeError: Illegal character`` — leaving the lockfile corrupted in place.
- Fix: extend the escape table with ``\b``/``\f`` and fall back to ``\uXXXX`` for any remaining C0 or DEL byte. ASCII-clean values are unchanged.

## Reproduction
```python
import tomllib
from ouroboros.plugin.lockfile import _toml_str
v = _toml_str("sha256:abc\x0b123")
tomllib.loads(f"x = {v}")
# Before: TOMLDecodeError: Illegal character '\x0b'
# After:  parses cleanly via  escape
```

## Test plan
- [x] new ``test_toml_str_escapes_all_control_chars`` parametrized over ``\x01 \x05 \x08 \x0b \x0c \x1f \x7f``
- [x] all 14 existing lockfile tests pass

### DIFF
--- a/src/ouroboros/plugin/lockfile.py
+++ b/src/ouroboros/plugin/lockfile.py
@@ -59,22 +59,44 @@ class LockEntry:
         return lines
 
 
+_TOML_BASIC_ESCAPES: dict[str, str] = {
+    "\\": "\\\\",
+    '"': '\\"',
+    "\n": "\\n",
+    "\t": "\\t",
+    "\r": "\\r",
+    "\b": "\\b",
+    "\f": "\\f",
+}
+
+
 def _toml_str(value: str) -> str:
-    """Serialize a string value as TOML basic string. Restricts content to
-    avoid escaping edge cases — the lockfile only stores names, paths, hashes
-    and timestamps, none of which contain control chars or non-ASCII in
-    practice."""
-    if any(ch == "\\" or ch == '"' or ord(ch) < 0x20 for ch in value):
-        # Use TOML's basic string escapes for the small set we actually need.
-        escaped = (
-            value.replace("\\", "\\\\")
-            .replace('"', '\\"')
-            .replace("\n", "\\n")
-            .replace("\t", "\\t")
-            .replace("\r", "\\r")
-        )
-        return f'"{escaped}"'
-    return f'"{value}"'
+    """Serialize a string value as a TOML basic string.
+
+    The lockfile normally stores names, paths, hashes and timestamps, none of
+    which contain control characters in practice. The escape table previously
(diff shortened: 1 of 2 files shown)


shaun0927:
@Q00 re-review requested.

I reviewed PR #795's intent, implementation direction, the current diff, and the bot design-note/review signal on the latest head (`6ce97052`). My conclusion is that this PR is valid and merge-ready.

## What this PR does

The PR fixes the plugin lockfile TOML serializer so `_toml_str()` no longer emits invalid bare control characters inside TOML basic strings. Previously, only `\\`, `\"`, `\n`, `\t`, and `\r` were escaped; other C0 bytes such as VT (`\x0b`), FF (`\x0c`), or arbitrary low control bytes could be written literally, causing the next `Lockfile.read()` / `tomllib` parse to fail and leaving the lockfile unreadable.

The current implementation:
- keeps standard TOML escapes for backslash, quote, newline, tab, carriage return, backspace, and form feed;
- emits `\uXXXX` for remaining C0 control characters and DEL (`0x7f`);
- preserves the previous output for printable ASCII/non-control values;
- adds regression coverage through the public `Lockfile.add()` → `Lockfile.read()` path.

## Review and validation

- Confirmed the effective diff is narrow: `src/ouroboros/plugin/lockfile.py` and `tests/unit/plugin/test_lockfile.py` only.
- Reviewed the serializer against TOML basic-string escaping requirements and `tomllib` parsing behavior.
- Ran a dependency-free manual roundtrip check by loading `lockfile.py` directly and verifying values containing sampled C0 controls/DEL serialize, parse with `tomllib`, and read back unchanged.
- Ran `git diff --check upstream/main...HEAD` successfully.
- Confirmed GitHub checks are green on the latest head: Ruff, MyPy, Bridge TypeScript, enforce-boundary, and Python 3.12/3.13/3.14 tests.
- Confirmed `ouroboros-agent[bot]` approved the latest head (`6ce97052`) with no in-scope blocking findings.

I did not find any blocking correctness, safety, or scope issues, and I did not make any code changes. The PR is mechanically clean (`MERGEABLE` / `CLEAN`) and review-ready for merge from my side.

PR #796, opened 2026-05-08
fix(plugin): deep-copy audit event in unwrap_plugin_event

Q00:
## Summary
- ``wrap_plugin_event`` defensively ``copy.deepcopy``s the audit event on the way into the envelope (ledger_adapter.py:125), but the inverse ``unwrap_plugin_event`` returned only ``dict(payload)`` — shallow.
- A caller mutating a nested container on the unwrapped value (e.g. ``result[\"plugin\"][\"name\"] = \"X\"``) wrote through to the envelope still referenced by the event store, silently corrupting the audit log.
- Fix: return ``copy.deepcopy(payload)`` to restore symmetry. ``envelope_to_base_event`` (line 269) already uses ``copy.deepcopy``.

## Reproduction
```python
env = wrap_plugin_event({\"event_type\":\"plugin.invoked\",\"plugin\":{\"name\":\"foo\"},...}, correlation_id=\"c\")
out = unwrap_plugin_event(env)
out[\"plugin\"][\"name\"] = \"MUTATED\"
env[\"payload\"][\"plugin\"][\"name\"]   # before: 'MUTATED' — corrupted; after: 'foo'
```

## Test plan
- [x] new ``test_unwrap_returned_value_is_isolated_from_envelope`` covers mutations on every nested container
- [x] all 22 existing ledger_adapter tests pass

### DIFF
--- a/src/ouroboros/plugin/ledger_adapter.py
+++ b/src/ouroboros/plugin/ledger_adapter.py
@@ -150,7 +150,12 @@ def unwrap_plugin_event(envelope: dict) -> dict:
     payload = envelope.get("payload")
     if not isinstance(payload, dict):
         raise ValueError("envelope payload is missing or not a dict")
-    return dict(payload)
+    # Deep copy to mirror ``wrap_plugin_event``'s defensive copy on the way
+    # in. ``dict(payload)`` is a shallow copy: mutating any nested dict on
+    # the returned value (e.g. ``result["plugin"]["name"] = "X"``) would
+    # write through to the live envelope still referenced by the event
+    # store, silently corrupting the audit log.
+    return copy.deepcopy(payload)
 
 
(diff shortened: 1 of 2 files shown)


Q00:
Re-labeling as **defensive-cleanup**, not a user-impacting bug fix.

To be fully honest: I did not find an actual caller that mutates the unwrapped audit-event dict. The change makes ``unwrap_plugin_event`` symmetric with ``wrap_plugin_event`` (which already deep-copies on the way in) and removes a foot-gun for future callers, but no current code path produces the corruption the test demonstrates.

Worth merging on the symmetry argument alone — ``copy.deepcopy`` of a small audit event is sub-microsecond and the test prevents future regressions — but please review with that framing rather than as a critical fix.

PR #797, opened 2026-05-08
fix(scripts): skip symlinks in check-auto-boundary scan

Q00:
## Summary
- ``Path.rglob(\"*.py\")`` follows symlinks; a symlink farm under SCAN_DIRS pulled vendored ``*.py`` into the scan and produced false positives.
- Fix: skip both leaf symlinks and any candidate whose parent walk hits a symlinked directory before reaching the scan root.

## Reproduction
```bash
mkdir -p src/ouroboros/auto/
ln -s /tmp/external src/ouroboros/auto/external
echo 'class GitHubVendor: pass' > /tmp/external/sdk.py
python3 scripts/check-auto-boundary.py   # before: fails; after: passes
```

## Test plan
- [x] new directory-symlink + file-symlink regressions
- [x] all 68 existing scripts/check-auto-boundary tests still pass

### DIFF
--- a/scripts/check-auto-boundary.py
+++ b/scripts/check-auto-boundary.py
@@ -209,38 +209,44 @@ def _resolve_scan_targets() -> tuple[list[Path], list[str]]:
     targets: list[Path] = []
     seen: set[Path] = set()
 
+    repo_root_resolved = REPO_ROOT.resolve()
     for d in SCAN_DIRS:
         root = REPO_ROOT / d
         if root.is_dir():
             for p in sorted(root.rglob("*.py")):
-                # Path.rglob follows symlinks by default. A symlink farm
-                # under SCAN_DIRS (e.g. ``src/ouroboros/auto/external ->
-                # ../../vendor/sdk``) would silently pull every ``*.py``
-                # under the link target into the scan, producing false
-                # positives for code we explicitly chose not to police —
-                # and training reviewers to add spurious allowlist markers
-                # to vendored files. Skip any symlink (file or directory
-                # link) so the scan stays anchored to in-tree files.
-                if p.is_symlink():
+                # ``Path.rglob`` follows symlinks by default. The risk a
+                # symlink farm introduces is *escape from the repo*: a
+                # link such as ``src/ouroboros/auto/external ->
+                # /home/user/vendor/sdk`` pulls every ``*.py`` under the
+                # link target into the scan, producing false positives
+                # for code we deliberately did not police.
+                #
+                # The meaningful architectural boundary here is "does
+                # this file still resolve inside the repo?", not "is
+                # this path a symlink?". A symlink that points to
+                # *another in-repo file* is just an implementation
+                # detail (``src/ouroboros/auto/legacy.py ->
+                # ../shared/legacy.py``) — Python imports it normally,
+                # so the boundary guard must scan it normally too.
+                # Earlier revisions skipped every symlink unconditionally
+                # and turned in-repo links into unscanned blind spots,
+                # weakening the guard's contract (Q00/ouroboros#797
+                # review feedback).
+                try:
+                    resolved = p.resolve(strict=False)
+                except OSError:  # broken symlink — treat as escape
                     continue
-                # Also skip files whose any parent on the scan path is a
-                # symlink: ``rglob`` traverses symlinked directories, and
-                # only the directory entry itself shows up as a symlink —
-                # the leaf ``*.py`` it descends into does not. Walk up
-                # from the file to ``root`` and reject if any intermediate
-                # directory is a symlink.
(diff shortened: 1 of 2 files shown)


shaun0927:
@Q00 — re-review ping. This PR is now bot-approved on `36588f60` with all 8 CI checks green and `mergeStateStatus: CLEAN`. Below is a self-contained merge rationale.

## What this PR does
`scripts/check-auto-boundary.py` is the mechanical guard for `ooo auto`'s product boundary (Q00/ouroboros#725). Before this PR, `Path.rglob("*.py")` followed symlinks unconditionally, so a symlink farm like `src/ouroboros/auto/external -> /home/user/vendor/sdk` pulled vendored `*.py` into the scan and produced false-positive failures on code we deliberately did not police.

The PR's goal: scan only **in-repo code**. The boundary the guard cares about is "does this path resolve inside REPO_ROOT?", **not** "is this path a symlink?" — Python imports an in-repo symlink normally, so the guard must too.

## Iteration trail (4 commits, 3 bot reviews)
The bot review process flushed out three distinct contract holes, each fixed by a follow-up commit:

| Commit | Bot finding | Fix |
|---|---|---|
| `799d12b` | Naive `is_symlink()` filter turned every in-repo symlink into an unscanned blind spot | Switched to "resolved target escapes REPO_ROOT" |
| `25ed0d7` | Repo-boundary rule applied only to `SCAN_DIRS`, not `SCAN_EXTRA_FILES` (extra-file symlinked-out → external file scanned, rc=1 false-positive) | Same rule applied in the extra-files loop |
| `351cadf` | Per-file rule was applied **after** `rglob` walked through a root-symlinked-out scan dir → every match filtered to nothing → `targets` empty, `missing == []`, `main()` silently returned 0; the entire `auto/` package could be moved out of tree and stop being policed | This commit |
| `36588f60` | **APPROVED** | — |

## What `36588f60` changed (the final fix)
1. **Single `_resolve_inside_repo(path)` helper.** Replaces three duplicated try/except blocks. The architectural boundary (resolves inside REPO_ROOT) is now enforced uniformly across the `SCAN_DIRS` loop, the `SCAN_EXTRA_FILES` loop, and the anchor check.
2. **Root-level escape check.** If a `SCAN_DIRS` entry itself resolves outside the repo, the rglob walk is skipped entirely.
3. **Anchor escape detection.** `is_file()` follows symlink chains, so an anchor reachable only through a symlinked-out parent used to look healthy on disk while contributing zero scanned code. Anchors are now reported missing unless they (a) `is_file()` *and* (b) resolve inside the repo. The bot's exact root-symlink scenario now fails loud with the existing missing-anchor message, naming each escaped anchor — verified: `targets=[cli/commands/auto.py]`, `missing=[8 anchors]`, `main()=1`.
4. **Test honesty.** Removed the descendant-dir-symlink test that was vacuous on Python 3.13+ (where `rglob` doesn't recurse into descendant symlinks anyway, so the test passed without exercising the actual failure mode). Added two replacements that pin the real contract: `test_scan_root_symlink_to_outside_repo_fails_loud` and `test_anchor_behind_outside_repo_symlink_dir_is_missing`. Decoupled `test_scan_extra_file_symlink_to_outside_repo_is_skipped` from the anchor contract (separate non-anchor extra-file slot) so it isolates the SCAN_EXTRA_FILES skip semantics.

## Why this is merge-ready

- **Bot verdict:** `APPROVED` on `36588f60` — *"The patch tightens the boundary guard in the right place: it treats 'resolves inside the repo' as the contract, closes the root-symlink coverage escape, and adds targeted regression tests for both in-repo and out-of-repo symlink cases."*
- **CI:** Ruff Lint, MyPy, Bridge TypeScript, enforce-envelope, enforce-boundary, Test Python 3.12 / 3.13 / 3.14 — all `SUCCESS`.
- **Merge state:** `MERGEABLE`, `CLEAN` (no conflicts, no required reviews outstanding).
- **Test coverage symmetry:** every contract direction exercised — out-of-repo leaf symlink (skipped), out-of-repo dir-symlink at scan root (fails loud), in-repo symlink with clean target (still scanned), in-repo symlink with dirty target (caught), SCAN_EXTRA_FILES out-of-repo (skipped), SCAN_EXTRA_FILES in-repo dirty (caught), anchor behind out-of-repo dir (missing).
- **Behavior on real repo:** `python3 scripts/check-auto-boundary.py` → `OK (18 files scanned, 0 findings)` — production scan unchanged.
- **No surface area expansion:** SCAN_DIRS, SCAN_EXTRA_FILES, ANCHOR_FILES, FORBIDDEN_PATTERNS, and the allowlist contract are all unchanged. The PR is purely a hardening of `_resolve_scan_targets()` plus tests.

Ready for your re-review and merge whenever convenient.

PR #799, opened 2026-05-08
fix(hook): structural mcpServers check in is_mcp_configured

Q00:
## Summary
- ``is_mcp_configured`` did ``\"ouroboros\" in mcp.read_text()`` — substring match that returned True for any unrelated mention (description fields, path strings, comments).
- Setup gate then skipped, sending the user into MCP-required skills that hadn't actually been registered.
- Fix: parse JSON, look for a real key under ``mcpServers`` equal to ``ouroboros`` or starting with ``ouroboros-``.

## Reproduction
```bash
echo '{"mcpServers": {"unrelated": {"description": "compatible with ouroboros workflows"}}}' > ~/.claude/mcp.json
echo "ooo run x" | python scripts/keyword-detector.py
# Before: emits /ouroboros:run (gate skipped — false positive)
# After:  emits /ouroboros:setup
```

## Test plan
- [x] 6 new regressions: substring/description/path false positives, malformed JSON, real-key match, alias-key match
- [x] all 39 existing keyword-detector tests pass

### DIFF
--- a/scripts/keyword-detector.py
+++ b/scripts/keyword-detector.py
@@ -156,14 +156,36 @@ KEYWORD_MAP = [
 
 
 def is_mcp_configured() -> bool:
-    """Check if MCP server is registered in ~/.claude/mcp.json."""
+    """Check if the Ouroboros MCP server is registered in ~/.claude/mcp.json.
+
+    Earlier this did ``"ouroboros" in mcp_path.read_text()`` — a substring
+    match that returns True whenever the file mentions ``ouroboros``
+    anywhere, even inside a comment, an unrelated server's ``description``
+    field (``"compatible with ouroboros workflows"``), or a path string.
+    The setup gate then silently skipped and routed the user into an
+    MCP-required skill that wasn't actually registered.
+
+    Now we parse the JSON and check that ``ouroboros`` (or any
+    ``ouroboros``-prefixed alias such as ``ouroboros-dev``) is a real key
+    under the ``mcpServers`` object — the structural location Claude
+    Code's MCP loader actually reads.
+    """
     try:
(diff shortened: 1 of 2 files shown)


Q00:
Converting to draft pending verification of an unverified assumption.

## What needs verifying before merge

This PR replaces a substring match with a structural lookup at `data["mcpServers"][<key>]`. **I assumed `mcpServers` is the structural location Claude Code's MCP loader actually reads, but I did not verify this against Claude Code's documentation or source.**

If the actual config shape is different (e.g. `data["servers"]` array, top-level keys, a different file path, or a per-project shape under `~/.claude/projects/<hash>/mcp.json`), this PR breaks the gate entirely:

```python
servers = data.get("mcpServers")   # → None on a different shape
if not isinstance(servers, dict):
    return False                    # → every user redirected to /ouroboros:setup
```

Every existing user with a working MCP setup would be told to run setup again, indefinitely. That is a much bigger regression than the false-positive substring match it fixes.

## Concrete checks to perform

1. Run `claude --help` / `claude mcp --help` and confirm the canonical config shape.
2. Read a real `~/.claude/mcp.json` from a working install (or claude.com docs) and verify the `mcpServers` key path.
3. Check whether `~/.claude/settings.json` (different file) is also a valid location — in which case the path itself needs to be widened.
4. Add a test pinning the actual on-disk shape, not just the assumed shape.

Holding as draft until those land.

Q00:
Closing — multi-persona lateral analysis converged on "the affected population is approximately zero and every fix introduces a larger regression class". Tracking the architectural follow-up (Ouroboros owns its own setup-state, not Claude's config files) as a separate issue.

PR #801, opened 2026-05-08
fix(plugin): cap argv length in firewall audit events

Q00:
## Summary
- ``_event_envelope`` copied ``argv`` verbatim into every emitted event with no length bound. ``ledger_adapter.wrap_plugin_event`` then deep-copies it again. A megabyte argv element multiplies into 6× memory and 3× ledger writes.
- Schema header documents argv as bounded; the absence of an actual cap is a local DoS / log-bloat surface.
- Fix: per-element 4 KiB cap and cumulative 16 KiB cap, with ``<truncated:N>`` sentinels recording the elided size so audits can detect the truncation.

## Test plan
- [x] new ``test_argv_per_element_byte_cap_truncates_into_sentinel``
- [x] new ``test_argv_total_byte_cap_truncates_remaining_tail``
- [x] all 19 existing firewall tests pass

### DIFF
--- a/src/ouroboros/plugin/firewall.py
+++ b/src/ouroboros/plugin/firewall.py
@@ -49,6 +49,54 @@ EventSink = Callable[[dict], None]
 ConfirmFn = Callable[[str], bool]
 
 
+# Per-element and total argv length caps for audit events. The schema
+# header documents argv as bounded; without explicit caps a megabyte-
+# scale argv is copied into every event the firewall emits (invoked,
+# permission_used, completed) and then deep-copied again by
+# ``ledger_adapter.wrap_plugin_event``. That is a local DoS / log-bloat
+# vector with no upside — operators do not benefit from megabyte-long
+# argv records in the audit log. The caps below are conservative for
+# real CLI usage (no realistic plugin command needs 4 KB in a single
+# argv element, and no realistic invocation needs more than 16 KB
+# total). When the caps trip we replace the offending bytes with the
+# ``<truncated:N>`` sentinel so consumers can detect that truncation
+# happened without having to re-derive lengths from the log.
+ARGV_ELEMENT_BYTE_LIMIT = 4 * 1024
+ARGV_TOTAL_BYTE_LIMIT = 16 * 1024
+
+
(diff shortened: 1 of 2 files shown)


Q00:
Converting to draft pending validation of the cap values.

## What needs verifying before merge

The caps `ARGV_ELEMENT_BYTE_LIMIT = 4 KiB` and `ARGV_TOTAL_BYTE_LIMIT = 16 KiB` are **values I picked**, not values I measured against actual plugin invocations. Risk: a normal real-world invocation gets its argv folded into `<truncated:N>` sentinels, hiding the actual command line from the audit log when it most matters.

Concrete failure modes I have not ruled out:

- A plugin command that takes a JSON blob on argv (~10 KiB+) for a code-review or PR diff — bounded but realistic. Total cap eats it.
- A plugin invoked with a base64-encoded artifact ID (~2 KiB element) plus a few flags — single element under per-element cap, but two such elements blow the total cap.
- Long file paths on monorepos where users routinely paste 200+ char paths.

The audit log's value is being able to reconstruct what happened. A cap that silently elides realistic argv content makes the log less useful precisely when an operator needs it.

## What I should do before merge

1. Sample real `plugin.invoked` argv from any Ouroboros user who has the plugin firewall enabled (or from the integration test fixtures), measure the p99 element-size and total-size distribution.
2. Set caps at p99 + headroom (e.g. p99 × 2) so normal usage is never truncated and only pathological argv hits the limiter.
3. Or: keep the per-event payload bound, but log the full argv to a separate sized-rotated stream (debug log) so audit completeness is preserved.
4. Add a test that asserts realistic plugin invocation argvs (paths, JSON blobs, command flags) survive verbatim through the bound.

Holding as draft until measurements support the cap values.

Q00:
Closing — the truncate-with-sentinel approach hides exactly the data audit logs exist to provide, and the 4KiB/16KiB caps are arbitrary without measurement. Replacing with an observation-only PR that adds argv metadata (sha256, byte-length, count) alongside the existing argv field, so we can measure the real distribution before deciding any cap or spill.

PR #802, opened 2026-05-08
fix(providers): pop CLAUDECODE from os.environ for SDK call, not mask

Q00:
## Summary
- ``env_overrides[\"CLAUDECODE\"] = \"\"`` masked the value but kept the key. SDK env build is ``{**os.environ, ..., **options.env, ...}`` — overlay merges values, cannot remove a key.
- Nested-session checks that use key presence (``in env`` / ``${CLAUDECODE+x}``) still fire, producing the empty-response failure mode the original strip was meant to prevent.
- Fix: pop ``CLAUDECODE`` from ``os.environ`` for the SDK call duration; restore in ``finally`` (and ``CancelledError`` path).

## Reproduction
```python
import os
os.environ[\"CLAUDECODE\"] = \"1\"
overlay = {\"CLAUDECODE\": \"\"}
merged = {**os.environ, **overlay}
print(\"CLAUDECODE\" in merged)   # True — key still present
```
Bundled CLI's nested-session detection observes the key and returns empty.

## Test plan
- [x] new ``test_claudecode_var_is_popped_for_subprocess_call`` captures ``os.environ`` at SDK ``query`` time
- [x] all 45 adapter tests pass

### DIFF
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -619,8 +619,22 @@ class ClaudeCodeAdapter:
         # The bundled CLI refuses to start when CLAUDECODE is set (nested
         # session check).  The Agent SDK sets CLAUDE_CODE_ENTRYPOINT=sdk-py
         # to signal it's an SDK call, but the older check on CLAUDECODE fires
-        # first, causing silent empty responses.  Strip it via env override.
-        claudecode_present = bool(os.environ.get("CLAUDECODE"))
+        # first, causing silent empty responses.
+        #
+        # Earlier this set ``env_overrides["CLAUDECODE"] = ""`` to mask the
+        # value, but the SDK builds the subprocess env via
+        # ``{**os.environ, ..., **options.env, ...}``
+        # (claude_agent_sdk/_internal/transport/subprocess_cli.py around
+        # line 348). An overlay value of ``""`` keeps the KEY present in the
+        # final env dict — so any nested-session check that uses key
+        # presence (``"CLAUDECODE" in env`` / ``${CLAUDECODE+x}``) rather
+        # than value truthiness still fires. To make the unset robust
+        # regardless of which check the CLI performs, we temporarily pop
+        # the var from ``os.environ`` while the SDK overlays — that
+        # removes it from the final subprocess env entirely. The pop is
+        # restored in a ``try/finally`` after the awaitable completes so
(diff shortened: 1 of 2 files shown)


Q00:
Closing this PR — the premise that motivated it is unverified, and the fix introduces real new risk.

## Why I'm closing

**1. The "nested-session refusal on CLAUDECODE" check does not appear to exist in the current bundled CLI.**

I disassembled `~/.local/bin/claude` (the active bundled CLI) and searched every reference to `CLAUDECODE`. The only patterns I find are:

- A provider-env capture function (`O$_()`) that explicitly value-truthy-checks: `if (q !== void 0 && q !== "") H.push(...)`. Empty string is treated the same as undefined here.
- A list `_$3 = [..., "CLAUDECODE", ...]` of env keys to forward.
- Subprocess spawn calls that **set** `CLAUDECODE=1` on outbound child processes.

There is no `if (process.env.CLAUDECODE) refuse()` / nested-session abort path. The existing `env_overrides["CLAUDECODE"] = ""` masking was either solving a different (older) CLI's behavior or was never load-bearing.

**2. The actual root cause of the empty-response bug was max_turns, not CLAUDECODE.**

Project memory note: *"Root cause 2 (REAL): max_turns=3 too low → CLI uses Read/Glob internally (2-4 turns) → exhausts turns before text output → empty response."* That fix landed separately. The CLAUDECODE strip was a parallel guess that never had a confirmed reproducer against the current CLI.

**3. The proposed fix introduces a real regression: process-global env mutation visible to sibling code.**

`os.environ.pop("CLAUDECODE")` for the duration of an SDK call:
- Briefly hides the var from any other code in the same process (logger branches, hooks, plugins) that reads `os.environ`.
- Is not safe across threads — `os.environ` mutation in CPython has no internal lock; concurrent `pop`/`restore` from `anyio.to_thread` or `ThreadPoolExecutor` will race and may leave the var permanently missing.
- Trades a theoretical-and-unverified attack on key-presence for a concrete-and-verifiable race surface.

## What I'd want before reopening

- A reproduction against `~/.local/bin/claude` (or the SDK-bundled CLI) showing that `CLAUDECODE=""` (key present, empty value) actually causes the empty-response failure mode. Without that, the strip is unnecessary.
- If reproduced, a fix that doesn't touch process-global state — e.g. an upstream SDK feature for "remove this key from the merged subprocess env."

Marking as not-merging.

PR #803, opened 2026-05-08
fix(providers): close catalog-drift gap for empty allowed_tools envelope

Q00:
## Summary
- ``disallowed = all_tools - allowed_tools`` against a hardcoded SDK-tool snapshot leaves new SDK built-ins permitted on the strict ``allowed_tools=[]`` interview path.
- Two-prong fix: ``allowed_tools=[]`` → ``disallowed`` includes ``\"*\"`` wildcard sentinel + the static list. Non-empty allowlists now also fold ``dangerous_tools`` into ``disallowed`` so future ``dangerous_tools`` additions propagate without a second edit.

## Test plan
- [x] regression on ``test_explicit_empty_allowed_tools_blocks_all_sdk_tools`` asserting ``\"*\"`` in disallowed
- [x] all 44 adapter tests pass

### DIFF
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -596,6 +596,36 @@ class ClaudeCodeAdapter:
 
         # If allowed_tools is explicitly set (even empty []), compute disallowed
         # from it (strict mode). None = permissive (only block dangerous).
+        #
+        # The hardcoded ``all_tools`` list below is a snapshot of the
+        # SDK's known built-in tools at the time this adapter was
+        # written. When the SDK adds a new built-in (a hypothetical
+        # future ``ShellSession``, ``BrowserControl``, ``RunCommand``,
+        # …), it does NOT appear in ``all_tools`` and therefore does
+        # not appear in ``disallowed`` — so the new tool stays
+        # permitted by the SDK's permissive default even when the
+        # caller explicitly requested ``allowed_tools=[]`` for a
+        # strictly read-only envelope (e.g. the interview path).
+        # That is a silent policy gap that widens with every SDK
+        # release.
+        #
+        # Two-prong fix:
+        #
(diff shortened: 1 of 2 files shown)


Q00:
Closing this PR — the wildcard sentinel does not actually do what the change claims.

## Why I'm closing

**The bundled CLI does not interpret `"*"` as a wildcard in `--disallowedTools`.**

I traced the path the SDK takes:

```
src/ouroboros/providers/claude_code_adapter.py
  → ClaudeAgentOptions(disallowed_tools=[...])
  → claude_agent_sdk/_internal/transport/subprocess_cli.py:204-205
      cmd.extend(["--disallowedTools", ",".join(self._options.disallowed_tools)])
  → /Users/jaegyu.lee/.local/bin/claude --disallowedTools "*,Read,Write,..."
```

The CLI receives `"*"` as a literal tool name in a comma-separated list. There is no documented or observable wildcard semantic — it is matched by exact tool-name string compare against the SDK's tool catalog. A tool literally named `*` does not exist, so the sentinel does nothing useful: a future SDK built-in not in the static `all_tools` list will still be permitted. The change ships a misleading defense-in-depth promise without the actual depth.

## The real concern is still real

The hardcoded `all_tools` list at `claude_code_adapter.py:600-615` does drift from the SDK's evolving built-in catalog, and `allowed_tools=[]` plus an incomplete `disallowed_tools` set could permit a new built-in. But the right fix is one of:

1. **Use only `allowed_tools`** as the policy surface, and trust the SDK to enforce it (the field exists for exactly this reason — `allowed_tools=[]` should mean "no tools," authoritatively, regardless of disallowed list).
2. **Auto-discover the SDK's tool catalog** at runtime (introspect `claude_agent_sdk` for tool names) instead of hardcoding the list.
3. **Pin and version-test against the SDK** so a new built-in surfaces in CI before reaching prod.

Each of those is a different patch with a different reviewer. This PR's `"*"` shortcut is not any of them.

Marking as not-merging.

PR #804, opened 2026-05-08
fix(auto): exact-match the canonical key in safe-default rollback

Q00:
## Summary
- ``_revert_safe_default_entries`` used ``entry.key.endswith(\".safe_default_finalization\")``. The only producer of that suffix writes the exact key ``{section}.safe_default_finalization``, but the suffix match would also delete unrelated entries (e.g. an answerer-authored ``constraints.my.safe_default_finalization``).
- Fix: switch to exact equality ``entry.key == f\"{section_name}.safe_default_finalization\"``.

## Test plan
- [x] new ``test_revert_safe_default_entries_preserves_user_keys_with_matching_suffix``
- [x] all 128 interview-pipeline tests pass

### DIFF
--- a/src/ouroboros/auto/interview_driver.py
+++ b/src/ouroboros/auto/interview_driver.py
@@ -646,11 +646,20 @@ def _revert_safe_default_entries(
         section = ledger.sections.get(section_name)
         if section is None:
             continue
-        section.entries = [
-            entry
-            for entry in section.entries
-            if not entry.key.endswith(".safe_default_finalization")
-        ]
+        # Match the EXACT key the safe-default policy writes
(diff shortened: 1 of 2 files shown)


Q00:
Re-labeling as **defensive-cleanup**, not a user-impacting bug fix.

To be fully honest: I did not find an actual ledger entry whose key ends with ``.safe_default_finalization`` outside the safe-default policy itself. ``finalize_safe_defaultable_gaps`` is the only writer of that suffix in the current codebase, so the ``endswith`` form was operationally equivalent to ``==`` for every input that exists today.

The change is correct and the test pins the invariant for the future, but if you would rather hold off until a second writer of that suffix actually appears, that is a defensible call too.

PR #807, opened 2026-05-08
feat(plugin): TrustStore concurrency primitives + LockEntry subject helper + manifest tuple ordering

shaun0927:
## Summary

Foundation work for the mutating ``ooo plugin {disable,trust,...}`` commands. Replaces the closed #751 with a focused, reviewable slice:

- **TrustStore concurrency** — every revocation entry point (``remove``, ``write_disable``, ``clear_disable``) now runs inside the per-plugin ``_grant_lock``. Adds two atomic primitives the CLI needs: ``apply_disable`` (disable record + trust file removal in one critical section) and ``grant_and_clear_disable`` (grant + disable clear in one critical section). ``wipe_subject`` now does both deletions inside a single critical section via lock-free internal bodies (``_remove_locked``, ``_write_disable_locked``, ``_clear_disable_locked``) — re-entering ``_grant_lock`` from a holder would deadlock under POSIX flock semantics.
- **Pre-RFC trust record handling** — ``_subject_matches`` treats blank subject columns on the existing record as a hard mismatch when the caller plumbs the corresponding column. Without this, a single new scope can silently rehydrate every legacy scope under the new subject.
- **Disable surface name validation** — every disable API (``is_disabled``, ``is_disabled_for_subject``, ``read_disable``, ``write_disable``, ``clear_disable``, ``apply_disable``) calls ``_validate_plugin_name``. Lockfile rows are operator-editable and ``Lockfile.read()`` does not enforce the name regex, so a malformed name like ``../../x`` would otherwise escape ``DEFAULT_TRUST_ROOT``.
- **``LockEntry.subject_for_disable()``** — canonical ``(source_type, source_identity)`` fallback. Both the disable write side and every read side (dispatcher, ``inspect``, ``list``) MUST agree on the same fallback for legacy lockfile rows or a ``disable`` write lands under one subject and the dispatcher's check runs against another.
- **Manifest tuple ordering** — ``PluginManifest.{capabilities,permissions}`` are tuples again so iteration order matches the manifest's declaration. ``frozenset`` reintroduces nondeterministic order in CLI output and emitted ``plugin.permission_used`` events.

No CLI surface change. Pure infrastructure.

## Test plan

- [x] ``test_disable_apis_validate_plugin_name`` — every disable API rejects malformed identifiers
- [x] ``test_apply_disable_is_atomic`` — post-condition observable in one step
- [x] ``test_grant_and_clear_disable_is_atomic`` — same for the trust side
- [x] ``test_grant_resets_legacy_unbound_record_on_subject_bound_grant`` — prevents legacy-record rehydration
- [x] ``test_revocation_serializes_with_grant`` — concurrent disable + remove + grant produces no scope inheritance
- [x] ``test_wipe_subject_atomic_against_concurrent_grant`` — wipe + grant cannot leave the ``trusted, enabled`` state
- [x] ``test_subject_for_disable_*`` — populated columns, legacy local rows, legacy git rows
- [x] All 1163 plugin + CLI tests pass; lint and format clean

## Context

This replaces #751 (closed because the diff had grown too large for the review bot to converge on a clean approve). The full iteration is preserved on ``backup/pr-751-full-iteration``. The next companion PR (firewall fail-closed hardening) and the original PR scope (CLI mutating commands) will follow as separate PRs that stack on this once it merges.

### DIFF
--- a/src/ouroboros/plugin/lockfile.py
+++ b/src/ouroboros/plugin/lockfile.py
@@ -89,6 +89,30 @@ class LockEntry:
             lines.append(f"artifact_digest = {_toml_str(self.artifact_digest)}")
         return lines
 
+    def subject_for_disable(self) -> tuple[str, str]:
+        """Canonical ``(source_type, source_identity)`` for disable lookup.
+
+        Older lockfile rows that pre-date the trust-subject contract have
+        empty ``source_type`` / ``source_identity`` columns. Both the
+        disable write path (``ooo plugin disable``) and every read path
+        (dispatcher, ``inspect``, ``list``) MUST agree on the same
+        fallback for those rows; otherwise a ``disable`` write lands
+        under one subject and the dispatcher's ``is_disabled_for_subject``
+        check runs against another, leaving the user with a "disabled"
+        plugin that ``ooo <name> ...`` still happily invokes.
+
+        Fallback rules:
+          - ``source_type``: stored value if set; else ``"plugin_home"``
+            for git installs, ``"local_path"`` otherwise.
+          - ``source_identity``: stored value if set; else the git
+            ``repository`` URL or the local ``plugin_home`` path.
+        """
+        source_type = self.source_type or (
+            "plugin_home" if self.source_kind == "git" else "local_path"
+        )
+        source_identity = self.source_identity or (self.repository or self.plugin_home)
+        return source_type, source_identity
+
 
 _TOML_BASIC_ESCAPES: dict[str, str] = {
     "\\": "\\\\",
--- a/src/ouroboros/plugin/manifest.py
+++ b/src/ouroboros/plugin/manifest.py
@@ -159,8 +159,12 @@ class PluginManifest:
     version: str
     source: SourceSpec
     commands: tuple[CommandSpec, ...]
-    capabilities: frozenset[Capability]
-    permissions: frozenset[Permission]
+    # Capabilities and permissions are stored as ordered tuples so iteration
+    # order matches the manifest's declaration. ``frozenset`` would
+    # reintroduce nondeterministic order in CLI output and emitted
+    # ``plugin.permission_used`` events for multi-entry manifests.
+    capabilities: tuple[Capability, ...]
+    permissions: tuple[Permission, ...]
     entrypoint: Entrypoint
     description: str = ""
     audit: AuditSpec = field(default_factory=AuditSpec.standard_four_events)
@@ -441,11 +445,11 @@ def load_manifest(path: str | Path) -> PluginManifest:
     )
 
     commands = tuple(_build_command(c) for c in raw["commands"])
-    capabilities = frozenset(
+    capabilities = tuple(
         Capability(name=c["name"], access=c["access"], reason=c.get("reason", ""))
         for c in raw["capabilities"]
     )
-    permissions = frozenset(
+    permissions = tuple(
         Permission(
             scope=p["scope"],
             risk=p["risk"],
--- a/src/ouroboros/plugin/trust_store.py
+++ b/src/ouroboros/plugin/trust_store.py
@@ -353,6 +353,76 @@ class TrustStore:
                 artifact_digest=artifact_digest,
             )
 
+    def grant_and_clear_disable(
+        self,
+        *,
+        plugin: str,
+        version: str,
+        scope: str,
+        granted_by: str,
+        source_type: str = "",
+        source_identity: str = "",
+        artifact_digest: str = "",
+        when: datetime | None = None,
+    ) -> TrustRecord:
+        """Atomic ``ooo plugin trust`` step: grant ``scope`` AND clear
+        any existing disable record inside a single per-plugin
+        critical section.
+
+        Calling ``grant()`` then ``clear_disable()`` from the CLI takes
+        and releases the lock twice. A concurrent ``ooo plugin
+        disable`` (which uses the atomic ``apply_disable``) could run
+        between them: grant writes ``trust.json``, disable atomically
+        writes ``disabled.json`` and removes ``trust.json``, then
+        trust's late ``clear_disable`` removes the freshly written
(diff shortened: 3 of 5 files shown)


shaun0927:
@ouroboros-agent please review. This is the first of three PRs that decompose the closed #751 into smaller, focused units. This PR is foundation work (TrustStore concurrency, LockEntry subject helper, manifest tuple ordering) with no CLI surface change. The companion PR-B (#808) for firewall fail-closed hardening is independent of this one — both can be reviewed in parallel.

shaun0927:
@Q00 ready for re-review and merge.

## What this PR is

Foundation work split out of the closed #751. Pure infrastructure — no CLI surface change. Three internal contracts the mutating ooo plugin {disable,trust,...} commands need:

1. **TrustStore concurrency** — every revocation entry point (remove / write_disable / clear_disable) now runs inside the per-plugin _grant_lock. Two atomic primitives (apply_disable, grant_and_clear_disable) let the CLI's disable-and-trust transactions land in a single critical section instead of split across two lock acquisitions where a concurrent grant/clear could interleave. Lock-free internal bodies (_remove_locked, _write_disable_locked, _clear_disable_locked) let wipe_subject reuse the primitives without re-entering flock (which Linux/macOS treat as separate locks per FD and would deadlock).

2. **Pre-RFC trust-record fail-closed** — _subject_matches now treats blank source_type / source_identity / artifact_digest columns on the existing record as a hard mismatch when the caller plumbs the corresponding column on the grant. Without this, a single new scope silently rehydrated every legacy scope under the new install subject — defeating the trust-subject binding.

3. **Disable surface name validation + LockEntry.subject_for_disable()** — every disable API now calls _validate_plugin_name, closing the path-traversal seam where a malformed lockfile name like ../../x could escape DEFAULT_TRUST_ROOT. LockEntry.subject_for_disable() is the canonical fallback both write-side and read-side must use for legacy lockfile rows so a disable write and the dispatcher's lookup key the same subject.

## Why merge is safe

- **Bot verdict: APPROVE** on first review (commit a22a4e22, no blocking findings, no non-blocking suggestions).
- **CI: 8/8 green** — Ruff Lint, MyPy, Bridge TypeScript, Tests on Python 3.12/3.13/3.14, enforce-boundary, enforce-envelope.
- **Tests added** — 9 new regression cases targeting exactly the contract edges the redesign enforces:
  - test_disable_apis_validate_plugin_name
  - test_apply_disable_is_atomic, test_grant_and_clear_disable_is_atomic
  - test_grant_resets_legacy_unbound_record_on_subject_bound_grant
  - test_revocation_serializes_with_grant, test_wipe_subject_atomic_against_concurrent_grant
  - test_subject_for_disable_uses_populated_columns_when_set, test_subject_for_disable_falls_back_for_legacy_local_row, test_subject_for_disable_falls_back_for_legacy_git_row
- **Surface area: 5 files, +566 / −15 LoC**, kept entirely inside src/ouroboros/plugin/ and tests/unit/plugin/. No CLI command, no public API rename, no schema/migration.
- **Backwards compatibility** — every loosening test fixture or call site that doesn't plumb the install subject keeps the legacy version-only path. Production CLI install paths always pass the full triple, so prod records become bound on every new grant; legacy records do NOT auto-upgrade.

The companion PR #808 (firewall fail-closed hardening) reviews independently and is also ready. Once both land, the mutating CLI commands (the original #751 scope) follow as PR-C.

shaun0927:
@Q00 ready for re-review and merge.

I independently re-reviewed PR #807 at current head `a22a4e2` and I think the PR is a valid foundation slice, not an incomplete behavioral change.

## Scope / intent verified

This PR decomposes the closed larger plugin-mutation work into internal contracts only:

- `TrustStore` now serializes revocation/trust transitions through the same per-plugin lock and adds atomic primitives for the later CLI integration (`apply_disable`, `grant_and_clear_disable`).
- Legacy/pre-RFC trust records now fail closed when a new subject-bound grant supplies populated subject fields, avoiding silent cross-subject scope carry-over.
- `LockEntry.subject_for_disable()` defines the canonical legacy fallback that later write/read paths can consume consistently.
- Manifest capabilities/permissions preserve declaration order by using tuples instead of `frozenset`, which matches deterministic CLI/event ordering expectations.

The bot's design note is correct: CLI/read-path call sites still need to switch to these primitives in the follow-up PR. I do not see that as a blocker for this PR because this PR intentionally limits itself to lower-level primitives and regression coverage; it does not partially rewrite the CLI path in a way that would create a mixed/unsafe behavior boundary.

## Review / validation evidence

- Bot review: `ouroboros-agent` **APPROVED** the current head (`a22a4e2`) with no blocking findings and no non-blocking suggestions.
- GitHub merge state: `MERGEABLE`, `CLEAN`, non-draft, review decision `APPROVED`.
- GitHub checks: all 8 required checks are green (`Ruff Lint`, `MyPy Type Check`, `Bridge TypeScript`, `Test Python 3.12/3.13/3.14`, `enforce-boundary`, `enforce-envelope`).
- Local focused validation on the checked-out PR head passed:
  - `PYTHONPATH=src python3 -m pytest tests/unit/plugin/test_lockfile.py tests/unit/plugin/test_trust_store.py -q` → `74 passed`
  - `PYTHONPATH=src python3 -m ruff check ...` → passed
  - `PYTHONPATH=src python3 -m ruff format --check ...` → passed
  - `git diff --check f9c54ed6...HEAD` → passed

## Why I think it is safe to merge

The locking changes are internally consistent: public operations validate plugin names before path derivation, nested atomic operations use lock-free internal helpers to avoid re-entering `flock`, and the new tests exercise atomic disable/trust/wipe races. The legacy-subject change is fail-closed only when callers provide the new subject columns, while legacy callers that omit those columns retain version-only behavior. The tuple-ordering change is localized to manifest representation and aligns with deterministic output/event ordering.

I did not push any remediation commits because I did not find a current-head blocker. The remaining CLI/read-path consumption is accurately documented as follow-up work rather than required in this foundation PR.

PR #812, opened 2026-05-08
feat(unstuck): debate mode for ooo lateral

Q00:
## Summary
- `ooo lateral` (no args) now defaults to a **5-persona parallel debate** — each lateral persona (hacker/researcher/simplifier/architect/contrarian) runs as an independent sub-agent and the user picks the verdict.
- Solo mode (`ooo lateral <persona>`) preserved with zero behavior change.
- SKILL.md-only change, **zero handler edits**. Uses the existing `_subagents` dispatch envelope that `ouroboros_lateral_think` already emits.
- Adds runtime-specific fan-out guidance: Claude Code Task tool, OpenCode plugin pane, Codex CLI sub-agent, or inline fallback — sub-agent fan-out is now visible to the user where the runtime supports it.
- Autonomous chain documented: a SKILL operating in another persona (e.g. `socratic-interviewer` during `ooo interview`) may invoke `ooo lateral` on its own self-assessment when a question needs multi-perspective deliberation.
- Synthesis defers to the user — main session presents options + disagreements + a labeled recommendation; never auto-emits a verdict or auto-progresses.

## Argument grammar (added)

| Input | Mode |
|---|---|
| `ooo lateral` | debate (5 personas) |
| `ooo lateral <persona>` | solo |
| `ooo lateral debate <p1> <p2> ...` | debate (explicit members) |
| `ooo lateral @all` | debate preset (only preset in Phase 1) |
| invalid persona / preset / multi-token without `debate` keyword | reject with guidance |

## Out of scope (tracked as follow-up issues)
1. Symposium primitive RFC (naming, unified envelope, cross-SKILL adoption)
2. Verdict envelope standardization
3. Stateful pool joining policy (evaluator/qa-judge/ontologist as debate members)
4. `evaluate` consensus migration from score-average to debate
5. Interview stagnation auto-trigger (code-level hook)
6. Streaming progress fallback for inline path (Layer 2)
7. Persistent transcript (Layer 3)

## Test plan
- [ ] `ooo lateral` (no args) → 5 sub-agents fan out in parallel, "Running 5 agents…" UI visible
- [ ] `ooo lateral hacker` → solo mode, regression-equivalent to prior behavior
- [ ] `ooo lateral debate architect contrarian` → exactly 2 sub-agents fan out
- [ ] `ooo lateral @all` → 5 sub-agents; `ooo lateral @unknown` → rejection with guidance
- [ ] `ooo lateral nobody` → rejection naming valid personas
- [ ] `ooo lateral hacker contrarian` (no `debate` keyword) → rejection suggesting `ooo lateral debate hacker contrarian`
- [ ] After debate Round 1, main session presents options + disagreements + recommendation; no auto-progression
- [ ] MCP unavailable: SKILL falls back to reading `src/ouroboros/agents/<persona>.md` and dispatches Task calls directly — functionally equivalent

### DIFF
--- a/skills/unstuck/SKILL.md
+++ b/skills/unstuck/SKILL.md
@@ -93,15 +93,26 @@ The handler picks one of two response shapes based on `should_dispatch_via_plugi
 | Solo (`persona=...`) | single `_subagent` envelope (one object) — `evaluation_handlers.py:1536-1563` | single `# Lateral Thinking: <approach>` block in `content` |
 | Debate (`personas=[...]`) | `_subagents` array (N objects) — `evaluation_handlers.py:1414+` | N blocks joined by `\n\n---\n\n` in `content`, **plus** an appended hidden dispatch block carrying the same canonical N payloads (see "Inline dispatch block" below) |
 
-**Inline dispatch block (debate, inline response only).** The handler appends a versioned, sentinel-bracketed JSON dispatch block to the end of `content` so that callers can recover the canonical structured payloads even though the FastMCP adapter drops `meta` on the wire (`src/ouroboros/mcp/server/adapter.py:923`, `src/ouroboros/mcp/tools/subagent.py:141-144`). Format:
+**Inline dispatch block (debate, inline response only).** The handler appends a versioned, sentinel-bracketed dispatch block to the end of `content` so that callers can recover the canonical structured payloads even though the FastMCP adapter drops `meta` on the wire (`src/ouroboros/mcp/server/adapter.py:923`, `src/ouroboros/mcp/tools/subagent.py:141-144`). Format:
 
 ```
-<!-- ouroboros-lateral-inline-dispatch-v1
-{"dispatch_mode": "inline_fallback", "persona_count": N, "payloads": [...]}
+<!-- ouroboros-lateral-inline-dispatch-v1 base64
+<base64-encoded-JSON>
 -->
 ```
 
-The block is a HTML comment, so markdown viewers ignore it. The sentinel `ouroboros-lateral-inline-dispatch-v1` is namespaced and versioned, so user-supplied `problem_context` cannot accidentally collide with it. To recover the dispatch struct, find the substring between `<!-- ouroboros-lateral-inline-dispatch-v1\n` and `\n-->` at the end of `content`, then `JSON.parse` the captured body.
+Two pieces matter:
+
+- **Hidden HTML comment** — markdown viewers render nothing for `<!-- ... -->`, so the block doesn't pollute the human-visible output.
+- **Base64 body** — base64's alphabet is `[A-Za-z0-9+/=]`, which can never produce the sequence `-->`. So even if a user-supplied `problem_context` or `current_approach` contains `-->` (HTML/JS debugging is the obvious case), the encoded body cannot prematurely close the wrapper and leak the dispatch into the visible markdown.
+
+Decoded, the body is JSON:
+
+```json
+{"dispatch_mode": "inline_fallback", "persona_count": N, "payloads": [...]}
+```
+
+To recover: locate the substring between `<!-- ouroboros-lateral-inline-dispatch-v1 base64\n` and `\n-->` at the end of `content`, base64-decode it, then `JSON.parse`. (See `tests/unit/mcp/tools/test_lateral_think_handler.py::_extract_inline_dispatch` for the canonical extraction helper.)
 
 #### Shape A — `dispatch_mode = "plugin"` (OpenCode plugin mode only)
 
@@ -138,7 +149,7 @@ Driving fan-out from this dispatch block — instead of from the joined human-di
 - **Separator collision** — `\n\n---\n\n` can legitimately appear inside a user-supplied `problem_context` or `current_approach`, so splitting the joined text would over-fragment and corrupt prompts. The dispatch block uses a unique versioned sentinel that user content cannot collide with.
 - **Behavioral drift across runtimes** — the dispatch block carries the same canonical payloads `_subagents` carries, so debate results don't diverge by environment.
 
-1. Locate the dispatch block at the end of the MCP response's `content` text — the substring between `<!-- ouroboros-lateral-inline-dispatch-v1\n` and `\n-->`. `JSON.parse` it to get `{dispatch_mode, persona_count, payloads}`. If the block is missing (older handler that pre-dates the v1 sentinel), fall through to the constrained-runtime path below — *do not* split the joined text.
+1. Locate the dispatch block at the end of the MCP response's `content` text — the substring between `<!-- ouroboros-lateral-inline-dispatch-v1 base64\n` and `\n-->`. Base64-decode the captured body, then `JSON.parse` to get `{dispatch_mode, persona_count, payloads}`. If the block is missing (older handler that pre-dates the v1 sentinel), fall through to the constrained-runtime path below — *do not* split the joined text.
 2. Surface a short "what the lateral toolkit suggests" header to the user, with the markdown above the dispatch block, so the MCP call is visible as a real product surface, not silent.
 3. In a **single message**, emit N parallel `Task` calls (`general-purpose` subagent) — one per entry in `payloads`. Each Task receives the payload's `prompt` verbatim plus the payload's `context` so the persona is grounded. Strict isolation per Task. The user sees "Running N agents…".
 4. Wait for all N to return.
@@ -212,9 +223,9 @@ with 2 tables, you haven't found the core feature yet.
 [ToolSearch loads ouroboros_lateral_think]
 [Call ouroboros_lateral_think(personas=[hacker,researcher,simplifier,architect,contrarian], ...)]
 [Handler returns: content = N persona blocks joined by ---,
-                  followed by hidden <!-- ouroboros-lateral-inline-dispatch-v1
-                  {payloads:[...]} --> sentinel block]
-[Extract dispatch_blob via the sentinel; JSON.parse → payloads[]]
+                  followed by hidden <!-- ouroboros-lateral-inline-dispatch-v1 base64
+                  <base64 body> --> sentinel block]
+[Extract via the sentinel; base64-decode then JSON.parse → payloads[]]
 [Single message: 5 parallel Task calls — one per payloads[] entry,
  each Task receives payload.prompt + payload.context verbatim;
  the visible markdown is shown to the user as the lateral scaffold]
--- a/skills/unstuck/SKILL.md
+++ b/skills/unstuck/SKILL.md
@@ -1,21 +1,28 @@
 ---
 name: unstuck
-description: "Break through stagnation with lateral thinking personas"
+description: "Break through stagnation with lateral thinking personas — single or multi-persona debate"
 ---
 
 # /ouroboros:unstuck
 
-Break through stagnation with lateral thinking personas.
+Break through stagnation with lateral thinking personas. Two modes:
+- **Solo** — one persona reframes the problem (fast, cheap).
+- **Debate** — multiple personas run in parallel as sub-agents and the user picks the verdict (visual, thorough).
 
 ## Usage
 
 ```
-/ouroboros:unstuck [persona]
+ooo lateral                       # debate (default) — all 5 lateral personas
+ooo lateral <persona>             # solo — single persona
+ooo lateral debate <p1> <p2> ...  # debate with explicit members
+ooo lateral @<preset>             # debate with preset (Phase 1: only @all = 5 personas)
 ```
 
-**Trigger keywords:** "I'm stuck", "think sideways"
+Trigger keywords: "I'm stuck", "think sideways", "ooo lateral", "/ouroboros:unstuck".
 
-## Available Personas
+## Personas (Lateral Pool)
+
+The lateral pool is **stateless mindset personas only** — five reframing lenses. Stateful roles (evaluator, qa-judge, ontologist, socratic-interviewer, etc.) are NOT mixed into this pool; they have their own SKILLs.
 
 | Persona | Style | When to Use |
 |---------|-------|-------------|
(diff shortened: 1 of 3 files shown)


Q00:
Thanks @ouroboros-agent — the BLOCKING finding was correct and the SKILL has been rewritten in `de286887`.

**What was wrong**
The previous Step 3 told Claude Code and Codex CLI to receive a `_subagents` envelope and fan it out manually, but the handler only emits that envelope when `should_dispatch_via_plugin(...)` is true (OpenCode plugin mode). In every other runtime the inline path returns one markdown blob, so the SKILL would have hung waiting for an envelope that never arrived.

**What changed**
Debate mode is now three explicit paths, picked **in Step 1** before any MCP call:

- **Path A — Direct sub-agent fan-out** (Claude Code, Codex CLI, any sub-agent-capable runtime). No MCP call. Read each persona file from `src/ouroboros/agents/<persona>.md` and emit N parallel `Task` calls in a single message. This is the default for the great majority of users.
- **Path B — MCP plugin dispatch** (OpenCode plugin only). Call `ouroboros_lateral_think(personas=[...])`; the plugin auto-fans-out via the `_subagents` envelope.
- **Path C — Inline** (constrained runtimes without sub-agent dispatch). MCP returns one markdown blob; present it directly. No per-persona visualization, no Round 2 cross-attack.

Step 1 selects the path so Step 2's MCP load is skipped entirely when Path A is chosen. Step 3 begins with a runtime-fact note that quotes the exact code paths (`subagent.py:186-218`, `evaluation_handlers.py:1414+`) so future readers can verify.

Solo mode and the autonomous-chain guidance are unchanged from the previous push.

CI was already green; no handler edits were introduced. Re-review please.

shaun0927:
@ouroboros-agent — both BLOCKING findings on `de286887` are addressed in `42ebeabb`. Single-file SKILL.md change, no handler edits, CI green (Lint / Tests 3.12-3.14 / MyPy / Bridge TS / enforce-envelope / enforce-boundary).

**1. Contract violation (Path A bypassing `ouroboros_lateral_think`)**

Removed. Step 2 is now "always call `ouroboros_lateral_think`" — solo *and* debate, every runtime, no exceptions. The previous "Path A skips MCP" branch is gone. The codex routing contract (`src/ouroboros/codex/ouroboros.md:21`) is now honored literally for every `ooo lateral` invocation.

**2. Fallback incoherence (`use Path A` deadend in Step 2 for Path B/C)**

Fixed by collapsing the path-selection-up-front logic. There is now exactly one authoritative entry path (call MCP), and Step 3 branches on the **handler's** response shape, not on a runtime guess made before the call:

- Shape A — `dispatch_mode = "plugin"` → envelope, plugin fans out.
- Shape B — inline response → split on `\n\n---\n\n`, fan out via Task on sub-agent-capable runtimes, or present concatenated text on constrained runtimes.

`ToolSearch` failure is now a hard stop with the same "MCP dispatch surface broken" rule the contract applies to `ooo auto`. No silent substitution.

**Why visual debate UX still works in Claude Code / Codex without bypassing MCP**

The handler's inline path is a deterministic prompt builder (`LateralThinker.generate_alternative` at `evaluation_handlers.py:1444+`) — no LLM rollout, just N reframing scaffolds. Step 3 splits the returned markdown on `\n\n---\n\n` and uses the scaffolds as inputs to N parallel `Task` calls. The MCP call has already happened (Step 2), so the contract is satisfied; the Task fan-out is a presentation/runtime layer over the MCP result, not a substitute for it. Same headline UX ("Running 5 agents…"), without crossing layers.

Re-review please.

shaun0927:
@ouroboros-agent — BLOCKING on `42ebeabb` addressed in `c3740bce`. Single-file SKILL.md change, no handler edits.

Step 3 Shape A (plugin mode) was incorrectly homogeneous — it documented only the `_subagents` array. Now split by mode to match the handler:

- **Solo (plugin)** — single `_subagent` envelope from `build_subagent_result` (`evaluation_handlers.py:1554+`); plugin spawns one Task pane.
- **Debate (plugin)** — `_subagents` array from `build_multi_subagent_result` (`evaluation_handlers.py:1419+`); plugin spawns N Task panes.

Added a response-shape table at the top of Step 3 so the solo/debate × plugin/inline asymmetry is visible at a glance and an implementer can't accidentally branch on the wrong key.

Re-review please.

shaun0927:
@ouroboros-agent — both BLOCKING findings on `c3740bce` addressed in `5a8c3b2f`. Single-file SKILL.md change, no handler edits.

**1. Empty-context hard-fail before MCP** (line 65 finding)

New **Step 2 — "Gather required context before the MCP call"** runs before the MCP routing step. It:
- Reuses session state for `problem_context` / `current_approach` / `failed_attempts` when present.
- Asks the user one short combined question when either field is unrecoverable: *"what are you stuck on right now, and what have you already tried?"*
- Forbids fabricating either field — `current_approach="not specified"` is acceptable, made-up content is not.
- Applies to solo *and* debate. Step 3 (MCP call) now uses the context populated by Step 2, never bare empty strings.

**2. Unsafe split on `\n\n---\n\n`** (line 110 finding)

Removed entirely. The Step 4 debate fan-out is now driven by the **personas list parsed in Step 1**, which is data we fully control. Each Task is built from:
- The persona name + style line from the SKILL's own **Personas (Lateral Pool)** table.
- The `problem_context` / `current_approach` / `failed_attempts` gathered in Step 2.
- A short ≤200-word response instruction.

No parsing of the MCP response is required for fan-out, so a separator inside user content cannot over-fragment anything. The MCP response is still surfaced to the user as the canonical scaffold the system computed, so the routing contract is honored as a real product surface — not just a checkbox.

The constrained-runtime path (no Task surface) presents the concatenated inline text as-is, with no parsing needed.

Re-review please.

shaun0927:
@ouroboros-agent — BLOCKING on `5a8c3b2f` addressed in `65db80ce`. Two-file change this round: SKILL.md plus a small additive handler edit (necessary, see below).

**The finding**

The inline debate path was rebuilding per-persona prompts locally (persona name + style line + a fresh `≤200 words` instruction) instead of reusing the canonical `build_lateral_multi_subagent` prompts (which include the `plan` / `assumption` / `verdict` task wrapper). That made Claude Code / Codex / OpenCode-subprocess debate behavior diverge from OpenCode plugin mode by environment — a real runtime bug for the default path.

**The fix**

The handler's multi-persona inline fallback already had the canonical `payloads` in scope (built by `build_lateral_multi_subagent` for the plugin branch a few lines above). It just wasn't being surfaced to non-plugin runtimes. Two-line additive change:

```python
meta={
    "persona_count": len(sections),
    "dispatch_mode": "inline_fallback",
    "payloads": [p.to_dict() for p in payloads],   # NEW
}
```

Backward-compat preserved: the joined text under `content` is unchanged, and existing tests on `meta["dispatch_mode"]` / `meta["persona_count"]` still pass (121 unit tests green).

SKILL Step 4 now drives the inline-debate fan-out from `meta.payloads`, passing each `payload.prompt` and `payload.context` verbatim to a parallel `Task`. This:
- Eliminates the unsafe `\n\n---\n\n` split — `meta.payloads` is structured, no separator collision is possible.
- Keeps debate prompts byte-identical to plugin mode (same canonical builder), so debate results no longer diverge by environment.

**Why a handler edit was necessary, despite the original "SKILL.md-only" goal**

The previous three rounds of bot findings collectively foreclosed every pure-SKILL path:
- Bypassing MCP entirely (early Path A) → contract violation.
- Splitting the joined text → unsafe separator collision with user context.
- Rebuilding prompts locally → divergence from canonical plugin behavior.

The only remaining clean answer is to expose the structured payloads the handler *already builds* on the inline response too. Two-line, additive, no behavior change for existing consumers.

**Also done**

Step 2's combined context question reworded to be mode-neutral (the non-blocking suggestion from the prior review).

Re-review please.

shaun0927:
@ouroboros-agent — BLOCKING on `65db80ce` addressed in `e5873f04`. The transport-side concern was correct and important; thank you for catching it.

**The finding**

`meta.payloads` cannot reach MCP callers — the FastMCP adapter forwards only `tool_result.text_content` to the wire (`src/ouroboros/mcp/server/adapter.py:923`), and `meta` loss is already documented in `src/ouroboros/mcp/tools/subagent.py:141-144`. The SKILL's new debate path therefore could not actually run on Claude Code / Codex / OpenCode subprocess.

**The fix**

The structured payloads now ride inside `content` itself — the same pattern plugin mode already uses for `_subagents` (serialized JSON in the response text). To preserve the existing human-readable markdown view of inline debate (and not break the existing test asserting `text.count(\"\\n\\n---\\n\\n\")` on the joined sections), the dispatch struct is appended as a hidden HTML-comment block with a versioned sentinel:

\`\`\`
<persona blocks joined by --->


\`\`\`

- The HTML comment is invisible in any markdown viewer.
- The sentinel `ouroboros-lateral-inline-dispatch-v1` is namespaced and versioned, so user-supplied `problem_context` / `current_approach` cannot collide with it.
- The JSON body carries the same canonical payloads that `_subagents` carries in plugin mode (same `build_lateral_multi_subagent` builder).

SKILL Step 4 now:
1. Locates the substring between `<!-- ouroboros-lateral-inline-dispatch-v1\\n` and `\\n-->` in the response `content`.
2. `JSON.parse`s the captured body.
3. Dispatches one parallel `Task` per `payloads[i]`, using `payloads[i].prompt` and `payloads[i].context` verbatim — the same prompt text plugin mode dispatches.

Result: byte-identical canonical prompts across all runtimes, no separator collision possible, no transport-layer field used that the FastMCP adapter would drop.

**Tests** — 121 passed (lateral_think_handler, subagent, round5_fixes). The pre-existing assertion `text.count(\"\\n\\n---\\n\\n\") == 1` for 2 personas still holds: the appended sentinel block does not contain that separator. `meta.payloads` is retained for in-process Python callers but is no longer the source of truth — the dispatch block in `content` is.

Re-review please.

shaun0927:
@ouroboros-agent — both BLOCKING findings on `e5873f04` addressed in `0bdd932f`. Both points were on the money — thank you for the close read on the wire format.

**1. Escaping safety**

The dispatch JSON is now base64-encoded inside the HTML comment. Sentinel updated to `ouroboros-lateral-inline-dispatch-v1 base64`:

\`\`\`

\`\`\`

Because `-->` cannot appear inside a base64 alphabet, an adversarial `problem_context` like \`I'm debugging \`\`...\` cannot prematurely close the wrapper or leak the structured payload into the visible markdown. The constrained-runtime "present text as-is" path is now safe under arbitrary user input.

**2. Regression test for the wire-side contract**

Two new tests in `tests/unit/mcp/tools/test_lateral_think_handler.py` exercise the `payload.content[0].text` boundary directly:

- `test_inline_fallback_carries_dispatch_block_in_content` — extracts the sentinel block, base64-decodes, JSON-parses, and asserts `dispatch_mode`, `persona_count`, and the per-persona payloads (`title`, `prompt` containing the canonical "Task for you (subagent)" wrapper, `context.persona`/`context.problem_context`/`context.current_approach`) all survive transport.
- `test_inline_fallback_dispatch_survives_html_close_in_user_context` — adversarial test with `-->` and `<!-- -->` and IE-style conditional comments scattered through `problem_context` and `current_approach`. Asserts the comment region (after the opening sentinel) contains exactly one `-->` (its legitimate closer) and that the adversarial context round-trips byte-perfect through the base64 boundary.

The shared `_extract_inline_dispatch` helper mirrors what a SKILL implementer must do on the wire side, so an escaping or formatting regression in the handler now fails a real test rather than shipping green.

Tests: **123 passed** (up from 121).

SKILL.md updated to document the base64 sentinel format and reference the test helper as the canonical extraction reference.

Re-review please.

shaun0927:
@Q00 — re-review ping for PR #812 current head `0bdd932fd25f656ee6166118d4d5a916f470a878`.

I re-read the PR body, prior bot design notes/reviews/comments, GitHub checks, and the effective GitHub diff. The PR direction is valid: it fixes the `ooo lateral` debate-mode runtime contract by preserving the required MCP routing while making non-plugin runtimes recover the same canonical per-persona dispatch payloads that plugin mode receives. The final shape avoids the earlier blockers around bypassing MCP, splitting visible markdown on `---`, rebuilding prompts locally, losing `meta` across FastMCP transport, and unsafe raw JSON inside an HTML comment.

What changed in the final head:
- `src/ouroboros/mcp/tools/evaluation_handlers.py` appends a versioned hidden inline dispatch block to non-plugin debate content: `<!-- ouroboros-lateral-inline-dispatch-v1 base64 ... -->`.
- The block carries base64-encoded JSON containing `dispatch_mode`, `persona_count`, and canonical `payloads` from the same builder used for plugin `_subagents`, so prompt behavior stays consistent across runtimes.
- `tests/unit/mcp/tools/test_lateral_think_handler.py` adds wire-boundary regression coverage for extracting the dispatch block from `content[0].text`, validating payload shape, and proving user text containing `-->` cannot prematurely close the HTML comment.
- `skills/unstuck/SKILL.md` documents the same response shapes, base64 sentinel format, extraction flow, and safe fallback behavior.

Validation performed:
- GitHub PR state: mergeable `MERGEABLE`, merge state `CLEAN`, review decision `APPROVED`.
- Bot review on current head: `ouroboros-agent` APPROVED with no blocking findings or non-blocking suggestions remaining.
- GitHub checks on current head: Ruff Lint, MyPy Type Check, Python tests 3.12/3.13/3.14, Bridge TypeScript, enforce-envelope, and enforce-boundary all passed.
- Local validation on a clean worktree at the PR head:
  - `python3 -m ruff check src/ouroboros/mcp/tools/evaluation_handlers.py tests/unit/mcp/tools/test_lateral_think_handler.py skills/unstuck/SKILL.md` → passed.
  - `python3 -m ruff format --check src/ouroboros/mcp/tools/evaluation_handlers.py tests/unit/mcp/tools/test_lateral_think_handler.py` → passed.
  - `PYTHONPATH=src /opt/hermes/.venv/bin/python -m pytest tests/unit/mcp/tools/test_lateral_think_handler.py` → 12 passed.
- Patch application/whitespace sanity: the GitHub effective patch applies cleanly to base `f9c54ed66b21` and passes whitespace check.

No remediation commit was needed: I found no merge-blocking issue in the current head. Based on the approved bot verdict, green checks, clean mergeability, focused local validation, and the transport-safety tests covering the previously risky edge case, this PR looks safe to merge when you are ready.

PR #824, opened 2026-05-10
docs(interview): add Refine and Restate gates to SKILL.md

Q00:
## Summary
Three additions to `skills/interview/SKILL.md` Path A so the main session sends rich, structured answers to the interview MCP and gates the seed transition with a one-line goal restatement.

| Step | Change |
|---|---|
| **3** | Multi-section payload by default. Single-line answers permitted only for PATH 1a auto-confirms and short PATH 2 answers. Free-text answers are sent as Decision / Reasoning / Constraints / Out of scope / Codebase context. |
| **4 (new)** | **Refine before forwarding** — single AskUserQuestion confirms the structured payload preserves every reasoning, constraint, and scope element. The user can accept, add to a section, or rewrite. |
| **9 (new)** | **Restate gate (post-Acceptance)** — main session collapses agreed answers into a one-sentence goal and confirms before suggesting `ooo seed`. The one place where one-line compression is the goal. |

Path B (agent-mode fallback) inherits both gates by reference.

## Why
The interview MCP is a pure question generator with no tools. The text the main session sends is the only context driving the next question. Compressing answers to a label degraded both question quality and ambiguity scoring.

The Step 3 example payloads in the existing SKILL.md showed only single-line answers (e.g. `\"[from-user] Stripe Billing\"`), which set the implicit expectation that all answers should be a label. Combined with the 800-char cap (#823), this collapsed the dialectic into a label-confirmation loop.

## Why "Refine" is reinterpreted, not copied
The standalone `refine` skill collapses meaning candidates into one shared-meaning line. Applying that literally inside the interview would re-introduce the compression problem. The underlying principle of `refine` is "miss no nuance" — preserved by Step 4 via structure preservation rather than line consolidation. The one-line restatement happens once at Step 9, where compression is the explicit goal.

## RFC
See `docs/rfc/interview-hardening.md` Change 2 (introduced in #822).

## Recommended merge order
Depends on #823 (raised prompt caps): without raised caps, multi-section payloads would still be truncated at ~800 chars before reaching the LLM. Recommend: #822 → #823 → this.

## Test plan
- [x] Documentation rendered locally; step numbering 1–11 consistent in both Path A and Path B.
- [ ] Reviewer: confirm payload examples are clear and the Refine/Restate distinction is unambiguous.
- [ ] Reviewer: read the "Why Refine is reinterpreted" rationale and push back if Step 4's gate text feels redundant with Step 9.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #823, opened 2026-05-10
feat(interview): raise prompt budget caps for richer answers

Q00:
## Summary
- `_MAX_TOTAL_PROMPT_CHARS`:  4,800 → 16,000
- `_MAX_USER_RESPONSE_CHARS`:   800 →  4,000

## Why
The 800-char per-answer cap was a workaround for an Agent SDK CLI empty-response quirk, but was applied uniformly across all adapters and rounds. The result: multi-section structured answers were **stored intact** (security validator permits 10 KB) but **silently truncated to ~800 chars** when `_build_conversation_history` built the next-question prompt. Only the headline of each answer reached the LLM. Constraints, codebase context, and scope decisions were invisible to the question generator, collapsing the dialectic into a label-confirmation loop.

## Verification
End-to-end with `claude-sonnet-4-5` via `ClaudeCodeAdapter`:
- 1,223-char structured answer preserved through `record_response` (`full preserved: True`)
- Q2 references content from the answer's body (e.g. "monthly/annual recurring plans") that would have been past the 800-char cut
- 5-round stress test: budget trim guards the total via existing `_trim_messages_to_budget`

| | Q1 | Q2 |
|---|---|---|
| Old caps | label-confirmation only | label-confirmation only |
| New caps | 149 chars, "What recurring billing intervals…" | 358 chars, quotes "Stripe Billing", points out missing tier structure |

## Risk
The 800 cap protected against an Agent SDK CLI empty-response bug. We did not encounter regressions during testing on `claude-sonnet-4-5`. If a specific older adapter (likely `gemini-cli`) regresses, follow-up is per-adapter capability rather than rollback — see RFC §"Future Work".

## RFC
See `docs/rfc/interview-hardening.md` Change 1 (introduced in #822).

## Recommended merge order
This PR depends conceptually on #822 (adapter MCP isolation) but the file diffs do not conflict. Recommend merging in order: #822 → this → #(refine/restate gates).

## Test plan
- [x] Single-round LLM round trip with 1,223-char answer — full preservation, follow-up question quotes the body.
- [x] 5-round budget trim test — most recent rounds fully preserved, older rounds trimmed by total budget guard.
- [ ] Reviewer: confirm no `gemini-cli` adapter regression if that adapter is tested.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #827, opened 2026-05-11
fix(interview): close Restate gate bypass for short PATH 2 answers

shaun0927:
## Summary

Follow-up to #824. Closes one of the three follow-ups flagged on the final APPROVE review at commit `b783fdd`.

Adds an explicit carve-out so the **Step 9 Restate correction** flow never inherits the **Step 4 short-PATH-2 Refine skip** rule.

| Step | Change |
|---|---|
| **4** | New "Exception — Restate corrections never skip Refine, regardless of length" block under the skip list, naming the exact failure mode (`[from-user]` instead of `[from-user][refined]`, no `last_question` reopen). |
| **9** | One-sentence reminder inserted at the start of the "Route the follow-up text through the Refine gate" paragraph: the Step 4 short-PATH-2 skip rule does NOT apply here, even for single-sentence corrections. Uses the same `"Exclude retry scheduling from the seed."` example as Step 4 so both gates point at the same canonical failure case. |

## Why

Step 4 currently lists three Refine skip cases, including *"Short PATH 2 answers (e.g., \"Yes\" / single proper noun) with no reasoning attached"*. Step 9's *"Adjust wording"* / *"Missing scope"* corrections are routed through the Refine gate, but the Step 4 short-PATH-2 exemption is generic and length-based — so a perfectly reasonable Restate correction such as:

> `"Exclude retry scheduling from the seed."`

matches the exemption, lets the agent skip Refine, and forwards `[from-user]` instead of `[from-user][refined]`. That misses two things the Step 9 contract requires:

1. The `[refined]` marker (MCP downgrades ambiguity weight without it).
2. The `last_question` reopen parameter (the interview handler rejects reopen calls that lack it — see `src/ouroboros/mcp/tools/authoring_handlers.py:1682-1702`).

The net effect: the seed is generated from **stale pre-correction MCP state** — the exact failure mode the Restate gate was added to prevent in #824.

## What this PR does NOT touch

The same final-APPROVE review left two other follow-ups that are **documentation-consistency** issues rather than runtime bypass risks:

- Step 9 payload field names (`answer_summary` / `open_questions`) vs. Step 3 schema (`Decision` / `Reasoning` / `Constraints` / `Out of scope` / `Codebase context`).
- 2nd-pass Refine handling of `Add context` (currently undefined for repeated selection).

Those are intentionally left for a separate follow-up PR so this fix can ship fast on its own scope.

## Recommended merge order

This PR is independent of #823/#822/#824 (which are merged) and depends on nothing else. Safe to merge as a hotfix.

## Test plan

Documentation-only change — reviewer checklist:

- [ ] Step 4 skip list now contains an explicit Restate-corrections exemption.
- [ ] Step 9 correction paragraph explicitly says the Step 4 short-PATH-2 skip rule does NOT apply.
- [ ] Both Step 4 and Step 9 reference the same canonical example (`"Exclude retry scheduling from the seed."`) — confirms the two gates are wired to the same contract.
- [ ] The skip-list bullets for PATH 1a / PATH 1b / PATH 4 / short PATH 2 are unchanged (no regression in the other Refine skip cases).
- [ ] `git diff main -- skills/interview/SKILL.md` shows only additive changes (+22 / -4 net) in two contiguous regions; no unrelated edits.

### DIFF
--- a/skills/interview/SKILL.md
+++ b/skills/interview/SKILL.md
@@ -325,6 +325,18 @@ MCP (question generator) ←→ You (answerer + router) ←→ User (human judgm
    - Short PATH 2 answers (e.g., "Yes" / single proper noun) with no
      reasoning attached
 
+   **Exception — Restate corrections never skip Refine, regardless of length.**
+   Any free-text follow-up collected by the Step 9 Restate gate
+   ("Adjust wording" / "Missing scope") must go through Refine before being
+   forwarded to MCP. Even a short correction like
+   `"Exclude retry scheduling from the seed."` carries scope/boundary
+   information that MCP needs in structured form during the reopen call, so
+   the short-PATH-2 exemption above does NOT apply to Restate corrections.
+   A single-line Restate correction that bypasses Refine would forward
+   `[from-user]` instead of `[from-user][refined]` and would not trigger the
+   `last_question` reopen, leaving MCP on stale pre-correction state when
+   the seed is generated.
+
    Refine-passed answers count as direct user judgment — they reset the
    Dialectic Rhythm Guard counter to 0 (see below).
 
@@ -425,8 +437,13 @@ MCP (question generator) ←→ You (answerer + router) ←→ User (human judgm
 
    Treat the follow-up text as a real interview correction, not a local-only
    wording tweak. Route the follow-up text through the Refine gate before
-   forwarding it: build the same structured multi-section payload from Step 3
-   using the canonical five-section schema —
+   forwarding it. **The Step 4 short-PATH-2 skip rule does NOT apply here,
+   even if the correction is a single sentence** — a bare reply like
+   `"Exclude retry scheduling from the seed."` still encodes a boundary
+   decision that must reach MCP as `[from-user][refined]` with the reopen
+   `last_question`; sending it as `[from-user]` would leave MCP on the stale
+   pre-correction state. Then build the same structured multi-section
+   payload from Step 3 using the canonical five-section schema —
    - **Decision**: the corrected one-line goal verbatim.
    - **Reasoning**: why the wording/scope changed (carry over the user's
      correction text, do not paraphrase).


shaun0927:
Reviewed PR #827 end-to-end, including the bot design notes and the surrounding interview reopen/Restate flow.

This PR is conceptually valid and merge-ready. It is intentionally narrow: it updates `skills/interview/SKILL.md` so that Restate-gate corrections cannot be misclassified as Step 4 short-PATH-2 answers. That closes a real instruction-level bypass where a short correction such as a missing non-goal or scope exclusion could otherwise be forwarded as a plain `[from-user]` answer instead of being refined and replayed into MCP as a reopened post-seed-ready answer.

Why the direction is sound:
- The change matches the existing `last_question` reopen contract: post-completion/seed-ready corrections must be attached to the local Restate follow-up question, not to the stale prior MCP-generated question.
- The patch reinforces both relevant instruction points: the normal short-answer Refine exception and the Restate correction flow.
- The resulting flow remains bounded: user rejects Restate -> assistant asks for the exact correction -> assistant Refines the correction -> MCP receives `[from-user][refined]` plus `last_question` -> seed-ready/Restate runs again before seed generation.
- The bot has already approved the current head commit and explicitly noted that the change is coherent with the interview workflow and handler-side `last_question` contract.

Additional validation I performed:
- Confirmed the PR is docs/skill-only and scoped to `skills/interview/SKILL.md`.
- Checked the surrounding handler semantics in `src/ouroboros/mcp/tools/authoring_handlers.py`, where already-answered reopen answers require `last_question` in the in-process path.
- Checked the interview reopen behavior in `src/ouroboros/bigbang/interview.py` and the existing unit coverage for the `last_question` reopen path.
- Ran local validation:
  - `git diff --check <base>...HEAD`
  - targeted invariant checks for the new Restate exception wording, `[from-user][refined]`, `last_question`, and returning to Step 7
  - `PYTHONPATH=src python3 -m pytest tests/unit/mcp/tools/test_interview_reopen_seed_ready.py -q` (`2 passed`)
- Confirmed GitHub state is clean: PR is `MERGEABLE`, `mergeStateStatus` is `CLEAN`, review decision is `APPROVED`, and all reported checks are passing.

I did not push any additional commits because I did not find a blocking or PR-scope issue. Two broader follow-ups remain possible but should not block this PR: plugin-mode parity around `last_question` enforcement, and the pre-existing naming inconsistency between prose payload headings and schema-like field names. Neither is introduced by this PR, and neither undermines the targeted Restate-gate bypass fix.

Based on the scope, bot approval, local validation, passing CI, and clean mergeability, I consider this PR safe to merge.

PR #832, opened 2026-05-11
feat(cli/plugin): surface trust_read_error in `ooo plugin list --json` (#806)

shaun0927:
## Summary

- Adds a per-row `trust_read_error: str | None` field to `ooo plugin list --json` so operators can deterministically tell which plugin's `trust.json` is malformed and what the parser actually saw.
- Schema stays stable: the field is set to `None` on every healthy row (unreadable-manifest branch, first-party branch, normal branch) and to the caught `ValueError`/`OSError` message on the corrupt-trust branch.
- Table view is unchanged in this PR — the operator-facing stderr warning is intentionally split out (see Follow-up below) so the JSON surface can land on its own.

## Why

#806 (the follow-up tracked from PR #798's bot review) flagged that the resilient corrupt-`trust.json` branch overshoots: a row whose trust file is malformed currently reports `trust_state: "trust_unreadable"` (good) but otherwise looks identical to "installed, no grants yet" (bad — `--json` consumers cannot recover the parse error). This PR closes the diagnostic gap on the JSON side without touching the human view.

## Acceptance criteria from the issue

- [x] For a row whose `trust.json` is malformed JSON / missing required keys, `ooo plugin list --json` surfaces that fact deterministically (`trust_read_error` non-null, `trust_state == "trust_unreadable"`).
- [x] Healthy rows expose `trust_read_error: null` so consumers can rely on `row[\"trust_read_error\"]` always being present.

## Test plan

- [x] `tests/unit/cli/test_plugin_command.py::test_list_survives_corrupt_trust_with_readable_manifest` extended to assert both the new field and the existing `trust_state` flip.
- [x] Full plugin command suite green (`pytest tests/unit/cli/test_plugin_command.py` → 27 passed).
- [x] `ruff check` / `ruff format --check` clean on both touched files.

## Out of scope (follow-up PR)

- Stderr warning in the human/table view when any row carries a non-null `trust_read_error` — will stack on this PR so reviewers can evaluate the JSON contract independently of the UX change.

Closes #806 (partial — JSON surface).

PR #833, opened 2026-05-11
feat(cli/plugin): warn on stderr when `ooo plugin list` row has unreadable trust.json (#806)

### LINKED ISSUE #806: follow-up(cli/plugin): row-level diagnostic when trust.json is unreadable
Tracking the **non-blocking follow-up** raised by ouroboros-agent[bot] on PR #798 (review of 4e13fbe).

## Context

PR #798 made ``ooo plugin list`` resilient to a corrupt ``trust.json``: instead of crashing the whole listing, the offending row is now degraded to ``record = None`` and the listing continues. That fixes the reported bug (one bad plugin can no longer hide every other installed plugin).

But the bot correctly pointed out that the fix overshoots: a corrupt ``trust.json`` is now indistinguishable from "no trust file exists." The row reports ``trust_state: installed`` with empty ``granted_scopes``, exactly like an untrusted-but-otherwise-fine plugin. Operators using ``--json`` consumers cannot tell which plugin's trust file is damaged, even though that's the diagnostic they came to ``list`` for.

## Suggested change

Continue the resilience behaviour, but surface a per-row diagnostic signal when the read failed. Options to consider:

- Add a top-level row field like ``trust_read_error: <message>`` to the ``--json`` output (and a stderr warning in the human view) whenever ``trust.read(...)`` raised inside the resilient branch.
- Or: split ``trust_state`` into ``installed_no_trust`` vs ``installed_trust_unreadable`` so the existing field carries the distinction without a new key.

Either way the regression to land is: for a row whose ``trust.json`` is malformed JSON / missing required keys, ``ooo plugin list --json`` MUST surface that fact deterministically, not collapse it onto the untrusted state.

## References

- Bot review: https://github.com/Q00/ouroboros/pull/798
- Related test: ``tests/unit/cli/test_plugin_command.py::test_list_survives_corrupt_trust_with_readable_manifest``

## Why this is a separate issue

PR #798 is APPROVED and lands the user-visible bug fix (no more crash). The diagnostic enrichment is a strictly additive UX improvement, so it does not need to block #798's merge.

shaun0927:
## Summary

- Adds a per-row stderr warning to `ooo plugin list` (human/table view) whenever a row's `trust.json` failed to parse. The warning names the offending plugin, surfaces the underlying parser message, and points the operator at the recovery command (`ooo plugin trust <name> --scope <...>`).
- Stdout stays free of the warning so the structured table is still safe to pipe to `less` / `column` / a screenshot.
- Uses `typer.echo(..., err=True)` (not a module-level `rich.Console(stderr=True)`) so `CliRunner` can observe stderr separately — see commit body for why.

## Why

Companion to PR #832 (which lands `trust_read_error` on `--json`). Together they close #806: machine consumers get the deterministic field, humans get the side-channel signal. This PR keeps the table view's stdout contract unchanged while making sure operators reading the human output actually notice that one of their trust files is malformed.

## Stack

Stacked on **#832**. Until #832 merges, the diff here will show both commits; when reviewing this PR, focus on the latest commit (`67c4affd feat(cli/plugin): warn on stderr when …`). After #832 merges this PR's mergeable diff will collapse to the +92 lines added here.

## Test plan

- [x] New `test_list_table_warns_on_unreadable_trust_via_stderr` pins:
  - the warning lands on stderr and names only the offending plugin,
  - the recovery command (`ooo plugin trust bad-plugin --scope ...`) appears in stderr,
  - stdout renders the full table (both clean and degraded rows),
  - stdout carries no `warning:` prefix.
- [x] Full plugin command suite green (`pytest tests/unit/cli/test_plugin_command.py` → 28 passed).
- [x] `ruff check` / `ruff format --check` clean on both touched files.

Closes #806 (human surface — final piece).

### DIFF
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -1131,6 +1131,29 @@ def list_command(
         typer.echo(json.dumps(rows, indent=2))
         return
 
+    # Operator-facing diagnostic: when any row's trust file failed to
+    # parse, the table column flips to ``trust_unreadable`` but does NOT
+    # carry the underlying parser message. Surface it on stderr so the
+    # operator sees both (a) which plugin's trust file is broken and
+    # (b) the literal recovery action, without polluting stdout (which
+    # remains the structured table for ``less`` / ``column`` /
+    # screenshotting). The ``--json`` path already exposes the same
+    # information via the ``trust_read_error`` field, so we only emit
+    # the stderr warning on the human surface. ``typer.echo(..., err=True)``
+    # resolves ``sys.stderr`` at call time, which is what
+    # ``CliRunner(mix_stderr=False)`` needs to capture the warning under
+    # test.
+    unreadable = [row for row in rows if row.get("trust_read_error") is not None]
+    for row in unreadable:
+        typer.echo(
+            f"warning: trust file for {row['name']!r} is unreadable "
+            f"({row['trust_read_error']}); the row is shown with "
+            f"trust_state=trust_unreadable and empty scopes. Inspect or "
+            f"replace the file, then re-grant scopes via "
+            f"`ooo plugin trust {row['name']} --scope <...>`.",
+            err=True,
+        )
+
     table = create_table(title="Installed UserLevel plugins")
     for column in ("name", "version", "source", "trust", "scopes"):
         table.add_column(column)
--- a/tests/unit/cli/test_plugin_command.py
+++ b/tests/unit/cli/test_plugin_command.py
@@ -1030,6 +1030,75 @@ def test_list_survives_corrupt_trust_with_readable_manifest(
     assert bad_row["trust_read_error"]
 
 
+def test_list_table_warns_on_unreadable_trust_via_stderr(runner: CliRunner, tmp_path: Path) -> None:
+    """Regression for #806: when ``ooo plugin list`` (no ``--json``) walks
+    a row whose ``trust.json`` is malformed, the table column already
+    flips to ``trust_unreadable`` but the parser's reason is lost. The
+    table view must surface that fact on **stderr** so an operator
+    reading the human output sees both which plugin's trust file is
+    broken and the literal recovery command, while stdout stays clean
+    for piping/screenshotting.
+
+    The ``--json`` path is covered by
+    ``test_list_survives_corrupt_trust_with_readable_manifest`` (the
+    ``trust_read_error`` field assertion); this test pins the human
+    surface that consumes the same signal.
+    """
+    clean_home = tmp_path / "clean_home"
+    _write_manifest(clean_home, {**REFERENCE_MANIFEST, "name": "clean-plugin"})
+    bad_home = tmp_path / "bad_home"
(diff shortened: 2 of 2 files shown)


shaun0927:
## Hermes merge-readiness review

I reviewed PR #833 against its stated intent in #806 and the current bot design notes.

### What this PR changes

This PR keeps `ooo plugin list` resilient when an individual plugin has an unreadable or malformed `trust.json`, while making that condition distinguishable from the normal "no trust grant yet" state:

- `ooo plugin list --json` now emits a per-row `trust_read_error` field.
  - Normal rows, first-party rows, and manifest-degraded rows carry `null`.
  - Rows whose trust record cannot be read carry a non-empty diagnostic string and use `trust_state="trust_unreadable"`.
- Human table output keeps stdout as the table and emits a targeted warning on stderr for each unreadable trust row, including the affected plugin name and the recovery command.
- The behavior remains fail-closed: unreadable trust records do not expose grants and required scopes are still treated as missing unless an applicable trust record can be read.

### Why the direction is sound

The intent is valid because a malformed trust file is operationally different from an untrusted plugin with no grant file. Collapsing both cases into the same row state would hide the exact diagnostic signal an operator needs from `ooo plugin list`. The PR addresses that gap without reintroducing the original failure mode: one bad trust file no longer aborts the whole listing, and the new diagnostic is scoped to the affected row.

The implementation is also appropriately narrow and backwards-compatible for this change:

- Existing JSON fields are preserved; `trust_read_error` is additive and consistently present.
- Machine output receives the diagnostic in JSON, while human output receives it on stderr so stdout remains usable for table piping/screenshotting.
- First-party and unreadable-manifest branches were updated to keep the new row schema stable with `trust_read_error: null`.

### Review and validation performed

- Reviewed the PR body/issue #806 intent and the bot design notes.
- Reviewed the full effective diff from base `e3a0112913cad303e3542bdb8b40aa01d3781721` to head `67c4affd47b2039b387010fee90ca74e4dde7e75`.
- Ran targeted local validation in an isolated worktree:
  - `PYTHONPATH=src python3 -m pytest tests/unit/cli/test_plugin_command.py -q` → `28 passed`
  - `ruff check src/ouroboros/cli/commands/plugin.py tests/unit/cli/test_plugin_command.py` → passed
  - `ruff format --check src/ouroboros/cli/commands/plugin.py tests/unit/cli/test_plugin_command.py` → passed
  - `git diff --check e3a0112913cad303e3542bdb8b40aa01d3781721...HEAD` → passed
- Ran an independent Codex review pass focused on semantic correctness, backwards compatibility, and test coverage; it found no blocking issues.
- Re-checked GitHub state for the current head:
  - Bot review: `ouroboros-agent[bot]` approved current head `67c4affd`.
  - Merge state: `MERGEABLE` / `CLEAN`.
  - Status checks: Ruff Lint, Python 3.12/3.13/3.14 tests, MyPy, Bridge TypeScript, and repository enforcement checks are all successful.

### Verdict

I did not find any issue that warrants a remediation commit. PR #833 is conceptually valid, implements the requested diagnostic improvement at the right surface boundaries, has focused regression coverage for both JSON and human table output, and is merge-ready from both code-review and current GitHub status perspectives.

Issue #831, opened 2026-05-11
interview MCP: hang via empty stop_reason=tool_use after length-guard in-band response (repro trace)


shaun0927:
## Summary

The interview MCP path (`ouroboros_interview`) intermittently induces Claude Code to produce a malformed assistant turn — `stop_reason=tool_use` with **zero `tool_use` blocks** and an empty `thinking` block — causing the user to see a multi-minute hang requiring repeated `continue` prompts. Reproduced in a long-form brownfield interview session; full event trace below.

This issue files the evidence and asks for design feedback before any code change is attempted. **No PR is attached yet** — there are at least two viable fix directions and the right one is not obvious.

## Repro / Evidence

Captured from a real Claude Code session against an ~8KB `initial_context` (brownfield repo cleanup interview). Source JSONL: `~/.claude/projects/-Users-jh0927-openchrome/026a2e19-….jsonl`.

### Trace timeline (key events)

```
ev[327] 08:48:39  assistant tool_use=ouroboros_interview (start, initial_context ~8KB)
ev[329] 08:48:39  tool_result is_error=False:
                  "Interview started. Session ID: …
                   Your saved initial context is too long to safely send to
                   the interview model without risking CLI prompt failure.
                   Please reply with a concise summary…"
                  ← Length-guard hit (MAX_PROMPT_SAFE_INITIAL_CONTEXT_CHARS=3500),
                    delivered as in-band TEXT in the question slot,
                    with NO structured signal (is_error / meta.recoverable).

ev[345] 08:50:16  assistant tool_use=ouroboros_interview (answer=summary, ~3KB)
ev[350] 08:50:28  tool_result: "Should OpenChrome's scope remain…"
                  ← Normal question, ~250 chars.
ev[354] 08:50:42  assistant stop_reason=tool_use
                  content=[{type:"thinking", thinking:""}]
                  ← ANOMALY: stop_reason=tool_use but tool_use blocks = 0,
                    thinking is empty string. API contract violation in
                    output shape. User perceives this as a hard hang.

ev[372] 09:12:02  next MCP response (after user typed `continue` repeatedly)
ev[426] 11:41:11  next genuine assistant action — 2h 29m gap.
```

### Two structural quirks that correlate with the hang

1. **In-band length-guard "question"** (`bigbang/interview.py:40-46, 392, 404`):
[...]

PR #834, opened 2026-05-11
feat(mcp): structured envelope for interview length-guard branch (#831)

shaun0927:
## Summary

**Direction A from #831 only.** Stops the interview length-guard branch from masquerading as a normal interview question, so a programmatic client can branch on `meta.reason` instead of mis-routing the meta-directive to a human via `AskUserQuestion`.

**This PR is a meta-only envelope:** `is_error` stays `False`, the text body is preserved verbatim, and the new structured information lives entirely in `meta`. *Earlier commits in this branch flipped `is_error` to `True` on the length-guard branch — that turned out to break `HandlerInterviewBackend.start` (the auto driver) and the follow-up commit on this branch walks the flip back and adds a regression test that locks the auto driver semantics.* The wire success/failure axis is intentionally NOT flipped.

## What this PR does

- When `initial_context` exceeds `MAX_PROMPT_SAFE_INITIAL_CONTEXT_CHARS`, the response (in all three return sites: `start`, `resume_pending`, `answer`) merges these keys into `meta`:

  ```
  meta.recoverable     = True
  meta.reason          = "initial_context_too_large"
  meta.expected_action = "resend_with_summary"
  meta.max_chars       = MAX_PROMPT_SAFE_INITIAL_CONTEXT_CHARS
  ```

- `is_error` stays `False`.
- Text body is byte-identical to today.
- `skills/interview/SKILL.md` gains a one-paragraph clause under the existing *Retry on Failure* section telling the routing layer: if `meta.reason == "initial_context_too_large"`, summarize `initial_context` locally (≤ `meta.max_chars`) and re-call `ouroboros_interview` with the summary as `answer`. Do not route it via `AskUserQuestion`.

## Why this is NOT a hang fix

On re-reading the JSONL trace in #831, the empty-`thinking` / `stop_reason=tool_use` hangs at `ev[354]` and around `ev[426]` both occurred *after normal MCP question responses*, not after the length-guard `ev[329]` response. That anomaly is a `claude-agent-sdk` / Claude Code layer issue, not ouroboros. This PR addresses a related but *separate* contract design issue surfaced during the same investigation. PR #837 (diagnostics-only event) is the companion that will give us data to confirm/deny the response-shape correlation for the actual hang.

## Test plan

- [x] `tests/unit/mcp/tools/test_interview_length_guard_envelope.py`:
  - Length-guard response: `is_error=False`, all documented `meta` keys present, text body preserved verbatim.
  - Normal first question: `is_error=False`, no length-guard `meta` keys leak.
  - **`test_auto_driver_does_not_raise_on_length_guard_response`** (new): drives `HandlerInterviewBackend.start` end-to-end with an oversized `initial_context`. Asserts no exception is raised and the returned `InterviewTurn` carries the summarize-prompt text as its question — locks the auto driver contract in so a future `is_error` flip would fail this test instead of surfacing as a runtime regression in `ooo auto`.
- [x] `pytest tests/unit/mcp/tools/ tests/unit/auto/ tests/unit/bigbang/test_interview.py tests/unit/bigbang/test_seed_generator.py tests/unit/bigbang/test_pm_interview.py` — 1322 passing.
- [x] `ruff format` clean; `ruff check` clean.

## Out of scope (still open in #831 for maintainer decision)

- **Direction B** (server-side auto-summarization LLM call) — deliberately not implemented; the in-band "ask the human to summarize" question is a deliberate trade-off and an automatic LLM call would change that contract silently.
- **Direction C** (SKILL.md-only handler via literal-string match) — partly subsumed by the structured `meta` field added here.
- No `answer` length cap.
- `_trim_messages_to_budget`, `should_dispatch_via_plugin`, and the `(ambiguity: …)` display prefix are not touched.

## Notes for reviewer

- Two commits on the branch: the initial structured envelope, and a follow-up that walks back the `is_error` flip after the regression on the auto driver was caught. Each commit message explains the rationale; squash-on-merge is fine if maintainers prefer one combined commit.

Refs #831 (Direction A).

### DIFF
--- a/skills/interview/SKILL.md
+++ b/skills/interview/SKILL.md
@@ -480,6 +480,15 @@ If MCP returns `is_error=true` with `meta.recoverable=true`:
 3. If still failing: "MCP is having trouble. Switching to direct interview mode."
    Then switch to Path B and continue from where you left off.
 
+**Special case — `meta.reason == "initial_context_too_large"`**: When the
+response carries this reason (also `is_error=true`, `recoverable=true`), the
+text body is a meta-directive asking *you* to re-send a shorter context — it
+is **NOT** an interview question. Do not route it via AskUserQuestion.
+Instead, produce a concise summary of the original `initial_context`
+(≤ `meta.max_chars` characters; covers goal, constraints, success criteria)
+and re-call `ouroboros_interview` with `session_id=<from meta>` and the
+summary as `answer`. The next response will contain the real first question.
+
 **Advantages of MCP mode**: State persists to disk, ambiguity scoring, direct `ooo seed` integration via session ID. Code-enriched confirmation questions reduce user burden — only human-judgment questions require user input.
 
 ### Path B: Plugin Fallback (No MCP Server)
--- a/skills/interview/SKILL.md
+++ b/skills/interview/SKILL.md
@@ -481,9 +481,11 @@ If MCP returns `is_error=true` with `meta.recoverable=true`:
    Then switch to Path B and continue from where you left off.
 
 **Special case — `meta.reason == "initial_context_too_large"`**: When the
-response carries this reason (also `is_error=true`, `recoverable=true`), the
-text body is a meta-directive asking *you* to re-send a shorter context — it
-is **NOT** an interview question. Do not route it via AskUserQuestion.
+response carries this `meta.reason` (with `meta.recoverable=true` and
+`is_error=false` — the wire success/failure axis is intentionally **not**
+flipped, to keep existing callers like the auto driver working), the text
+body is a meta-directive asking *you* to re-send a shorter context — it is
+**NOT** an interview question. Do not route it via AskUserQuestion.
 Instead, produce a concise summary of the original `initial_context`
 (≤ `meta.max_chars` characters; covers goal, constraints, success criteria)
 and re-call `ouroboros_interview` with `session_id=<from meta>` and the
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -1479,9 +1479,13 @@ class InterviewHandler:
                 }
                 if is_length_guard:
                     # Q00/ouroboros#831: surface the length-guard meta-directive
-                    # as a structured envelope so clients do not mis-route it
-                    # to a human via AskUserQuestion.  Text body is preserved
-                    # verbatim so the CLI interview UX is unchanged.
+                    # via structured meta keys so clients can branch on
+                    # ``meta.reason`` instead of mis-routing the text body to a
+                    # human via AskUserQuestion.  ``is_error`` is intentionally
+                    # left ``False`` -- the wire success/failure axis must not
+                    # flip or ``HandlerInterviewBackend.start`` (auto driver)
+                    # would raise on every oversized ``initial_context`` where
+                    # it previously delivered the summarize question.
                     start_meta.update(_length_guard_meta_fields())
                 return Result.ok(
                     MCPToolResult(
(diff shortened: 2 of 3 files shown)


shaun0927:
Pushed a follow-up commit on this branch that walks back the `is_error` flip on the length-guard branch.

**Why**: on closer review I noticed `HandlerInterviewBackend.start` (`src/ouroboros/auto/adapters.py:73-87`) raises `PartialInterviewStartError` whenever `value.is_error` is true with `meta.session_id` present; `answer()` / `resume()` raise `HandlerError` via `_unwrap` on the same axis. Flipping `is_error` to true on the length-guard branch therefore broke `ooo auto` for every oversized `initial_context` — the auto driver previously delivered the summarize prompt as a normal interview turn and would now hard-fail instead. Real regression, missed in the first round.

**Fix**: `is_error` stays `False`. The new structured signal lives entirely in `meta` (`meta.reason="initial_context_too_large"` etc.). That's enough for a programmatic client to branch on, which was the original goal. SKILL.md clause is updated to make this explicit. New regression test `test_auto_driver_does_not_raise_on_length_guard_response` drives `HandlerInterviewBackend.start` end-to-end with an oversized context and asserts no exception is raised and the turn carries the summarize-prompt text — locks the auto driver contract in so a future `is_error` flip would fail the test instead of surfacing as a runtime regression.

The PR description is updated to reflect the meta-only envelope shape.

shaun0927:
I reviewed this PR end-to-end and believe it is now safe to merge.

What this PR does:
- Adds a structured, machine-readable envelope for the oversized `initial_context` interview length guard.
- Keeps the existing human-readable summarize prompt intact, but adds `meta.recoverable=true`, `meta.reason="initial_context_too_large"`, `meta.expected_action="resend_with_summary"`, and `meta.max_chars` so MCP clients can route the directive deterministically instead of treating it as an ordinary AskUserQuestion prompt.
- Documents the client contract in `skills/interview/SKILL.md`.

Why the final shape is correct:
- The first implementation direction (`is_error=true`) was not compatible with the in-repo auto caller: `HandlerInterviewBackend.start()` / `resume()` / `answer()` treat the MCP error axis as a hard failure. That would have turned a recoverable length-guard directive into an unrecoverable `ooo auto` failure for oversized contexts.
- The revised implementation keeps `is_error=false` and carries the new signal entirely in `meta`. That preserves the existing success-path contract while still giving programmatic clients the structured information they need.
- The handler now applies the same meta-only length-guard envelope consistently across all modified response-shaping paths: initial start, resuming a pending length-guard round, and generating the length-guard prompt after an answer.

Additional hardening added:
- Added regression coverage for the resume path and post-answer path, directly addressing the bot design note that only the start path had been locked down.
- Existing start-path / auto-driver coverage remains in place, ensuring the auto driver still receives the summarize directive as a normal turn and does not raise on oversized contexts.

Validation:
- Local targeted validation passed: `ruff check tests/unit/mcp/tools/test_interview_length_guard_envelope.py` and `PYTHONPATH=src python3 -m pytest tests/unit/mcp/tools/test_interview_length_guard_envelope.py tests/unit/auto/test_surface.py tests/unit/auto/test_interview_pipeline.py` (`207 passed`).
- CI is green on the PR head: Ruff Lint, MyPy Type Check, Bridge TypeScript, enforce-boundary, enforce-envelope, and Python 3.12 / 3.13 / 3.14 tests all passed.
- `ouroboros-agent[bot]` approved the latest commit (`452d23d`) with no blocking or non-blocking findings, and explicitly notes that the regression coverage is now focused on the three touched return sites plus the auto-driver start contract.

Given the compatibility correction, the added regression coverage for the previously identified gaps, green CI, and the bot approval on the current head, I consider this PR merge-ready.

PR #837, opened 2026-05-11
feat(mcp): diagnostic event for interview response shape (#831)

shaun0927:
## Summary

Follow-up to #831 (and to PR #834). Adds a new diagnostic event
`interview.response.emitted` fired every time `InterviewHandler` returns
an MCP response carrying an interview question — start, resume-pending,
and answer build paths. **Pure observability; no behaviour change.**

## Why

The hang trace attached to #831 shows two instances of
`stop_reason=tool_use` with an empty `thinking` block and zero `tool_use`
blocks, both *after normal MCP question responses*. That anomaly almost
certainly originates in the `claude-agent-sdk` / Claude Code layer, but
the trigger surface correlates with the response payload ouroboros sends
back. Today we have no way to verify whether response size, transcript
pressure, prefix presence, or round number actually correlate with hang
frequency — every analysis is speculation against a single trace.

This event makes the correlation queryable. After this lands, the next
hang report can be joined against `interview.response.emitted` rows in
the event store and we can confirm or rule out the response-shape
hypothesis with data instead of more guessing.

## Payload

```
interview.response.emitted {
  interview_id,
  response_kind:           "start" | "resume_pending" | "answer",
  round_number,
  payload_chars,           # full response text body length
  transcript_chars,        # sum of question+user_response over rounds
  ambiguity_prefix_present # response starts with "(ambiguity: …)"
  is_length_guard,         # response is INITIAL_CONTEXT_SUMMARY_QUESTION
}
```

## What this PR is NOT

- Not a fix for the hang. The empty-`tool_use` anomaly is in the
  Claude Code / claude-agent-sdk layer and needs a separate upstream
  report.
- Not a contract change. Existing `MCPToolResult` shape (text, `is_error`,
  `meta`) is identical to today's behaviour.
- Not coupled to PR #834 (Direction A). This PR branches from
  `upstream/main` and works whether or not #834 lands. The
  `is_length_guard` field here is computed locally by direct comparison
  with `INITIAL_CONTEXT_SUMMARY_QUESTION`; it does not depend on the
  structured envelope introduced in #834.

## Test plan

- [x] New regression test `tests/unit/mcp/tools/test_interview_response_diagnostics.py`:
  - Start path with a normal question — asserts `response_kind="start"`,
    `round_number=1`, payload/transcript char fields populated,
    `is_length_guard=False`.
  - Start path with the length-guard meta-directive — asserts
    `is_length_guard=True`.
  - Resume-pending path — asserts `response_kind="resume_pending"` and
    that `transcript_chars` reflects the pending round's question
    length.
  - Uses an in-memory `_CapturingEventStore` that mirrors only the
    `initialize` / `append` surface `_emit_event` actually touches.
- [x] `pytest tests/unit/mcp/tools/ tests/unit/bigbang/test_interview.py` — 623 passing.
- [x] `ruff format` clean; `ruff check` clean.

## Notes for reviewer

- Event helper: `interview_response_emitted` in
  `src/ouroboros/events/interview.py`, following the existing
  `dot.notation.past_tense` naming convention.
- Three emission sites in `src/ouroboros/mcp/tools/authoring_handlers.py`
  mirror the three `MCPToolResult` build paths in `InterviewHandler.handle`.
  The response text body is now bound to a local variable at each site so
  `len(response_text)` can be captured without re-formatting.
- A tiny module-level helper `_compute_transcript_chars` keeps the
  per-site code one-line short and the test surface single-call.
- Fire-and-forget via the existing `self._emit_event_bg` path; failures
  to emit are swallowed by the existing `_emit_event` warning path, so
  the event store being temporarily unavailable cannot break interview
  flow.

Refs #831.

### DIFF
--- a/src/ouroboros/events/interview.py
+++ b/src/ouroboros/events/interview.py
@@ -71,3 +71,49 @@ def interview_failed(
             "phase": phase,
         },
     )
+
+
+def interview_response_emitted(
+    interview_id: str,
+    *,
+    response_kind: str,
+    round_number: int,
+    payload_chars: int,
+    transcript_chars: int,
+    ambiguity_prefix_present: bool,
+    is_length_guard: bool,
+) -> BaseEvent:
+    """Diagnostic event recording the shape of an MCP question-bearing response.
+
+    No behaviour change: emitted alongside the existing lifecycle events so
+    a later investigation can correlate hang reports (e.g. claude-agent-sdk
+    producing an empty-``thinking`` / ``stop_reason=tool_use`` turn after
+    receiving an interview question) with response payload characteristics.
+    See Q00/ouroboros#831 comment thread for the hang trace context.
+
+    Args:
+        interview_id: Session id.
+        response_kind: One of ``"start"`` / ``"resume_pending"`` /
+            ``"answer"`` -- which build path produced the response.
+        round_number: Round count at emission time (``len(state.rounds)``).
+        payload_chars: Total characters in the response text body.
+        transcript_chars: Sum of ``len(question) + len(user_response)`` for
+            every completed round in ``state.rounds``.
+        ambiguity_prefix_present: ``True`` when the response text begins
+            with ``(ambiguity: ...)``.
+        is_length_guard: ``True`` when the response carries the
+            ``INITIAL_CONTEXT_SUMMARY_QUESTION`` meta-directive.
+    """
+    return BaseEvent(
+        type="interview.response.emitted",
+        aggregate_type="interview",
+        aggregate_id=interview_id,
+        data={
+            "response_kind": response_kind,
+            "round_number": round_number,
+            "payload_chars": payload_chars,
+            "transcript_chars": transcript_chars,
+            "ambiguity_prefix_present": ambiguity_prefix_present,
+            "is_length_guard": is_length_guard,
+        },
+    )
--- a/src/ouroboros/events/interview.py
+++ b/src/ouroboros/events/interview.py
@@ -98,7 +98,8 @@ def interview_response_emitted(
         round_number: Round count at emission time (``len(state.rounds)``).
         payload_chars: Total characters in the response text body.
         transcript_chars: Sum of ``len(question) + len(user_response)`` for
-            every completed round in ``state.rounds``.
+            every round in ``state.rounds``, including pending questions with
+            no response yet.
         ambiguity_prefix_present: ``True`` when the response text begins
             with ``(ambiguity: ...)``.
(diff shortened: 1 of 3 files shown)


shaun0927:
I reviewed this PR's intent and direction, then did a second pass against the bot design notes and the implementation/test coverage.

This PR is the right shape for #831: it adds a pure-observability `interview.response.emitted` event for the three question-bearing MCP interview response paths (`start`, `resume_pending`, and `answer`) so future hang reports can be correlated with response shape data (`response_kind`, round count, payload size, transcript size, ambiguity-prefix shape, and length-guard usage). It does not change the public MCP response contract or the interview control flow.

I found the PR direction valid, but the bot's follow-up notes identified two issues worth fixing before merge:

1. `ambiguity_prefix_present` was documented as describing the emitted response body, but the implementation checked only the embedded displayed question. Because the actual MCP response body starts with `Interview started...` or `Session ...`, that could have made the diagnostic misleading.
2. The regression suite covered `start` and `resume_pending`, but not the `answer` path even though the PR also emits diagnostics there.

I pushed follow-up fixes to resolve those points:

- `3a375b9e` now computes `ambiguity_prefix_present` from the full emitted response text at all three emission sites.
- The `transcript_chars` event documentation now explicitly matches the implementation: it includes all rounds in `state.rounds`, including pending questions without responses.
- Added regression coverage for the `answer` branch and for the full-response ambiguity-prefix behavior.
- Formatted the updated test file so CI lint passes.

Validation performed:

- `PYTHONPATH=src /opt/hermes/.venv/bin/python -m pytest tests/unit/mcp/tools/test_interview_response_diagnostics.py -q` → 5 passed
- `PYTHONPATH=src /opt/hermes/.venv/bin/python -m pytest tests/unit/mcp/tools/ tests/unit/bigbang/test_interview.py -q` → 625 passed
- `PYTHONPATH=src /opt/hermes/.venv/bin/python -m ruff check src/ouroboros/events/interview.py src/ouroboros/mcp/tools/authoring_handlers.py tests/unit/mcp/tools/test_interview_response_diagnostics.py` → passed
- `PYTHONPATH=src /opt/hermes/.venv/bin/python -m ruff format --check ...` → passed
- `git diff --check` → passed
- GitHub branch checks are now all green: Ruff, Python 3.12/3.13/3.14 tests, MyPy, boundary/envelope, and Bridge TypeScript.

Current PR state is mergeable (`MERGEABLE`, `CLEAN`) and review status remains approved. With the bot's follow-up findings addressed and the CI matrix green, I consider this PR merge-ready.

PR #845, opened 2026-05-11
feat(orchestrator): durable cancel signal for AgentProcess (#518)

shaun0927:
## Summary

Slice 4 of M6 (#518). Makes the cooperative cancel signal survive process restarts by persisting it through the existing ``CheckpointStore`` (#338). The in-memory flag remains the authoritative source; the durable record is a best-effort hint for the restart path.

## Behavior

- ``AgentProcessHandle.cancel(reason=..., checkpoint_store=...)`` now writes a ``agent_process_cancelled`` checkpoint keyed by ``process_id``. Save failures are logged at WARNING and never raised.
- New classmethod ``AgentProcessHandle.load_persisted_cancel(process_id, store=...)`` returns ``(True, reason)`` iff the latest checkpoint for that process_id is a cancel record. A restart loop calls this once at startup and jumps straight to teardown when True.
- Cancellation semantics remain cooperative (per #627). No forced kill, no behavior change for in-memory consumers.

## Test plan

- ``test_cancel_persists_state_via_checkpoint_store`` — cancel(reason='X'), load returns (True, 'X').
- ``test_cancel_swallows_checkpoint_save_error`` — injected store raises; in-memory transition still completes, no exception leaks.
- ``test_load_persisted_cancel_returns_false_when_no_checkpoint`` — fresh process_id.
- ``test_load_persisted_cancel_returns_false_without_store`` — ``store=None``.
- Full agent_process suite: 25 passed.
- ``ruff check`` + ``ruff format --check`` clean.

## Companion

Companion to PR-A (``feat/518-agent-process-durable-pause`` — durable pause/resume). Both share the ``CheckpointStore`` instance and the ``process_id``-as-seed-id convention. They are additive at the API level (different phase strings, different classmethods) so they can land in either order with at most a trivial textual merge.

Part of #518.

### DIFF
--- a/src/ouroboros/mcp/job_manager.py
+++ b/src/ouroboros/mcp/job_manager.py
@@ -7,14 +7,17 @@ from dataclasses import dataclass, field, replace
 from datetime import UTC, datetime, timedelta
 from enum import StrEnum
 import inspect
+import logging
 from typing import Any
 from uuid import uuid4
 
 from ouroboros.events.base import BaseEvent
+from ouroboros.orchestrator.agent_process import AgentProcessHandle
 from ouroboros.orchestrator.events import create_execution_terminal_event
 from ouroboros.orchestrator.heartbeat import is_holder_alive, is_owned_by_current_process
 from ouroboros.orchestrator.runner import clear_cancellation, request_cancellation
 from ouroboros.orchestrator.session import SessionRepository, SessionStatus
+from ouroboros.persistence.checkpoint import CheckpointStore
 from ouroboros.persistence.event_store import EventStore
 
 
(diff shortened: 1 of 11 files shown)


shaun0927:
I reviewed this PR’s intent, the accumulated bot design notes, and the current diff after rebasing it onto latest `main`.

My conclusion is that the PR is valid and now merge-ready.

What this PR does:
- Adds a durable cooperative-cancel signal for `AgentProcess` using `CheckpointStore`.
- Keeps the live lifecycle `process_id` attempt-unique while introducing a restart-stable `cancel_key` for MCP job cancellation (`mcp_job:<job_id>`), avoiding lineage/session-level cancel-channel collisions.
- Persists the cancel marker at the user-facing `JobManager.cancel_job()` boundary, so a crash after accepting cancellation but before the runner observes it does not lose the operator’s cancel request.
- Wires the production Ralph / evolve_step / execute_seed start surfaces so restarted work checks `handle.should_cancel()` before entering side-effectful workflow logic.

What was improved during this pass:
- Rebased the PR onto current `main`, preserving the already-merged AgentProcess pause/replay/evolve-step slices.
- Addressed the latest blocking design issue: `run_with_agent_process()` no longer consumes durable cancel markers on ordinary live cancellation. Markers are consumed only when a restarted process has actually loaded and observed a persisted cancel.
- Tightened rollback semantics for cancel checkpoints: a rotated `agent_process_cancel_consumed` marker with no current level is treated as ambiguous failed-write state and fails closed instead of silently resuming work.
- Added/kept focused regression coverage for job-scoped cancel keys, restart teardown before work begins, live-cancel crash-window behavior, corrupt/ambiguous checkpoint artifacts, and production start-surface wiring.

Validation evidence:
- Local targeted tests: `248 passed`
  - `tests/unit/orchestrator/test_agent_process.py`
  - `tests/integration/orchestrator/test_agent_process_three_surface_acceptance.py`
  - `tests/unit/mcp/test_job_manager.py`
  - `tests/unit/mcp/tools/test_job_instance_cancel_keys.py`
  - `tests/unit/mcp/tools/test_definitions.py`
  - `tests/unit/mcp/tools/test_start_execute_seed_idempotency.py`
  - `tests/unit/evolution/test_evolve_step_agent_process_integration.py`
- Local static checks passed:
  - `ruff check` on changed Python files/tests
  - `ruff format --check` on changed Python files/tests
  - `mypy` on changed production modules
- GitHub CI is green across lint, MyPy, Bridge TypeScript, and Python 3.12 / 3.13 / 3.14 tests.
- The latest `ouroboros-agent` review on commit `20dffac` is `APPROVE` with no blocking findings.

Given those points, the original design concern has been resolved: durable cancellation is now best-effort at the live boundary, restart-safe at the job boundary, isolated from unrelated job attempts, and conservative when checkpoint state is corrupt or ambiguous. I think this is safe to merge.

PR #847, opened 2026-05-11
feat(orchestrator): implement AgentProcess.replay() from control directive events (#518)

shaun0927:
## Summary

Slice 3 of M6 (#518). Replaces the ``NotImplementedError`` placeholder in ``AgentProcessHandle.replay()`` with a working implementation that folds persisted ``control.directive.emitted`` events into an ``AgentProcessSnapshot`` via the existing projector (``project_agent_process_snapshot`` from #628).

## Wiring

- New ``_ReplayableEventStore`` Protocol so this module has zero runtime import dependency on the persistence layer.
- ``AgentProcess.spawn`` automatically threads the event store into the handle as ``_replay_store`` when the store implements a ``replay`` method.
- Without an event store, ``replay()`` raises a structured ``RuntimeError`` (not ``NotImplementedError``) so callers can tell the feature exists but needs wiring.

## Folding rules (pinned in tests)

| Scenario                       | Snapshot.status | is_terminal |
|-------------------------------|-----------------|-------------|
| No events                     | RUNNING         | False       |
| Cancel                        | CANCELLED       | True        |
| Complete                      | COMPLETED       | True        |
| Pause without subsequent resume | PAUSED        | False       |

The projector handles all interleaving; no folding logic was duplicated.

## Test plan

- ``test_replay_without_event_store_raises_runtime_error`` (replaces the old ``test_replay_is_not_yet_implemented``)
- ``test_replay_after_completed_returns_completed_snapshot``
- ``test_replay_after_cancel_returns_cancelled_snapshot``
- ``test_replay_during_pause_returns_paused_snapshot``
- Full ``test_agent_process.py``: 24 passed.
- ``ruff check`` / ``ruff format --check`` clean.

## Companions

Independent of PR-A (durable pause) and PR-C (durable cancel) — those add new methods, this implements ``replay()``. All three target ``src/ouroboros/orchestrator/agent_process.py`` but at non-overlapping line ranges.

Part of #518.

### DIFF
--- a/src/ouroboros/orchestrator/agent_process.py
+++ b/src/ouroboros/orchestrator/agent_process.py
@@ -3,22 +3,12 @@
 Issue #518 — M6 of the Phase-2 Agent OS RFC. The five verbs ``spawn``,
 ``pause``, ``resume``, ``cancel``, and ``replay`` are the unified
 abstraction every long-running workflow consumes (ralph, evolve_step,
-execute_seed). This module is **slice 1 of #518** — the interface
-itself, an in-memory implementation that supports cooperative
-``cancel()``, ``pause()``, ``resume()``, and ``status()``, plus the
-lifecycle directive emission that lands ``control.directive.emitted``
-events with ``target_type="agent_process"``.
-
-The verbs whose durability is the headline of #518 are intentionally
-left for follow-up slices so this PR stays single-responsibility:
-
-* ``replay()`` raises :class:`NotImplementedError`. Slice 3 (#518)
-  reads the EventStore and reconstructs a timeline.
-* ``pause()`` / ``resume()`` are in-memory only here — they signal a
-  cooperative work loop via :meth:`AgentProcessHandle.should_pause`
-  but they do **not** persist a checkpoint. Slice 2 (#518) extends
-  the existing :class:`CheckpointStore` (#338) so pause survives a
-  process restart.
+execute_seed). This module is the AgentProcess spine for #518: the
+interface, cooperative ``cancel()``, ``pause()``, ``resume()``, and
+``status()`` operations, lifecycle directive emission that lands
+``control.directive.emitted`` events with ``target_type="agent_process"``,
+durable pause checkpoints, and ``replay()`` projection from the persisted
+lifecycle journal.
 
 Cooperative semantics, locked here:
 
@@ -71,6 +61,7 @@ logger = logging.getLogger(__name__)
 
 _TARGET_TYPE: Final[str] = "agent_process"
 _EMITTED_BY: Final[str] = "agent_process"
+_DIRECTIVE_EMIT_TIMEOUT_SECONDS: Final[float] = 1.0
 
 
 class AgentProcessStatus(StrEnum):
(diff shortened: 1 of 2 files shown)


shaun0927:
@ouroboros-agent please re-review the latest head (87d0095). The follow-up commits address the prior design notes by failing closed on empty and partial/stale AgentProcess lifecycle histories, adding lost terminal/pause regression tests, and updating the stale module docstring.

shaun0927:
@ouroboros-agent please re-review latest head 5818560. This update changes transition ordering so live status is updated only after the directive emit attempt completes, and adds a regression test proving replay during an in-flight pause append returns the last durable status instead of reporting corruption.

shaun0927:
@ouroboros-agent please re-review latest head b20cb47. The latest commit restores liveness by advancing live status and terminal waiters before awaiting journal I/O, while tracking a pending emit status so replay treats in-flight appends as transient durable lag and only fails stale histories after the emit attempt finishes.

shaun0927:
@ouroboros-agent please re-review latest head b8eb405. CI is green. The latest changes address the previous liveness finding by advancing status/completed_event before journal I/O, tracking pending emit status for transient replay lag, and updating the stale test header.

shaun0927:
@ouroboros-agent please re-review current head 318f8ec. All required checks are green on this head; the implementation now preserves liveness under stalled journal appends and fails closed only after an emit attempt completes with durable history still empty/stale.

shaun0927:
## Merge-readiness review

This PR implements the #518 AgentProcess replay slice: `AgentProcessHandle.replay()` now reconstructs an `AgentProcessSnapshot` from persisted `control.directive.emitted` lifecycle events through the existing projector, while keeping the orchestration layer decoupled from the concrete persistence implementation via a replay-capable protocol.

I reviewed the PR intent against #518 and I think the direction is valid: replay belongs in this slice, the projector reuse avoids duplicating folding rules, and the scope remains centered on AgentProcess lifecycle reconstruction rather than the later workflow migrations.

During review/repair, the PR was hardened beyond the initial implementation:

- Empty persisted lifecycle histories now fail closed instead of fabricating `RUNNING`.
- Partial/stale histories are detected even when the final projected status happens to match the live status, by validating durable directive counts as well as final status.
- Pending lifecycle emits distinguish transient journal lag from true corruption.
- Per-handle emit serialization preserves pause/resume directive order under high-latency stores.
- Directive append attempts are time-bounded so a wedged EventStore cannot leak the internal work task or keep replay in a permanent pending state.
- The stale module/test documentation was updated, and the branch was merged with current `upstream/main` to clear conflicts.

Validation evidence on the current head (`74226e3`):

- `ouroboros-agent[bot]`: **APPROVE** — no remaining blocking findings.
- GitHub checks: Ruff Lint, MyPy Type Check, Bridge TypeScript, Test Python 3.12, Test Python 3.13, Test Python 3.14, enforce-envelope, and enforce-boundary are all green.
- PR is mergeable.

Given the bot approval, green CI, resolved merge conflict, and the added failure-mode coverage for empty, partial, stale, in-flight, reordered, and wedged lifecycle journals, I consider this PR merge-ready.

PR #852, opened 2026-05-11
feat(auto): 3-step DomainProfile activation in ooo auto CLI (#809 P3, PR 3/6)

shaun0927:
## Summary

Wires the 3-step DomainProfile activation policy into the ``ooo auto`` CLI. Stacks on **#849** (PR-1: contracts).

## Activation order

1. ``--domain <name>`` explicit flag wins.
2. Otherwise, ``DEFAULT_REGISTRY.detect_best(cwd)`` picks the highest-confidence profile.
3. Otherwise, ``active_domain_profile_name`` stays ``None`` — the pre-P3 hardcoded coding logic remains in charge (the safety hatch PR-4/PR-5 will use).

## Changes

- ``src/ouroboros/auto/state.py``: new ``active_domain_profile_name: str | None`` field on ``AutoPipelineState``, JSON-safe (stores the name, not the instance, so resume works).
- ``src/ouroboros/cli/commands/auto.py``: ``--domain`` option + 3-step resolution block.
- ``tests/unit/cli/test_auto_domain_activation.py``: 4 tests covering the explicit flag, auto-detect, none-fallback, and unknown-domain error.

## Test plan

- 4 new tests covering each activation step.
- Broader ``pytest tests/unit/cli/ tests/unit/auto/`` → **1454 passed**, no regressions.
- ``ruff check`` / ``ruff format --check`` clean.

## Why behavior is unchanged for current users

Until PR-4 (``answerer.py`` routing) and PR-5 (``safe_defaults.py`` routing) land, the activated profile is informational only — ``answerer`` and ``safe_defaults`` keep using their hardcoded coding logic. This PR just makes the activation observable on ``AutoState`` so subsequent PRs have something to read.

## Stack

Stacks on **#849** (PR-1: DomainProfile contracts). Independent of PR-2 (#TBD — coding profile), PR-4, PR-5 — they consume the activated name but don't change the wiring this PR adds.

Part of #809 P3.

PR #850, opened 2026-05-11
feat(auto): second built-in research DomainProfile + plurality acceptance (#809 P3, PR 6/6)

shaun0927:
## Summary

- Adds `src/ouroboros/auto/profiles/research.py`: a second built-in `DomainProfile` registered via `DEFAULT_REGISTRY.register()` — zero core changes required
- Adds `src/ouroboros/auto/profiles/__init__.py`: registers `research` (and conditionally `coding` when PR-2 lands)
- Adds `tests/unit/auto/test_research_profile.py`: 10 unit tests covering detector, intent classifier, predicates, vague terms, and registry registration
- Adds `tests/integration/auto/test_domain_profile_plurality.py`: acceptance test proving plurality — `detect_best` routes `references.bib` repos to `research`, Python repos to `coding` (skipped until PR-2 merges), and no core file was touched

Stacks on PR #849 (PR-1, domain profile contracts).

## Test plan

- [ ] `PYTHONPATH=src python -m pytest tests/unit/auto/test_research_profile.py tests/integration/auto/test_domain_profile_plurality.py -v` → 14 passed, 1 skipped (coding profile skip is expected until PR-2 merges)
- [ ] `ruff check` + `ruff format --check` → all clean
- [ ] No files outside `profiles/` were modified

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #849, opened 2026-05-11
feat(auto): DomainProfile and VerifiablePredicate contracts (#809 P3, PR 1/6)

shaun0927:
## Summary

- Introduces `DomainProfile`, `VerifiablePredicate`, `IntentClassifier`, and `RepoContextExtractor` as the contract layer for the Pillar A refactor (#809 P3).
- Adds `DomainProfileRegistry` with `register`, `get`, `all`, `detect_best` (highest confidence, ties by registration order), and `union_predicates` (monorepo support).
- Exposes `DEFAULT_REGISTRY` as the module-level singleton that PR-2 will populate with the `coding` profile.
- **No existing caller is modified** — `answerer.py`, `safe_defaults.py`, and all other modules are untouched.

## Exported surface (`src/ouroboros/auto/domain_profile.py`)

| Symbol | Kind |
|---|---|
| `VerifiablePredicate` | `Protocol` (`code`, `matches`, `repair_template`) |
| `IntentClassifier` | `Protocol` (`classify`, `supported_intents`) |
| `RepoContextExtractor` | `Protocol` (`extract`) |
| `DomainProfile` | `frozen dataclass` (7 fields + `find_verifiable_predicate`) |
| `DomainProfileRegistry` | class (`register`, `get`, `all`, `detect_best`, `union_predicates`) |
| `DEFAULT_REGISTRY` | module-level singleton |

## Test plan

- [x] `test_verifiable_predicate_protocol_smoke` — minimal stub satisfies Protocol
- [x] `test_intent_classifier_protocol_smoke`
- [x] `test_repo_context_extractor_protocol_smoke`
- [x] `test_domain_profile_is_frozen` — mutation raises `FrozenInstanceError`
- [x] `test_find_verifiable_predicate_returns_first_match`
- [x] `test_registry_rejects_duplicate_name`
- [x] `test_registry_get_returns_none_for_unknown`
- [x] `test_registry_get_returns_registered_profile`
- [x] `test_registry_detect_best_picks_highest_confidence`
- [x] `test_registry_detect_best_breaks_ties_by_registration_order`
- [x] `test_registry_detect_best_returns_none_when_empty`
- [x] `test_registry_union_predicates_applies_threshold`
- [x] `test_default_registry_starts_empty` — singleton is empty at PR-1

All 13 tests pass; `ruff check` and `ruff format --check` clean.

## Why observational

[...]

Issue #809, opened 2026-05-08
RFC: ooo auto evolution to domain-agnostic self-healing E2E


Q00:
## Status

**Proposed (2026-05-09)** — companion to the accepted #725 UserLevel plugin RFC and `docs/rfc/userlevel-plugins.md`. Where #725 defines *what* sits above core, this RFC defines *how `ooo auto`, the first first-party UserLevel program, evolves to honor that vision*.

## Vision

`ooo auto` is not a standalone automation tool. It is the **first first-party UserLevel program** running on top of Ouroboros core (#725).

Target end state:

- One-line goal → Seed → execute → **self-evaluate → self-heal** until convergence
- Domain-agnostic (coding / design / research / planning / …)
- When verification itself becomes impossible due to environment constraints, lateral thinking reroutes the verification path
- Every decision is either deterministic, or — when determinism cannot decide — calls a human

This RFC closes 8 gaps between today's `ooo auto` and that target state.

---

## What Already Works (Preserve)

These are the foundation. Do not touch.

| Asset | Location | Role |
|---|---|---|
| Driver bounding | `interview_driver.py` | timeout + max_rounds caps LLM runaway |
| Deterministic AutoAnswerer | `answerer.py` | pattern-matched answers + source tagging |
| Deterministic A-grade gate | `grading.py` | vague-term ban, observable-AC enforcement, 5-metric scoring |
| SeedRepairer | `seed_repairer.py` | bounded deterministic repair + fingerprint loop guard |
| Ledger provenance | `ledger.py` | evidence-backed vs assumption-only tracking |
| Run-handoff safety | `pipeline.py` | unknown_no_handle / unknown_timeout duplicate-execution guard |
| Resume durability | `state.py` | every phase persisted, exact-point resume |

---

## Gap Matrix

| # | Gap | Priority | Phase |
|---|---|---|---|
| 5 | AutoAnswerer ignores user preferences | Quick win | P1 |
| 4 | \`ambiguity_score\` provenance unclear / determinism leak | Quick win | P1 |
| 2 | No self-evolutionary loop after RUN | Critical | P2 |
| 3 | No automated AC verification against run output | Critical | P2 |
| 1 | Heavy bias toward Python/coding domain | Critical (large) | P3 |
| 7 | AutoAnswerer 1834 LOC monolith | Long-term (folds into #1) | P3 |
| 8 | Ledger conflict resolution policy not centralized | Medium | P4 |
[...]

PR #853, opened 2026-05-11
feat(auto): route safe_defaults through DomainProfile (#809 P3, PR 5/6)

shaun0927:
## Summary

Routes ``safe_defaults.py``'s default lookups through the active ``DomainProfile`` while keeping the hardcoded coding defaults as the no-profile fallback. Stacks on **#849** (PR-1 contracts).

## Changes

- ``src/ouroboros/auto/safe_defaults.py``:
  - New ``_resolve_spec()`` helper that prefers ``active_profile.safe_defaults`` per-key, falling back to the existing private ``_SAFE_DEFAULTS`` dict for any key the profile does not provide.
  - ``finalize_safe_defaultable_gaps`` gains an ``active_profile: DomainProfile | None = None`` param. ``DomainProfile`` imported under ``TYPE_CHECKING`` to avoid a runtime import cycle.
- ``src/ouroboros/auto/state.py``: ``active_domain_profile_name: str | None = None`` field added so the pipeline can recover the profile by name on resume.
- ``src/ouroboros/auto/interview_driver.py``: at the ``finalize_safe_defaultable_gaps`` call site, look up ``state.active_domain_profile_name`` → ``DEFAULT_REGISTRY.get(name)`` and thread the result through.

## Test plan

- 6 new tests in ``tests/unit/auto/test_safe_defaults_domain_profile.py`` covering profile override, hardcoded fallback, no-profile, dict/str spec coercion, and state-field round-trip.
- Full ``pytest tests/unit/auto/`` → **150 passed**, no regressions.
- ``ruff check`` / ``ruff format --check`` clean.

## Safety hatch

When ``active_profile is None`` the function behaves byte-for-byte like the current ``main``. Behavior changes only when a registered profile (e.g. coding from #851, research from #850) is active.

## Stack

Stacks on **#849**. Independent of PR-3 (#852) and PR-4 — they touch ``state.py`` and ``pipeline.py`` adjacent surfaces, but at different line ranges. A trivial merge resolution may apply.

Part of #809 P3.

### DIFF
--- a/src/ouroboros/auto/domain_profile.py
+++ b/src/ouroboros/auto/domain_profile.py
@@ -257,8 +257,71 @@ class DomainProfileRegistry:
         return tuple(result)
 
 
-#: Module-level singleton registry.  PR-2 will register the ``coding`` profile here.
-DEFAULT_REGISTRY: DomainProfileRegistry = DomainProfileRegistry()
+class _CodingRepoContextExtractor:
+    """Minimal built-in coding profile repo context extractor.
+
+    Full repository fact extraction remains owned by the auto answerer / driver
+    integration.  The profile still needs a concrete extractor so the built-in
+    ``coding`` profile can satisfy the DomainProfile contract at production
+    bootstrap time.
+    """
+
+    def extract(self, cwd: Path) -> dict[str, Any]:  # noqa: ARG002
+        return {}
+
+
+class _CodingIntentClassifier:
+    """Minimal classifier surface for the built-in coding profile."""
+
+    _SUPPORTED = frozenset(
+        {
+            "acceptance_criteria",
+            "actor_io",
+            "failure_modes",
+            "non_goals",
+            "product_behavior",
+            "runtime_context",
+            "verification",
+            "verification_plan",
+        }
+    )
+
+    def classify(self, question: str) -> str | None:  # noqa: ARG002
+        return None
+
+    def supported_intents(self) -> frozenset[str]:
+        return self._SUPPORTED
+
+
+def _coding_detector(cwd: Path) -> float:
+    """Return a conservative confidence that *cwd* is a coding project."""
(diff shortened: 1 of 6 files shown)


shaun0927:
Ready for merge from my side.

I reviewed the PR intent and the bot design notes, then updated the branch to keep the DomainProfile routing contract intact on the latest base. The final head is `e7998aaaf78ee0ca82134d10977667abc516245a`.

What changed during the readiness pass:
- Rebased the PR onto the current upstream `main`, dropping already-landed stack commits while preserving the PR-specific safe-default routing changes.
- Routed safe-default finalization through the active `DomainProfile` instead of raw-writing profile-scoped assumptions, so profile invariants and unknown-domain semantics are preserved.
- Made missing requested profiles fail closed instead of silently falling back to coding defaults.
- Made malformed profile `safe_defaults` fail closed for that section rather than persisting empty default ledger entries.
- Preserved legacy state resume compatibility for `active_domain_profile_name`.
- Fixed frozen `MappingProxyType` safe-default mappings so profile defaults remain immutable and still resolve correctly.
- Bootstrapped the built-in `coding` profile in `DEFAULT_REGISTRY`, covering the production max-round finalization path without relying on test-only monkeypatch registration.

Validation performed:
- Local syntax/check validation: `py_compile` on the touched auto modules/tests and `git diff --check`.
- Focused unit coverage: `tests/unit/auto/test_domain_profile_contracts.py` and `tests/unit/auto/test_safe_defaults_domain_profile.py` passed (`28 passed`).
- GitHub checks are green on the final head.
- Bot review is approved on the final head.
- GitHub reports the PR as `MERGEABLE` / `CLEAN` with `reviewDecision: APPROVED`.

Given those results, the PR now matches the intended PR #809 P3/PR 5 behavior and is merge-ready.

PR #857, opened 2026-05-11
fix(plugin): prevent argv secret leaks across firewall outputs

### LINKED ISSUE #855: fix(plugin): prevent argv secret leaks in confirmation prompts and split Bearer redaction
## Problem

The plugin firewall's argv secret-safety contract is currently weaker on two user-visible paths than the RFC implies:

1. Destructive-command confirmation prompts interpolate raw `argv`:

```text
Action: merge --token hunter2-secret
```

2. The audit-event redactor handles one-token `Bearer ...` values, but misses the shell-tokenized form commonly produced by auth headers:

```text
--authorization Bearer opaqueSecret123
```

The current event argv becomes roughly:

```python
["--authorization", "[redacted]", "opaqueSecret123"]
```

so the opaque credential tail still reaches persisted audit events.

## Why this is a contract bug

`docs/rfc/userlevel-plugins.md` frames the UserLevel plugin layer as defense-oriented: success is that the boundary holds. In the bounded-payload section, argv secret handling is explicitly defense-in-depth:

- plugin authors must not accept secrets via argv;
- the firewall still must apply built-in redaction before ledger write;
- `Bearer ...` and well-known secret flags are listed as redaction targets.

A confirmation prompt is not the ledger, but it is still an output surface: terminal transcripts, tmux logs, CI captures, and support screenshots can persist it. It should not be less safe than emitted audit events.

## Local reproduction

On `upstream/main` at `e3a01129`, these desired-behavior tests fail:

```python
def test_confirmation_prompt_redacts_secret_argv_values(...):
    result = invoke_plugin(
        program,
        command_name="merge",
        argv=["--token", "hunter2-secret"],
        confirm=lambda prompt: prompts.append(prompt) or False,
        ...
    )
    assert "hunter2-secret" not in prompts[0]


def test_split_bearer_authorization_does_not_reach_audit_events(...):
    invoke_plugin(
        program,
        command_name="review",
        argv=["--authorization", "Bearer", "opaqueSecret123"],
        ...
    )
    assert "opaqueSecret123" not in json.dumps(events)
```

Observed failures:

```text
'hunter2-secret' is contained here: Action: merge --token hunter2-secret
'opaqueSecret123' is contained here: "argv": ["--authorization", "[redacted]", "opaqueSecret123"]
```

Baseline plugin/firewall tests still pass, so this is a gap in coverage rather than a covered behavior change:

```bash
uv run pytest tests/unit/plugin/test_firewall.py tests/unit/cli/test_plugin_dispatch.py -q
# 36 passed
```

## Suggested fix

- Reuse the firewall's redacted argv for destructive confirmation prompt text, or display only an argv summary.
- Extend `_redact_argv` for authorization-like flags so `--authorization Bearer <token>` redacts the bounded two-token Bearer value as one logical secret.
- Add regression tests for both cases.

## Out of scope

- Making argv a supported secret transport. The RFC still says plugins must reject argv-supplied secrets at their own boundary.
- Arbitrary argv truncation or spill storage. #805 intentionally chose observation-only `argv_summary`; this issue is about redaction correctness, not size policy.

shaun0927:
Closes #855.

## Summary

- Reuse the firewall's redacted argv when rendering the destructive-command confirmation prompt, so `--token <value>` is not echoed to terminal/session logs before the user answers.
- Extend `_redact_argv` to treat shell-tokenized `Bearer <opaque>` as one logical secret, including after secret flags such as `--authorization` and `--bearer=Bearer`.
- Preserve the existing observation-first audit shape: argv is still present, but secret-shaped values are replaced with `[redacted]` and `argv_summary` describes the redacted payload that lands in the ledger.

## Why this fits the contract

`docs/rfc/userlevel-plugins.md` says plugin argv secret handling is defense-in-depth: plugins must not accept secrets via argv, but the firewall still redacts known secret shapes before persistence. A confirmation prompt is not the ledger, yet it is still a durable output surface in tmux logs, terminal transcripts, CI captures, and screenshots. It should not be weaker than audit events.

## What this does not do

- Does not make argv a supported secret transport.
- Does not add argv truncation, spill storage, or arbitrary caps. #805 intentionally kept size policy observation-first; this PR only fixes redaction correctness.

## Test plan

- [x] `uv run pytest tests/unit/plugin/test_firewall.py tests/unit/cli/test_plugin_dispatch.py -q` — 37 passed
- [x] `uv run ruff check src/ouroboros/plugin/firewall.py tests/unit/plugin/test_firewall.py` — clean
- [x] `uv run ruff format --check src/ouroboros/plugin/firewall.py tests/unit/plugin/test_firewall.py` — clean

## Regression coverage

- `test_confirmation_prompt_uses_redacted_argv` proves destructive confirmation text no longer contains the raw token value.
- Extended `test_argv_redacts_secret_flags_and_high_confidence_tokens` proves split Bearer marker + opaque tails are redacted and do not appear anywhere in serialized audit events.

### DIFF
--- a/src/ouroboros/plugin/firewall.py
+++ b/src/ouroboros/plugin/firewall.py
@@ -167,18 +167,28 @@ def _redact_argv(argv: list[str]) -> tuple[list[str], bool]:
     redacted: list[str] = []
     fired = False
     pending_value_redact = False
+    pending_bearer_tail_redact = False
     for token in argv:
+        if pending_bearer_tail_redact:
+            redacted.append(_REDACTED)
+            fired = True
+            pending_bearer_tail_redact = False
+            continue
         if pending_value_redact:
             redacted.append(_REDACTED)
             fired = True
             pending_value_redact = False
+            if token == "Bearer":
+                pending_bearer_tail_redact = True
             continue
         # `--flag=value` form: split on first `=`.
         if token.startswith("-") and "=" in token:
-            flag, _, _ = token.partition("=")
+            flag, _, value = token.partition("=")
             if flag in _SECRET_FLAG_NAMES:
                 redacted.append(f"{flag}={_REDACTED}")
                 fired = True
+                if value == "Bearer":
+                    pending_bearer_tail_redact = True
                 continue
         # Bare flag form: value is the next argv element.
         if token in _SECRET_FLAG_NAMES:
@@ -186,7 +196,15 @@ def _redact_argv(argv: list[str]) -> tuple[list[str], bool]:
             pending_value_redact = True
             continue
         # High-confidence value-shaped match (Bearer …, gh*_…, sk-…,
-        # AKIA…, JWT-shaped).
+        # AKIA…, JWT-shaped).  Shells commonly split an Authorization
+        # header into ``Bearer`` and the opaque credential tail; redact
+        # both adjacent tokens as the same logical secret rather than
+        # letting the tail land in audit events.
+        if token == "Bearer":
+            redacted.append(_REDACTED)
+            fired = True
+            pending_bearer_tail_redact = True
+            continue
         if _is_secret_value(token):
             redacted.append(_REDACTED)
             fired = True
@@ -195,6 +213,10 @@ def _redact_argv(argv: list[str]) -> tuple[list[str], bool]:
     if pending_value_redact:
         # Trailing `--token` with no value: nothing to redact, but the
         # plugin will reject it at parse-time anyway. Emit it as-is.
+        pass
+    if pending_bearer_tail_redact:
+        # Trailing `Bearer` without a credential tail was already
+        # redacted above. Nothing else to consume.
         pass
     return redacted, fired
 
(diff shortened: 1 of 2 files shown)


shaun0927:
## Merge-readiness review

I reviewed the intent, implementation direction, bot design notes, and the final head of this PR.

This PR is valid and directionally correct: it tightens the plugin firewall's argv redaction boundary so secrets do not escape into either persisted audit events or the user-visible destructive-command confirmation prompt. The original change correctly addressed the main leak class: secret flag values, high-confidence token shapes, full `Bearer ...` values, and split Bearer credentials introduced through explicit auth/secret flags are redacted before they reach firewall outputs.

During review, the bot identified one blocking regression: treating every standalone literal `Bearer` token as secret context could hide unrelated following arguments such as `--force`, which would make both audit logs and confirmation prompts less truthful. I agreed that this was a real merge blocker and added commit `1dc41835` to narrow the split-Bearer rule. Split Bearer tails are still redacted when introduced by known auth/secret flags such as `--authorization Bearer <token>` or `--bearer=Bearer <token>`, but a standalone ordinary `Bearer` argument now remains visible and no longer consumes the next argv token.

Additional regression coverage was added for the bot finding:

- clean argv containing standalone `Bearer`, `--force`, and plain trailing arguments stays verbatim and does not get `argv_sha256` provenance;
- destructive confirmation prompts preserve visible non-secret arguments like `Bearer --force`;
- the existing secret-redaction tests continue to verify secret flags, full Bearer strings, split Bearer auth values under explicit secret flags, and high-confidence token patterns.

Validation performed:

- `ruff format --check src/ouroboros/plugin/firewall.py tests/unit/plugin/test_firewall.py`
- `ruff check src/ouroboros/plugin/firewall.py tests/unit/plugin/test_firewall.py`
- `pytest -q tests/unit/plugin/test_firewall.py` locally: `30 passed`
- `git diff --check`
- GitHub required checks are all green on head `1dc4183529b10e3a3eb5ea4ae3e66fd52d642abb`
- latest bot review on that head is `APPROVED`
- PR state is `CLEAN` and `APPROVED`

Given the narrowed redaction semantics, targeted regression tests, all-green CI, and current bot approval, I believe this PR is now merge-ready.

PR #868, opened 2026-05-11
fix(cli/plugin): make trust and disable transitions atomic

### LINKED ISSUE #867: bug(cli/plugin): trust and disable commands can split atomic TrustStore transitions
## Summary

`ooo plugin trust` and `ooo plugin disable` should use the TrustStore atomic state-transition APIs end-to-end. The store already documents the race, but the CLI path can still split the logical transition across multiple calls.

## Why this matters

The user-level plugin trust boundary is fail-closed: a plugin should end a trust/disable race in either a trusted/re-enabled state or a disabled/revoked state. It should not end in a third state where both `trust.json` and `disabled.json` have been removed.

That third state is possible when the CLI performs a grant and a disable-clear as separate critical sections while a concurrent disable uses the atomic disable primitive.

## Repro / evidence

A deterministic interleaving demonstrates the loss:

1. `ooo plugin trust` writes `trust.json` with `TrustStore.grant(...)`.
2. Concurrent `ooo plugin disable` calls the atomic disable primitive, which writes `disabled.json` and removes `trust.json`.
3. The first `ooo plugin trust` resumes and clears `disabled.json` as a separate late step.
4. Final state: `trust.json` is gone and `disabled.json` is gone.

The TrustStore comments describe this as the exact reason to keep grant+clear-disable and disable+remove-trust inside one per-plugin lock. The CLI should preserve that contract instead of rebuilding the transition out of lower-level calls.

## Expected behavior

- `ooo plugin trust --scope ...` grants the requested scope(s) and clears a matching disable record as one per-plugin TrustStore transition.
- `ooo plugin disable` writes `disabled.json` and removes `trust.json` as one per-plugin TrustStore transition.
- Regression tests pin the CLI to the atomic primitives so future refactors do not reintroduce split state transitions.

## Suggested fix

- Route scoped `ooo plugin trust` through a single multi-scope TrustStore method that grants all requested scopes and clears `disabled.json` while holding the per-plugin lock.
- Route `ooo plugin disable` through `TrustStore.apply_disable(...)` instead of composing lower-level remove/write calls in the CLI.
- Keep the existing bare re-enable path for zero-required-permission plugins as the clear-only case.

## Duplicate check

I checked the current open issue/PR set for plugin trust/disable/atomic/race overlap before filing. The nearby open items I found are about firewall subprocess timeouts, argv secret redaction, and `plugin list` trust-read diagnostics; I did not find an open issue covering this trust/disable atomicity race.

shaun0927:
## Summary

- Route scoped `ooo plugin trust` through one TrustStore transition that grants all requested scopes and clears `disabled.json` under the same per-plugin lock.
- Route `ooo plugin disable` through `TrustStore.apply_disable(...)` so `disabled.json` write and `trust.json` removal stay atomic.
- Add TrustStore and CLI regression tests that pin trust/disable command paths to the atomic primitives.

Closes #867.

## Rationale

Recent plugin-security merges keep this area narrow and contract-driven: small fail-closed changes with explicit regression coverage around the user-level plugin trust boundary. This PR follows that pattern rather than broadening the plugin trust model.

The bug is a split-transition race: `trust` can write `trust.json`, a concurrent `disable` can atomically write `disabled.json` and remove `trust.json`, then `trust` can clear the disable record as a late separate step. The final state has neither trust nor disable state. Keeping each logical command transition inside one TrustStore critical section removes that interleaving.

I kept this as one PR because the store method and CLI call-site changes form one atomicity boundary; splitting them would either leave the CLI unprotected or add unused API surface without the caller that closes the bug.

## Tests

- `uv run pytest tests/unit/plugin/test_trust_store.py tests/unit/cli/test_plugin_command_mutating.py -q`
- `uv run ruff check src/ouroboros/plugin/trust_store.py src/ouroboros/cli/commands/plugin.py tests/unit/plugin/test_trust_store.py tests/unit/cli/test_plugin_command_mutating.py`
- `git diff --check`

### DIFF
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -2473,6 +2473,7 @@ def trust_command(
     # MUST surface that state's own corruption clearly.
     try:
         was_disabled = trust.is_disabled(name)
+        existing_record = trust.read(name) if scopes else None
     except (ValueError, OSError) as exc:
         print_error(
             f"trust state for {name!r} is unreadable: {exc}. "
@@ -2480,6 +2481,17 @@ def trust_command(
             f"then re-run `ooo plugin trust {name} --scope <...>`."
         )
         raise typer.Exit(code=1) from exc
+    prior_granted_scopes = (
+        {g.scope for g in existing_record.granted_scopes}
+        if existing_record is not None
+        and existing_record.matches_subject(
+            version=manifest.version,
+            source_type=entry.source_type,
+            source_identity=entry.source_identity,
+            artifact_digest=entry.artifact_digest,
+        )
+        else set()
+    )
 
     # ``--audit-log`` is operator-supplied and routinely points at a
     # path the user expects to exist (e.g. inside a pre-created log
@@ -2519,9 +2531,11 @@ def trust_command(
                     f"re-run `ooo plugin trust {name} --scope <...>`."
                 )
                 raise typer.Exit(code=1) from exc
+        granted_for_event = set(prior_granted_scopes)
         for scope in scopes:
             assert record is not None
             print_success(f"Granted: {scope} ({len(record.granted_scopes)} total scope(s))")
+            granted_for_event.add(scope)
 
             # Audit `trust_state` must mirror the firewall's invokability
             # view, not the fact that *some* grant was just written. A
@@ -2534,11 +2548,10 @@ def trust_command(
             # one optional scope, where the user grants only the
             # optional one. Compute the same predicate the firewall
             # uses post-grant.
-            granted_after = {g.scope for g in record.granted_scopes}
             required_scopes = {p.scope for p in manifest.permissions if p.required}
             event_trust_state = (
                 "trusted"
-                if granted_after and not (required_scopes - granted_after)
+                if granted_for_event and not (required_scopes - granted_for_event)
                 else "installed"
             )
 
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -2667,7 +2667,7 @@ def disable_command(
     # a recovery hint instead of a raw traceback in the very command
     # operators run to repair that state.
     try:
-        removed_trust = (trust.root / name / "trust.json").is_file()
+        removed_trust = trust.has_trust_record(name)
         if not removed_trust:
             # Disabling an already-untrusted plugin is valid: the disable
             # record is an independent revocation signal. However, when
@@ -2682,7 +2682,7 @@ def disable_command(
                 candidate_root = candidate_root.expanduser()
                 if candidate_root == trust.root.expanduser():
                     continue
-                if (candidate_root / name / "trust.json").is_file():
+                if TrustStore(root=candidate_root).has_trust_record(name):
                     print_error(
                         f"no trust grant for {name!r} exists under {trust.root}, "
                         f"but a grant exists under {candidate_root}; pass the "
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -2500,12 +2500,13 @@ def trust_command(
         )
         raise typer.Exit(code=1) from exc
     try:
-        for scope in scopes:
+        record = None
+        if scopes:
             try:
-                record = trust.grant(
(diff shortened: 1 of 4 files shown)


shaun0927:
Autopilot review completed.

This PR is valid and merge-ready. Its intent is to close the trust/disable split-transition race from #867: `ooo plugin trust --scope ...` now grants the requested scopes and clears `disabled.json` through one TrustStore critical section, while `ooo plugin disable` writes `disabled.json` and removes `trust.json` through `TrustStore.apply_disable(...)`. That keeps the final state fail-closed and prevents the bad interleaving where both `trust.json` and `disabled.json` disappear.

I made one follow-up hardening commit (`20f32cba`) before approval: the new disable preflight was checking `(trust.root / name / "trust.json").is_file()` directly, which preserved the atomic transition but bypassed the TrustStore plugin-name/path-boundary validation used elsewhere. The fix adds `TrustStore.has_trust_record()` and routes those existence checks through it, with regression coverage for invalid plugin names.

Validation evidence:
- Local: `uv run pytest tests/unit/plugin/test_trust_store.py tests/unit/cli/test_plugin_command_mutating.py -q` → 116 passed.
- Local: targeted `ruff check` → passed.
- Local: `git diff --check upstream/main...HEAD` → passed.
- GitHub checks on the latest head (`20f32cba`) are all green: Ruff, MyPy, Bridge TypeScript, Python 3.12/3.13/3.14 tests, enforce-boundary, and enforce-envelope.
- PR merge state is `CLEAN`, and no bot/reviewer objections are present after polling the PR feedback.

Given the narrow scope, the regression tests pinning both CLI call paths to the atomic TrustStore primitives, the added path-boundary hardening, and the green CI, I consider this safe to merge.

Q00 (approved):
APPROVE

Reviewed locally in a fresh clone. Focused checks passed:

- `uv run pytest tests/unit/plugin/test_trust_store.py tests/unit/cli/test_plugin_command_mutating.py -q` → 117 passed
- focused `uv run ruff check ...` over changed trust/plugin files → passed

The trust/disable/remove transitions now use single critical-section primitives (`grant_many_and_clear_disable`, `apply_disable`, `wipe_subject`), and the CLI call sites use those primitives. GitHub CI is green and the PR is mergeable. No blocking issues found.

shaun0927:
Post-approval update after the final review cycle.

This PR was valid in scope and direction: it fixes the plugin trust/disable race from #867 by moving the CLI state transitions onto TrustStore single-critical-section primitives. Scoped `ooo plugin trust` now persists all requested grants and clears `disabled.json` through `grant_many_and_clear_disable(...)`, while `ooo plugin disable` writes the disable record and removes `trust.json` through `apply_disable(...)`. That removes the split-transition interleaving where a concurrent trust/disable sequence could leave both `trust.json` and `disabled.json` absent.

During review, two follow-up fixes were added:

1. `20f32cba` — kept disable preflight path checks behind a TrustStore method (`has_trust_record`) so plugin-name/path-boundary validation still runs before deriving trust-root paths.
2. `1f89e86` — resolved the bot's design note about multi-scope audit semantics. Persistence remains atomic, but each per-scope `plugin.trusted` audit event now computes `trust_state` at that event's conceptual grant step instead of reusing the final batched grant state for every event. A regression test now covers the two-required-scope case (`installed` for the first event, `trusted` for the second).

Final validation evidence:
- Local targeted tests: `uv run pytest tests/unit/cli/test_plugin_command_mutating.py::test_trust_multi_scope_audit_events_reflect_incremental_grant_state tests/unit/plugin/test_trust_store.py tests/unit/cli/test_plugin_command_mutating.py -q` → 117 passed.
- Local focused `ruff check` over the changed trust/plugin files → passed.
- `git diff --check upstream/main...HEAD` → passed before push.
- GitHub CI on `1f89e86` was green across Ruff, MyPy, Bridge TypeScript, Python 3.12/3.13/3.14 tests, enforce-boundary, and enforce-envelope.
- `ouroboros-agent[bot]` re-reviewed commit `1f89e86` and returned `APPROVE`; Q00 also approved after local validation.
- The PR reached `APPROVED` / mergeable state and has now been merged as `c6a985b`.

Given the narrow diff, the atomicity regression coverage, the audit-semantics regression coverage, and the bot/human approvals, this reached a merge-safe level.

PR #865, opened 2026-05-11
fix(security): contain auto Seed persistence paths

shaun0927:
## Summary

Fixes #861.

`ooo auto` saves generated Seeds under `~/.ouroboros/seeds`, but the filename was derived directly from `seed.metadata.seed_id`. Since generated/imported Seed metadata is not a filesystem trust boundary, this PR now separates the semantic Seed ID from the filesystem filename: the persisted YAML keeps the original `metadata.seed_id`, while the filename stem percent-encodes path-sensitive bytes before writing.

## Why this shape

Recent security fixes in this repo are narrow, boundary-focused, and regression-tested:

- #413 sanitized `CheckpointStore` seed IDs and enforced containment.
- #631 enforced containment for seed-encoded project paths.
- Recent plugin/firewall changes fail closed when trust inputs are malformed.

This PR applies the same boundary policy to the auto Seed persistence surface without changing the Seed schema or narrowing the existing `SeedMetadata.seed_id` contract.

## Changes

- Encode unsafe Seed ID bytes for the YAML filename instead of interpolating raw metadata into a path.
- Preserve the original semantic `metadata.seed_id` in the persisted YAML.
- Verify the resolved output path stays directly under the Seed directory before writing.
- Add regression coverage for traversal encoding, whitespace/metadata consistency, punctuation compatibility, Windows reserved basename handling, and normal generated Seed persistence.

## Verification

- `uv run pytest -q tests/unit/auto/test_surface.py::test_auto_save_seed_encodes_path_traversal_seed_id_inside_seed_dir tests/unit/auto/test_surface.py::test_auto_save_seed_encodes_whitespace_padded_seed_id_without_metadata_drift tests/unit/auto/test_surface.py::test_auto_save_seed_encodes_previously_accepted_punctuation_seed_id tests/unit/auto/test_surface.py::test_auto_save_seed_encodes_windows_reserved_basename_seed_id tests/unit/auto/test_surface.py::test_auto_save_seed_accepts_default_seed_id_inside_seed_dir`
- `uv run pytest -q tests/unit/auto/test_surface.py tests/unit/auto/test_interview_pipeline.py`
- `uv run pytest -q tests/unit/auto tests/unit/mcp/tools/test_definitions.py`
- `uv run ruff check src/ouroboros/auto/adapters.py tests/unit/auto/test_surface.py`
- `uv run ruff format --check src/ouroboros/auto/adapters.py tests/unit/auto/test_surface.py`

Issue #861, opened 2026-05-11
security(auto): contain auto Seed persistence paths derived from metadata.seed_id


shaun0927:
## Summary

`ooo auto` persists generated Seeds with `src/ouroboros/auto/adapters.py::save_seed()`. That function interpolates `seed.metadata.seed_id` directly into a filename:

```python
path = directory / f"{seed.metadata.seed_id}.yaml"
```

Because `SeedMetadata.seed_id` is a plain string, a generated or imported Seed can use traversal segments such as `../outside` and cause the auto Seed YAML write to escape `~/.ouroboros/seeds` (or the caller-provided `seeds_dir`).

## Why this fits the project direction

Recent merged work consistently treats Seed-derived paths and IDs as untrusted boundaries:

- #413 fixed the same class for `CheckpointStore` seed IDs.
- #631 enforced containment for seed-encoded project paths.
- Recent plugin/firewall work (#807, #808, #855/#857) follows the same fail-closed security posture.

This is the same boundary issue on a newer auto persistence surface, not a new policy direction.

## Reproduction

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from ouroboros.auto.adapters import save_seed
from ouroboros.core.seed import OntologySchema, Seed, SeedMetadata

with TemporaryDirectory() as td:
    root = Path(td)
    seeds_dir = root / "seeds"
    seed = Seed(
        goal="Demo",
        ontology_schema=OntologySchema(name="demo", description="demo ontology"),
        metadata=SeedMetadata(seed_id="../outside"),
    )
    returned = Path(save_seed(seed, seeds_dir=seeds_dir))
    assert (root / "outside.yaml").exists()
    assert seeds_dir.resolve() not in returned.resolve().parents
```

Observed locally on `upstream/main` (`e3a01129`, later rebased to current `main`):

- returned path: `<tmp>/seeds/../outside.yaml`
- outside file exists: `true`
- resolved path is not inside `seeds_dir`

## Expected behavior

Auto Seed persistence should fail closed (or otherwise guarantee containment) when `seed.metadata.seed_id` is not safe for filename interpolation.

A minimal fix is to allow only filename-safe Seed IDs and verify the resolved output path remains directly under the Seed directory before writing.

## Verification target

Add regression coverage that:

- `metadata.seed_id="../outside"` raises and does not create `<tmp>/outside.yaml`.
- normal generated IDs such as `seed_safe_123` still persist under the Seed directory.

PR #878, opened 2026-05-12
fix(auto): bound encoded Seed filenames

shaun0927:
## Summary

Follow-up to #865.

#865 correctly separated semantic `SeedMetadata.seed_id` from the filesystem filename and fixed the traversal boundary from #861, but the final bot review identified a compatibility regression: percent-encoding every non-ASCII UTF-8 byte can inflate a previously valid seed id beyond common 255-byte filename component limits.

This PR keeps the #865 security boundary and adds a bounded filename-stem fallback:

- short/readable generated ids still persist as before;
- unsafe bytes are still encoded so traversal and path syntax cannot escape the managed seed directory;
- if the encoded stem would exceed the filename component budget, the adapter preserves a readable prefix and appends a SHA-256 digest suffix derived from the original semantic seed id;
- persisted YAML still keeps the original `metadata.seed_id` unchanged.

## Why this shape

`SeedMetadata.seed_id` remains semantic data, not a filesystem trust boundary. The adapter should therefore bound the filename at the filesystem edge rather than mutating, truncating, or rejecting the semantic id.

## Regression coverage

Adds tests that:

- save an 80-character CJK seed id without `ENAMETOOLONG`;
- assert the resulting filename stays within a 255-byte component limit;
- verify the YAML metadata still contains the original Unicode seed id;
- verify digest suffixes distinguish long ids with the same encoded prefix.

Existing #865 traversal, whitespace, punctuation, Windows-reserved basename, and default-id tests continue to pass.

## Validation

- `PYTHONPATH=src pytest -q` on the seven targeted `save_seed` regression tests → passed
- `PYTHONPATH=src pytest -q tests/unit/auto/test_surface.py tests/unit/auto/test_interview_pipeline.py` → `209 passed`
- `PYTHONPATH=src ruff check src/ouroboros/auto/adapters.py tests/unit/auto/test_surface.py` → passed
- `PYTHONPATH=src ruff format --check src/ouroboros/auto/adapters.py tests/unit/auto/test_surface.py` → passed
- `PYTHONPATH=src uv run mypy src/ouroboros/auto/adapters.py` → passed

Refs #865.

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Callable
 from dataclasses import dataclass
+import hashlib
 from pathlib import Path
 from typing import Any
 
@@ -30,6 +31,13 @@ _WINDOWS_RESERVED_FILENAME_STEMS = frozenset(
     {"CON", "PRN", "AUX", "NUL"}
     | {f"COM{index}" for index in range(1, 10)}
     | {f"LPT{index}" for index in range(1, 10)}
+)
+_SEED_FILENAME_SUFFIX = ".yaml"
+_SEED_FILENAME_COMPONENT_MAX_BYTES = 255
+_SEED_FILENAME_DIGEST_HEX_LENGTH = 24
+_SEED_FILENAME_DIGEST_SEPARATOR = "--"
+_SEED_FILENAME_STEM_MAX_BYTES = _SEED_FILENAME_COMPONENT_MAX_BYTES - len(
+    _SEED_FILENAME_SUFFIX.encode("utf-8")
 )
 
 
@@ -736,7 +744,7 @@ def save_seed(seed: Seed, *, seeds_dir: Path | None = None) -> str:
     directory = seeds_dir or (Path.home() / ".ouroboros" / "seeds")
     directory.mkdir(parents=True, exist_ok=True)
     seed_id = _safe_seed_id_for_filename(seed.metadata.seed_id)
-    path = directory / f"{seed_id}.yaml"
+    path = directory / f"{seed_id}{_SEED_FILENAME_SUFFIX}"
     _require_path_inside_directory(path, directory)
     path.write_text(
         yaml.dump(seed.to_dict(), default_flow_style=False, allow_unicode=True, sort_keys=False),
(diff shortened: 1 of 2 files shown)


shaun0927:
@ouroboros-agent please re-review latest head 4dbdfada. I addressed the deterministic overwrite finding by putting digest-bounded stems in a truncation namespace () that an untruncated semantic seed ID cannot emit without encoding, and added a regression test proving the long-id fallback cannot overwrite the corresponding untruncated ID.

shaun0927:
@ouroboros-agent please re-review latest head 4dbdfada. I addressed the deterministic overwrite finding by putting digest-bounded stems in a truncation namespace (`--%TRUNC%...`) that an untruncated semantic seed ID cannot emit without encoding, and added a regression test proving the long-id fallback cannot overwrite the corresponding untruncated ID.

shaun0927:
@ouroboros-agent please re-review latest head 2e339f5a. I addressed the non-blocking coverage suggestion by adding a long ASCII-only seed_id regression test, so bounding is now covered for both UTF-8 percent-encoding expansion and already-safe overlength identifiers.

shaun0927:
## Merge-readiness review — latest head `2e339f5`

I opened this follow-up after re-reviewing #865 and confirming that its final percent-encoding strategy could still regress valid semantic `SeedMetadata.seed_id` values by expanding long Unicode IDs past common filename component limits.

### What this PR fixes

This PR keeps #865's security boundary intact while closing the compatibility gap:

- traversal and other path-sensitive bytes are still encoded before writing under the managed seed directory;
- the original semantic `metadata.seed_id` remains preserved in the YAML and is not narrowed or mutated;
- encoded filename stems are now bounded to fit common 255-byte filename component limits;
- long filenames use a readable prefix plus a digest suffix derived from the original semantic seed id.

### Review-loop improvements

The PR was improved through two review iterations:

1. `e068d02` added the initial bounded prefix+digest fallback and tests for long Unicode seed IDs.
2. `4dbdfad` fixed the bot's deterministic overwrite finding by moving truncated names into a marker namespace that untruncated semantic IDs cannot emit without encoding.
3. `2e339f5` addressed the bot's follow-up coverage suggestion by adding long ASCII-only seed-id coverage, so both percent-encoding expansion and already-safe overlength inputs are tested.

### Validation evidence

Local validation passed:

- targeted long Unicode / long ASCII / truncation-namespace `save_seed` regression tests passed;
- `tests/unit/auto/test_surface.py tests/unit/auto/test_interview_pipeline.py` → `211 passed`;
- targeted `ruff check` and `ruff format --check` passed;
- targeted `mypy src/ouroboros/auto/adapters.py` passed.

GitHub validation on latest head `2e339f5` is also clean:

- `ouroboros-agent[bot]` returned `APPROVE` with no blocking findings and no non-blocking suggestions;
- all checks passed: Ruff, MyPy, Bridge TypeScript, Python 3.12/3.13/3.14 tests, `enforce-envelope`, and `enforce-boundary`;
- GitHub reports the PR as `MERGEABLE` and `reviewDecision` is `APPROVED`.

Given the narrow boundary-focused fix, regression coverage for the exact #865 gap plus the bot-discovered overwrite case, green CI, and bot approval, I consider this PR ready to merge.

PR #848, opened 2026-05-11
test(orchestrator): pin three-surface AgentProcess acceptance contract (#518)

shaun0927:
## Summary

Slice 6 of M6 (#518). This PR now pins the three-surface AgentProcess lifecycle contract against the real production start surfaces instead of inline stand-ins.

## What's in this PR

- `src/ouroboros/orchestrator/agent_process.py`
  - Adds `run_with_agent_process(...)`, a shared helper that runs a production JobManager runner through `AgentProcess.spawn` while preserving the existing JobManager return/failure contract.
- `src/ouroboros/mcp/tools/evolution_handlers.py`
  - Routes `StartEvolveStepHandler` background work through `AgentProcess` with intent `evolve_step`.
- `src/ouroboros/mcp/tools/ralph_handlers.py`
  - Routes `RalphHandler` background work through `AgentProcess` with intent `ralph`.
- `src/ouroboros/mcp/tools/execution_handlers.py`
  - Routes `StartExecuteSeedHandler` background work through `AgentProcess` with intent `execute_seed`.
- `tests/integration/orchestrator/test_agent_process_three_surface_acceptance.py`
  - Calls the real public handler methods for all three production start surfaces.
  - Verifies each surface emits `continue` on spawn and `converge` on clean completion.
  - Verifies job-runner cancellation is mirrored into `cancel` lifecycle directives.

## Review-driven correction

The original version only exercised three trivial inline coroutines with different intent strings. The bot review correctly flagged that as insufficient because it tested `AgentProcess` itself, not production adoption. The latest commits fix that by moving the production start surfaces onto a shared `AgentProcess.spawn` boundary and making the acceptance test fail if any of those real handler paths stops using that boundary. Follow-up bot reviews also caught cancellation/error propagation and result-level failure risks in the shared wrapper; the wrapper now owns and cancels the spawned work task, re-raises original exceptions so JobManager keeps actionable error text, maps returned `MCPToolResult(is_error=True)` to AgentProcess failure while returning the original result to JobManager, applies the same cancel/join cleanup on timeout, and preserves interrupted/cancelled result semantics from either `meta.action` or `meta.status` instead of collapsing every `is_error=True` result into failure, and keeps JobManager terminal mapping aligned with that same vocabulary, and bounds cleanup joins so cancellation-resistant workers cannot wedge timeout/cancel callers.

## Test plan

[...]

PR #880, opened 2026-05-12
fix(orchestrator): keep AgentProcess cancellation owned until work exits

shaun0927:
## Summary

Follow-up to #848. This closes the cancellation-ownership gap found after merge: `run_with_agent_process()` must not publish terminal `cancelled` lifecycle state while the underlying production work task is still alive.

## What changed

- External cancellation now keeps the wrapper task attached to the spawned AgentProcess work task until that work actually exits, then re-raises cancellation to preserve JobManager semantics.
- Timeout cleanup still requests/cancels the work task, but if the work remains pending after the bounded grace period it returns the timeout without emitting terminal `cancelled`; the AgentProcess runner will publish terminal cancellation later only when the work exits.
- Acceptance coverage now proves:
  - stubborn timeout workers do **not** get a terminal cancel directive while still running;
  - external cancellation does not complete the wrapper until stubborn work stops.

## Why

Python asyncio cannot force-kill a task that swallows `CancelledError`. The safe contract is therefore: request cancellation immediately, but do not claim terminal lifecycle completion until the owned work task has actually stopped.

## Test plan

- `PYTHONPATH=src uv run pytest tests/integration/orchestrator/test_agent_process_three_surface_acceptance.py tests/unit/orchestrator/test_agent_process.py tests/unit/mcp/tools/test_ralph_handler.py tests/unit/mcp/tools/test_start_execute_seed_idempotency.py tests/unit/mcp/tools/test_round5_fixes.py -q` — 132 passed.
- `PYTHONPATH=src uv run pytest -q` — 7985 passed, 3 skipped.
- `PYTHONPATH=src uv run ruff check .` — clean.
- `PYTHONPATH=src uv run ruff format --check .` — clean.
- `uv run mypy src` — clean.

Fixes the post-merge blocker identified on #848.

### DIFF
--- a/src/ouroboros/orchestrator/agent_process.py
+++ b/src/ouroboros/orchestrator/agent_process.py
@@ -907,22 +907,50 @@ async def run_with_agent_process[T](
             error_box.append(exc)
             raise
 
-    process = AgentProcess(event_store=event_store)
-    handle = await process.spawn(intent=intent, work_fn=_work)
-    try:
-        final_status = await handle.wait_until_complete(timeout=timeout)
-    except (asyncio.CancelledError, TimeoutError):
+    async def _request_work_cancellation() -> asyncio.Task[None] | None:
         await handle.cancel(reason="cancelled by job runner")
         work_task = handle._work_task
         if work_task is not None and not work_task.done():
             work_task.cancel()
-            done, _pending = await asyncio.wait({work_task}, timeout=1.0)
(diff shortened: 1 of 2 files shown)


shaun0927:
Follow-up pushed in ee920ec: removed repeated work_task.cancel() on subsequent wrapper cancellations and added double-cancel coverage. Please re-review.

Issue #830, opened 2026-05-11
RFC v2: Thin Skill (YAML) + Fat Harness — Composable Execution Invariants


Q00:
## Status

**Proposed v2 (2026-05-12).** Reframed per [@shaun0927's review](https://github.com/Q00/ouroboros/issues/830#issuecomment-4422268207). v1 is superseded — same goal, structurally stronger implementation. Follow-up to the interview-hardening RFC ([#826](https://github.com/Q00/ouroboros/pull/826) / `docs/rfc/interview-hardening.md`).

## TL;DR

`ooo run`'s leaf execution feels weaker than fat-skill libraries (e.g. [obra/superpowers](https://github.com/obra/superpowers)) because:

- The leaf executor prompt is 8 lines (`src/ouroboros/agents/code-executor.md`).
- The decomposer is profile-blind (`parallel_executor.py:_try_decompose_ac`).
- The leaf is self-approving — no external verifier exists.
- Evidence is free prose, not a typed contract.

**v1 framed this as a prompt-assembly problem (compose L1+L2+L3 markdown).** v2 reframes it as a **harness-invariants** problem: per-domain artifacts shrink to thin YAML, and behavior enforcement lives in orchestrator code, not in markdown rules the LLM is asked to obey.

This reframe preserves the "one page per new domain" economics v1 promised, while making the norms genuinely enforceable, structurally cheaper per dispatch, and free of self-verification blind spots.

## Why the reframe (v1 → v2)

v1 proposed three layers of agent-facing markdown (`core/`, `phases/`, `profiles/`). The framing assumed the behavioral gap could be closed by giving the agent more text to read.

That framing has four structural problems:

1. **Markdown rule ≠ invariant.** L1 norms (verification, evidence, no-fabrication) only fire if the LLM chooses to obey them. There is no external check that they were followed.
2. **Composition has a per-dispatch token tax.** L1+L2+L3 is concatenated into every AC's system prompt. v1 Open Question #2 already flagged this; the structural answer is that the harness, not the skill, manages context.
3. **Self-verification has same-context bias.** L2's `phase-task-end` runs in the same agent context that produced the result. The result and its check share the same hallucination distribution. Only a separate pass with a separate context can catch this class.
4. **Prose decomposition axes are weak.** v1's L3 cards (TDD axis / source axis / perspective axis) only influence the decomposer if it reads them carefully. Structured parameters consumed by harness logic are non-negotiable.

[...]

PR #882, opened 2026-05-12
feat(mcp): add ouroboros_start_evaluate fire-and-forget handler

Q00:
## Summary

- New `ouroboros_start_evaluate` MCP tool: wraps `EvaluateHandler` in a `JobManager`-backed background task and returns a `job_id` immediately, mirroring the existing `ouroboros_start_execute_seed` / `ouroboros_start_evolve_step` async-start handlers.
- Plugin (OpenCode subagent) mode remains terminal — the envelope is emitted directly, no job is enqueued.
- The synchronous `ouroboros_evaluate` is untouched and stays available for short artifacts; callers that previously timed out can switch to the async variant.

## Why

The MCP tool-call layer (Claude Code) caps tool calls at ~120s. The three-stage evaluation pipeline (mechanical + semantic + optional consensus) regularly runs longer once the working directory is non-trivial, surfacing as:

```
ouroboros_evaluate → timed out awaiting tools/call after 120s
```

The pipeline keeps running in-process but the caller never sees the verdict. Fire-and-forget removes the dependency on the tool-call timeout — callers poll `ouroboros_job_status` / `ouroboros_job_wait` and read the verdict via `ouroboros_job_result`, the same path execute_seed and evolve_step already use.

## Files

- `src/ouroboros/mcp/tools/evaluation_handlers.py` — `StartEvaluateHandler` (plugin-gate parity + JobManager dispatch)
- `src/ouroboros/mcp/tools/definitions.py` — factory + tuple wiring + union type
- `src/ouroboros/mcp/tools/__init__.py` — public exports
- `src/ouroboros/mcp/server/adapter.py` — server-side handler instantiation with shared `job_manager`
- `tests/unit/mcp/tools/test_start_evaluate.py` — new (12 tests: definition, required args, background-job dispatch, plugin-mode terminal envelope, factory wiring)
- `tests/unit/mcp/tools/test_definitions.py`, `tests/unit/mcp/tools/test_round5_fixes.py`, `tests/unit/mcp/server/test_adapter.py`, `tests/integration/mcp/test_server_adapter.py` — registry expectations updated for the new tool (count 25 → 26, name set augmented, server-creation patch dict)

## Test plan

- [x] `uv run ruff check` + `uv run ruff format` on changed files — clean
- [x] `uv run mypy src/ouroboros/mcp/tools/evaluation_handlers.py src/ouroboros/mcp/tools/definitions.py src/ouroboros/mcp/server/adapter.py` — no issues
- [x] `uv run pytest tests/unit/mcp/tools/test_start_evaluate.py` — 12/12 pass
- [x] `uv run pytest tests/unit/mcp/` — 870/870 pass
- [x] `uv run pytest tests/integration/mcp/test_server_adapter.py` — passes
- [ ] Manual: invoke `ouroboros_start_evaluate` from a real MCP client against a slow evaluation and confirm the response returns under the 120s ceiling with a usable `job_id`


Q00:
Addressed the BLOCKING finding in `bac1b3ff` — the plugin-mode dispatch in `StartEvaluateHandler` now mirrors the AC-input normalisation that `EvaluateHandler.handle` performs on the non-plugin path:

- Prefer `acceptance_criteria` (plural list); fall back to `acceptance_criterion` (singular).
- Multi-AC input is rendered as a numbered checklist into the singular field that `build_evaluate_subagent` accepts, so no list items are dropped on the wire.
- Three new regression tests in `tests/unit/mcp/tools/test_start_evaluate.py`:
  - `test_multi_acceptance_criteria_preserved_in_plugin_payload` — checklist round-trips through the payload
  - `test_singular_acceptance_criterion_still_forwarded` — single-string input still works
  - `test_plural_takes_precedence_over_singular` — precedence matches the inner handler

Local: ruff + mypy clean, 152/152 targeted tests pass.

Q00:
On the second review against `bac1b3f` — the BLOCKING finding is on `src/ouroboros/orchestrator/profile_loader.py:123`, which is from `798debe5` (`feat(orchestrator): profile YAML schema + loader (#830 PR 1/9)`). That commit does not belong to this PR; it was accidentally grafted onto the branch during a multi-branch local workflow.

Rather than expand this PR's scope to fix unrelated code, I've reverted `798debe5` on this branch in `1a431c3c`. The PR now contains only the two commits intended for `ouroboros_start_evaluate`:

- `0756d4c1` feat(mcp): add ouroboros_start_evaluate fire-and-forget handler
- `bac1b3ff` fix(mcp): forward acceptance_criteria list through start_evaluate plugin path
- `1a431c3c` Revert "feat(orchestrator): profile YAML schema + loader (#830 PR 1/9)" — scope hygiene

`profile_loader.py` will land separately via its own PR; the BLOCKING finding belongs there.

Local: ruff + mypy clean, 152/152 targeted tests pass.

Q00:
@shaun0927 made Evaluate with fire and forget pattern

PR #889, opened 2026-05-12
feat(orchestrator): adaptive model/tool routing (#830 PR 7/9)

Q00:
## Summary

Seventh PR in the [#830](https://github.com/Q00/ouroboros/issues/830) RFC v2 stack. **Stacked on #887**.

Implements H5 (adaptive model/tool routing). Skills never pick a model; the harness chooses a tier per role + profile + H7 retry context.

## Routing rules

| Role | Tier | Tools | Notes |
|------|------|-------|-------|
| \`DECOMPOSER\` | \`HAIKU\` | none | Planning only; HAIKU keeps the per-AC fixed cost low |
| \`EXECUTOR\` | \`SONNET\` | \`profile.suggested_tools\` | Bumps to \`OPUS\` on \`fabrication_retry\` per H7's \`ESCALATE_MODEL\` |
| \`VERIFIER\` | One tier above executor (capped at \`OPUS\`) | \`Read\` / \`Glob\` / \`Grep\` only | Read-only so verifier cannot mutate state the executor produced |

## Why abstract tiers

\`ModelTier\` is an abstract band (\`HAIKU\` / \`SONNET\` / \`OPUS\`), not a concrete model ID. The adapter layer maps to vendor SKUs in PR 9. Profile authors think in cost/quality bands rather than SKU drift.

## Wiring scope

\`parallel_executor\` still uses its hardcoded adapter call until PR 9 plumbs \`decide_route\` through dispatch.

## Test plan

- [x] \`uv run pytest tests/unit/orchestrator/test_routing.py\` — 11 passed
- [x] \`uv run ruff check\` / \`ruff format --check\` — clean
- [x] \`uv run mypy src/ouroboros/orchestrator/routing.py\` — Success

Refs #830

### DIFF
--- a/src/ouroboros/orchestrator/routing.py
+++ b/src/ouroboros/orchestrator/routing.py
@@ -46,6 +46,15 @@ class ModelTier(StrEnum):
 
 _TIER_ORDER: tuple[ModelTier, ...] = (ModelTier.HAIKU, ModelTier.SONNET, ModelTier.OPUS)
 
+# Tools the verifier route MUST drop from the executor's profile tool
+# set. The H1 contract says the verifier is read-only with respect to
+# source files, so Edit/Write/NotebookEdit are filtered. Bash is *not*
+# in this set — code profile's verifier_focus says "Run the project's
+# test command", which requires Bash to be available. Whether a Bash
+# invocation is read-only is the verifier prompt's responsibility, not
+# the router's. (Bot finding on PR #889.)
+_VERIFIER_DENIED_TOOLS: frozenset[str] = frozenset({"Edit", "Write", "NotebookEdit", "MultiEdit"})
+
 
 @dataclass(frozen=True)
 class RouteDecision:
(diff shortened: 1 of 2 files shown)


Q00:
Addressed bot BLOCKING finding(s) and follow-ups: TypeError on non-DispatchRole inputs at decide_route boundary; verifier branch now explicit, not fall-through. Pushed force-push; please re-review.

shaun0927:
@Q00

Requesting changes on the read-only verifier invariant.

`decide_route()` now derives verifier tools from the profile and strips `Edit`/`Write`, which fixes the earlier issue where the code verifier could not run tests. But `Bash` is still a mutation-capable tool, and the router comment says whether Bash is read-only is "the verifier prompt's responsibility". That moves H1 back from a harness invariant to prompt obedience.

For fat-harness semantics, verifier command execution needs a structural guard: e.g. a read-only command allowlist/test-runner, sandboxed Bash mode, or a verifier-specific tool envelope that can run tests without arbitrary mutation. Otherwise a verifier with Bash can mutate the workspace while still being described as read-only.

Also: H5 in the RFC says routing inspects AC text plus profile-declared `suggested_model_tier`; the current API has no AC input and the profile has no tier field. If that is intentionally deferred, please mark it as an explicit follow-up gate before default flip.

PR #893, opened 2026-05-12
refactor(pm): replace hardcoded model strings with config-aware getters

cohemm:
## Motivation

PM components had model strings hardcoded directly in code, and `mcp/server/adapter.py` was tightly coupled to Claude Code via `if resolved_runtime_backend == "claude":` branching. This caused two problems:

1. Changing the model via config file or environment variable had no effect on PM components
2. When running with non-Claude backends (codex, gemini, copilot), execution/validation models were left unset

### `bigbang/question_classifier.py`
- Removed `_CLASSIFIER_MODEL = "claude-sonnet-4-20250514"` constant
- Replaced with `get_clarification_model()`

### `bigbang/pm_document.py`
- Removed `_FALLBACK_MODEL = "claude-opus-4-6"` constant
- Replaced with `get_clarification_model()`

### `bigbang/pm_interview.py`
- Removed `_FALLBACK_MODEL = "claude-opus-4-6"` constant
- Replaced with `get_clarification_model()`

### `config/loader.py`
- Added `get_execution_model()`: reads `OUROBOROS_EXECUTION_MODEL` env var first, falls back to `get_clarification_model()`
- Added `get_validation_model()`: reads `OUROBOROS_VALIDATION_MODEL` env var first, falls back to `get_clarification_model()`
- Both follow the same pattern as existing getters in the module

### `config/__init__.py`
- Added `get_execution_model` and `get_validation_model` to imports and `__all__`

### `mcp/server/adapter.py`
- Removed `if resolved_runtime_backend == "claude":` branching for execution/validation model resolution
- Replaced with `get_execution_model(llm_backend)` / `get_validation_model(llm_backend)`
- All runtime backends now handled uniformly

## Verification

```bash
uv run python -c "
from ouroboros.config import get_execution_model, get_validation_model, get_clarification_model
from ouroboros.bigbang.pm_document import PMDocumentGenerator
from ouroboros.bigbang.pm_interview import PMInterviewEngine
from ouroboros.bigbang.question_classifier import QuestionClassifier
print('OK')
"
# → OK
```


cohemm:
Addressed review feedback.

**Changes:**
- `pm_interview.py`: Removed `model=model` from `QuestionClassifier` constructor — passing it set `model_is_explicit=True`, silently bypassing the configured `question_classification` role profile
- `question_classifier.py`: Updated stale "Sonnet-grade" docstring to reflect the configured clarification model

Please take another look!

Q00:
@cohemm can you resolve the non blocking issue?

cohemm:
Addressed the non-blocking regression-coverage suggestion in a411620f.

What changed:
- Added regression coverage to prove the implicit default path keeps `model_is_explicit=False` for the PM flows that should remain profile-overridable.
- Fixed `PMInterviewEngine.create()` so the backend-normalized interview model is propagated to `QuestionClassifier` as an *implicit* model, rather than as an explicit pin.
- Updated the `QuestionClassifier` docstring to match the current behavior.

Tests added/updated:
- `tests/unit/bigbang/test_question_classifier.py`
  - verifies implicit default routing keeps `model_is_explicit=False`
  - verifies explicit classifier overrides still set `model_is_explicit=True`
- `tests/unit/bigbang/test_pm_interview.py`
  - verifies `PMInterviewEngine.create(model=...)` propagates the resolved model to `classifier.model` without marking it explicit
- `tests/unit/bigbang/test_pm_document_generator.py`
  - verifies the default document-generator path also remains implicit/profile-overridable

Validation:
- targeted new regression tests: passed
- broader related suite:
  - `tests/unit/bigbang/test_question_classifier.py`
  - `tests/unit/bigbang/test_pm_interview.py`
  - `tests/unit/bigbang/test_pm_document_generator.py`
  - result: `113 passed`

So this should now lock down the intended distinction between:
- backend-aware implicit defaults, and
- true explicit model pins.

PR #962, opened 2026-05-12
fix(auto): mutual-agreement closure gate for interview driver

Q00:
## Summary
- Replace two independent readiness gates (backend ambiguity score vs. driver 10-section ledger) with a single closure principle: an interview closes only when both agree on the same turn; disagreement is reframed as the next answer.
- Delete the auto-path safe-default finalization that short-circuited backend follow-ups — when the backend signals premature closure the driver now answers the first open gap so the backend re-scores instead of unilaterally winning.
- Emit a structured diagnostic blocker on max_rounds exhaustion: `backend_done=… (ambiguity_score=…), ledger_done=… (open_gaps=[…])` so callers can decide whether to raise the bound or sharpen the goal.

## Background
Both v0.38.0 \`ooo auto\` blocker paths were symmetric races:
- \`reached max rounds before backend marked the Seed ready\` — ledger full, backend's 2-streak ambiguity gate stuck (e.g. fuzzy success criteria).
- \`completed before auto ledger was ready: <gaps>\` — backend declared closure while 7 of 10 required sections (actors / inputs / outputs / constraints / non_goals / failure_modes / runtime_context) were still empty.

The two modules had drifted apart: the backend evolved a scalar ambiguity model and the driver a structural ledger model, with no synchronisation layer left. This PR collapses them into one gate (\`Closure = backend_done ∧ ledger_done\`).

## Test plan
- [x] \`uv run ruff check src tests\`
- [x] \`uv run ruff format --check src tests\`
- [x] \`uv run mypy src/ouroboros/auto/interview_driver.py src/ouroboros/auto/adapters.py\`
- [x] \`uv run pytest tests/unit/auto tests/unit/cli/test_auto_command.py tests/integration/auto\` — 844 passed
- [x] \`uv run pytest tests/unit/mcp tests/unit/auto/test_surface.py\` — 999 passed
- [ ] Manual: re-run \`ooo auto\` against the original failing goal (the one that produced \`auto_82859c9fd9c1\` / \`auto_5af92c62c3bd\`) and confirm the driver converges or emits the new diagnostic blocker instead of deadlocking on the contract gap.

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -811,11 +811,18 @@ def _turn_from_result(
     if not session_id:
         raise HandlerError("ouroboros_interview did not return a session_id")
     text = result.content[0].text if result.content else ""
+    raw_ambiguity = meta.get("ambiguity_score")
+    ambiguity_score: float | None
+    if isinstance(raw_ambiguity, (int, float)):
+        ambiguity_score = float(raw_ambiguity)
+    else:
+        ambiguity_score = None
     return InterviewTurn(
         question=_extract_interview_question(text, session_id=session_id),
         session_id=session_id,
         seed_ready=bool(meta.get("seed_ready")),
         completed=bool(meta.get("completed")),
+        ambiguity_score=ambiguity_score,
     )
 
 
--- a/src/ouroboros/auto/interview_driver.py
+++ b/src/ouroboros/auto/interview_driver.py
@@ -217,14 +217,44 @@ class AutoInterviewDriver:
             self._save(state)
 
             if backend_done and not ledger_done:
-                # Backend said done but ledger isn't — pick the first open gap
-                # and answer it. This drives the backend to reopen with
+                # Backend said done but ledger isn't — pick the first detected
+                # gap and answer it. This drives the backend to reopen with
                 # substantive new content; we never accept closure unilaterally.
-                first_gap = ledger.open_gaps()[0]
-                answer = self.answerer.answer_gap(first_gap, ledger, answer_context)
-                question_for_record = (
-                    f"[driver gap-reopen '{first_gap}': backend_completed=True ledger_done=False]"
-                )
+                # Mirror the safety guards that ``_answer_with_gap_steering``
+                # enforces so a backend-reported "done" against a CONFLICTING /
+                # BLOCKED / goal-missing ledger does NOT silently get a
+                # fabricated auto-answer appended — those terminal conditions
+                # must surface the unresolved conflict immediately.
+                detected_gaps = self.gap_detector.detect(ledger)
+                if not detected_gaps:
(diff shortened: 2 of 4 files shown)


Q00:
Thanks for the thorough review — addressing each finding:

**Finding 2 (BLOCKING, interview_driver.py:219) — fixed in f3034daa.**

You're right that the disagreement branch was bypassing the safety guards. The fix routes through ``gap_detector.detect()`` in that branch and applies the same terminal guards ``_answer_with_gap_steering`` already enforces:
- ``goal``-section gap → immediate blocker (cannot be auto-filled)
- ``CONFLICTING`` / ``BLOCKED`` state → immediate blocker (surface the conflict; never fabricate)

Added \`test_interview_driver_blocks_on_conflict_when_backend_signals_premature_closure\` which seeds a ``CONFLICTING`` actors entry, drives the backend to declare closure, and asserts the driver terminates with the conflict surfaced after exactly one prior backend.answer call (no fabricated gap-fill).

**Finding 1 (BLOCKING, interview_driver.py:292) — deliberate design decision, keeping as-is. Happy to discuss if there's a concrete failure mode I'm missing.**

The removal of \`finalize_safe_defaultable_gaps\` from the auto path is the change this PR is *about*, not collateral. The two readiness gates (backend ambiguity score, driver 10-section ledger) had drifted apart and were producing two symmetric deadlocks in v0.38.0 — the auto-path safe-default fallback was what let the driver paper over disagreement instead of resolving it. Tearing it out is the only way to make the mutual-agreement closure principle have teeth: if the driver can silently fabricate a ledger and declare seed_ready unilaterally, the contract collapses back to the original split-brain.

Concretely:
- \`safe_defaults.py\` and \`finalize_safe_defaultable_gaps\` remain fully exercised — every unit test for that module still passes (\`tests/unit/auto/test_safe_defaults*.py\`). The contract is unchanged at the library level; only the auto-driver call site goes away.
- Domain-profile-aware defaults still flow through any consumer that calls \`finalize_safe_defaultable_gaps\` directly (e.g. CLI or future opt-in surfaces). I deliberately did NOT touch those.
- Sessions that previously closed via safe-default fallback now do one of: (a) converge naturally as the backend re-scores once the driver answers the first open gap, (b) emit the new structured diagnostic blocker naming both readiness states + open gaps. Both are recoverable — the diagnostic tells the operator exactly whether to raise \`max_rounds\` or sharpen the goal.

The principle this PR establishes: \`Closure = backend_done ∧ ledger_done\`. Allowing safe-default to satisfy ledger_done while backend hasn't agreed is the exact bug we're fixing.

**CI**
\`Test Python 3.14\` failed on \`tests/unit/tui/test_app.py::test_subscription_reads_active_session_events\` (TUI event subscription, \`'running' != 'completed'\`) — unrelated to this PR's surface and passes locally. Likely a flake; will re-run.

Issue #869, opened 2026-05-11
[Bug] ouroboros_interview: "Reached maximum number of turns (1)" raised as unhandled Exception (v0.36.0 + main HEAD)


jcfernandez-890825:
## Summary

After applying PR #754 (`strict_mcp_config` fix) on top of v0.36.0, the recursion bug is gone but `ouroboros_interview` immediately fails with `Reached maximum number of turns (1)` raised as an unhandled `Exception` from `claude_agent_sdk._internal.query.receive_messages:852`.

The existing `error_max_turns` ResultMessage handling in `claude_code_adapter.py:854-886` / `_is_usable_max_turns_partial:1029` cannot catch this because the SDK raises the error as a generic `Exception` BEFORE the ResultMessage is yielded to the adapter.

## Environment

- `ouroboros-ai` 0.36.0 (PyPI)
- `claude-agent-sdk` 0.1.76
- Claude Code CLI 2.1.138 (Linux)
- Running INSIDE Claude Code (i.e., `CLAUDECODE=1` in parent env)
- Patched with PR #754 + post-#754 refinements from `main` (files: `mcp/server/adapter.py`, `mcp/tools/authoring_handlers.py`, `providers/claude_code_adapter.py`, `providers/factory.py`, `events/interview.py`)

## Reproduction

1. Install ouroboros 0.36.0: `uv tool install 'ouroboros-ai[mcp,claude]==0.36.0'`
2. Overwrite the 5 files above with their current `main` versions
3. Restart the MCP server so the patched code loads
4. From a running Claude Code session, call `mcp__plugin_ouroboros_ouroboros__ouroboros_interview` with any `initial_context`
5. Observe immediate failure (traceback below)

Same error occurs with tiny context (`"Simple test"`, ~30 chars) and large context (>2KB), suggesting size is NOT the trigger.

## Traceback

```
Claude Agent SDK request failed: Claude Code returned an error result: Reached maximum number of turns (1)
  File "ouroboros/providers/claude_code_adapter.py", line 806, in _execute_single_request
    async for sdk_message in _safe_query():
  File "ouroboros/providers/claude_code_adapter.py", line 798, in _safe_query
    yield await gen.__anext__()
  File "claude_agent_sdk/query.py", line 123, in query
  File "claude_agent_sdk/_internal/client.py", line 74, in process_query
  File "claude_agent_sdk/_internal/client.py", line 223, in _process_query_inner
[...]

PR #979, opened 2026-05-13
fix(auto): isolate sub-interview from parent adapter tool-call leak

Q00:
## Summary

`ooo auto` sub-interview still crashed with `Reached maximum number of turns (1)` after #869 / #974 because two new leak surfaces remained:

- `AutoHandler._authoring_interview_handler` reused the **parent's tool-capable `llm_adapter`** when constructing the nested `InterviewHandler` — the envelope patches from #869/#974 never reached this code path.
- The `socratic-interviewer` agent role and the `_build_perspective_panel_prompt` strategies contained **code-exploration / file-reading cues** that induced a `ToolUseBlock` under `max_turns=1`, consuming the only allowed turn before any text could be emitted.

## Root cause closure

Four surfaces closed with one isolation contract:

1. **Adapter inheritance** — `auto_handler._authoring_interview_handler` now forces `llm_adapter=None` and propagates `suppress_tool_use_prompt_cues=True` on every reuse path so the nested interview never inherits a tool-armed adapter from the auto driver.
2. **Authoring handler plumbing** — `InterviewHandler` accepts and forwards the suppress flag into `InterviewEngine`.
3. **Prompt cues** — `bigbang/interview.py` introduces `_TOOLLESS_INTERVIEW_BASE_PROMPT` and rewrites the answer-prefix, brownfield, and perspective-panel sections to drop file-reading / tool-use language when suppressed.
4. **Contract enforcement** — `claude_code_adapter` returns a `ToolUseBlockViolation` `ProviderError` when the SDK still emits a `ToolUseBlock` under `allowed_tools=[]`, so spy-style regression tests fail closed.

Out of scope (tracked separately): the `_ISOLATION_OVERRIDES (\"agents\", [])` SDK type mismatch, replacing the CLI completion path, and extending the same spy coverage to the standalone `ouroboros_interview` path.

## Test plan

- [x] `uv run pytest -q tests/unit/mcp/tools/test_auto_interview_construction.py` — **11 passed** (new, 1416 lines; entrypoint manifest, envelope isolation per surface — `allowed_tools/disallowed_tools/setting_sources/skills/agents/plugins/hooks`, prompt-content assertions, mocked auto flow returning plain-text question, spy adapter failing on any tool request)
- [x] `uv run pytest -q tests/unit/providers/test_claude_code_adapter.py` — **51 passed**
- [x] `uv run pytest -q tests/unit/bigbang/test_interview.py` — **66 passed** (standalone interview behavior unchanged)
- [x] `uv run pytest -q tests/unit/mcp/tools/ -k interview` — **104 passed**
- [x] `uv run mypy` on changed files — clean
- [x] **Live end-to-end**: `ooo auto` with `skip_run=true`, `max_interview_rounds=3` → 3 rounds completed, real Socratic question returned (\"Should acceptance be verified specifically as `ouroboros greet` exiting with code 0 …\"), `ambiguity_score=0.29`, **no `max_turns=1` exception, no empty responses**.

## Related

- #869 — original `ooo_interview` max_turns=1 exception
- #974 — closed the `allowed_tools=[]` SDK falsy-short-circuit leak

### DIFF
--- a/src/ouroboros/agents/socratic-interviewer.md
+++ b/src/ouroboros/agents/socratic-interviewer.md
@@ -8,9 +8,9 @@ You are an expert requirements engineer conducting a Socratic interview to clari
 - NEVER promise to build demos, write code, or execute anything
 - Another agent will handle implementation AFTER you finish gathering requirements
 
-## TOOL USAGE
-- You are a QUESTION GENERATOR. You do NOT have direct tool access.
-- The caller (main session) handles codebase reading and provides code context in answers.
+## CONTEXT BOUNDARIES
+- You are a QUESTION GENERATOR.
+- The caller provides any existing-system context in answers.
 - Your job: generate the single best Socratic question to reduce ambiguity.
 - Do NOT reference specific files or code unless they appear in previous answers.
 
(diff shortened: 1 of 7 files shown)


shaun0927:
@ouroboros-agent please review the current head `dd65baf3f`. Applied `ruff format` to the new `tests/unit/mcp/tools/test_auto_interview_construction.py` (only formatting; no behavior change). CI is now green across Ruff, MyPy, Bridge TS, enforce-boundary, enforce-envelope, and Python 3.12/3.13/3.14 tests.

Issue #978, opened 2026-05-13
Design spine: AgentOS evidence-gated delivery via TraceGuard


Q00:
## SSOT status

**Accepted as the AgentOS Evidence Gate spine, not as a new canonical substrate RFC.**

## Current implementation status (2026-05-18)

**P5 legacy self-report fallback removal is cleared.** #1086 removed the runner-managed legacy self-report acceptance fallback, #1093 hardened validation-only/docs-only evidence scope classification, and the post-#1093 broader observation passed: https://github.com/Q00/ouroboros/issues/978#issuecomment-4473197927

#978 should now remain open only as the AgentOS evidence-gate design spine / regression-monitoring anchor while remaining substrate follow-ups continue through #946, #960, #939, and #961. Future work should not restore legacy self-report acceptance; failures should be tracked as scoped evidence-gate regressions or hardening follow-ups.

This issue is a cross-cutting design anchor for AgentOS evidence-gated execution. It does not create a new standalone Track C surface under #961. Implementation is routed through the existing canonical issues:

- #920 — fat-harness / verifier loop / TraceGuard deliver gate
- #946 — Run / Step / Artifact projection and journal → manifest read model
- #960 — blocker / HITL WAIT-RESUME routing
- #939 — future plugin-defined evidence types
- #961 — hard gates and default-flip governance

The runner-managed P5 default/legacy-fallback removal is no longer blocked after the post-#1093 observation. The target product direction remains: no evidence, no completion.

## Summary

Define the AgentOS evidence-gated delivery spine that replaces unchecked **agent self-report** with a **journal-backed evidence claim → deterministic verification** loop. Adopt [rlm-forge](https://github.com/Q00/rlm-forge)'s **TraceGuard** as the verifier. The target default is: no evidence, no completion — but the default flip is gated by #961 acceptance metrics.

Related: #920 (fat harness), #946 (Run/Step/Artifact), #960 (HITL WAIT/RESUME), #961 (meta SSOT).

## Problem

Today `ooo run` accepts a leaf agent's summary-based completion report at face value. The evaluator scores the summary, not the work.

- Agent: \"AC done\" (summary)
- Harness: trusts the claim
- Evaluator: scores the summary

This contradicts Ouroboros's evidence-first identity. Fabrication, half-done ACs, and AC bypass are not caught. Self-report ≠ ground truth.

### Core loop

```
AC decomposition
→ leaf task execution (journal auto-recorded)
→ leaf agent submits typed evidence claim
[...]

Issue #961, opened 2026-05-12
AgentOS sequencing SSOT: canonical ownership and bounded slices


Q00:
# AgentOS sequencing SSOT

> [!NOTE]
> This is the compact living sequencing policy for AgentOS work. Last reconciled on 2026-08-03 against **main@525a06ab**.
>
> The historical #920-#960 inventory and its consolidation record remain available in this issue's comments and edit history. They are not copied into a manually maintained live dashboard here.

## Purpose

Keep one authoritative rule set for deciding where AgentOS work belongs and when an approval-gated design may become implementation. This issue sequences canonical RFCs, epics, and explicitly accepted slices; it does not restate their product or technical contracts.

### 1. Do not create restatement meta layers

Before opening a roadmap, epic, or RFC, check whether the same problem and decisions already belong to a canonical RFC, epic, or scoped issue.

- If the proposed issue only reframes existing scope, fold the information into the canonical owner.
- A new meta issue must own a genuinely new decision boundary that cannot be expressed in the existing canonical issue.
- Independent reproducible bugs remain independent bugs and do not wait behind roadmap metadata.

### 2. Only owner-approved bounded slices may leave an approval gate

**needs-design** or **needs-approval** describes proposed work, not authorization to materialize an implementation stack.

A bounded slice becomes accepted only when the repository owner records that acceptance on the canonical parent. The accepted slice must have a finite contract, explicit dependencies, testable acceptance criteria, and a clear completion disposition.

### 3. One active layer per approval-gated stack

For one decision boundary, keep only one implementation layer active.

- The canonical RFC or epic holds future sequencing as a checklist.
- Once an owner-approved slice is active, later dependent slices remain in that checklist rather than becoming open child issues.
- The next slice is materialized only after the active slice is disposed and the owner accepts the next boundary.
- Do not simultaneously present an RFC, restatement epic, speculative children, and their PRs as separate active work for the same contract.

## Current canonical status

Restoring this issue does not reopen any issue in the historical #920-#960 cluster. Those issues retain their individual completed, replaced, or scoped dispositions.

New AgentOS work must route to its existing canonical RFC, epic, or defect. This SSOT authorizes no implementation by itself.

## Reconciliation rule

When implementation lands:

1. reconcile the canonical parent's remaining acceptance criteria;
2. close completed children or fold unapproved remainder back into the parent;
3. remove stale active-state metadata before accepting another slice.

## Non-goals

- Maintaining a copied list of every open PR or issue.
[...]

PR #993, opened 2026-05-13
feat(harness): load AC manifests for TraceGuard deliver gate (#978 P2)

shaun0927:
## Summary

Adds the first A/B-safe #978 P2 bridge from EventStore journal reads to the evidence-manifest input expected by the future TraceGuard deliver gate.

This intentionally does **not** enforce TraceGuard verdicts or alter AC success semantics. It gives the P2 deliver-gate path a deterministic, read-only loader that can be used by observe-only wiring and later verdict enforcement.

## What changed

- Adds `src/ouroboros/harness/deliver_gate.py` with `load_ac_evidence_manifest()`.
- Prefers `execution_id` reads via `query_execution_related_events()` to avoid reused-session splicing.
- Supports session fallback for observe-only consumers that only know a session id.
- Uses `limit=None` by default so the manifest is not truncated before verification.
- Sorts queried events oldest-first before feeding the #978 P1 normalizer.
- Exports the loader from `ouroboros.harness`.
- Adds unit coverage for execution-scoped reads, session fallback, AC filtering, full-read default, chronological normalization, and input validation.

## Scope controls

- No default AC acceptance change.
- No TraceGuard verdict enforcement yet.
- No retry/redispatch/failure-taxonomy routing yet.
- No live LLM/API calls.

## Verification

- `uv run ruff check src/ouroboros/harness/deliver_gate.py src/ouroboros/harness/__init__.py tests/unit/harness/test_deliver_gate.py`
- `uv run mypy src/ouroboros/harness/deliver_gate.py tests/unit/harness/test_deliver_gate.py`
- `uv run pytest tests/unit/harness/test_deliver_gate.py tests/unit/harness/test_journal_normalizer.py -q`

Part of #978 P2 under the #961 SSOT C.2 sequence.

### DIFF
--- a/src/ouroboros/harness/__init__.py
+++ b/src/ouroboros/harness/__init__.py
@@ -5,6 +5,10 @@ Run / Stage / Step / Artifact / Verdict records for #946 plus the
 journal-to-evidence-manifest normalizer for #978.
 """
 
+from ouroboros.harness.deliver_gate import (
+    EventStoreEvidenceReader,
+    load_ac_evidence_manifest,
+)
 from ouroboros.harness.journal import (
     EvidenceEntry,
     EvidenceKind,
@@ -27,6 +31,7 @@ __all__ = [
     "ArtifactRecord",
     "EvidenceEntry",
     "EvidenceKind",
+    "EventStoreEvidenceReader",
     "EvidenceManifest",
     "RunRecord",
     "StageKind",
@@ -36,5 +41,6 @@ __all__ = [
     "VerdictOutcome",
     "VerdictRecord",
     "filter_events_for_ac",
+    "load_ac_evidence_manifest",
     "normalize_events",
 ]
--- a/src/ouroboros/harness/deliver_gate.py
+++ b/src/ouroboros/harness/deliver_gate.py
@@ -74,24 +74,36 @@ async def load_ac_evidence_manifest(
     if not normalized_ac_id:
         msg = "load_ac_evidence_manifest requires a non-blank ac_id"
         raise ValueError(msg)
-    if execution_id is None and session_id is None:
+    normalized_execution_id = _normalize_optional_anchor("execution_id", execution_id)
+    normalized_session_id = _normalize_optional_anchor("session_id", session_id)
+    if normalized_execution_id is None and normalized_session_id is None:
         msg = "load_ac_evidence_manifest requires execution_id or session_id"
         raise ValueError(msg)
 
-    if execution_id is not None:
+    if normalized_execution_id is not None:
         events = await event_store.query_execution_related_events(
-            execution_id,
+            normalized_execution_id,
             limit=limit,
         )
     else:
-        assert session_id is not None
+        assert normalized_session_id is not None
         events = await event_store.query_session_related_events(
-            session_id,
-            execution_id=execution_id,
+            normalized_session_id,
+            execution_id=normalized_execution_id,
             limit=limit,
         )
 
     return normalize_events(_chronological_events(events), ac_id=normalized_ac_id)
+
+
+def _normalize_optional_anchor(name: str, value: str | None) -> str | None:
+    if value is None:
+        return None
+    stripped = value.strip()
+    if not stripped:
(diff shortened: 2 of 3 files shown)


shaun0927:
Addressed review b4292457 on head commit 63992113. `_chronological_events()` no longer uses `BaseEvent.id` for timestamp ties; it now applies event-phase ordering (`tool.call.started`/`llm.call.requested` before matching returned events), and the new regression test covers identical timestamps with returned-first query order. Re-review requested for 63992113.

shaun0927:
Addressed review 98fb3839 on head commit db93729d. `execution_id` and `session_id` now go through strict `.strip()` validation before any EventStore query, and blank caller-provided anchors fail fast rather than falling through or querying with unusable IDs. Added regression tests for blank `execution_id` with a valid `session_id` and blank `session_id`. Re-review requested for db93729d.

shaun0927:
Addressed review c6239106 on head commit f5acd708. `load_ac_evidence_manifest()` now accepts a separate `scope_id` so production-shaped recorder events can be filtered by runtime aggregate/phase tokens even when they do not carry `data["ac_id"]`; the returned `EvidenceManifest.ac_id` remains the public AC id. Added a regression test using events with no `ac_id` payload and `aggregate_id="ac_runtime_scope"` to match the existing recorder contract. Re-review requested for f5acd708.

shaun0927:
Addressed review 467f7181 on head commit 0e4fc7bc. When both `session_id` and `execution_id` are available, the loader now calls `query_session_related_events(session_id, execution_id=...)` so session-aggregate I/O events correlated to the execution are included before normalization. `query_execution_related_events()` is now used only when no session anchor is known. Added a regression test proving session-related query is preferred when both anchors exist. Re-review requested for 0e4fc7bc.

shaun0927:
Addressed review 8512789f on head commit 20e3d38b. The loader now post-filters all queried events against the required `session_id` and/or `execution_id` anchors before chronological sorting and normalization, so the OR-shaped EventStore query cannot splice another execution/session into the deliver-gate manifest. Added a regression test with wrong-execution and wrong-session events returned by the fake store; only the event matching both anchors survives. Re-review requested for 20e3d38b.

shaun0927:
Addressed review 7da49f33 on head commit bab2a169. Session-only reads are now rejected; `execution_id` is required for every deliver-gate manifest load, and `session_id` is only an additional ownership filter when available. Updated tests to assert session-only loading fails before any EventStore query. Re-review requested for bab2a169.

shaun0927:
Addressed the approved review's non-blocking documentation suggestion on head commit 06dba6f1. The `Args`/`Raises` text now states that `execution_id` is required and that session_id is only an additional ownership filter. Re-review requested for the final head commit 06dba6f1.

shaun0927:
Merge-readiness note for #978 P2 loader slice:

This PR is intentionally scoped to a read-only, A/B-safe bridge from EventStore journal data into the #978 evidence-manifest input shape. It does not enforce TraceGuard verdicts, alter AC success semantics, add retry/redispatch routing, or flip any default path.

The bot review cycle found several real deliver-gate input safety issues, all addressed:
- timestamp ties no longer use UUID/id ordering; start/requested phases sort before returned phases;
- blank execution/session anchors now fail fast before queries;
- production-shaped events without `data["ac_id"]` are supported via a separate runtime `scope_id`;
- session-related queries are used when both session and execution anchors are known;
- OR-shaped EventStore results are post-filtered against required ownership anchors;
- session-only reads are rejected so one manifest cannot splice multiple executions/retries.

Current validation evidence:
- `ouroboros-agent` approved final head commit 06dba6f with no blocking findings.
- GitHub checks are green: Ruff, MyPy, Python 3.12/3.13/3.14 tests, boundary/envelope checks, and Bridge TypeScript.
- Local targeted checks passed for the deliver-gate loader and journal normalizer.

This is safe to merge as the #978 P2 read-only loader foundation. Actual TraceGuard verdict enforcement and failure routing remain explicitly sequenced follow-up work.

PR #1010, opened 2026-05-14
fix(harness): require current-run proof for atomic file evidence

shaun0927:
## Summary

- Tightens the newly merged #1006 atomic verifier boundary so `files_touched` claims require current-run runtime evidence, not merely an existing file under the active workspace.
- Keeps active-workspace bounds, but only accepts basename-only file proof when the transcript plausibly reports a write/update/create/generate action.
- Pins regressions for stale preexisting files, read-only file references, and targeted failed test output.

## Why this follow-up exists

While executing the #961/#920 recommended action review after #1006, I found that the merged verifier could still accept a stale workspace file as `files_touched` proof. That weakens the intended "typed evidence + verifier PASS" invariant because a prior-run file could support a current-run claim.

This PR is a narrow hardening follow-up. It does not change the #961 sequencing state, does not remove legacy self-report fallback, and does not start #978 P5.

## Validation

- `PYTHONPATH=src pytest -q tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_verifier.py tests/unit/orchestrator/test_runner.py tests/unit/cli/test_run_qa.py` → 237 passed, 1 skipped
- `uv run mypy src/ouroboros/orchestrator/verifier.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/parallel_executor_models.py`
- `uv run ruff check src/ouroboros/orchestrator/verifier.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/parallel_executor_models.py tests/unit/orchestrator/test_parallel_executor.py`
- `uv run ruff format --check src/ouroboros/orchestrator/verifier.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/parallel_executor_models.py tests/unit/orchestrator/test_parallel_executor.py`

Refs #920
Refs #961
Follow-up to #1006

PR #1006, opened 2026-05-14
Require verifier PASS for fat-harness AC acceptance

shaun0927:
## Summary

Clean replacement for #1004. Completes the remaining #920 atomic acceptance invariant by requiring a separate verifier PASS after profile typed-evidence validation in the live `parallel_executor` atomic AC boundary.

## Scope

- Adds atomic verifier verdict telemetry to `ACExecutionResult` and `execution.ac.typed_evidence.observed`.
- Fat-harness acceptance now requires:
  1. runtime success,
  2. profile typed evidence present,
  3. profile evidence validation passing,
  4. a separate `VerifierVerdict.passed == true`.
- Keeps observe-only mode isolated: injected verifiers do not run unless fat-harness enforcement is active.
- Converts operational verifier failures (`TimeoutError`, `OSError`, subprocess errors) into typed STALL verifier verdicts while still surfacing verifier programming bugs.
- Default verifier rejects final-message-only self-reporting and supports evidence only through non-final runtime transcript evidence or active-workspace file evidence.
- Bounds file proof to the active workspace (`task_cwd` or adapter working directory), rejecting absolute paths and parent traversal.
- Verifies `tests_passed` per claim against the adjacent transcript chunk of a backed successful test command; failed mixed output such as `1 failed, 3 passed` remains rejected.

## Milestone boundaries

- This is a narrow #920 completion slice for the unchecked success criterion: typed evidence **and verifier PASS** before atomic AC acceptance.
- It does not remove the legacy self-report fallback.
- It does not start #978 P5; legacy removal remains blocked on the #961 release-cycle observation window.
- #1004 is superseded only to give bots a clean, single-commit review surface after the iterative review loop became stale.

## Validation

- `PYTHONPATH=src pytest -q tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_verifier.py tests/unit/orchestrator/test_runner.py tests/unit/cli/test_run_qa.py` → 233 passed, 1 skipped
- `uv run mypy src/ouroboros/orchestrator/verifier.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/parallel_executor_models.py`
- `uv run ruff check src/ouroboros/orchestrator/verifier.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/parallel_executor_models.py tests/unit/orchestrator/test_parallel_executor.py`
[...]

Issue #920, opened 2026-05-12
Agent OS roadmap: make `ooo run` trustworthy with a fat harness execution path


shaun0927:
## Why this matters

`ooo run` should not accept work simply because a leaf agent says it is done. For the Agent OS framing to be real, each acceptance-criterion execution needs a harness-owned contract: typed evidence, external verification, bounded retry, clear blocker handling, and measurable quality gates.

This is the human-facing roadmap issue for finishing the #830 thin-profile / fat-harness work after the current substrate stack lands.

## Current state

Progress already exists in #830 and the stacked PRs around profile YAML, evidence parsing, verifier loop, failure taxonomy, routing, context governance, and ProfileBackedStrategy.

The remaining gap is that much of this is still substrate. The real `parallel_executor` / AC execution path still needs to consume these pieces coherently before the default execution behavior can change.

## Desired outcome

A maintainer can run `ooo run` through a profile-backed fat-harness path where:

- AC decomposition is profile-aware.
- Leaf outputs emit typed evidence.
- Evidence is parsed and validated by the harness.
- A separate verifier pass accepts or rejects the result.
- Failures are classified as retry, redispatch, model escalation, or blocker.
- Context is bounded and deterministic.
- The default flip is supported by before/after metrics, not vibes.

## Key decisions / open questions

- What structured profile field should describe verifier capability? For example: read-only discovery, subprocess test runner, external source checker.
- How should legitimate blocked preconditions be represented in evidence without being misclassified as missing evidence?
- Which metric harness is good enough before default flip?

## Success criteria

- [x] Built-in execution profiles contain enough structured metadata for decomposition, routing, verifier capability, and evidence obligations.
- [x] Verifier routing is structural and does not infer behavior from `verifier_focus` prose.
- [x] `parallel_executor` uses profile-aware decomposition in the live path.
- [x] Atomic AC execution requires typed evidence and verifier PASS before acceptance.
- [x] Context governance is wired into leaf dispatch.
- [x] A fat-harness mode is available as opt-in before default flip.
- [x] Baseline metrics exist for legacy vs fat-harness execution.

## Implementation slices to discuss before PRs

Likely PR count: **8**.

1. ExecutionProfile schema expansion.
2. Verifier capability routing.
3. Typed blocked evidence support.
4. Profile-aware decomposition wiring.
5. Evidence + verifier loop wiring.
6. Context governor wiring.
7. Opt-in fat-harness mode.
8. Baseline metrics / default-flip gate.

[...]

PR #1021, opened 2026-05-14
fix(harness): harden Bash file evidence proof after #1010

shaun0927:
## Summary

Follow-up to #1010 / #1006 / #920.

This PR closes one remaining current-run proof gap in the default atomic verifier: a read-only Bash command such as `grep updated src/preexisting.py` could previously contain both a claimed file path and a mutation verb, causing `files_touched` to pass even though the command only read an existing file.

Changes:

- split file-proof search text from generic claim search text
- for Bash tool invocations, ignore command text as file mutation proof and trust only result/output fields
- keep dedicated mutation tools (`Edit`, `Write`, `NotebookEdit`) as valid file proof through their tool inputs
- add a regression test for read-only Bash command text containing a write verb

## Scope

- Narrow hardening follow-up to #1010
- No #978 P5 legacy self-report removal
- No default behavior flip
- No new verifier architecture or shell-command ontology

## Validation

- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q` → 75 passed
- Targeted regression/positive checks:
  - `test_fat_harness_verifier_rejects_read_only_bash_command_with_write_word`
  - `test_fat_harness_verifier_allows_bash_generated_file_and_whole_suite_test`
  - `test_fat_harness_verifier_rejects_read_only_file_reference`
  → 3 passed
- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py` → passed
- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed

Refs #920
Refs #961
Follow-up to #1010

### DIFF
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -239,7 +239,9 @@ def _runtime_message_supports_file_reference(reference: str, message: AgentMessa
     normalized_reference = reference.strip().lower()
     if not normalized_reference:
         return False
-    text = _runtime_message_search_text(message)
+    text = _runtime_message_file_proof_text(message)
+    if not text:
+        return False
     reference_pattern = re.compile(rf"(?<![\w./-]){re.escape(normalized_reference)}(?![\w./-])")
     if not reference_pattern.search(text):
         return False
@@ -255,6 +257,25 @@ def _runtime_message_supports_file_reference(reference: str, message: AgentMessa
             text,
         )
     )
+
+
+def _runtime_message_file_proof_text(message: AgentMessage) -> str:
+    """Return text that can prove a file was touched by the current run.
+
+    For Bash tool invocations, command text is not proof by itself: read-only
+    commands such as ``grep updated src/app.py`` can contain both the claimed
+    path and mutation verbs. Trust Bash result/output fields instead. Dedicated
+    edit/write tools still expose their tool inputs because their tool identity
+    supplies the mutation semantics.
+    """
+    if message.tool_name == "Bash":
+        parts: list[str] = []
+        for key in ("result_preview", "output", "stdout", "stderr"):
+            value = message.data.get(key)
+            if isinstance(value, str):
+                parts.append(value)
+        return "\n".join(parts).lower()
+    return _runtime_message_search_text(message)
 
 
 def _looks_like_test_command(command: str) -> bool:
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -224,26 +224,37 @@ def _runtime_messages_support_file_claim(
         resolved.relative_to(base)
     except ValueError:
         return False
-    if any(_runtime_message_supports_file_reference(value, message) for message in messages):
+    if any(
+        _runtime_message_supports_file_reference(value, message, messages=messages, index=index)
+        for index, message in enumerate(messages)
+    ):
         return True
     if not resolved.exists():
         return False
     basename = candidate.name.strip().lower()
     return bool(basename) and any(
-        _runtime_message_supports_file_reference(basename, message) for message in messages
+        _runtime_message_supports_file_reference(basename, message, messages=messages, index=index)
+        for index, message in enumerate(messages)
     )
 
 
-def _runtime_message_supports_file_reference(reference: str, message: AgentMessage) -> bool:
+def _runtime_message_supports_file_reference(
+    reference: str,
+    message: AgentMessage,
+    *,
+    messages: tuple[AgentMessage, ...],
+    index: int,
+) -> bool:
     """Return True when one message plausibly reports touching a file reference."""
     normalized_reference = reference.strip().lower()
     if not normalized_reference:
         return False
     text = _runtime_message_file_proof_text(message)
     if message.tool_name == "Bash":
-        return _text_supports_file_mutation_reference(
-            text, normalized_reference
-        ) or _bash_command_mutates_file_reference(message, normalized_reference)
+        return _text_supports_file_mutation_reference(text, normalized_reference) or (
+            _bash_command_mutates_file_reference(message, normalized_reference)
+            and _runtime_message_has_following_success(messages, index)
+        )
     if message.tool_name in {"Edit", "Write", "NotebookEdit"}:
         return bool(text and _file_reference_pattern(normalized_reference).search(text))
     return _text_supports_file_mutation_reference(text, normalized_reference)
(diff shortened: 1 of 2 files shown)


shaun0927:
@ouroboros-agent Please review this narrow #1010 follow-up. It prevents read-only Bash command text containing mutation words (for example `grep updated src/preexisting.py`) from proving `files_touched`, while preserving Edit/Write tool proof and Bash result/output proof.

shaun0927:
@ouroboros-agent Please re-review after commit 269dd8366. The prior blocking finding about shell edits with no path-specific output was addressed by allowing Bash command text only when the claimed file is the explicit shell-write target (for example touch/redirection/in-place edit), while keeping read-only probes such as `grep updated src/preexisting.py` rejected.\n\nLocal validation after the fix:\n- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q` → 79 passed\n- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py` → passed\n- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed\n- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed

shaun0927:
Fresh #1010/#961 alignment follow-up review for current commit `269dd8366`:

- #1010 was directionally correct: it narrowed `files_touched` proof to current-run transcript support and avoided stale workspace/read-only path evidence, matching the #961/#920 AgentOS evidence-gate spine.
- The remaining edge after #1010 was Bash command text: a read-only command such as `grep updated src/preexisting.py` could pair a path with a mutation word. This PR keeps the #1010 boundary and hardens that case without starting #978 P5 or changing the legacy fallback policy.
- The earlier bot concern on commit `dcdd058` is addressed: current code accepts Bash command text only when the claimed file is the explicit shell-write target (`touch`, redirection, `tee`, `sed/perl -i`), while read-only Bash output must come from result/output text instead of command text.

Fresh local validation on this worktree:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py -k 'fat_harness_verifier or message_contains_test_success'` → 20 passed, 59 deselected
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_verifier.py tests/unit/orchestrator/test_runner.py tests/unit/cli/test_run_qa.py` → 250 passed, 1 skipped
- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed
- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed
- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/verifier.py src/ouroboros/orchestrator/parallel_executor_models.py` → passed

GitHub checks are currently green and the merge state is `CLEAN`; the only remaining blocker appears to be the stale `CHANGES_REQUESTED` review from the pre-fix commit. @ouroboros-agent please re-review current head `269dd8366`.

shaun0927:
@ouroboros-agent Please re-review current head bd69d1333. The previous blocking finding on commit 269dd8366 is addressed: explicit Bash shell-write syntax is now accepted as file evidence only when the tool call is followed by a successful result before the next tool call. A failed explicit write regression was added for the denied/exit-code-1 case.\n\nFresh local validation:\n- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q` → 80 passed\n- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py` → passed\n- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed\n- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed

shaun0927:
@ouroboros-agent Please re-review current head cb29761a8. The basename fallback finding is addressed: Bash command-text proof is now allowed only for exact claimed paths; basename fallback remains available for output/result text but not for command text. Added a regression where `touch generated.py` cannot prove `src/generated.py`.\n\nFresh local validation:\n- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q` → 81 passed\n- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py` → passed\n- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed\n- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed

shaun0927:
Merge-readiness review for current head `cb29761`:

This PR is valid and worth merging. Its intent is a narrow follow-up to #1010/#961/#920: close a current-run evidence gap where read-only Bash command text could combine a claimed path with mutation language (for example `grep updated src/preexisting.py`) and incorrectly satisfy `files_touched`. That is aligned with the AgentOS evidence-gate direction because `files_touched` should be backed by actual current-run transcript evidence, not by stale workspace files or self-report text.

During review, the implementation was improved through three bot-driven iterations:

1. The first version over-corrected by ignoring all Bash command text, which would reject legitimate direct shell edits such as `touch`, `sed -i`, or redirection with no path-specific output. The PR now allows Bash command text only when the claimed file is the explicit shell-write target.
2. The next version was syntax-only, so a failed shell edit could still prove `files_touched`. The PR now also requires a successful following result before Bash command text can count as file-touch proof.
3. The basename fallback could have attributed `touch generated.py` to `src/generated.py` when both shared a basename. The PR now keeps basename fallback for output/result text, but not for Bash command-text proof.

The final test coverage pins the important boundaries: read-only Bash probes are rejected, explicit successful shell writes are accepted, failed shell writes are rejected, wrong-file mutations are rejected, and basename fallback cannot be abused by Bash command text.

Validation evidence:
- Local: `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q` → 81 passed
- Local: `uv run mypy src/ouroboros/orchestrator/parallel_executor.py` → passed
- Local: `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed
- Local: `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py` → passed
- GitHub checks: Ruff, MyPy, Bridge TypeScript, boundary/envelope, and Python 3.12/3.13/3.14 tests are all green
- GitHub merge state: `CLEAN`
- `ouroboros-agent[bot]`: `APPROVE` on `cb29761`

The bot left only a non-blocking note that the shell-write detector remains regex-fragile for some equivalent Bash syntax variants. That is acceptable for this PR because the scope is conservative hardening, not a broad shell parser/ontology. The current behavior is safer than the original gap and avoids the regressions found during review, so this PR is merge-ready.

PR #1026, opened 2026-05-14
fix(orchestrator): align atomic verifier with Codex runtime evidence

shaun0927:
## Summary

Fixes the post-#1025 #978 observation blocker where typed evidence is present, schema-valid, and verifier-invoked, but atomic verification rejects real Codex runtime-backed claims as unsupported.

The controlled post-#1025 run failed at:

- `files_touched: hello.py`
- `files_touched: test_hello.py`
- `tests_passed: test_hello.py::test_hello_returns_hello`

This PR aligns verifier support matching with the real Codex CLI transcript shape while preserving anti-fabrication guardrails.

Refs #978
Refs #961

## Changes

- Preserve Codex `command_execution` output metadata (`output` / `stdout` / `stderr` / `exit_code` / `status`) on Bash `AgentMessage`s so verifier success checks can inspect the command result.
- Convert multi-file Codex `file_change.changes[]` payloads into one file-proof message per changed path instead of only the first path.
- Let `files_touched` claims compare workspace-relative typed evidence paths against absolute runtime file-change paths.
- Let same-message successful Bash events support explicit shell mutation claims when the command/result are emitted together.
- Let successful broad pytest commands (`pytest`, `python -m pytest`) support an existing pytest node-id claim, while preserving targeted-test rejection.

## Scope guardrails

- No legacy self-report fallback removal.
- No #978 P5 readiness claim.
- No default behavior expansion.
- No new AgentOS substrate surface.
- Final self-report remains excluded from verifier support; evidence still must come from non-final runtime transcript messages.

## Validation

- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -q` → 88 passed
- `uv run pytest tests/unit/orchestrator/test_codex_cli_runtime.py -q` → 53 passed
- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/codex_cli_runtime.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_codex_cli_runtime.py` → passed
- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/codex_cli_runtime.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_codex_cli_runtime.py` → passed
- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/codex_cli_runtime.py` → success
[...]

PR #1025, opened 2026-05-14
fix(harness): resolve #978 typed evidence blocker

shaun0927:
## Summary

Fixes the #978 blocker observed in batch 2b by keeping fat-harness prompt wiring, typed-evidence parsing, and acceptance-gate routing aligned:

- route fat-harness fresh runs through the profile-backed strategy so leaves see the active evidence contract
- remove the atomic prose/[TASK_COMPLETE] completion instruction when fat-harness evidence gating is active
- parse the explicit `json` evidence fence even when final output contains earlier prose/code fences
- reject the observed `{ "type": "command_result", ... }` wrapper as schema-invalid instead of treating it as parse failure
- fail closed on fat-harness resume before the legacy direct adapter path can bypass typed evidence + verifier PASS

## Root cause addressed

The observed final message contained prose plus a Python code fence before `Validation evidence:`. `extract_evidence()` selected the earliest fence, so it tried to parse `def hello():` as JSON and reported `Evidence is not valid JSON: Expecting value (line 1, col 1)`. The actual JSON fence was later in the same final message.

## Tests

- `uv run pytest -q tests/unit/orchestrator/test_evidence_schema.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_verifier.py tests/unit/orchestrator/test_runner.py` — 273 passed, 1 skipped
- `uv run pytest -q tests/unit/orchestrator/test_runner.py::TestOrchestratorRunner::test_fat_harness_resume_is_blocked_before_legacy_direct_execution tests/unit/orchestrator/test_runner.py::TestOrchestratorRunner::test_fat_harness_uses_profile_backed_prompt_contract tests/unit/orchestrator/test_evidence_schema.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_verifier.py` — 161 passed
- `uv run mypy src/ouroboros/orchestrator/evidence_schema.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/runner.py`
- `uv run ruff check src/ouroboros/orchestrator/evidence_schema.py src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/runner.py tests/unit/orchestrator/test_evidence_schema.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_runner.py`
[...]

PR #1032, opened 2026-05-14
docs(agentos): add #978 observation protocol before P5

shaun0927:
## Summary

Adds a reproducible #978 / #961 observation protocol for the post-#1026 evidence gate check before any #978 P5 fallback-removal work.

Refs #978
Refs #961

## Why

The previous two-AC controlled seed exposed two separate facts:

- post-#1025 typed evidence can become present and schema-valid;
- the two-AC seed overlaps because AC1 may create both `hello.py` and `test_hello.py`, leaving AC2 unable to produce fresh `files_touched` evidence for the same file.

This doc defines a non-overlapping single-AC controlled seed and an EventStore summary query so clean-main observations are reproducible.

## Scope guardrails

- Documentation only.
- No #978 P5 fallback removal.
- No runtime/default behavior change.
- No new AgentOS substrate surface.
- A positive controlled run remains only a signal; #961 still requires a later release/usage observation cycle before P5.

## Validation

- Documentation-only.
- Protocol reviewed against current #961/#978 gates.

### DIFF
--- a/docs/agentos/traceguard-observation-protocol.md
+++ b/docs/agentos/traceguard-observation-protocol.md
@@ -80,8 +80,7 @@ mkdir -p /tmp/character-chat-978-observation
 cd /tmp/character-chat-978-observation
 git init
 cat > controlled-hello-seed-978.yaml <<'YAML'
-name: "Controlled hello #978 observation"
-description: "Single-AC observation seed for post-#1026 TraceGuard evidence gate."
+goal: "Create a minimal Python hello function with a pytest verification."
 
 constraints:
   - "Use Python only."
@@ -194,7 +193,7 @@ A full controlled pass is stronger and should also show the CLI run succeeds.
 
 Post the result on #978, and summarize the state on #961 if it changes the SSOT.
 
-```md
+````md
 ## #978 Observation batch N — post-#1026 clean-main controlled run
 
 Date: YYYY-MM-DD TZ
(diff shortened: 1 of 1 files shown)


shaun0927:
Pushed follow-up fix `be51b53` for the ouroboros-agent reproducibility blocker.

Change made:
- The run-command block now writes `controlled-hello-seed-978.yaml` with a here-doc after resetting and recreating `/tmp/character-chat-978-observation`, so following the protocol in order no longer deletes the seed immediately before `ouroboros run`.

Validation:
- Manual doc inspection of the run-command sequence confirms the seed is created after `rm -rf` and before the run command.

@ouroboros-agent please re-review current head `be51b53`.

shaun0927:
Pushed follow-up fix `f41bff5ed` for the latest ouroboros-agent seed-schema blocker.

Changes made:
- The executable here-doc seed now uses the required top-level goal field while preserving the controlled single-AC observation body.
- The reporting template uses a four-backtick outer fence so the nested bash fence remains copy-pastable.

Validation:
- Manual inspection against existing seed examples with top-level goal + acceptance_criteria.

@ouroboros-agent please re-review current head `f41bff5ed`.

shaun0927:
Merge-readiness review completed for the current approved head.

This docs-only PR is a valid #978/#961 observation protocol before any P5 fallback-removal work. It documents a reproducible clean-main controlled run and was corrected so the executable seed is recreated after tmp reset, uses the required top-level goal field, and keeps the reporting template copy-pastable.

The bot review is APPROVED, checks are green, and the PR is mergeable. It strengthens the observation gate without authorizing P5/default-path changes.

PR #1054, opened 2026-05-16
fix(auto): pass last_question when reopening seed-ready interviews

shaun0927:
## Summary
- extend the auto interview backend contract with optional `last_question`
- pass the driver-originated gap-reopen probe through `AutoInterviewDriver` when answering a backend-completed but ledger-incomplete interview
- forward `last_question` from `HandlerInterviewBackend` to `ouroboros_interview`
- keep legacy test/function backends compatible by forwarding `last_question` only when their callable accepts it

## Why
A live `ooo auto` Codex observation reached `ouroboros_auto`, then blocked before Seed generation with:

```
Cannot record answer - the previous round is already answered and no follow-up question was provided. When reopening a completed interview (Seed-ready challenge), pass the new probe question as last_question alongside answer.
```

That means dispatch worked, but auto answered a driver-created closure/gap probe without passing the probe text to the MCP interview handler.

## Test plan
- `uv run pytest tests/unit/auto/test_interview_pipeline.py tests/unit/auto/test_adapters_client_gates.py -q`
- `uv run ruff check src/ouroboros/auto/interview_driver.py src/ouroboros/auto/adapters.py tests/unit/auto/test_interview_pipeline.py tests/unit/auto/test_adapters_client_gates.py`
- `uv run ruff format --check src/ouroboros/auto/interview_driver.py src/ouroboros/auto/adapters.py tests/unit/auto/test_interview_pipeline.py tests/unit/auto/test_adapters_client_gates.py`

Refs #1045

PR #509, opened 2026-04-29
fix(interview): allow reopening seed-ready interviews to honor session challenges

Q00:
## Summary
- The Seed-ready Acceptance Guard in \`skills/interview/SKILL.md\` instructs the main session to challenge MCP's seed-ready signal when material decisions remain unresolved, but \`record_response\` refused new answers once stored ambiguity ≤ 0.2 — a deadlock with no MCP path to record the missing decision.
- \`can_reopen\` now treats any completed interview as reopenable; the main session is the gate.
- On reopen, \`clear_stored_ambiguity()\` invalidates the now-stale ambiguity snapshot so the next seed-ready evaluation runs against the extended round history.

## Changes
- \`src/ouroboros/bigbang/interview.py\`
  - Removed unused \`_SEED_READY_THRESHOLD\` constant
  - Simplified \`can_reopen\` to \`is_complete\` with updated docstring citing the Seed-ready Acceptance Guard
  - \`record_response\` deadlock-recovery branch: clear ambiguity, log \`interview.reopened\` with \`prior_ambiguity_score\`
- \`tests/unit/bigbang/test_interview.py\`
  - Replaced \`test_record_response_when_complete\` (which asserted rejection) with \`test_record_response_reopens_seed_ready_and_invalidates_ambiguity\` covering the new behavior
  - Existing long-context summary reopen test still passes unchanged

## Checks passed
- [x] ruff check
- [x] ruff format
- [x] mypy (no issues in 2 source files)
- [x] pytest tests/unit/bigbang/ — 518 passed

## Test plan
- [x] Unit: seed-ready completed interview accepts new answer, status flips to IN_PROGRESS, ambiguity_score and breakdown reset to None
- [x] Regression: long-context summary reopen path (\`test_record_response_reopens_completed_long_context_for_summary\`) still green
- [x] Manual: real MCP session with ambiguity ≤ 0.2 — challenge seed-ready, send extra answer, confirm round appended and ambiguity recomputed on next \`generate_seed\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #405, opened 2026-04-15
bug(interview): explicit 'done' path bypasses completion_candidate_streak requirement


shaun0927:
## Summary

The interview auto-complete path correctly requires `completion_candidate_streak >= 2` (two consecutive qualifying turns) before ending the interview. However, the explicit "done" path (when the user types "done") only checks `qualifies_for_seed_completion()` without any streak requirement, allowing premature interview completion on a single lucky low-ambiguity score.

## Root Cause

```python
# authoring_handlers.py — auto-complete path (line ~1028)
if (
    live_score is not None
    and qualifies_for_seed_completion(live_score, ...)
    and state.completion_candidate_streak >= AUTO_COMPLETE_STREAK_REQUIRED  # ✅ streak checked
):
    return await self._complete_interview_response(...)

# authoring_handlers.py — explicit done path (line ~947)
if exit_score is not None and qualifies_for_seed_completion(exit_score, ...):
    return await self._complete_interview_response(...)  # ❌ no streak check
```

## Impact

- User can type "done" every turn until the ambiguity score happens to dip below threshold
- A single fluctuation produces an incomplete Seed that fails during execution
- Defeats the purpose of the streak requirement introduced in #342

## Suggested Fix

Add a minimum streak check to the done path (at least `streak >= 1`), or use the same `AUTO_COMPLETE_STREAK_REQUIRED` gate.

## Introduced In

v0.28.3 — commit `b6e53ea` (#342)

shaun0927:
**Labels**: `bug` `interview`

PR #1077, opened 2026-05-17
fix(interview): scope completion-signal heuristic to user-prefix answers

ASak1104:
## Summary

- Scope `_is_interview_completion_signal` (`mcp/tools/authoring_handlers.py:206`) so only `[from-user]` and prefix-less answers trip the heuristic.
- `[from-auto]` / `[from-code]` / `[from-research]` answers carry facts or auto-synthesised fillers, not user closure intent — they now early-return False, preventing the auto driver's echoed-question answers from accidentally triggering the shortfall branch.
- Add `tests/unit/mcp/tools/test_interview_completion_signal_prefix_guard.py` (22 cases across 5 axes) to lock the contract.
- No driver / adapter / answerer / `safe_defaults` edits; no new imports; no MCP wire-contract change.

## Discovery

This bug was found during a hands-on `ooo auto` session — it is not filed against an existing GitHub issue, so all the context a reviewer needs is embedded inline below (symptom, reproduction, environment, on-disk evidence, root-cause chain, alternatives considered, contracts preserved). If anything is missing, please ask in a review comment rather than looking for an external tracking issue.

## Why

`ooo auto "Create hello.py that prints Hello, Ouroboros!"` deterministically blocks at round 5/12 before Seed generation on Ouroboros 0.38.2 with:

```
Cannot record answer - the previous round is already answered and no follow-up
question was provided. When reopening a completed interview (Seed-ready
challenge), pass the new probe question as 'last_question' alongside 'answer'.
```

PR #1054 (`04dcd8af`) already wired `last_question` through `backend.answer` in the driver — verified to be on HEAD `a4a6f3f4`. The blocker still reproduces because the **predicate that classifies an answer as "user wants to close"** does not distinguish a human typing "no remaining ambiguity" from the auto driver echoing the LLM's closing prompt verbatim into its own answer.

Reproduced twice on HEAD:

| Surface | Session | Outcome |
|---|---|---|
| `mcp__ouroboros__ouroboros_auto` | `auto_a8e3ebe65406` / `interview_888633884693452a` | blocked at round 5, exact error verbatim |
| CLI subprocess `ooo auto` | `auto_9a2fc293080e` / `interview_4dbb737622994a69` | blocked at round 5, exact error verbatim |

## Reproduction environment

| | |
|---|---|
| Ouroboros | `0.38.2` (`pyproject.toml` `name = "ouroboros-ai"`) |
| Git HEAD | `a4a6f3f4` (latest on `main` at branch time) |
| Python | 3.14 (uv-managed source checkout via `.python-version`) |
| Claude Code | 2.1.133, in-process authoring backend (`agent_runtime_backend="claude"`, `opencode_mode="subprocess"`) |
| OS | macOS Darwin 25.5.0 |
| Goal used | `Create hello.py that prints Hello, Ouroboros! when run with python3` |

Both reproductions used the **default** `max_interview_rounds=12` and the in-process Claude SDK path (no OpenCode plugin involved). Verified by diff that `pyproject.toml` / `auto/*.py` / `mcp/tools/authoring_handlers.py` are byte-identical between the cached plugin install (`~/.claude/plugins/cache/ouroboros/ouroboros/0.38.2/`) and `origin/main` at `a4a6f3f4`.

## Root cause

The chain — every step verified against the persisted state files in `~/.ouroboros/data/`:

1. **LLM emits closing-mode prompt at r3.** Around the second qualifying score, the question generator phrases its r3 with closure language ("**Interview complete. Ambiguity: 0.00** There are no remaining questions...", "Based on the three rounds of clarification, the requirements are now fully unambiguous. ... There is no remaining ambiguity to reduce.").

2. **Auto driver's acceptance answerer echoes the question.** `auto/answerer.py:480` synthesises every conservative-default acceptance answer as

   ```
   Acceptance for {echoed_question} must cover the requested behavior directly: ...
   ```

   So the driver's r3 answer body literally contains "no remaining ambiguity" — verbatim from the LLM's prompt.

3. **Handler's heuristic misclassifies the echo.** `_is_interview_completion_signal` (`mcp/tools/authoring_handlers.py:206`) was designed for *human* "done" / "no remaining ambiguity" intent (`_INTERVIEW_COMPLETION_PHRASES` line 159, tokens check line 229). It runs on the driver's prefixed answer and returns `True`, so the handler takes the **shortfall branch** (line 1944): it does *not* call `record_response`, bumps `completion_candidate_streak` 0→1, persists the streak, and returns

   ```
   Ambiguity looks low (score=...). Stability check: 1/2. Type 'done' once more to confirm...
   ```

   with `meta.seed_ready=False`. **The r3 answer is silently dropped on the floor.**

4. **Next driver iteration finalises the interview.** Driver iter 4 sees `backend_done=False`, falls into `_answer_with_gap_steering`, answers the meta-prompt with `[from-auto][assumption] Assume a single local user...`. That body does *not* contain a completion-trigger phrase. Handler records it as r3 (using the placeholder question as `pending_question`), the subsequent score qualifies, streak 1→2, and `_complete_interview_response` (line 2107) closes the interview at `total_rounds=3`.

5. **Gap-reopen probe lands on a COMPLETED state.** Driver iter 5 now sees `turn.completed=True` but `ledger.is_seed_ready()=False` (non_goals empty). It enters the mutual-agreement gap-reopen branch (PR #962, `auto/interview_driver.py:228-266`) and sends `last_question="[driver gap-reopen 'non_goals': ...]"`. Handler at line 2046 sees `state.is_complete=True` and `state.rounds[-1].user_response is not None`, falls into the else branch, and surfaces the user-visible error.

**On-disk evidence** (`interview_*.json` vs `auto_*.json` of the same blocked session, both persisted under `~/.ouroboros/data/`) initially looked as if `last_question` were silently dropped between driver and handler — the driver's `auto_answer_log` recorded the gap-reopen probe as the question, but the handler-side interview state stored a different question text. That first hypothesis was tested by adding DIAG logging at both `auto/adapters.py:113-118` (where `last_question` is added to the `arguments` dict) and `mcp/tools/authoring_handlers.py:1305` (where the handler reads it back). The logs showed `last_question` is in fact passed truthfully end-to-end. The contradiction resolves once you observe that the shortfall branch at step 3 drops the answer **before** any reopen path even runs — the persisted r3 question text matches the LLM placeholder because the handler never popped it and the next driver iteration's answer landed against that same placeholder. The "dropped `last_question`" hypothesis was a downstream artefact of the shortfall trigger, not an actual leak.

## Fix

```python
def _is_interview_completion_signal(answer: str | None) -> bool:
    ...
    if answer is None:
        return False

    stripped = answer.lstrip().lower()
    if stripped.startswith("[from-") and not stripped.startswith("[from-user]"):
        return False

    normalized = _normalize_interview_answer(answer)
    ...
```

Plus a comment block in the docstring directing future contributors to audit the guard when introducing a new `[from-<source>]` prefix.

Rationale:

- **Prefix is the de-facto contract today.** `skills/interview/SKILL.md` enumerates the four prefixes; `bigbang/interview.py:745-754` documents three of them in the LLM-facing system prompt; `auto/safe_defaults.py:161` pins `AUTO_ANSWER_PREFIX = "[from-auto]"`; `auto/answerer.py:113` builds `f"[from-auto][{self.source.value}] {self.text}"`. This fix consumes that contract rather than introducing a new one.
- **Default-deny on unknown prefixes is the safe direction.** A new prefix that turns out to be human-origin still gives the user a path forward (type a prefix-less `done`); the docstring nudges contributors to update the guard explicitly.
- **`.lower()` covers case variation** (`[From-Auto]`, `[FROM-CODE]`).

## Why this approach (and not the alternatives I considered)

| Alternative | Why rejected |
|---|---|
| Re-introduce driver-side safe-default closure (one early candidate fix) | Direct regression of PR #962, which explicitly removed it because it "short-circuited the conversation the user actually wanted — letting the backend keep asking." |
| Source-side strip in `_feature_acceptance_answer` | Changes acceptance-answer semantics (echoing the question is a deliberate context-preservation choice in the answerer template); does not protect future answerers from the same trap; touches `auto/` instead of fixing the predicate that should never have matched in the first place. |
| Typed `AnswerEnvelope(source: Source, body: str)` Pydantic model | Cleanest long-term, but a wire-contract change touching every MCP client (CLI, plugin, opencode bridge, tests at `test_adapters_client_gates.py`, `test_interview_pipeline.py`, ...). Deferred to a follow-up RFC; this PR consumes the existing prefix contract for minimal-blast-radius. |
| Explicit `is_driver_answer: bool` kwarg through `backend.answer` and `arguments` | Same wire-contract cost as the envelope, with redundant information (the prefix already encodes source). |

## Contracts and prior PRs explicitly preserved

| Contract | Source | This PR's interaction |
|---|---|---|
| `last_question` required on post-seed-ready challenge reopen | PR #509 (`79ef2cf5`), handler `authoring_handlers.py:2046` | **Unchanged.** The fix prevents the chain that exposes this contract for auto-driver answers; the contract itself stays intact for human Seed-ready challenges. |
| Mutual-agreement closure gate, no driver-side safe-default | PR #962 (`c85feb68`), driver `auto/interview_driver.py:200-266` | **Unchanged.** Fix lives entirely in the handler. |
| Two-signal completion gate (`AUTO_COMPLETE_STREAK_REQUIRED=2`) | PR #405 (`5eb338ff`), handler line 1896 | **Unchanged.** Human "done" still bumps streak and closes after two qualifying signals. |
| Driver passes `last_question=question_for_record` | PR #1054 (`04dcd8af`), `auto/interview_driver.py:305-308` | **Unchanged.** Already on HEAD; this fix is orthogonal and preempts the trigger chain that would otherwise re-expose its symptom. |
| Plugin-mode dispatch envelope (`should_dispatch_via_plugin`) | `mcp/tools/subagent.py:241` | **Unchanged.** Plugin branch (`authoring_handlers.py:1356-1505`) never calls `_is_interview_completion_signal`. |
| PMInterviewHandler completion logic | `mcp/tools/pm_handler.py` | **Unchanged.** PM handler has no reference to `_is_interview_completion_signal`. |

## Single caller, single function, zero new imports

```
$ grep -rn _is_interview_completion_signal src/
src/ouroboros/mcp/tools/authoring_handlers.py:206:def _is_interview_completion_signal(answer: str | None) -> bool:
src/ouroboros/mcp/tools/authoring_handlers.py:1862:                    if _is_interview_completion_signal(answer):
```

Single production caller, in the same file. No cross-module import added.

### New regression suite (22 cases across 5 axes)

`tests/unit/mcp/tools/test_interview_completion_signal_prefix_guard.py`:

1. **Human completion intent still signals** — `"done"`, `"DONE"`, `"  done  "`, `"complete"`, `"no ambiguity remains"`, `"ready for seed generation"`, `"[from-user] done"`, `"[from-user][refined] No remaining ambiguity"`.
2. **Driver and fact-source prefixes never signal** — exhaustive coverage of `[from-auto][conservative_default]` with echoed closure phrase, `[from-auto][assumption]`, `[from-auto] done` (forced), `[from-code] auth uses JWT; no remaining ambiguity to reduce`, `[from-code][auto-confirmed] Python 3.12, FastAPI (pyproject.toml)`, `[from-research] Stripe rate limit is 100 rps; ready for seed generation`.
3. **Case-insensitive prefix guard** — `[From-Auto]`, `[FROM-AUTO]`, `[From-Code]`, `[FROM-RESEARCH]` all still block.
4. **Negation rule survives the prefix guard** — `[from-user] not done`, `[from-user] don't close` both stay False.
5. **Boundary cases** — `None`, `""`, leading whitespace before prefix.

### Existing regression preserved

Pre-existing parametric coverage at `tests/unit/mcp/tools/test_definitions.py:1893-1903` (five prefix-less / natural-prose cases — `"done"`, `"Yes. Close now."`, `"Correct. No remaining ambiguity. Close the interview."`, `"Yes. Lock it. Documentation-only outcomes. Done."`, `"Not done yet."`) passes unchanged.

### Full verification

```bash
uv run ruff check src/ tests/              # All checks passed!
uv run ruff format --check src/ tests/     # 824 files already formatted
uv run mypy src/ouroboros                  # Success: no issues found in 364 source files
uv run pytest tests/unit/ -q               # 8947 passed, 2 skipped in 110.21s
uv run pytest tests/unit/mcp/tools/ tests/unit/auto/ -q
                                           # 1510 passed in 16.55s
```

### Live e2e (against local source via `PYTHONPATH`)

```bash
$ cd /tmp/ooo-verify && \
  PYTHONPATH=/path/to/local/src \
  python -m ouroboros auto "Create hello.py that prints Hello, Ouroboros!..."
...
Auto session: auto_0b393e503c37
Status: complete
Seed grade: A
Interview session: interview_be08d1335ee94c3d
Seed: ~/.ouroboros/seeds/seed_e2d615190fce.yaml
Run handoff status: started
```

`Status: complete` on **both** of two consecutive runs. Pre-fix: 100% block rate on the same goal across the two reproductions captured today (one MCP, one CLI subprocess) and several earlier hands-on attempts before the root cause was isolated.

### Cached install confirms regression target

Calling `mcp__ouroboros__ouroboros_auto` against the cached `uvx --from ouroboros-ai[mcp,claude]` install (which does NOT pick up this fix) still reproduces the exact blocker — `auto_e840acfb933d` / `interview_356c2692b96b4a53` — confirming the fix targets the right code path. Once this PR merges, the cached install needs `uv tool upgrade ouroboros-ai` to pick it up.

### MCP-layer integration probe (handler entry point)

Beyond the unit tests, I wrote an integration harness that pre-builds the exact `InterviewState` the buggy moment produces (2 answered rounds + 1 placeholder whose question contains the closure phrase), then calls `InterviewHandler.handle({session_id, answer="[from-auto]... no remaining ambiguity ...", last_question=...})` — the exact dispatch the MCP server performs. With the fix: handler proceeds normally, `r3.user_response` is recorded, `meta.seed_ready=True`. Without the fix (verified earlier): handler enters shortfall, r3 stays as `user_response=None`, blocker chain unfolds.

## Not tested

- **Live `ooo auto` against an installed MCP server.** The cached `uvx --from ouroboros-ai[mcp,claude]` install does not load local source, so MCP e2e validation requires either a fresh release on PyPI or a manual `uv tool install --reinstall --editable` against this branch. CLI-via-`PYTHONPATH` and the handler-level integration harness cover the same code path that MCP would execute.
- **OpenCode plugin dispatch path.** Confirmed via grep that the plugin branch never reaches `_is_interview_completion_signal`, so no behavioural change is possible — but I did not exercise it live.
- **Long-form Codex observation run.** Out of scope for this PR; the trigger is goal-text- and LLM-output-shaped, and the prefix guard is purely additive at function top, so a Codex observation should at minimum not be worse and very likely confirms the fix.

## Out of scope (separate PRs)

- **`ouroboros_interview` resume from `COMPLETED`.** `bigbang/interview.py:430-437` still errors with `Interview is already complete` for stranded sessions. Once the shortfall trigger is closed by this PR, new sessions should not strand in COMPLETED, so the resume-from-COMPLETED hardening is best handled in its own narrowly-scoped PR.
- **Typed `AnswerEnvelope` RFC.** See "Why this approach" above — recommended as the long-term direction but out of scope here.
- **The initially suspected "`last_question` dropped between driver and handler" hypothesis.** Could not be reproduced under DIAG logging (see "Root cause" above for the test setup). The predicate gate in this PR pre-empts the chain that exposed the symptom; if a true `last_question` leak exists under a different trigger, it would now be diagnosable in isolation rather than tangled with the shortfall path.

## Review hints for `ouroboros-agent[bot]`

| Bot-typical concern | Where to look |
|---|---|
| Over-broad change applied beyond described scope | Single function (`authoring_handlers.py:206`), additive guard at function top, +20/-1 prod lines. |
| Silent regression in other callers of changed function | `grep -rn _is_interview_completion_signal src/` → single caller at `authoring_handlers.py:1862`. |
| Cross-call-path impact (PM / plugin / CLI) | PMInterviewHandler does not reference the predicate; plugin branch returns before it; CLI path goes through the same handler and is verified by live e2e. |
| Comment vs implementation mismatch | Docstring describes `[from-user]` + prefix-less as the only human-intent inputs; implementation is `startswith("[from-")` AND NOT `startswith("[from-user]")` → default-deny, matching the close-world docstring. |
| SDK / dependency contract violation | No new imports, no new MCP wire fields, no SDK kwargs. |
| Test coverage of contract boundary | 22 cases × 5 axes (above) plus pre-existing 5-case parametric coverage preserved. |
| Scope creep beyond bug | Fix B + typed envelope + H1 mystery all explicitly deferred. |
| Hidden persistence/state regression | No state shape change; existing on-disk format unchanged. |

## Footer (PR #1054-style structured fields)

- **Constraint:** `_is_interview_completion_signal` is the sole entrypoint to the handler's `_INTERVIEW_COMPLETION_*` heuristic; the guard is additive at function top and preserves the existing branches.
- **Rejected:** Source-side stripping in `_feature_acceptance_answer` (silently changes acceptance-answer semantics); safe-default driver closure (regresses PR #962); explicit `is_driver_answer` kwarg (changes MCP wire contract for every client).
- **Confidence:** high (single predicate, single caller, additive guard, exhaustive parametric coverage, live e2e ×2).
- **Scope-risk:** narrow (handler only, +20/-1 prod lines, no cross-module imports, no new MCP fields).
- **Directive:** Driver-synthesised answers (`[from-auto]`) and factual answers (`[from-code]`, `[from-research]`) must never trip the human-intent completion heuristic; only `[from-user]` and prefix-less answers do.
- **Tested:** ruff / ruff-format / mypy / pytest full suite (8947 passed); live `ooo auto` pipeline pass×2 against local source; handler-level integration harness; cached-install bug reproduction confirms regression target.
- **Not tested:** OpenCode plugin dispatch (no plugin invocation of the predicate); long-form Codex observation against an installed MCP server (cached uvx install must be upgraded separately to pick up this fix).

No external tracking issue — this is a self-contained bug report. All context the reviewer needs is in the sections above.

### DIFF
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -204,8 +204,27 @@ def _normalize_interview_answer(answer: str) -> str:
 
 
 def _is_interview_completion_signal(answer: str | None) -> bool:
-    """Return True when the answer explicitly asks to end the interview."""
+    """Return True when the answer explicitly asks to end the interview.
+
+    Only ``[from-user]`` and prefix-less answers represent human intent to close.
+    The auto driver's ``_feature_acceptance_answer`` echoes the LLM question into
+    its answer text, which can accidentally include phrases like "no remaining
+    ambiguity" and trip the shortfall branch. Gate the heuristic by prefix so
+    ``[from-auto]`` / ``[from-code]`` / ``[from-research]`` answers — which carry
+    facts or auto-generated fillers, not user intent — never enter completion.
+
+    If a new ``[from-<source>]`` prefix is introduced in
+    ``skills/interview/SKILL.md`` (or the LLM-facing prompt at
+    ``bigbang/interview.py``), audit whether that source represents direct human
+    intent and update the guard below — otherwise it is silently default-denied,
+    which is safe (the user can still close with a prefix-less ``"done"``) but
+    may surprise callers.
+    """
     if answer is None:
+        return False
+
+    stripped = answer.lstrip().lower()
+    if stripped.startswith("[from-") and not stripped.startswith("[from-user]"):
         return False
 
     normalized = _normalize_interview_answer(answer)
--- a/tests/unit/mcp/tools/test_interview_completion_signal_prefix_guard.py
+++ b/tests/unit/mcp/tools/test_interview_completion_signal_prefix_guard.py
@@ -1,0 +1,105 @@
+"""Regression tests for the prefix-aware completion-signal gate.
+
+The auto driver's ``_feature_acceptance_answer`` echoes the LLM question text
+into its answer. When the LLM emits closing-mode phrasing such as "no remaining
+ambiguity", the echoed phrase used to trip ``_is_interview_completion_signal``
+inside ``InterviewHandler.handle`` and drive the shortfall branch, which in
+turn produced a ``Cannot record answer - the previous round is already
+answered`` deadlock in ``ooo auto``.
+
+These tests pin the contract: only ``[from-user]`` and prefix-less answers
+represent human intent to close. ``[from-auto]`` / ``[from-code]`` /
+``[from-research]`` answers MUST return False regardless of their text.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from ouroboros.mcp.tools.authoring_handlers import _is_interview_completion_signal
+
+
+@pytest.mark.parametrize(
+    "answer",
+    [
+        "done",
+        "DONE",
+        "  done  ",
+        "complete",
+        "no ambiguity remains",
+        "ready for seed generation",
+        "[from-user] done",
+        "[from-user][refined] No remaining ambiguity",
+    ],
+)
+def test_human_typed_completion_intent_still_signals(answer: str) -> None:
+    """Raw user input and ``[from-user]`` answers preserve the original heuristic."""
+    assert _is_interview_completion_signal(answer) is True
+
+
+@pytest.mark.parametrize(
+    "answer",
+    [
+        "[from-auto][conservative_default] Acceptance for r3 must cover the requested "
+        "behavior; there is no remaining ambiguity to reduce.",
+        "[from-auto][assumption] Assume a single local user; ready for seed generation.",
+        "[from-auto] done",
+        "[from-code] auth uses JWT; no remaining ambiguity to reduce",
+        "[from-code][auto-confirmed] Python 3.12, FastAPI (pyproject.toml)",
+        "[from-research] Stripe rate limit is 100 rps; ready for seed generation",
+    ],
+)
+def test_driver_and_factual_prefixes_never_signal_completion(answer: str) -> None:
+    """Driver and fact-source prefixes never represent user closure intent.
+
+    Regression for the ``ooo auto`` deadlock where the auto driver's
+    acceptance template echoed the LLM question into the answer, accidentally
+    matching the human-intent heuristic and shunting the round into the
+    shortfall branch (state.rounds left with an unanswered placeholder).
+    """
+    assert _is_interview_completion_signal(answer) is False
+
+
+def test_none_answer_returns_false() -> None:
+    """A None answer must short-circuit before any string slicing."""
+    assert _is_interview_completion_signal(None) is False
+
+
+def test_empty_string_returns_false() -> None:
+    """An empty string has no completion intent."""
+    assert _is_interview_completion_signal("") is False
+
+
+def test_leading_whitespace_does_not_bypass_prefix_guard() -> None:
+    """Whitespace before the prefix must not allow ``[from-auto]`` to trip the heuristic."""
+    answer = "   [from-auto] no remaining ambiguity"
+    assert _is_interview_completion_signal(answer) is False
+
+
+@pytest.mark.parametrize(
+    "answer",
+    [
+        "[From-Auto] no remaining ambiguity",
+        "[FROM-AUTO] done",
+        "[From-Code] done",
+        "[FROM-RESEARCH] ready for seed generation",
+    ],
+)
+def test_prefix_guard_is_case_insensitive(answer: str) -> None:
+    """Prefix case must not weaken the guard.
+
+    If a future caller (or typo) emits ``[From-Auto]`` instead of the canonical
+    lowercase form, the guard must still block — otherwise the same shortfall
+    trigger reopens under a trivial case variation.
+    """
+    assert _is_interview_completion_signal(answer) is False
+
+
+def test_negation_still_blocks_human_completion_intent() -> None:
+    """Negations applied to a [from-user] prefix must keep completion=False.
+
+    Regression for the heuristic itself: the prefix guard widens the surface
+    that passes through to the heuristic; ensure the negation rule still wins.
+    """
+    assert _is_interview_completion_signal("[from-user] not done") is False
+    assert _is_interview_completion_signal("[from-user] don't close") is False


src/ouroboros/mcp/tools/authoring_handlers.py
```diff
@@ -204,10 +204,29 @@ def _normalize_interview_answer(answer: str) -> str:
...
 def _is_interview_completion_signal(answer: str | None) -> bool:
-    """Return True when the answer explicitly asks to end the interview."""
+    """Return True when the answer explicitly asks to end the interview.
+
+    Only ``[from-user]`` and prefix-less answers represent human intent to close.
+    The auto driver's ``_feature_acceptance_answer`` echoes the LLM question into
+    its answer text, which can accidentally include phrases like "no remaining
+    ambiguity" and trip the shortfall branch. Gate the heuristic by prefix so
+    ``[from-auto]`` / ``[from-code]`` / ``[from-research]`` answers — which carry
+    facts or auto-generated fillers, not user intent — never enter completion.
```
Q00, line 214: good findings. 
completion heuristic only applies to human intent prefix or pure prompt. 
let's see after applying this logic!

Q00 (approved):

PR #1156, opened 2026-05-20
fix(persistence): sanitize Windows-reserved chars in checkpoint seed_id (fixes #1155)

### LINKED ISSUE #1155: [Bug] Checkpoint write fails on Windows: _sanitize_seed_id leaves reserved char ':' in filename (WinError 123)
## Summary

On Windows, a checkpoint whose `seed_id` contains a Windows-reserved character (notably `:`) fails to write. `CheckpointStore._sanitize_seed_id` only neutralizes `/`, `\`, `..` and null bytes, so the colon survives into the filename and Windows rejects the path with `WinError 123` (The filename, directory name, or volume label syntax is incorrect).

This is hit **deterministically** during MCP background-job execution: the agent-process *cancel* checkpoint uses a `seed_id` like `ouroboros_agent_process_cancel:mcp_job:job_<id>`, which contains colons. So `execute_seed` over MCP always fails on Windows, even though interview -> seed generation succeeds.

## Environment

- Windows 11 (10.0.26200)
- `ouroboros-ai` 0.39.1 (bug also present on `main`)
- Runtime / LLM backend: claude / claude_code

## Error

```
Failed to inspect checkpoint artifact: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\Users\<user>\.ouroboros\data\checkpoints\checkpoint___ouroboros_agent_process_cancel__:mcp_job:job_19a927b41098.json' (details: {'seed_id': 'ouroboros_agent_process_cancel:mcp_job:job_19a927b41098', 'level': 0})
```

## Root cause

`src/ouroboros/persistence/checkpoint.py` -> `CheckpointStore._sanitize_seed_id`:

```python
sanitized = seed_id.replace("\x00", "")
sanitized = sanitized.replace("..", "")
sanitized = re.sub(r"[/\\]", "_", sanitized)   # only / and \
```

The regex strips POSIX path separators but not Windows-reserved filename characters `: * ? " < > |`. On Linux/macOS `:` is a legal filename character, so this never surfaces there.

## Reproduction

```python
from ouroboros.persistence.checkpoint import CheckpointStore
sid = "ouroboros_agent_process_cancel:mcp_job:job_19a927b41098"
print(CheckpointStore._sanitize_seed_id(sid))
# -> 'ouroboros_agent_process_cancel:mcp_job:job_19a927b41098'  (colon survives)
# 'checkpoint_<that>.json' is then rejected by Windows with WinError 123
```

## Suggested fix

Extend the sanitizer to also replace Windows-reserved characters:

```python
sanitized = re.sub(r'[/\\:*?"<>|]', "_", sanitized)
```

(Optionally also guard against reserved DOS device names like `CON` / `PRN` / `NUL` and trailing dots/spaces, but the colon is the one that breaks normal MCP runs on Windows.)

Happy to send a PR.

Jun-0913:
## Summary

`CheckpointStore._sanitize_seed_id` only replaced POSIX path separators (`/`, `\`), leaving Windows-reserved filename characters (`:` `*` `?` `"` `<` `>` `|`) intact. On Windows the resulting checkpoint path is invalid and the write fails with `WinError 123`.

This is hit **deterministically** during MCP background-job execution: the agent-process *cancel* checkpoint uses a `seed_id` like `ouroboros_agent_process_cancel:mcp_job:job_<id>`, whose colons break the path. `interview` -> seed generation works, but `execute_seed` over MCP always fails on Windows.

Fixes #1155.

## Change

- Extend the sanitization regex from `r"[/\\]"` to `r'[/\\:*?"<>|]'` so reserved characters become `_`.
- Update the docstring to mention Windows-reserved characters.

## Tests

Added to `TestCheckpointStorePathTraversal`:
- `test_colon_in_seed_id_is_sanitized` — the exact failing seed_id now saves and round-trips through `load()`.
- `test_windows_reserved_chars_are_sanitized` — parametrized over `: * ? " < > |`.

`tests/unit/persistence/test_checkpoint.py`: **45 passed** locally (CPython 3.14, Windows 11).

### DIFF
--- a/src/ouroboros/persistence/checkpoint.py
+++ b/src/ouroboros/persistence/checkpoint.py
@@ -183,8 +183,9 @@ class CheckpointStore:
     def _sanitize_seed_id(seed_id: str, *, max_len: int | None = None) -> str:
         """Sanitize seed_id to prevent path traversal attacks.
 
-        Strips null bytes, removes path separators and parent-directory
-        sequences, and caps length so that the **full** checkpoint
+        Strips null bytes, removes path separators, Windows-reserved
+        filename characters and parent-directory sequences, and caps
+        length so that the **full** checkpoint
         filename (prefix + seed + suffix) stays within the 255-byte
         filesystem limit.  When truncation is needed a SHA-256 hash
         fragment is appended to keep the mapping collision-resistant.
@@ -211,8 +212,13 @@ class CheckpointStore:
         # so that inputs like "x/../../PWNED" are neutralised.
         sanitized = sanitized.replace("..", "")
 
-        # Replace path separators with underscores
-        sanitized = re.sub(r"[/\\]", "_", sanitized)
+        # Replace path separators AND Windows-reserved filename characters
+        # (: * ? " < > |) with underscores. On POSIX only "/" is illegal in a
+        # filename, but Windows forbids these too; without stripping them a
+        # colon-bearing seed_id (e.g. an MCP cancel-checkpoint id like
+        # "..._cancel:mcp_job:job_<id>") yields an invalid path and the write
+        # fails with WinError 123. See https://github.com/Q00/ouroboros/issues/1155
+        sanitized = re.sub(r'[/\\:*?"<>|]', "_", sanitized)
 
         # Cap length to the remaining filename budget.
         # When truncation is required, append a hash suffix for collision resistance.
--- a/tests/unit/persistence/test_checkpoint.py
+++ b/tests/unit/persistence/test_checkpoint.py
@@ -231,6 +231,38 @@ class TestCheckpointStorePathTraversal:
         expected = checkpoint_store._base_path / "checkpoint_x___PWNED.json"
         assert expected.exists()
 
+    def test_colon_in_seed_id_is_sanitized(self, checkpoint_store: CheckpointStore) -> None:
+        """Regression (#1155): a colon-bearing seed_id (e.g. an MCP cancel-checkpoint
+        id) is sanitized so the checkpoint is writable on Windows (WinError 123)."""
+        seed_id = "ouroboros_agent_process_cancel:mcp_job:job_19a927b41098"
+        cp = CheckpointData.create(seed_id, "phase1", {"a": 1})
+        result = checkpoint_store.save(cp)
+        assert result.is_ok
+        # Colons -> underscores -> a filename valid on every platform.
+        expected = (
+            checkpoint_store._base_path
+            / "checkpoint_ouroboros_agent_process_cancel_mcp_job_job_19a927b41098.json"
+        )
+        assert expected.exists()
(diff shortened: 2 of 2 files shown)


shaun0927:
**Recommendation: APPROVE / merge as-is.**

### What I verified

- **Regex set is exact.** `r'[/\\:*?"<>|]'` matches the Windows `CreateFile` reserved set per MSDN — no over- or under-stripping.
- **Operation ordering is safe.** Null-byte strip → `..` removal → reserved-char replacement → length cap. I walked through the `.<reserved>.` → `._.` case and confirmed the broadened replacement cannot re-create `..` sequences that the prior strip already neutralized. The `..` strip running *before* the new replacement is the right call.
- **Path-traversal posture is strengthened, not weakened.** More attacker-influenced characters are neutralized before `_validate_path_containment`. Placing the new tests under `TestCheckpointStorePathTraversal` is the right home.
- **Namespace guard still holds.** `_validate_process_id_namespace` sanitizes both sides of the prefix comparison with the same function, so widening the replacement set leaves the namespace check correct.
- **Collision claim checks out.** I traced the three cancel-checkpoint callsites (`execution_handlers.py:1274`, `ralph_handlers.py:429`, `evolution_handlers.py:914`) — all construct ids of the form `mcp_job:job_<uuid4_hex>`, so the `job_<hex>` suffix preserves identity after `:` → `_` collapse. No real-world collision surface.
- **Regression test is genuine.** `test_colon_in_seed_id_is_sanitized` uses the exact real-world seed_id format from `agent_process.py` and verifies the full save→load round-trip, not just the sanitizer output. That's the right level.

### Non-blocking notes (follow-up, not for this PR)

1. **MEDIUM — latent collision footgun.** `a:b` and `a_b` now both map to `a_b`. Safe for every current caller, but worth one comment-line at the regex noting the many-to-one property so future callers don't introduce structurally ambiguous seed_id pairs.
2. **LOW — missing edge tests.** Three useful additions: (a) multiple reserved chars in one id (`a:b*c?d`), (b) a long seed_id that hits the truncation+hash path *and* contains reserved chars, (c) DOS device names (`CON`/`PRN`/`NUL`). (c) is pre-existing and orthogonal to this fix — better as a separate hardening issue if Windows CI ever cares.
3. **NIT — docstring wording.** "removes path separators, Windows-reserved filename characters and parent-directory sequences" reads as if all three are removed; in fact separators and reserved chars are *replaced* with `_`, only `..` and `\x00` are removed. A small phrasing tweak (e.g. "strips null bytes and parent-directory sequences; replaces path separators and Windows-reserved filename characters with underscores") would be more precise.

None of the above blocks this PR. The fix is minimal, surgical, well-tested, fixes a deterministic Windows MCP `execute_seed` failure, and doesn't touch any Track C / #978 substrate surface — so the AgentOS roadmap sequencing in #961 is unaffected. Ship it.

PR #1162, opened 2026-05-21
fix(codex): accept official Rust CLI binaries

### LINKED ISSUE #1159: bug(codex): _WRAPPER_MAGIC_HEADERS misidentifies OpenAI Codex CLI v0.132+ Rust binary as a wrapper
## Summary

`ouroboros.codex.cli_policy._WRAPPER_MAGIC_HEADERS` flags every Mach-O / ELF
binary as a "wrapper", which now matches the official OpenAI Codex CLI itself.
Since OpenAI v0.132+ ships codex as a Rust binary (Mach-O 64-bit on macOS),
ouroboros logs `codex_cli_runtime.cli_wrapper_detected` + `cli_no_fallback`
on every `ouroboros run workflow` invocation, then refuses to use the only
codex on PATH because the fallback search keeps rejecting Mach-O candidates
too. The result: `ouroboros run` cannot dispatch any AC to codex on a fresh
macOS install that follows OpenAI's official install path
(`brew install --cask codex` or the latest Homebrew formula).

## Impact

- Affects every user whose codex CLI was installed via the standard Homebrew
  Cask `codex` (currently 0.132.0) — `/opt/homebrew/bin/codex` is a Mach-O
  binary symlinking into `/opt/homebrew/Caskroom/codex/0.132.0/...`.
- All ACs in a parallel run fail in `Analyzing complexity...` because
  `resolve_codex_cli_path` returns the binary path but the runtime still
  treats it as a wrapper without fallback, and downstream AC dispatch then
  short-circuits with `LLM call failed: Failed to start Codex CLI`.
- `ouroboros setup --runtime codex --non-interactive` succeeds (registers
  profiles, skills, MCP) but does not change `_WRAPPER_MAGIC_HEADERS`, so
  every subsequent `ouroboros run` reproduces the failure.

## Steps to Reproduce

1. Fresh macOS, install codex via `brew install --cask codex`
   (yields `/opt/homebrew/bin/codex` → Mach-O 64-bit binary v0.132.0).
2. `pipx install 'ouroboros-ai[mcp]'`
3. `ouroboros setup --runtime codex --non-interactive`
4. Create any minimal `seed.yaml` and run
   `ouroboros run workflow seed.yaml --dry-run`.
5. Observe the stderr stream:

   ```
   [warning] codex_cli_runtime.cli_wrapper_detected
       wrapper_path=/opt/homebrew/bin/codex
       hint='Searching PATH for the real Node.js codex CLI.'
   [warning] codex_cli_runtime.cli_no_fallback
       wrapper_path=/opt/homebrew/bin/codex
   ```

   followed by `0 succeeded / N failed`.

## Expected Behavior

Ouroboros should recognise the official OpenAI Codex CLI (Rust binary) as
the actual codex executable and not search for a non-existent "real
Node.js codex" — the Node.js codex CLI was deprecated by upstream when they
rewrote codex in Rust.

## Actual Behavior

`is_wrapper_binary()` (`ouroboros/codex/cli_policy.py`) returns `True` for
any Mach-O / ELF file because the magic header set was scoped for an older
Zeude-shim era. `find_real_cli()` then rejects every other codex on PATH
(also Mach-O), and falls back to the wrapper path with `cli_no_fallback`.
The runtime continues with that path but downstream AC dispatch refuses to
treat the binary as usable.

## Suggested Fix

`_WRAPPER_MAGIC_HEADERS` should be tightened to actually identify
historical wrappers (Zeude shim, etc.) rather than every compiled binary.
Two plausible approaches:

1. Drop the magic-header heuristic entirely and rely on a small content
   signature (e.g. a known wrapper-specific symbol like `zeude` or a
   `wrapper:codex` marker string).
2. Keep the magic-header set but additionally check whether the binary
   embeds a `codex-rs` / `OpenAI Codex` version string before classifying
   it as a wrapper.

Either fix should add a regression test that the official OpenAI Rust codex
CLI (v0.132+) on `PATH` is treated as a real codex by `resolve_codex_cli_path`.

## Acceptance Criteria

- [ ] `is_wrapper_binary()` returns `False` for OpenAI Codex CLI v0.132+ on macOS/Linux.
- [ ] `ouroboros run workflow seed.yaml --dry-run` no longer emits
      `cli_wrapper_detected` / `cli_no_fallback` for the system codex.
- [ ] Regression test covers the official Codex CLI Rust binary in
      `tests/unit/codex/`.

## Environment

- Ouroboros: `0.39.1` (also reproduced on 0.29.2+ where the helper exists).
- Codex CLI: `0.132.0` (Homebrew Cask, Mach-O 64-bit arm64).
- macOS: 14+ (Apple Silicon).
- Python: 3.14.

## Notes

A local workaround is to set `_WRAPPER_MAGIC_HEADERS = ()` in
`ouroboros/codex/cli_policy.py`; once that's patched in-place, `ouroboros
run workflow` proceeds to real codex calls (verified locally — codex
process spawns, `Analyzing complexity` phase progresses past 60 s instead
of failing in <2 s).

Q00:
## Summary

- Tighten Codex wrapper detection so Mach-O/ELF magic alone is not treated as a wrapper.
- Require a compiled-binary header plus a wrapper-specific signature before falling back to another CLI.
- Add regression coverage for official OpenAI Codex Rust binaries (`OpenAI Codex codex-rs 0.132.0`) resolving without wrapper warnings.

Closes #1159

## Test plan

- `uv run pytest tests/unit/codex/test_cli_policy.py -q`
- `uv run pytest tests/unit/codex/test_cli_policy.py tests/unit/orchestrator/test_codex_cli_runtime.py tests/unit/providers/test_codex_cli_adapter.py -q`
- `uv run ruff check src/ouroboros/codex/cli_policy.py tests/unit/codex/test_cli_policy.py tests/unit/orchestrator/test_codex_cli_runtime.py tests/unit/providers/test_codex_cli_adapter.py`

## Notes

This replaces the stale design-blocked assessment with a focused compatibility fix: the Codex CLI policy files are present on current `main`, so the reported bug is directly actionable.

### DIFF
--- a/src/ouroboros/codex/cli_policy.py
+++ b/src/ouroboros/codex/cli_policy.py
@@ -11,11 +11,18 @@ DEFAULT_CODEX_CLI_NAME = "codex"
 DEFAULT_MAX_OUROBOROS_DEPTH = 5
 DEFAULT_CODEX_CHILD_ENV_KEYS = ("OUROBOROS_AGENT_RUNTIME", "OUROBOROS_LLM_BACKEND")
 DEFAULT_CODEX_CHILD_SESSION_ENV_KEYS = ("CODEX_THREAD_ID",)
-_WRAPPER_MAGIC_HEADERS = (
+_COMPILED_BINARY_MAGIC_HEADERS = (
     b"\xcf\xfa\xed\xfe",  # Mach-O 64-bit
     b"\xce\xfa\xed\xfe",  # Mach-O 32-bit
     b"\x7fELF",  # ELF
 )
+_WRAPPER_SIGNATURES = (
+    b"zeude",
+    b"Zeude",
+    b"wrapper:codex",
+    b"codex-wrapper",
+)
+_WRAPPER_SIGNATURE_SCAN_BYTES = 128 * 1024
 
 
 @dataclass(frozen=True, slots=True)
@@ -38,8 +45,8 @@ def resolve_codex_cli_path(
 ) -> CodexCliResolution:
     """Resolve the safest Codex CLI path for nested automation.
 
-    When the configured candidate is a compiled wrapper (for example a Zeude
-    shim), prefer the next real ``codex`` binary on ``PATH`` instead.
+    When the configured candidate is a known compiled wrapper (for example a
+    Zeude shim), prefer the next real ``codex`` binary on ``PATH`` instead.
     """
     if explicit_cli_path is not None:
         candidate = str(Path(explicit_cli_path).expanduser())
@@ -57,7 +64,7 @@ def resolve_codex_cli_path(
     logger.warning(
         f"{log_namespace}.cli_wrapper_detected",
         wrapper_path=resolved,
-        hint="Searching PATH for the real Node.js codex CLI.",
+        hint="Searching PATH for the real Codex CLI.",
     )
     fallback = find_real_cli(default_cli_name=default_cli_name, skip=resolved)
     if fallback is not None:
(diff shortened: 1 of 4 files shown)


shaun0927:
@ouroboros-agent Re-review requested for latest HEAD b85c5b5e. The prior blocker was addressed with ELF/Linux official Rust Codex regression coverage for both is_wrapper_binary() and resolve_codex_cli_path(), and all GitHub checks are green.

shaun0927:
Re review ping

Issue #1164, opened 2026-05-21
[Bug] bug(executor): implementer subtype=success is rejected by parallel_executor without logging the reason, then cascades dependency_failed to every downstream AC


YaniM:
### Summary

 When ouroboros_start_execute_seed runs a seed with multiple ACs, the implementation session for the first AC returns subtype=success from the Claude adapter and reports tests passing in its result content — but parallel_executor.ac.completed is then emitted with success=False. No log line records why the executor declined the implementer's verdict. The
  executor next marks every downstream AC dependency_failed (the dependency graph is inferred from AC text, no explicit depends_on edges in the seed), and the job exits with 0/N success.

### Impact

- Any seed with multiple ACs where downstream ACs reference an artifact introduced in an earlier AC will fail after only the first AC runs.
  - The orchestrator's worktree is left containing valid partial work (file edit applied, tests passing), but the overall job is reported as failed, blocking ooo evaluate and downstream automation.
  - The implementer↔executor disagreement is silent: the only diagnostic is the AC's terminal success=False flag and the bulk skip events for the rest.

### Environment

 - Ouroboros: 0.39.1
  - MCP server: ouroboros-ai 0.39.1 (uv tool archive)
  - Claude Code: 2.1.146
  - Runtime backend: claude, LLM backend: claude_code
  - macOS: 26.4.1 (Apple Silicon)
  - Python: 3.14

### Steps to reproduce

 1. Author a seed with N ACs (N > 1) where AC2..ACN textually reference an artifact (interface method, class, schema field) that AC1 introduces.
  2. Run via the MCP surface:
  mcp.ouroboros_start_execute_seed(
      seed_content=<above seed>,
      model_tier="medium",
      max_iterations=10,
  )
  3. Poll with ouroboros_job_wait. The job terminates failed after AC1 finishes.
  4. Inspect ~/.ouroboros/logs/ouroboros.log around the run. The implementer's result_message shows subtype=success; the executor's ac.completed for the same AC shows success=False; every other AC shows ac.skipped reason=dependency_failed.

### Expected behavior

- If the adapter returns subtype=success for an AC and the result body indicates tests passed, the executor should either accept it or emit a structured log line that names the verifier criterion it failed (e.g. verifier.rejected reason=<...>).
  - The dependency graph that drives the cascade should be visible — at minimum log the inferred depends_on edges at AC-start time so the cascade is not invisible to the operator.

### Actual behavior

[...]

PR #1073, opened 2026-05-17
Respect docs-only evidence scope

shaun0927:
## Summary
- Make the AC-specific effective evidence schema authoritative for verifier success/failure.
- Ignore non-empty evidence fields that were removed by the effective schema for docs-only ACs, instead of failing the AC on out-of-scope `tests_passed` noise.
- Persist `ignored_out_of_scope_evidence_fields` in typed-evidence events so over-reported fields remain observable.

## Why
The broader post-#1069 #978/#961 observation completed the implementation and passed `python -m unittest test_todo.py`, but AC4 was documentation-only and still emitted individual `tests_passed` names. The docs-only effective schema required only `files_touched` and `commands_run`, yet the verifier re-added the excluded `tests_passed` field and failed with `FABRICATION_SUSPECTED`.

## Scope boundaries
- Code/implementation ACs still verify `tests_passed` strictly.
- This does not broadly accept individual unittest names from aggregate unittest output.
- This does not restore legacy/self-report fallback.
- This only ignores fields excluded by the AC-specific effective schema and records them for audit.

## Validation
- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -k 'docs_only_ac_ignores_out_of_scope_test_id_bleed or test_not_covered_by_success_chunk or docs_only_ac_uses_effective_schema' -q`
- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_profile_strategy.py tests/unit/orchestrator/test_phase_wrappers.py -q`
- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py`
- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py && git diff --check`
- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py`

## Not tested
- Live broader #978 observation rerun after this patch.

Refs #978, #961, #1069.

PR #1072, opened 2026-05-17
Harden docs-only evidence prompts

shaun0927:
## Summary
- Strengthen the docs-only AC prompt to say `tests_passed` must not be emitted at all.
- Clarify that sanity-check tests should not become individual test-name claims for docs-only ACs.
- Keep docs-only evidence focused on current-session docs evidence and validation commands.

## Why
The broader post-#1069 observation showed that schema omission alone was not enough: a docs-only README AC still over-reported individual unittest names in `tests_passed`. This prompt change reduces recurrence while the verifier policy remains the enforcement boundary.

## Scope boundaries
- Prompt-only preventative change.
- Does not change verifier acceptance semantics.
- Does not relax code AC `tests_passed` verification.
- Does not reintroduce legacy/self-report fallback.

## Validation
- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -k 'docs_only_ac_uses_effective_schema or docs_only_ac_allows_code_subject_documentation or markdown_code_ac_keeps_test_evidence_required' -q`
- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py`
- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py tests/unit/orchestrator/test_parallel_executor.py && git diff --check`
- `uv run mypy src/ouroboros/orchestrator/parallel_executor.py`

## Not tested
- Live broader #978 observation rerun after this patch.

Refs #978, #961, #1069.

PR #1069, opened 2026-05-17
Normalize workspace file evidence claims

shaun0927:
## Summary
- Normalize `files_touched` evidence claims to workspace-relative paths before runtime verification.
- Accept absolute `files_touched` claims only when they resolve inside the active task workspace and remain backed by transcript mutation evidence.
- Strengthen evidence prompts to ask for workspace-relative `files_touched` paths and forbid `/tmp/...`, `/private/tmp/...`, or outside-workspace paths.

## Why
The broader post-#1068 #978/#961 observation reached runtime-backed unittest success and avoided exploratory `commands_run` contamination, but AC3 still failed the verifier because typed evidence cited an absolute path under the target repo:

`/private/tmp/.../target-repo/test_todo.py`

The verifier expected workspace-relative file claims, so it rejected the otherwise workspace-local touched file as unsupported evidence.

## Boundaries
- Does not accept arbitrary absolute paths; outside-workspace paths remain unsupported.
- Does not allow file existence alone to prove `files_touched`.
- Does not trust final prose, read-only command mentions, or legacy/self-report fallback.
- Does not change `commands_run` or `tests_passed` semantics beyond preserving existing coverage.

## Validation
- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py -k 'absolute_file_claim or unscoped_file or preexisting_file or read_only_file_reference or typed_evidence_prompt' -q`
- `uv run pytest tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_profile_strategy.py tests/unit/orchestrator/test_phase_wrappers.py -q`
- `uv run ruff check src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/profile_strategy.py src/ouroboros/orchestrator/phase_wrappers.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_profile_strategy.py tests/unit/orchestrator/test_phase_wrappers.py`
- `uv run ruff format --check src/ouroboros/orchestrator/parallel_executor.py src/ouroboros/orchestrator/profile_strategy.py src/ouroboros/orchestrator/phase_wrappers.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_profile_strategy.py tests/unit/orchestrator/test_phase_wrappers.py && git diff --check`
[...]

PR #1166, opened 2026-05-22
fix(orchestrator): credit transcript test commands for tests_passed claims

nkjunbc:
## Problem

In fat-harness mode (the default since the `legacy` execution mode was removed), the atomic AC verifier rejects legitimate, green work as `FABRICATION_SUSPECTED`, failing the whole AC even though the code is written and the tests pass.

Observed in production (Windows, Claude Code runtime): every sub-AC reported success (`[AC_COMPLETE: N]`, full suite green) yet `parallel_executor.ac.completed` logged `success=False`, cascading to `dependency_failed` skips and a failed session. The recorded verdict:

```
typed_evidence_present: True
typed_evidence_valid:   True
verifier_ran:           True
verifier_passed:        False
verifier_failure_class: FABRICATION_SUSPECTED
verifier_reasons: unsupported evidence claims:
  commands_run: python -m ruff check src/poc/structure_extractor.py ...;
  tests_passed: tests/test_structure_and_draft_substance.py::test_numbering_system_classifies_roman_and_arabic; ...
```

## Root cause

`_verify_atomic_evidence_against_runtime_messages` builds `backed_commands` **only** from `commands_run` evidence entries, and `_runtime_messages_support_test_claim` uses that list as the **sole** source of candidate test commands.

So a `tests_passed` node-id claim can only be verified if the corresponding `pytest` invocation was also echoed into `commands_run`. When the agent ran `pytest <file>` (clearly present in the runtime transcript, with a passing result) but listed only a lint command in `commands_run`, there is **no** candidate test command, so every node-id claim is marked unsupported → `FABRICATION_SUSPECTED` → the AC fails on correct work.

The transcript is the ground truth here, but the verifier never consults the Bash messages' own commands for test-claim support.

## Fix

`_runtime_messages_support_test_claim` now also treats each Bash message's own recorded command (`_runtime_message_command_values`) as a candidate test command, in addition to the backed `commands_run` entries.

That command is backed by definition — it is the literal invocation recorded in the transcript — so a real, transcript-proven test run can support a node-id `tests_passed` claim even when the agent did not duplicate it into `commands_run`.

**Anti-fabrication is preserved.** A claim is still credited only when:
- the command is a real Bash invocation in the transcript (`_runtime_message_supports_command_claim`),
- the chunk shows test success (`_message_contains_test_success`), and
- the claim is targeted by the command (`_test_command_targets_claim`).

The two existing "rejects … missing from runtime" tests still reject.

## Tests

- Added `test_fat_harness_verifier_accepts_pytest_node_id_claim_backed_by_transcript_command`, which reproduces the bug: lint in `commands_run`, `pytest <file>` only in the transcript, node-id `tests_passed` claim. It fails on `main` (verifier rejects) and passes with this change.
- Existing fat-harness accept/reject tests in `tests/unit/orchestrator/test_parallel_executor.py` continue to pass.

> Note: 3 pre-existing failures in this file (`*_normalizes_workspace_absolute_file_claim`, `*_rejects_unscoped_file_and_failed_test_command`, `*_accepts_wrapped_broad_pytest_for_current_test_file`) are unrelated Windows-only test-fixture issues: they embed `tmp_path` directly into a JSON evidence literal, and on Windows the backslash path separators are invalid JSON escapes (`Evidence is not valid JSON: Invalid \escape`), so the evidence fails to parse before any verifier logic runs. They pass on POSIX CI (forward-slash paths) and fail identically on unmodified `main`; this change does not touch them.

## Context

Relates to #1164 (implementer `subtype=success` rejected by the executor, then cascades `dependency_failed`). This PR addresses one root cause of that rejection in the verifier's matching logic. Companion: #1168 covers the sibling `commands_run` redirection/pipe matching gap. The two are independent and can land in either order.

### DIFF
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -1276,10 +1276,17 @@ def _runtime_messages_support_test_claim(
         # latter is backed by definition — it is the literal invocation in the
         # transcript — so a real ``pytest <file>`` run can support a node-id
         # ``tests_passed`` claim even when the agent did not also echo that
-        # exact command into its ``commands_run`` evidence. Anti-fabrication is
-        # preserved: the command must still be a real Bash invocation, the
-        # chunk must show test success, and the claim must be targeted by the
-        # command (see ``_test_command_targets_claim``).
+        # exact command into its ``commands_run`` evidence.
+        #
+        # These are NOT three independent checks. For a per-message candidate
+        # the ``_runtime_message_supports_command_claim`` gate below is
+        # tautological — the candidate is that message's own command, so it
+        # trivially supports itself. The anti-fabrication guarantee is carried
+        # entirely by the downstream gates: ``_message_contains_test_success``
+        # (reads only structured runtime output, never agent narration) and
+        # ``_test_command_targets_claim`` (anchors the claim's node-id/file to
+        # the recorded command + proof text). The ``_looks_like_test_command``
+        # filter still excludes non-test commands from this candidate source.
         candidate_commands = (*backed_commands, *_runtime_message_command_values(message))
         matching_commands = tuple(
             candidate
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -1270,9 +1270,20 @@ def _runtime_messages_support_test_claim(
     for index, message in enumerate(messages):
         if message.tool_name != "Bash":
             continue
+        # Candidate test commands are drawn from two transcript-grounded
+        # sources: (1) ``commands_run`` evidence entries already proven against
+        # the transcript, and (2) the Bash message's own recorded command. The
+        # latter is backed by definition — it is the literal invocation in the
+        # transcript — so a real ``pytest <file>`` run can support a node-id
+        # ``tests_passed`` claim even when the agent did not also echo that
+        # exact command into its ``commands_run`` evidence. Anti-fabrication is
+        # preserved: the command must still be a real Bash invocation, the
+        # chunk must show test success, and the claim must be targeted by the
+        # command (see ``_test_command_targets_claim``).
+        candidate_commands = (*backed_commands, *_runtime_message_command_values(message))
         matching_commands = tuple(
             candidate
-            for candidate in backed_commands
+            for candidate in candidate_commands
             if _looks_like_test_command(candidate)
             and _runtime_message_supports_command_claim(candidate, message)
         )
--- a/tests/unit/orchestrator/test_parallel_executor.py
+++ b/tests/unit/orchestrator/test_parallel_executor.py
@@ -5113,6 +5113,84 @@ class TestParallelACExecutor:
         assert evidence_event.data["verifier_passed"] is True
 
     @pytest.mark.asyncio
+    async def test_fat_harness_verifier_rejects_node_id_claim_backed_only_by_non_test_command(
+        self, tmp_path
+    ) -> None:
+        """The new candidate source must not let a non-test command back a test claim.
+
+        Guards the message-command candidate path added for node-id ``tests_passed``
+        support: a Bash message whose command merely prints a fake success line
+        (``cat fake_results.txt`` whose output is ``test_x.py::test_y passed``) is
+        not a test command, so ``_looks_like_test_command`` must exclude it and the
+        node-id claim stays unsupported (FABRICATION_SUSPECTED) — even though the
+        recorded output literally contains the node-id-plus-"passed" marker.
+        """
+        source_file = tmp_path / "string_utils.py"
+        source_file.write_text("def slugify(text):\n    return text\n", encoding="utf-8")
+
+        fake_command = "cat fake_results.txt"
+        event_store, appended_events = _make_replaying_event_store()
+        executor = ParallelACExecutor(
+            adapter=_FinalMessageRuntime(
+                "```json\n"
+                "{\n"
+                '  "files_touched": ["string_utils.py"],\n'
+                f'  "commands_run": ["{fake_command}"],\n'
+                '  "tests_passed": ["test_slugify.py::test_spaces"]\n'
+                "}\n"
+                "```",
+                native_session_id="session-node-id-non-test-command-only",
+                support_messages=(
+                    AgentMessage(
+                        type="assistant",
+                        content=f"Calling tool: Edit: {source_file}",
+                        tool_name="Edit",
+                        data={"tool_input": {"file_path": str(source_file)}},
+                    ),
+                    AgentMessage(
+                        type="assistant",
+                        content=f"Calling tool: Bash: {fake_command}",
+                        tool_name="Bash",
+                        data={
+                            "tool_input": {"command": fake_command},
+                            "output": "test_slugify.py::test_spaces passed",
+                            "exit_code": 0,
+                        },
+                    ),
+                ),
+                cwd=str(tmp_path),
+            ),
+            event_store=event_store,
+            console=MagicMock(),
+            enable_decomposition=False,
+            execution_profile=load_profile("code"),
+            fat_harness_mode=True,
+            task_cwd=str(tmp_path),
+        )
+
+        result = await executor._execute_atomic_ac(
+            ac_index=0,
+            ac_content="Create slugify and pytest coverage.",
+            session_id="orch_123",
+            tools=["Read", "Edit", "Bash"],
+            tool_catalog=(MCPToolDefinition(name="Read", description="Read a file."),),
+            system_prompt="system",
+            seed_goal="Ship string utilities",
+            depth=0,
+            start_time=datetime.now(UTC),
+        )
+
+        assert result.success is False
+        assert result.error is not None
+        assert "tests_passed:" in result.error
+        evidence_event = next(
+            event
+            for event in appended_events
+            if event.type == "execution.ac.typed_evidence.observed"
+        )
+        assert evidence_event.data["verifier_failure_class"] == "FABRICATION_SUSPECTED"
+
+    @pytest.mark.asyncio
     async def test_fat_harness_verifier_rejects_unittest_summary_missing_from_runtime(
         self, tmp_path
     ) -> None:
--- a/tests/unit/orchestrator/test_parallel_executor.py
+++ b/tests/unit/orchestrator/test_parallel_executor.py
@@ -5008,6 +5008,111 @@ class TestParallelACExecutor:
         assert evidence_event.data["verifier_failure_class"] == "FABRICATION_SUSPECTED"
 
     @pytest.mark.asyncio
+    async def test_fat_harness_verifier_accepts_pytest_node_id_claim_backed_by_transcript_command(
+        self, tmp_path
+    ) -> None:
+        """A transcript ``pytest <file>`` run backs node-id ``tests_passed`` claims.
+
+        Regression: candidate test commands were sourced only from
+        ``commands_run`` evidence. When the agent listed lint in
+        ``commands_run`` but ran ``pytest`` (recorded in the transcript) without
+        echoing it into ``commands_run``, every node-id ``tests_passed`` claim
+        was rejected as FABRICATION_SUSPECTED even though the run is real and
+        green. The Bash message's own command is now also a candidate, so a
+        transcript-proven test run supports the claim.
+        """
+        source_file = tmp_path / "string_utils.py"
+        test_file = tmp_path / "test_slugify.py"
+        source_file.write_text(
+            "def slugify(text):\n    return text.lower().replace(' ', '-')\n",
+            encoding="utf-8",
+        )
+        test_file.write_text(
+            "from string_utils import slugify\n\n"
+            "def test_spaces():\n"
+            "    assert slugify('Hello World') == 'hello-world'\n",
+            encoding="utf-8",
+        )
+
+        lint_command = "python -m ruff check string_utils.py test_slugify.py"
+        pytest_command = "python -m pytest test_slugify.py -q"
+        event_store, appended_events = _make_replaying_event_store()
+        executor = ParallelACExecutor(
+            adapter=_FinalMessageRuntime(
+                "```json\n"
+                "{\n"
+                '  "files_touched": ["string_utils.py", "test_slugify.py"],\n'
+                f'  "commands_run": ["{lint_command}"],\n'
+                '  "tests_passed": ["test_slugify.py::test_spaces"]\n'
+                "}\n"
+                "```",
+                native_session_id="session-pytest-node-id-transcript-only",
+                support_messages=(
+                    AgentMessage(
+                        type="assistant",
+                        content=f"Calling tool: Edit: {source_file}",
+                        tool_name="Edit",
+                        data={"tool_input": {"file_path": str(source_file)}},
+                    ),
+                    AgentMessage(
+                        type="assistant",
+                        content=f"Calling tool: Edit: {test_file}",
+                        tool_name="Edit",
+                        data={"tool_input": {"file_path": str(test_file)}},
+                    ),
+                    AgentMessage(
+                        type="assistant",
+                        content=f"Calling tool: Bash: {lint_command}",
+                        tool_name="Bash",
+                        data={
+                            "tool_input": {"command": lint_command},
+                            "output": "All checks passed!",
+                            "exit_code": 0,
+                        },
+                    ),
+                    AgentMessage(
+                        type="assistant",
+                        content=f"Calling tool: Bash: {pytest_command}",
+                        tool_name="Bash",
+                        data={
+                            "tool_input": {"command": pytest_command},
+                            "output": "1 passed in 0.01s",
+                            "exit_code": 0,
+                        },
+                    ),
+                ),
+                cwd=str(tmp_path),
+            ),
+            event_store=event_store,
+            console=MagicMock(),
+            enable_decomposition=False,
+            execution_profile=load_profile("code"),
+            fat_harness_mode=True,
+            task_cwd=str(tmp_path),
+        )
+
+        result = await executor._execute_atomic_ac(
+            ac_index=0,
+            ac_content="Create slugify and pytest coverage.",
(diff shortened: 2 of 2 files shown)


shaun0927:
Thanks @nkjunbc — approving. This lands cleanly as a narrow Track A / #978 verifier follow-up, outside Track C tier gates (same lineage as #1006 / #1018 / #1020 / #1026).

I traced the verifier path end-to-end before approving. A few notes for the record, since this touches anti-fabrication semantics:

**Why this is safe.** Adding `_runtime_message_command_values(message)` to the candidate set means `_runtime_message_supports_command_claim(candidate, message)` is tautologically satisfied for per-message candidates (the candidate is drawn from the same message it is checked against). That gate is effectively a no-op here. The actual anti-fabrication protection now rests on two downstream invariants:

1. `_message_contains_test_success` consults `_runtime_message_test_proof_text`, which extracts only from structured runtime fields (`output` / `stdout` / `tool_result`) — never from assistant narration or `content`. An agent cannot inject a fake success marker into those fields; they are populated by the tool runtime itself.
2. For node-id claims, `_test_command_targets_claim` requires the file part of the node id to appear in the chunk, anchored to the recorded command. For summary-style claims, `_claim_summary_matches_runtime_chunk` requires the summary text to appear in the structured proof text. I confirmed `test_fat_harness_verifier_rejects_shell_wrapped_unittest_summary_missing_from_runtime` and `test_fat_harness_verifier_rejects_unittest_summary_missing_from_runtime` still reject under this code path for exactly that reason.

Also confirmed: `_runtime_message_command_values` reads from `tool_input` / `input` / `arguments` containers and never from `output` or `content`, so the new candidate source is structurally trustworthy.

**Two small follow-ups (non-blocking, fine in a separate PR):**

- The 9-line comment is helpful for the *why*, but it would be worth one extra line noting that the first gate is tautological for per-message candidates and that the security guarantee is carried by the success + targeting gates. Right now a future reader could mistake it for three independent checks.
- A negative test for the new code path would round out coverage: a Bash message whose `tool_input.command` is a non-test command (e.g., `echo "test_foo.py::test_bar passed"` or a `cat` of a fake summary) — the `_looks_like_test_command` filter should reject it, and an explicit test would lock that invariant in.

**One housekeeping ask before I merge:** could you link a CI run on unmodified `main` showing the three test failures you flagged (`*_normalizes_workspace_absolute_file_claim`, `*_rejects_unscoped_file_and_failed_test_command`, `*_accepts_wrapped_broad_pytest_for_current_test_file`)? Your description says they fail identically on `main` and are unrelated Windows path-resolution issues — I'd like that on the record so we can split them into a separate `needs-info` issue rather than letting them stay implicit in the test file.

Thanks again for the careful root-cause writeup — the production transcript with `verifier_failure_class: FABRICATION_SUSPECTED` despite `[AC_COMPLETE]` made the bug very easy to confirm.

nkjunbc:
Thanks @shaun0927 — both follow-ups are in `e6911da`:

- **Comment clarification.** Expanded the `_runtime_messages_support_test_claim` comment to spell out that the per-message `_runtime_message_supports_command_claim` gate is tautological (the candidate is that message's own command) and that the anti-fabrication guarantee is carried by `_message_contains_test_success` + `_test_command_targets_claim`, not three independent checks — matches your trace.
- **Negative test.** `test_fat_harness_verifier_rejects_node_id_claim_backed_only_by_non_test_command`: a Bash message running `cat fake_results.txt` whose recorded **output** is literally `test_slugify.py::test_spaces passed` must not back the node-id claim — `_looks_like_test_command` excludes it and it stays `FABRICATION_SUSPECTED`. (Used `cat` instead of `echo "...::... passed"` to avoid unescaped quotes in the JSON evidence literal; same intent, and it also proves a fake marker in the *output* alone isn't enough.)

**On the housekeeping ask — plus a correction to my own PR text.** There's no CI run on `main` showing those three failures, because they don't fail on CI. When I dug in, my original "Windows path-resolution" wording turned out to be wrong. The real cause is a test-fixture portability bug: the three tests embed `tmp_path` directly into a JSON evidence literal, e.g.

```python
f'{{"files_touched":["{touched_file}"], ...}}'
```

On Windows `tmp_path` is `C:\Users\...\test_todo.py`, so the backslashes are invalid JSON escapes and the evidence never parses — it fails before any verifier logic runs:

```
typed_evidence_error = 'Evidence is not valid JSON: Invalid \escape (line 1, col 22)'
```

On POSIX the paths are forward-slash, the JSON is valid, and the tests pass — which is exactly why this PR's Linux CI is green (Test 3.12/3.13/3.14 ✓). Local Windows run on unmodified `main` (`git checkout main`, clean tree):

```
FAILED tests/unit/orchestrator/test_parallel_executor.py::TestParallelACExecutor::test_fat_harness_verifier_normalizes_workspace_absolute_file_claim
FAILED tests/unit/orchestrator/test_parallel_executor.py::TestParallelACExecutor::test_fat_harness_verifier_rejects_unscoped_file_and_failed_test_command
FAILED tests/unit/orchestrator/test_parallel_executor.py::TestParallelACExecutor::test_fat_harness_verifier_accepts_wrapped_broad_pytest_for_current_test_file
3 failed, 186 deselected
# all three fail with: Evidence is not valid JSON: Invalid \escape
```

I've corrected the PR description to reflect this. The fix for those tests is to encode the path in the fixtures (`json.dumps(str(path))` or `Path.as_posix()`) rather than f-string interpolation. Happy to open a separate `needs-info` issue with this writeup if that's the preferred home for it — just say the word.

shaun0927:
## Merge-readiness re-audit — reviewer pass from a separate worktree

Re-audited this PR end-to-end from an isolated worktree (`/tmp/ouro-pr1166` at head `e6911da`) against the **Issue #961 Meta SSOT**, the `ooo auto` / `ooo run` direction, and an over-engineering lens. Posting the reasoning so the audit trail is self-contained.

### 1. SSOT alignment (Issue #961)

This PR is **explicitly named** in the Track A row of the SSOT body and re-confirmed in five freshness syncs on 2026-05-22 (warden 11:20 / 13:06 / 15:56 / 20:27 / 20:59 KST):

> "**#1166** and #1168 are open, green, and **APPROVED** as fat-harness verifier evidence follow-ups."

It sits in the same `#978 / #920` fat-harness verifier lineage as the already-merged hardening series **#1006 / #1018 / #1020 / #1026 / #1068 / #1069 / #1072 / #1073**, outside Track C tier gates. There is no roadmap drift.

### 2. Direction check vs. `ooo auto` / `ooo run`

The bug is the **upstream root cause** of a real `ooo auto` failure mode documented in #1164: every sub-AC reports `[AC_COMPLETE]` and `subtype=success`, but the verifier records `verifier_passed=False` / `verifier_failure_class=FABRICATION_SUSPECTED`, which cascades into `dependency_failed` skips and fails the session. That directly blocks the Track B "self-healing `ooo auto`" narrative. Fixing it here is on-strategy, not a tangent.

### 3. Over-engineering check — **surgical, not over-engineered**

Production-code delta is literally one effective line of logic:

```python
candidate_commands = (*backed_commands, *_runtime_message_command_values(message))
# ...
for candidate in candidate_commands  # was: for candidate in backed_commands
```

Everything else in the +202/-1 diff is:
- **+16 lines of comment** spelling out the anti-fabrication contract (added at owner's explicit request in review).
- **+2 tests** — one positive regression (the actual production transcript shape) and one negative guard (`cat fake_results.txt` whose output literally contains `test_x.py::test_y passed` is **still** rejected because `_looks_like_test_command` filters it out).

No new types, no new abstractions, no new public API, no helper added — the change reuses six pre-existing helpers (`_runtime_message_command_values`, `_looks_like_test_command`, `_runtime_message_supports_command_claim`, `_message_contains_test_success`, `_test_command_targets_claim`, `_runtime_message_test_proof_text`). This is the minimum-blast-radius shape this fix can take.

### 4. Anti-fabrication invariants — preserved

Re-traced the gates after the change:

1. **Candidate source is structural, not prose.** `_runtime_message_command_values` reads only from `tool_input` / `input` / `arguments` / `args` containers (lines 545–553 of `parallel_executor.py`), never from `output` or assistant `content`. Adapters populate these; the agent cannot inject into them.
2. **Test-command filter still in force.** `_looks_like_test_command` (line 903) gates every candidate — proven by the new negative test.
3. **Success marker is structural, not prose.** `_message_contains_test_success` → `_runtime_message_test_proof_text` only consults structured runtime fields (`result_preview` / `output` / `stdout` / `stderr`), never `content`.
4. **Targeting still anchored.** `_test_command_targets_claim` requires the node-id file part to appear in the recorded command + proof text.

The two existing FABRICATION_SUSPECTED reject tests (`*_rejects_shell_wrapped_unittest_summary_missing_from_runtime`, `*_rejects_unittest_summary_missing_from_runtime`) plus the new negative guard all still reject — verified locally:

```
tests/unit/orchestrator/test_parallel_executor.py
  test_fat_harness_verifier_accepts_pytest_node_id_claim_backed_by_transcript_command   PASSED
  test_fat_harness_verifier_rejects_node_id_claim_backed_only_by_non_test_command       PASSED
  test_fat_harness_verifier_rejects_shell_wrapped_unittest_summary_missing_from_runtime PASSED
  test_fat_harness_verifier_rejects_unittest_summary_missing_from_runtime               PASSED
4 passed, 187 deselected in 0.77s
```

(Run on `e6911da` against local `src/` on `PYTHONPATH` to bypass the globally-installed `ouroboros` shim.)

### 5. Review-cycle state

- Owner @shaun0927 approved on 2026-05-22 02:07 KST with an end-to-end trace.
- Both non-blocking follow-ups he asked for landed in `e6911da` (comment clarification + negative test).
- The author also **corrected** their own PR description: the three pre-existing test failures are not "Windows path-resolution" but a JSON-escape portability bug from f-string-interpolating `tmp_path` into a JSON evidence literal; root-cause is on record in-thread.
- CI: 9/9 green (Lint Ruff/MyPy/Bridge-TS, Tests 3.12/3.13/3.14, enforce-envelope, enforce-boundary).
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `reviewDecision: APPROVED`.

### Verdict

**Merge-ready as-is.** No fix commits added by this audit — the diff is already at the right scope, the anti-fabrication contract is preserved, the SSOT explicitly lists this PR as a clean Track A follow-up, and every concern raised in the original review thread has been addressed in `e6911da`. A separate `/pr-review` skill pass with full mutation-survivor / CRAP-risk analysis is being posted next.

shaun0927:
# PR Review Summary

> Posted via the `/pr-review` skill (intent/scope trace + changed-logic-to-test mapping + mutation-test thinking + CRAP-style risk + security/operational risk). Inputs were collected against head `e6911da` from an isolated worktree (`/tmp/ouro-pr1166`), base `main`.

## Verdict
**Approve**

## Scope Reviewed
- **PR intent:** Fix fat-harness atomic-AC verifier so a transcript-proven `pytest <file>` run can back a node-id `tests_passed` claim even when the agent did not duplicate the exact `pytest` invocation into `commands_run` evidence. Without this, legitimate green ACs are stamped `FABRICATION_SUSPECTED` and cascade `dependency_failed`.
- **Main changed areas:**
  - `src/ouroboros/orchestrator/parallel_executor.py` — `_runtime_messages_support_test_claim` (one-line candidate-set expansion + 16-line anti-fabrication contract comment).
  - `tests/unit/orchestrator/test_parallel_executor.py` — one positive regression test (`*_accepts_pytest_node_id_claim_backed_by_transcript_command`) + one negative guard test (`*_rejects_node_id_claim_backed_only_by_non_test_command`).
- **Tests reviewed:**
  - New positive: agent puts `ruff` in `commands_run`, runs `pytest test_slugify.py -q` only in transcript → claim accepted (was rejected on `main`).
  - New negative guard: agent puts `cat fake_results.txt` in `commands_run`, recorded **output** literally contains `test_slugify.py::test_spaces passed` → claim still rejected via `_looks_like_test_command`.
  - Existing reject tests preserved: `*_rejects_shell_wrapped_unittest_summary_missing_from_runtime`, `*_rejects_unittest_summary_missing_from_runtime`.
  - Local pytest run on `e6911da` (with `PYTHONPATH=src` to bypass the globally-installed shim): `4 passed, 187 deselected in 0.77s`.
- **Checks considered:** GitHub status checks 9/9 SUCCESS (Lint Ruff/MyPy/Bridge-TS, Tests 3.12/3.13/3.14, enforce-envelope, enforce-boundary); mergeStateStatus `CLEAN`; reviewDecision `APPROVED`; head commit `e6911da` already addresses owner's two follow-ups.

## Blocking Issues
None.

## Warnings
None.

## Mutation-Test Thinking

- **Likely mutants that should be killed:**
  - Drop `_runtime_message_command_values(message)` from the candidate union → positive test fails (regression coverage proven).
  - Drop `_looks_like_test_command(candidate)` filter → negative guard test fails (`cat fake_results.txt` would back the claim through its fake output marker).
  - Replace `tuple(candidate for candidate in candidate_commands if ...)` with `(backed_commands[0],)` style narrowing → positive test fails (claim becomes unsupported again).
  - Make `_runtime_message_supports_command_claim` skip the message-self check → no functional regression in the current test surface (this gate is **tautological** for per-message candidates, as both owner trace and the in-code comment document), so this mutant is intentionally *uncovered* — the in-code comment now explicitly explains why that is by design.
- **Mutants current tests may not catch:**
  - Loosening `_message_contains_test_success` to read from assistant `content` instead of structured runtime fields. This would silently weaken anti-fabrication but is **out of this PR's scope** — the helper isn't touched. `needs verification` whether a standalone test asserts the structural-only read.
  - Loosening `_test_command_targets_claim` to drop the file-part anchor. Same comment — outside this diff; covered indirectly by the two preserved reject tests, but no targeted single-mutant test exists.
- **Additional tests recommended:** None as a merge blocker. The two helpers above are pre-existing and untouched by this PR; harvesting targeted mutant-killing tests for them belongs in a separate `#978` verifier-hardening follow-up alongside #1068/#1069/#1072/#1073, not here.

- **High-risk functions/modules:** `_runtime_messages_support_test_claim` is on a security-relevant path (anti-fabrication). Cyclomatic complexity is unchanged by this PR: still one outer `for` + one inner generator + three guard checks; the diff only widens the iterable feeding the generator.
- **Complexity increase:** Zero net complexity added. One identifier (`candidate_commands`) introduced; one assignment line; one renamed variable in the comprehension. Branch count, nesting depth, and side-effect surface are identical.
- **Test coverage concern:** None for the changed surface — both the new acceptance path and the new candidate source's failure mode (non-test command) are exercised by the two new tests, and the prior reject tests remain green.
- **Refactoring recommendation:** None. The 16-line contract comment is the right shape for this kind of fix — the change in the candidate set is one line, but the *semantic* change is "this gate is now tautological for per-message candidates; the security guarantee shifts to the downstream success+targeting gates," which is exactly what the comment captures. Removing or shortening the comment would be a regression.

## Test Quality Assessment (6/7)

- **Strong tests:**
  - The positive regression mirrors the **actual production transcript shape** (lint in `commands_run`, `pytest` only in `tool_input.command`, node-id claim) — not a synthetic minimal repro.
  - The negative guard is deliberately adversarial: `cat fake_results.txt` whose **output** literally contains `test_slugify.py::test_spaces passed`. This proves the new candidate source does **not** open a fake-marker-in-output attack vector, because the filter is on the *command*, not the output.
  - Both tests assert at the *event* level (`execution.ac.typed_evidence.observed` → `verifier_passed` / `verifier_failure_class`), which is the contract surface, not internal state.
- **Weak tests:** None in this PR's surface.
- **Missing edge cases:** None blocking. Possible future hardening (out of scope): a Goose-style `argv` list form for `tool_input.command`, e.g. `["python", "-m", "pytest", "test_x.py"]`; this is already covered by `_runtime_command_value_to_text` (line 557) and the `_test_command_invocation` shell-wrapper path (line 908), but a single explicit Goose-shape test would be a tidy addition in a follow-up.
- **Mocking concerns:** Tests use `_FinalMessageRuntime` + `_make_replaying_event_store` — same shape as every other fat-harness verifier test in the file. No new mock layers introduced.

- **Anti-fabrication contract (security-equivalent for this codebase):** preserved end-to-end after re-tracing:
  1. `_runtime_message_command_values` extracts only from `tool_input` / `input` / `arguments` / `args` structured containers (lines 545–553) — never from agent narration or `content`. The runtime adapter populates these, not the LLM.
  2. `_looks_like_test_command` filter still gates every candidate (verified by the new negative guard).
  3. `_message_contains_test_success` → `_runtime_message_test_proof_text` reads only `result_preview` / `output` / `stdout` / `stderr` structured fields (lines 895–899).
  4. `_test_command_targets_claim` still anchors the node-id file part to the recorded command + proof text.
- **Operational risk:** None. The change can only *widen* the set of legitimate transcripts the verifier accepts; it cannot make the verifier accept any transcript the prior gates would have accepted, because the candidate added is gated by the same `_looks_like_test_command` filter and the same downstream success/targeting gates.
- **Backward compatibility / migration:** None. Pure in-process verifier logic; no schema, no DB, no on-disk format, no event-stream contract changes.
- **Logging / secrets:** No changes.

## Looks Good

- Diff is the minimum-blast-radius shape this fix can take — one-line logic change, zero new types/abstractions/helpers.
- Anti-fabrication contract is documented *in the code itself*, not just in PR text, so the rationale survives the PR being closed.
- Negative guard test is genuinely adversarial (fake marker in *output*) and locks the new candidate source's safety invariant in.
- Author corrected their own PR description after deeper investigation (Windows path-resolution → JSON-escape portability bug from f-string `tmp_path` interpolation) — good intellectual honesty on the side issue.
- SSOT (Issue #961) explicitly names this PR as a clean Track A fat-harness verifier follow-up across five 2026-05-22 freshness syncs, same lineage as #1006/#1018/#1020/#1026/#1068/#1069/#1072/#1073.

## Final Recommendation

**Approve and merge.** This is a surgical, well-tested, on-strategy fix for a production-observed verifier bug. The diff is the minimum shape it can take, the anti-fabrication contract is preserved and now documented in-code, the two added tests target both the bug and the new attack vector the bug-fix could have opened, all 9 CI checks are green, mergeStateStatus is `CLEAN`, and the owner has already approved with both review follow-ups addressed in `e6911da`. No blocking issues, no warnings, no merge-time risk identified.

Issue #1157, opened 2026-05-21
Vision SSOT: ooo auto reaches verified product completion without intervention


shaun0927:
## Current north star

Given a product request and the authority required to execute it, `ooo auto` should autonomously produce a usable artifact, verify the declared product behavior, and leave durable evidence that explains the terminal result.

## Terminal contract

Recoverable process or interview non-convergence must not end as `BLOCKED`. Safety, authority, explicit human confirmation, missing credentials or production permission, runtime cancellation, and irrecoverable execution failure may still end as `BLOCKED` or `FAILED`.

`PRODUCT_COMPLETE` requires both implementation evidence and the declared product-reality smoke contract. Code-complete, partial, detached, or degraded outcomes do not satisfy that terminal.

## Current implementation status

The original implementation lanes are substantially shipped: product inference and acceptance policy, execution and recovery substrate, durable event/status surfaces, focused regressions, and the canonical `cli-todo` product-reality smoke harness. Historical component names, PR queues, watchdog vocabulary, and the global Product-or-Die proposal are not current requirements.

Last reconciled: 2026-08-03, `main@cb658aa`.

## Remaining close gate

- [ ] Two consecutive live `cli-todo` runs from independent fresh workdirs, stores, and session IDs reach `PRODUCT_COMPLETE` on the same recorded current-`main` SHA without intervention, and both generated artifacts pass the declared add/list/done and invalid-input smoke commands.

The committed #1450 evidence is retained as diagnostic evidence, but it is inconclusive: its baseline blocked and its treatment detached. It does not satisfy this gate.

## Evidence checklist

- [ ] Record the exact `main` SHA, Ouroboros version, Python/runtime/provider versions, and command.
- [ ] Run `OUROBOROS_RUN_CANONICAL=1 uv run pytest tests/canonical/ -k cli-todo -v` twice consecutively.
- [ ] Use independent fresh workdirs, stores, and session IDs for the two runs.
- [ ] Confirm both runs reach the canonical `PRODUCT_COMPLETE` terminal without human input after invocation.
- [ ] Confirm both artifacts pass the declared add/list/done and invalid-input smoke commands and persist the declared output.
- [ ] Archive the auto, execution, run, and EventStore correlation IDs, artifact paths, smoke commands, exit codes, wall times, and terminal evidence.

## Failure handling

[...]

PR #1151, opened 2026-05-20
feat(auto): canonical stop_reason_code for interview-layer blockers

shaun0927:
## Summary
- Add two interview-layer canonical codes: ``interview_max_rounds_exhausted`` and ``interview_phase_deadline``.
- Add optional keyword-only ``error_code`` to ``state.mark_blocked`` and ``last_error_code`` field on ``AutoPipelineState``.
- Add ``stop_reason_code: str | None = None`` to ``AutoPipelineResult`` and populate it in ``_result()``.
- Document the 7-code total taxonomy in ``skills/auto/SKILL.md``.

## Why this scope
``skills/auto/SKILL.md`` documents 5 Ralph-layer canonical stop reasons but the interview layer uses two free-form blocker strings that are not in the documented taxonomy. Live ``ooo auto`` runs against the #821 acceptance matrix scored 0/7 on the "stop in COMPLETE or one of the canonical 5" criterion partly because the interview layer was outside the taxonomy entirely. This PR closes that surface gap so observers and result-envelope consumers can programmatically classify the interview-layer terminal conditions.

## What is NOT done here
- Only the 2 interview-layer call sites are wired. The other ~28 ``state.mark_blocked`` call sites remain code-less in this PR — Ralph-layer code marking (the existing 5 stop_reason texts already match the canonical names) is a follow-up.
- No change to any existing blocker text string.
- No new closure-mode field (PR-B1).
- No defaulted_sections surface (PR-C).

## Test Plan
- [x] ``uv run pytest tests/unit/auto/test_stop_reason_code.py -v`` (3 new tests pass)
- [x] ``uv run pytest tests/unit/auto tests/integration/auto -q`` (no regression: 897 → 900)
- [x] ``uv run ruff check`` on touched files
- [x] ``uv run mypy`` on touched files

Refs #821, #908, #1138.

PR #1148, opened 2026-05-20
fix(auto): close interview on ledger-only consensus at max_rounds

shaun0927:
## Summary
- Close the auto interview when ``max_rounds`` is exhausted with ledger-side completeness already met but the backend still refusing to declare closure.
- Record the closure path on the result envelope via ``interview_closure_mode="ledger_only"`` so callers can audit which interviews used the new path.
- Preserve the genuine-deadlock blocker (both sides refusing) unchanged.

## Why this scope
Live ``ooo auto`` runs against the #821 acceptance matrix observed this exact case via the ``auto.interview.mutual_agreement_deadlock_at_max_rounds`` event shipped in #1138 (PR-A instrumentation). The #962 mutual-agreement closure gate was designed to avoid fabricated unilateral closures; in this case there is no fabrication — the auto answerer has filled the ledger via explicit answers and the backend simply refuses to update its own ambiguity reading. The #821 autonomy contract requires that ``ooo auto`` reach COMPLETE for locally-reproducible, safe-defaultable goals; this PR closes that contract for the ledger-only consensus case.

## What is NOT done here
- No change to the safe-default finalization fallback (PR-B2 will address the ``ledger_done=False`` cases #3-6 after the next live re-observation).
- No new envelope fields besides ``interview_closure_mode``.
- No taxonomy/stop_reason change (PR-E).
- No source-tag promotion (deferred PR-C2).

## Test Plan
- [x] ``uv run pytest tests/unit/auto/test_interview_pipeline.py -v -k "ledger_only or genuine_deadlock"`` (2 new tests pass)
- [x] ``uv run pytest tests/unit/auto tests/integration/auto -q`` (baseline 897 + 2 new = 899, no regression)
- [x] ``uv run ruff check`` on touched files
- [x] ``uv run mypy`` on touched files

Refs #821, #962, #1138.

PR #1138, opened 2026-05-19
instrument(auto): emit structured-log events at safe-default decision points

shaun0927:
## Summary
- Add structured-log events at the safe-default finalization decision points in `interview_driver.py` and the unsafe-context gate in `safe_defaults.py`.
- Keep the change instrumentation-only: no new closure policy, result-envelope surface, state field, or control-flow branch.
- Tighten unsafe-context observability so structured logs remain content-free: the unsafe-context match event now emits only categorical/positional metrics (`pattern_name`, `context_length`, `match_start`, `match_end`, `matched_length`) and no raw prefixes, matched text, or raw-text-derived digests.

## Scope relative to #961 / AgentOS
This PR should be read as a narrow Track B auto-observability slice that supports the post-#1122/#1129 safe-default re-observation work. It is **not** an AgentOS Track C implementation and should not be treated as resolving the #961 roadmap/meta SSOT.

## Why this scope
Live `ooo auto` runs from the #821 acceptance set blocked at INTERVIEW, and the safe-default fallback added by #1122 did not expose enough structured evidence to distinguish:

1. no remaining defaultable sections / mutual-agreement deadlock with `ledger_done=True`,
2. unsafe-context gate trips, or
3. synthesis push timeout / non-closure.

This PR makes those decision points observable so a later live re-run can produce actionable evidence. No fix is attempted here.

## What is NOT done here (deliberate)
- No closure-policy change.
- No AgentOS Track C orchestration/runtime architecture change.
- No surface/envelope change to result meta.
- No taxonomy/stop_reason change.
- No claim that these unit tests are #821/#809 live/full-chain smoke evidence; they are synthetic instrumentation regression coverage only.
- No live-replay regression suite (deferred until the relevant closure fixes land).

## Test Plan
- [x] `uv run ruff format --check src/ouroboros/auto/safe_defaults.py tests/unit/auto/test_safe_default_observability.py`
- [x] `uv run ruff check src/ouroboros/auto/safe_defaults.py tests/unit/auto/test_safe_default_observability.py`
- [x] `uv run pytest tests/unit/auto/test_safe_default_observability.py -q` (4 tests)
- [x] CI: Ruff Lint, MyPy Type Check, Bridge TypeScript, Python 3.12/3.13/3.14 tests, boundary/envelope checks all pass on `f049f267`.

Refs #821, #809, #961, #1122, #1129.

PR #1167, opened 2026-05-22
feat(auto): safe-default closure mode + partial-unsafe blocker code (PR-B2)

shaun0927:
## Summary

PR-B2 of the L4 Auto Envelope v2 freeze (#1157, #821, #1138).

Closes two gaps observed in live ``ooo auto`` runs against the #821 acceptance matrix where the safe-default policy could have produced a clean closure but the result envelope did not reflect what happened:

- The existing safe-default success path closed the interview but never set ``state.interview_closure_mode``. Callers could not distinguish ``mutual_agreement`` (default ``None``) from a safe-default-applied closure. PR-B2 tags the envelope with ``interview_closure_mode = "safe_default"`` on that path.
- The **partial safe-default** case — some required gaps were safely defaultable but at least one remained unsafe (e.g. CONFLICTING ledger entry, per-section unsafe-context flag) — used to fall through to the generic ``interview_max_rounds_exhausted`` blocker with no structured event and no rollback. PR-B2 routes it to a dedicated event ``auto.interview.safe_default_partial_unsafe_gaps`` and a new typed code ``interview_unsafe_gaps_remain``. Partial defaults are rolled back via the existing ``_revert_safe_default_entries`` helper because synthesis was never pushed to the backend transcript — same invariant as the synthesis-failure rollback already in place.

The genuine-deadlock case (nothing defaultable at all) is **unchanged** — it continues to emit ``interview_max_rounds_exhausted``.

## Why this scope

This is the smallest first slice that closes the documented PR-B2 gap from the #1157 living SSOT and pushes the L4 Envelope v2 lane from 🟡 partial to closer-to-🟢-complete. After this lands, `closure_mode` taxonomy has four values (`None` / `ledger_only` / `safe_default`; blockers do not carry it), and `stop_reason_code` taxonomy has eight codes (3 interview-layer + 5 Ralph-layer).

## What is NOT done here

- No schema change, no manifest change, no new envelope field. Pure behavior + envelope-tag refinement on fields shipped by #1148 (PR-B1, ``interview_closure_mode``) and #1151 (PR-E, ``stop_reason_code``).
- No change to ``finalize_safe_defaultable_gaps`` or the safe-default policy itself — only the driver-level routing of its outcome.
- PR-C2 (``assumptions[].source`` provenance promotion) remains a separate follow-up.

## Scope

- ``src/ouroboros/auto/interview_driver.py``
  - Tag ``safe_default`` closure_mode on the existing success branch.
  - Insert the new partial-unsafe branch with rollback + typed code, ahead of the existing ``mutual_agreement_deadlock_at_max_rounds`` event emission.
- ``skills/auto/SKILL.md``
  - Extend the canonical ``stop_reason_code`` taxonomy table to 8 codes (adds ``interview_unsafe_gaps_remain``).
  - Add an Interview closure mode taxonomy table covering ``None`` / ``ledger_only`` / ``safe_default``.
- ``tests/unit/auto/test_interview_pipeline.py``
  - Extend the existing safe-default success test to assert ``state.interview_closure_mode == "safe_default"``.
  - Add a new test for the partial-unsafe path that constructs a CONFLICTING ledger entry on one section and asserts the new typed code, the rollback invariant, and the preserved user-recorded entry.
  - Extend the existing unsafe-everything test (case C) with a regression guard that ``last_error_code`` stays ``interview_max_rounds_exhausted``.

## Test plan

- [x] ``uv run pytest tests/unit/auto/test_interview_pipeline.py -k "safe_default or unsafe_gaps or finalize" -q`` → 39 passed
- [x] ``uv run pytest tests/unit/auto tests/integration/auto -q`` → 908 passed (baseline 907 + 1 new)
- [x] ``uv run ruff check`` on touched files → clean
- [x] ``uv run ruff format --check`` on touched files → clean
- [x] ``uv run mypy src/ouroboros/auto/interview_driver.py`` → clean

Refs #1157 (L4 lane), #821 (autonomy acceptance matrix), #1138 (PR-A instrumentation), #1148 (PR-B1 ``ledger_only``), #1151 (PR-E ``stop_reason_code``).

### DIFF
--- a/skills/auto/SKILL.md
+++ b/skills/auto/SKILL.md
@@ -71,7 +71,8 @@ When the user types `ooo auto` with CLI-style flags inside chat, translate to MC
 
 | Layer | Code | Surface | Meaning |
 |---|---|---|---|
-| Interview | `interview_max_rounds_exhausted` | `last_error_code`, `result.stop_reason_code` | Auto interview ran `max_interview_rounds` without ledger+backend mutual closure or safe-default fallback. |
+| Interview | `interview_max_rounds_exhausted` | `last_error_code`, `result.stop_reason_code` | Auto interview ran `max_interview_rounds` without ledger+backend mutual closure, no section was safely defaultable, and no partial defaults applied — i.e. genuine deadlock with nothing the policy could close. |
+| Interview | `interview_unsafe_gaps_remain` | `last_error_code`, `result.stop_reason_code` | Auto interview ran `max_interview_rounds` with at least one section safely defaultable and at least one section remaining unsafe (e.g. CONFLICTING ledger entry, production/credential context). Partial defaults are rolled back so the persisted transcript and ledger stay aligned; resume can address the unsafe gap and re-run. |
 | Interview | `interview_phase_deadline` | `last_error_code`, `result.stop_reason_code` | Interview phase exceeded its per-phase timeout. |
 | Ralph | `iteration_timeout` | blocker text + (future) `result.stop_reason_code` | A single Ralph iteration exceeded its per-iteration timeout. |
 | Ralph | `wall_clock_exhausted` | blocker text + (future) `result.stop_reason_code` | The Ralph wall-clock budget was exhausted before convergence. |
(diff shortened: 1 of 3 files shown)


Q00:
Follow-up policy note from #1219 / PR #1220: if the safe-default synthesis is successfully written to the persisted interview transcript but the backend returns a non-terminal follow-up turn, the auto driver should now fail forward as a `safe_default` closure. The ledger is already structurally complete and mirrored into the transcript, so rolling defaults back creates the observed cli-todo false block. The remaining fail-closed path is transcript-sync failure before the synthesis is persisted.

_Posted by [agentos-roadmap-warden](https://github.com/Q00/ouroboros/blob/main/agents/agentos-roadmap-warden.md) — bot. Reply with `/warden ignore` to suppress further comments on this thread._

PR #1168, opened 2026-05-22
fix(orchestrator): match command claims wrapped in output redirection/pager pipes

nkjunbc:
## Problem

In fat-harness mode the atomic verifier matches `commands_run` evidence against the runtime transcript through **exact** command aliases (`_normalized_command_claim_aliases` → `_runtime_message_supports_command_claim`).

Agents routinely run a command with output plumbing — e.g. `python -m ruff check src/foo.py tests/test_foo.py 2>&1 | tail -20` — while their evidence cites the clean `python -m ruff check src/foo.py tests/test_foo.py`. The trailing `2>&1` redirection and the `| tail -20` pager pipe make the two strings differ, so the claim is rejected as `FABRICATION_SUSPECTED` and the whole AC fails on correct work.

Observed in production: a sub-AC ran (and the transcript records)
```
python -m ruff check src/poc/structure_extractor.py tests/test_structure_and_draft_substance.py 2>&1 | tail -20
```
but the evidence claim omitted the `2>&1 | tail -20`, and the verifier reported `unsupported evidence claims: commands_run: python -m ruff check src/poc/structure_extractor.py ...`.

## Fix

`_normalized_command_claim_aliases` now also emits a **plumbing-stripped** alias via a new `_strip_command_output_plumbing` helper that removes:
- trailing output redirections — `2>&1`, `> log`, `2> err`, `&> out`, etc.
- a trailing pipe into a known output-filter command — `tail`, `head`, `cat`, `less`, `more`, `tee`, `wc`, `grep`, `egrep`, `fgrep`.

Because the stripped alias is added symmetrically to both the claim and the runtime command, `<cmd>` and `<cmd> 2>&1 | tail -20` now share an alias and match.

**Conservative by design / anti-fabrication preserved:**
- Alias matching stays **exact** (set intersection), not substring containment — the existing "no partial-substring proof" invariant holds.
- Only trailing output redirections and pipes into a known output-filter command are dropped. Meaningful pipelines such as `python gen.py | python process.py` are kept, so a partial `python gen.py` claim is still not proven.

## Tests

- `test_command_claim_supports_command_with_output_redirection_and_pager_pipe` — the clean claim now matches a `... 2>&1 | tail -20` run (fails on `main`, passes with this change).
- `test_command_claim_keeps_meaningful_pipeline_segments` — a real `a | python process.py` pipeline is **not** over-matched by a partial claim (guards against widening proof).
- Existing command-claim and fat-harness verifier tests continue to pass.

## Context

Relates to #1164 (implementer `subtype=success` rejected by the executor, then cascades `dependency_failed`). This PR addresses one root cause of that rejection in the verifier's command-matching logic. Companion to #1166, which lets a transcript-proven test run back node-id `tests_passed` claims; this PR lets a redirected/piped command back its clean `commands_run` claim. They are independent and can land in either order.

> Note: 3 pre-existing failures in `tests/unit/orchestrator/test_parallel_executor.py` (`*_normalizes_workspace_absolute_file_claim`, `*_rejects_unscoped_file_and_failed_test_command`, `*_accepts_wrapped_broad_pytest_for_current_test_file`) are unrelated Windows path-resolution issues in the local test harness; they fail identically on unmodified `main` and are untouched here.

### DIFF
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -906,8 +906,17 @@ def _looks_like_test_command(command: str) -> bool:
 
 
 def _test_command_invocation(command: str) -> str | None:
-    """Return the backed inner test invocation for a direct or wrapped command."""
-    normalized = command.strip().lower()
+    """Return the backed inner test invocation for a direct or wrapped command.
+
+    Output plumbing (``2>&1``, ``| tail -20``) is peeled first so a clean
+    invocation can be extracted from a ``<cmd> 2>&1 | tail -20`` runtime
+    command. ``_strip_command_output_plumbing`` is deliberately narrow — only
+    presentation-only tails are peeled — so an evidence-altering filter such
+    as ``| grep passed`` survives the strip and is rejected downstream by
+    ``_test_invocation_from_prefix`` rather than being silently dropped.
+    """
+    stripped = _strip_command_output_plumbing(command)
+    normalized = stripped.lower()
     if not normalized:
         return None
 
@@ -1012,13 +1021,24 @@ def _strip_env_prefix(parts: list[str]) -> list[str]:
 
 
 def _test_invocation_from_prefix(command: str) -> str | None:
-    """Return a normalized test invocation only when it starts the command text."""
+    """Return a normalized test invocation only when it starts the command text.
+
+    Refuses to extract from commands that still contain a residual shell pipe
+    after presentation plumbing has been peeled (``pytest x | grep passed``).
+    A residual pipe means the runtime command is followed by an
+    evidence-transforming filter (``grep`` / ``wc`` / ``tee``); treating the
+    bare prefix as the clean test invocation there would let a filtered run
+    silently back a clean ``tests_passed`` / ``commands_run`` claim via the
+    ``startswith`` widening in ``_runtime_message_supports_command_claim``.
+    """
     try:
         parts = shlex.split(command)
     except ValueError:
         parts = command.replace('"', "").replace("'", "").split()
     parts = _strip_env_prefix(parts)
     if not parts:
+        return None
+    if "|" in parts:
         return None
 
     if parts[0] in {"pytest", "py.test", "tox", "nox"}:
@@ -1058,9 +1078,15 @@ def _looks_like_unittest_command(command: str) -> bool:
 
 # Output-only shell filters: a trailing pipe into one of these is presentation
 # or paging, not the work an evidence claim is about.
-_OUTPUT_FILTER_COMMANDS = frozenset(
-    {"tail", "head", "cat", "less", "more", "tee", "wc", "grep", "egrep", "fgrep"}
-)
+#
+# Deliberately narrow: only filters that pass the output stream through (or
+# truncate it positionally) are allowed. ``grep``/``egrep``/``fgrep`` are
+# excluded because they can hide failure lines and make a filtered run back a
+# clean ``commands_run`` / ``tests_passed`` claim (e.g. ``pytest ... | grep
+# passed``). ``tee`` and ``wc`` are excluded for the same reason: ``tee`` can
+# redirect the stream and ``wc`` collapses it to a count, both of which alter
+# what the runtime would have observed and so weaken anti-fabrication.
+_OUTPUT_FILTER_COMMANDS = frozenset({"tail", "head", "cat", "less", "more"})
 
 # Trailing shell output redirection (``2>&1``, ``> log``, ``2> err``, ``&> out``).
 _TRAILING_REDIRECT_RE = re.compile(
@@ -1073,10 +1099,21 @@ def _strip_command_output_plumbing(command: str) -> str:
 
     Agents routinely run ``<cmd> 2>&1 | tail -20`` while their ``commands_run``
     evidence cites the clean ``<cmd>``. The trailing redirection and the
-    output-only filter pipe are presentation plumbing, not the work being
-    claimed, so they must not block a match. Deliberately conservative: only
-    trailing output redirections and pipes into a known output-filter command
-    are dropped, so meaningful pipelines (``a | python process.py``) are kept.
+    output-only pager pipe are presentation plumbing, not the work being
+    claimed, so they must not block a match. Deliberately conservative:
+
+    - Only trailing output redirections (``2>&1``, ``> log``, ``2> err``,
+      ``&> out``) and pipes into a pager-style filter listed in
+      ``_OUTPUT_FILTER_COMMANDS`` (``tail``/``head``/``cat``/``less``/``more``)
+      are dropped. These pass the underlying stream through (or truncate it
+      positionally), so the runtime evidence is unchanged in kind.
+    - Filters that *transform* the stream — ``grep`` family, ``wc``, ``tee`` —
+      are intentionally not stripped, because they can hide failure lines,
+      collapse the stream to a count, or divert it to a file, which would let
+      a filtered run back a clean ``commands_run`` / ``tests_passed`` claim.
+    - Meaningful pipelines such as ``a | python process.py`` are kept, so a
+      partial ``a`` claim is still not proven by an ``a | python process.py``
+      runtime command.
     """
     text = command.strip()
     if not text:
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -1056,6 +1056,47 @@ def _looks_like_unittest_command(command: str) -> bool:
     return _unittest_command_invocation(command) is not None
 
 
+# Output-only shell filters: a trailing pipe into one of these is presentation
+# or paging, not the work an evidence claim is about.
+_OUTPUT_FILTER_COMMANDS = frozenset(
+    {"tail", "head", "cat", "less", "more", "tee", "wc", "grep", "egrep", "fgrep"}
+)
+
+# Trailing shell output redirection (``2>&1``, ``> log``, ``2> err``, ``&> out``).
+_TRAILING_REDIRECT_RE = re.compile(
+    r"\s*(?:[0-9]*>{1,2}\s*(?:&[0-9]+|[^\s|]+)|&>{1,2}\s*[^\s|]+)\s*$"
+)
+
+
+def _strip_command_output_plumbing(command: str) -> str:
+    """Return a command with trailing output redirection and pager pipes removed.
+
+    Agents routinely run ``<cmd> 2>&1 | tail -20`` while their ``commands_run``
+    evidence cites the clean ``<cmd>``. The trailing redirection and the
+    output-only filter pipe are presentation plumbing, not the work being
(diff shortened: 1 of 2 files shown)


shaun0927:
## Merge readiness — follow-up commits land the bot's blocker

### What this PR does (after follow-up commits)

This PR fixes a fat-harness verifier false-rejection where a `commands_run` /
`tests_passed` evidence claim like

```
python -m ruff check src/foo.py tests/test_foo.py
```

was rejected as `FABRICATION_SUSPECTED` whenever the agent actually ran the
same command with trailing output plumbing such as

```
python -m ruff check src/foo.py tests/test_foo.py 2>&1 | tail -20
```

The aliases differed by the redirection and pager pipe, so the exact
set-intersection in `_normalized_command_claim_aliases` →
`_runtime_message_supports_command_claim` failed and a whole AC went red on
correct work. See #1164 for the upstream cascade.

### What the follow-up commits improved

The first commit (5460477d) introduced `_strip_command_output_plumbing` and
applied it symmetrically to claim and runtime aliases, so a clean claim and a
`<cmd> 2>&1 | tail -20` run share a stripped alias.

`ouroboros-agent[bot]` then flagged a real blocker on that commit: the
allowlist included `grep`/`egrep`/`fgrep`, `wc`, and `tee`, which are *not*
presentation-only. Treating them as removable plumbing would have widened
the anti-fabrication boundary — e.g. `pytest tests/foo.py | grep passed`
could have backed a clean `pytest tests/foo.py` claim while hiding failure
lines. The bot also pointed out that the same alias path is consumed by
`tests_passed` matching at `parallel_executor.py:1322`, so the leak would
have applied to test-run evidence as well.

Commits 4eddff8e and 668127b0 close that gap on three matching paths:

1. **Allowlist narrowing.** `_OUTPUT_FILTER_COMMANDS` now contains only
   `{"tail", "head", "cat", "less", "more"}` — filters that pass the stream
   through or truncate it positionally. The inline comment documents
   *why* `grep` / `wc` / `tee` are excluded (hide / collapse / divert the
   stream).
2. **`_test_command_invocation` peels first, then extracts.** It calls
   `_strip_command_output_plumbing` so `<cmd> 2>&1 | tail -20` still yields
   the clean test invocation, but a non-presentation pipe survives the
   strip and is rejected downstream.
3. **`_test_invocation_from_prefix` rejects residual pipes.** After plumbing
   is peeled, any remaining un-quoted `|` token short-circuits the
   extraction to `None`. This kills the `startswith(claim + " ")` widening
   in `_runtime_message_supports_command_claim` (lines 674–683), which was
   the pre-existing path a grep-filtered run could have used to back a
   `tests_passed` claim. Quoted pipes (`pytest "tests/foo|bar.py"`) are
   safe because `shlex.split` keeps them as one token.

Four new negative tests guard the boundary against `| grep`, `| grep PASSED`
(targeting the `tests_passed` startswith path specifically), `| wc -l`, and
`| tee pytest.log`. The positive `2>&1 | tail -20` regression and the
`a | python process.py` non-output-pipeline guard from the original commit
continue to pass.

### Why this is mergeable

- **SSOT alignment.** #961 explicitly lists this PR as an open, green,
  APPROVED fat-harness verifier evidence follow-up under Track A. Scope
  stays within that mandate; no `ooo auto` or Track C surface is touched.
- **Direct fix for the bot's blocker.** The `grep`/`wc`/`tee` leak is closed
  on the alias set-intersection path *and* on the `startswith` widening
  path, with explicit negative tests for both `commands_run` and
  `tests_passed` consumers.
- **Anti-fabrication invariant preserved and strengthened.** Alias matching
  remains an exact set intersection. The only behavioral change is that
  presentation-only tails (`tail`/`head`/`cat`/`less`/`more`) and trailing
  output redirections (`2>&1`, `> log`, `&> out`) are now symmetrically
  ignored on both sides. Filters that can transform the runtime evidence
  stream are explicitly *not* stripped, with an inline rationale.
- **No collateral damage.** All 195 tests in
  `tests/unit/orchestrator/test_parallel_executor.py` pass locally; CI is
  green on Ruff, MyPy, Bridge TypeScript, Python 3.12/3.13/3.14 tests, and
  the `enforce-boundary` / `enforce-envelope` gates. `mergeStateStatus`
  reports `CLEAN`.
- **Surface area is minimal.** Two files, ~140 net additions; no public
  API changes; no plugin contract or workflow IR touched.

Net effect: a stuck-on-correct-work false rejection is fixed without
weakening the verifier's trust boundary — the anti-fabrication invariant is
actually tighter now than before the PR was opened, because the residual
`|` guard also closes the pre-existing `startswith` widening path for
test-evidence claims.

shaun0927:
# PR Review Summary

> Posted by `/pr-review` skill. HEAD reviewed: `668127b0` (3 commits since `upstream/main`, +207/-3 across 2 files).

## Verdict
**Approve**

## Scope Reviewed
- **PR intent:** fix a fat-harness verifier false-rejection (`FABRICATION_SUSPECTED`) where `commands_run` / `tests_passed` claims that cite a clean command are rejected when the runtime ran the same command with trailing output plumbing (`2>&1 | tail -20`). Tracked under #1164. Listed in #961 SSOT as a Track A fat-harness verifier evidence follow-up.
- **Main changed areas:**
  - `src/ouroboros/orchestrator/parallel_executor.py` — adds `_strip_command_output_plumbing`, narrows `_OUTPUT_FILTER_COMMANDS` to presentation-only filters, peels plumbing inside `_test_command_invocation`, and adds a residual-`|` guard inside `_test_invocation_from_prefix`.
  - `tests/unit/orchestrator/test_parallel_executor.py` — 1 positive regression test (`2>&1 | tail -20`), 1 pipeline-preservation guard (`a | python process.py`), 4 anti-fabrication negative tests for `| grep`, `| grep PASSED` (tests_passed path), `| wc -l`, `| tee out.log`.
- **Tests reviewed:** all 6 new tests + the existing 189-test suite in `tests/unit/orchestrator/test_parallel_executor.py`. 195 / 195 pass locally; CI shows Ruff Lint, MyPy Type Check, Bridge TypeScript, Python 3.12 / 3.13 / 3.14 tests, `enforce-boundary`, `enforce-envelope` all green at `668127b0`.
- **Checks considered:** `mergeStateStatus = CLEAN`; PR is `MERGEABLE`; SSOT alignment with #961 Track A; security boundary preservation of the verifier's anti-fabrication invariant.

## Blocking Issues
None.

## Warnings
None.

## Mutation-Test Thinking
- **Likely mutants that should be killed:**
  - Adding `grep` / `wc` / `tee` back into `_OUTPUT_FILTER_COMMANDS` (line 1089). Killed by `test_command_claim_rejects_grep_filtered_run_as_clean_command`, `…_as_tests_passed_claim`, `…_wc_collapsed_run_…`, `…_tee_redirected_run_…`.
  - Removing the `"|" in parts` guard in `_test_invocation_from_prefix` (line 1041). Killed by the same four negative tests via the `_test_command_invocation` → `_runtime_message_supports_command_claim` startswith path.
  - Removing the `_strip_command_output_plumbing(command)` call inside `_test_command_invocation`. Killed by the positive regression test, which would then fail to match `<cmd> 2>&1 | tail -20`.
  - Replacing `set.intersection` with `set.issubset` / containment in `_runtime_message_supports_command_claim:667`. Not a new mutation introduced here, but the existing `test_command_claim_keeps_meaningful_pipeline_segments` guards against widening to substring containment.
  - Loop direction flip in `_strip_command_output_plumbing` (using `partition` instead of `rpartition`). Killed by the pipeline-preservation test, which requires *trailing* segment inspection only.
- **Mutants current tests may not catch:**
  - Adding a *new* filter to the allowlist (e.g., `awk`, `cut`, `sed`). The current negative tests only enumerate `grep` / `wc` / `tee`, so a future PR could quietly expand `_OUTPUT_FILTER_COMMANDS` without breaking these tests. The inline comment at lines 1079–1088 mitigates this by making the allowlist policy explicit.
- **Additional tests recommended (non-blocking):** a parametrized test that asserts the allowlist is exactly `{"tail", "head", "cat", "less", "more"}` would lock the policy at the data level. Optional follow-up, not required for merge.

- **High-risk functions/modules:** `_strip_command_output_plumbing` (new, ~25 LOC), `_test_command_invocation` (modified, low complexity), `_test_invocation_from_prefix` (modified, added one branch).
- **Complexity increase:** minimal. Two short while-loops with explicit fixed-point guards and a single `"|" in parts` early-return. Cyclomatic complexity stays low; no nested decision logic.
- **Test coverage concern:** none — every new branch has at least one positive and one negative regression test.
- **Refactoring recommendation:** none. The helper is appropriately localized; pulling it into a shared module would be premature.

## Test Quality Assessment (6/7)
- **Strong tests:**
  - `test_command_claim_rejects_grep_filtered_run_as_tests_passed_claim` exercises the `_test_command_invocation` startswith widening path explicitly with `-x`, which is the specific path the bot's original blocker flagged at `parallel_executor.py:1322`.
  - `test_command_claim_keeps_meaningful_pipeline_segments` is a true anti-widening guard: it would fail if the implementation ever degraded to substring containment.
- **Weak tests:** none material.
- **Missing edge cases:**
  - Quoted-pipe edge case (`pytest "tests/foo|bar.py"`) is not tested directly, but the implementation uses `shlex.split` and the `"|" in parts` check operates on tokens, so a quoted `|` becomes one token and is correctly *not* rejected. Non-blocking; would be a nice belt-and-suspenders test.
  - Chained plumbing (`<cmd> 2>&1 | head -50 | tail -10`) is not tested, but the while-loop peels rightmost-first and terminates correctly. Non-blocking.
- **Mocking concerns:** none — tests use real `AgentMessage` instances and exercise the public matching helper end-to-end.

None. The verifier's anti-fabrication trust boundary is *strengthened*, not weakened:
- Alias matching remains strict set intersection (no widening to substring containment).
- Only filters that pass the stream through or truncate it positionally are treated as removable plumbing. Filters that hide (`grep`), collapse (`wc`), or divert (`tee`) the runtime stream are explicitly excluded, with documented rationale.
- The new residual-`|` guard also closes a pre-existing soft spot in the `_test_command_invocation` startswith path that could have backed a `tests_passed` claim from a grep-filtered run, even before this PR existed.
- No public API change; no plugin contract / workflow IR / event schema touched; no migration concerns.

## Looks Good
- Direct, narrow fix for a real production failure observed at `python -m ruff check … 2>&1 | tail -20` vs the cited clean claim.
- The bot's REQUEST_CHANGES blocker on commit `5460477d` (`grep`/`wc`/`tee` in allowlist) is fully resolved on both matching paths, with explicit negative tests.
- Inline policy comment at `parallel_executor.py:1079–1088` documents *why* each excluded filter is excluded, which prevents accidental regressions when future maintainers see the short allowlist.
- CI fully green; `mergeStateStatus = CLEAN`; surface area is two files, ~140 net additions; no scope creep into `ooo auto` or Track C surfaces.

## Final Recommendation
**Approve and merge.** The PR delivers a tight fix that aligns with #961's Track A fat-harness verifier evidence mandate, resolves the bot's only blocker on the prior HEAD, and leaves the anti-fabrication invariant strictly tighter than it was before. No blocking issues, no warnings.

— `/pr-review`

PR #1194, opened 2026-05-23
fix(auto): clear stale stop reason codes

Q00:
## Summary
- Clear `AutoPipelineState.last_error_code` whenever the session transitions out of `BLOCKED`.
- Add regression coverage for recovery and later hard-failure transitions so `AutoPipelineResult.stop_reason_code` cannot report a stale interview blocker.

## Review context
This follows up the post-v0.39.1 `stop_reason_code` surface from PR #1151. The public result envelope copies `state.last_error_code`; without clearing it on non-blocked transitions, a resumed session could recover or fail for a different reason while still exposing an old canonical interview blocker code.

## Validation
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run pytest tests/unit/auto/test_stop_reason_code.py tests/unit/auto/test_pipeline_watchdog_integration.py tests/unit/auto/test_surface.py -q` -> 96 passed
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run pytest tests/unit/auto tests/unit/mcp/tools/test_generate_seed_force_flag.py tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_runtime_evidence.py tests/unit/plugin tests/unit/runtime tests/canonical -q` -> 1677 passed, 2 skipped
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run ruff check src/ouroboros/auto/state.py tests/unit/auto/test_stop_reason_code.py` -> passed
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run ruff format --check src/ouroboros/auto/state.py tests/unit/auto/test_stop_reason_code.py` -> passed
- `git diff --check` -> passed

### DIFF
--- a/src/ouroboros/auto/state.py
+++ b/src/ouroboros/auto/state.py
@@ -598,6 +598,8 @@ class AutoPipelineState:
         self.updated_at = now
         self.last_progress_message = message
         self.last_error = error
+        if next_phase is not AutoPhase.BLOCKED:
+            self.last_error_code = None
         # Authoring-backend attribution is scoped to the most recent
         # authoring failure; reset on every transition so a later
         # non-authoring blocker (grade_gate, seed_saver, run_starter)
--- a/tests/unit/auto/test_stop_reason_code.py
+++ b/tests/unit/auto/test_stop_reason_code.py
@@ -176,3 +176,33 @@ async def test_blockers_without_canonical_code_leave_stop_reason_code_none(tmp_p
     assert result.stop_reason_code is None
     assert result.blocker is not None
     assert "pipeline budget exhausted" in result.blocker
+
+
+def test_recovery_transition_clears_stale_stop_reason_code(tmp_path) -> None:
+    """A recovered session must not keep reporting an old blocker code."""
+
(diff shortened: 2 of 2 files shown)


shaun0927 (approved):
Cross-checked against #1151 (the merged `stop_reason_code` envelope) and #1157 (the `ooo auto` SSOT, which names this PR as the remaining L2 follow-up for stop-reason-code clearing). The two-line guard in `state.transition()` is the right scope: it clears `last_error_code` exactly when the session leaves BLOCKED, leaves the BLOCKED→BLOCKED self-edge untouched, and the two added tests pin both the `recover()` and `mark_failed()` paths. No regression risk vs. the documented `last_error_code` semantics.

FYI #1205 introduces a *competing* hunk at the same location with condition `if error is None`, which diverges on `transition(BLOCKED, msg, error=None)` and on non-BLOCKED→non-BLOCKED with an explicit error string — I left a change-request on #1205 to drop its hunk so this PR can own the contract cleanly. Land this before #1205 rebases.

PR #1188, opened 2026-05-22
feat(auto): Seed AC injection + active_task_class envelope (L1-d, L1-e)

shaun0927:
## Summary

L1-d / L1-e slices of #1171 — consumption side of L1-a (#1173) catalog + L1-b (#1177) inference.

After Seed generation, the pipeline now:

1. Calls \`derive_domain_from_ledger\` on the standardized ledger.
2. On a **single-class** inference, prepends any non-duplicate entries from the catalog's \`default_ac_template\` to the Seed's \`acceptance_criteria\` via \`Seed.model_copy\`.
3. Persists \`state.active_task_class\` for resume + envelope.
4. Surfaces \`active_task_class\` on \`AutoPipelineResult\`.

Ambiguous / unmatched inferences are **no-ops** — neither AC injection nor envelope tagging fires. The interview-driver disambiguation hook (L1-c) is a separate evidence-driven follow-up.

## What lands

- \`src/ouroboros/auto/task_class_application.py\` (new): \`AppliedTaskClassDefaults\` + \`apply_default_ac_template\` helper (prepend + dedupe + no-op fast path).
- \`src/ouroboros/auto/state.py\`: add \`active_task_class\` field + serialization wiring.
- \`src/ouroboros/auto/pipeline.py\`: import + call after seed_generator; add \`active_task_class\` to \`AutoPipelineResult\`; populate in \`_result\`.
- Tests:
  - \`tests/unit/auto/test_task_class_application.py\` (9 helper tests).
  - \`tests/unit/auto/test_pipeline_task_class_envelope.py\` (2 end-to-end tests: single-match populates envelope + injects AC; ambiguous → both None).
  - \`tests/unit/auto/test_interview_pipeline.py\`: \`test_pipeline_repairs_b_seed_to_a_and_starts_run\` updated to locate the repaired user AC by content rather than index.

## What is NOT in this PR

- Interview-driver disambiguation hook (L1-c) — ambiguous cases remain no-ops here.
- \`domain_unmatched\` telemetry — added when a real catalog gap surfaces.
- Runtime probe binding consumption — separate L3-2 lane.

## Test plan

- [x] \`uv run pytest tests/unit/auto -q\` → 916 passed.
- [x] \`uv run ruff check\` on touched files → clean.
- [x] \`uv run ruff format\` on touched files → clean.
- [x] \`uv run mypy\` on touched files → clean.

Refs #1157, #1171, #1173 (L1-a catalog), #1177 (L1-b matcher).

PR #1196, opened 2026-05-23
fix(auto): expose active task class in MCP meta

Q00:
## Summary
- Surface `AutoPipelineResult.active_task_class` through MCP `_result_meta`.
- Preserve the field as `null` when inference was ambiguous or unavailable, so clients can distinguish an empty computed value from a missing protocol field.
- Add regression coverage for populated and empty active-task-class metadata.

## Review context
This follows up the post-v0.39.1 L1 task-class sequence. PR #1188 added `active_task_class` to `AutoPipelineResult` and persisted state, but the public MCP metadata envelope omitted it. That meant MCP clients could not observe the L1-e result-envelope surface without reaching into local state files.

## Validation
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run pytest tests/unit/auto/test_surface.py tests/unit/auto/test_pipeline_task_class_envelope.py -q` -> 91 passed
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run ruff check src/ouroboros/mcp/tools/auto_handler.py tests/unit/auto/test_surface.py` -> passed
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.39.2 uv run ruff format --check src/ouroboros/mcp/tools/auto_handler.py tests/unit/auto/test_surface.py` -> passed
- `git diff --check` -> passed

### DIFF
--- a/src/ouroboros/mcp/tools/auto_handler.py
+++ b/src/ouroboros/mcp/tools/auto_handler.py
@@ -1205,6 +1205,7 @@ def _result_meta(result: AutoPipelineResult) -> dict[str, Any]:
         "blocker": result.blocker,
         "seed_path": result.seed_path,
         "seed_origin": result.seed_origin,
+        "active_task_class": result.active_task_class,
         "grade": result.grade,
         "last_grade": result.last_grade,
         "interview_session_id": result.interview_session_id,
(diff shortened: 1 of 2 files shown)


shaun0927 (approved):
Cross-checked against #1188 (which added `active_task_class` to `AutoPipelineResult` + persistence) and #1157 (which names this PR as the remaining L1 MCP-metadata cleanup). The unconditional `meta["active_task_class"] = result.active_task_class` is exactly the right shape — PR body justifies it explicitly: clients need to distinguish "inference was ambiguous/unavailable" (null) from "protocol field absent" (key missing). Two tests cover populated + null. No risk.

Heads-up: #1205 silently writes the same field with `if ... is not None`, which would downgrade this contract back to "key omitted on null". I requested changes on #1205 to drop that hunk so this PR can own the surface. Recommend landing this one before #1205 rebases.

PR #1193, opened 2026-05-22
Keep plugin runtime writes outside trusted homes

shaun0927:
## Summary

Closes #1192 by giving UserLevel plugin subprocesses a safe runtime environment while preserving `cwd=plugin_home` import resolution.

## Problem

The plugin dispatcher runs entrypoints from `plugin_home` so commands like `python -m superpowers_ouroboros` resolve installed modules. But `plugin_home` is also the immutable trust subject verified before every invocation. Without a runtime env contract:

- Python can create `__pycache__` under `plugin_home`.
- Plugins may default runtime artifacts to `./...` under `plugin_home`.
- The next invocation fails closed with trust-subject drift.

## Changes

`firewall.invoke_plugin(...)` now passes plugin subprocesses and lifecycle hooks:

- `PYTHONDONTWRITEBYTECODE=1`
- `OUROBOROS_PLUGIN_HOME=<plugin_home>` when available
- `OUROBOROS_PLUGIN_WORKDIR=<caller cwd>`
- `OUROBOROS_PLUGIN_OUTPUT_DIR=<caller cwd>/.ouroboros/plugin-artifacts/<plugin-name>`

This keeps code in plugin homes and runtime artifacts in an Ouroboros-native user/workspace artifact root. It intentionally avoids an `.omx` default in core so non-OMX Ouroboros users get a native path.

## Validation

```bash
PYTHONPATH=src pytest -q \
  tests/unit/plugin/test_firewall.py::test_plugin_subprocess_receives_immutable_home_runtime_environment \
  tests/unit/plugin/test_firewall.py::test_subprocess_failure_emits_failed_with_exit_code
```

Result: `2 passed`.

Cross-repo smoke with the companion `ouroboros-plugins` branch:

- `ouroboros plugin add ... --plugin superpowers`
- `ouroboros plugin trust superpowers --scope filesystem:read --scope filesystem:write`
- `ouroboros superpowers list`
- `ouroboros superpowers test-driven-development ...`
- `ouroboros superpowers verification-before-completion ...`

Repeated dispatch succeeded without plugin-home digest drift when paired with the plugin-side output-dir fix.

PR #1191, opened 2026-05-22
feat(tests): L0 live-wire + L1 catalog cross-validate (P1)

shaun0927:
## Summary

L0 follow-up of #1170 — wires the \`OUROBOROS_RUN_CANONICAL=1\` path to programmatically invoke the \`ouroboros_auto\` MCP tool against a canonical scenario, and adds the L1 catalog cross-validation tests deferred from L0-a (#1174) until L1-a (#1173) was on main.

## What lands

- **L1 catalog cross-validate (always runs, hermetic):**
  - \`test_scenario_domain_class_resolves_in_l1_catalog\` — pins \`expected.yaml\` \`domain_class\` to a real \`TaskClass\` value.
  - \`test_scenario_completion_mode_matches_l1_catalog_default\` — pins \`completion_mode\` to the L1 catalog default for the scenario's class.
  - \`test_scenario_runtime_probe_kinds_subset_of_l1_catalog\` — pins \`runtime_probe_kinds\` to a subset of the catalog's declared probes.
- **Live-run wiring (opt-in via env var):** when \`OUROBOROS_RUN_CANONICAL=1\` is set, the test programmatically invokes \`AutoHandler.handle(...)\` with a goal/cwd/skip_run/pipeline_timeout_seconds arguments dict built from the scenario fixture. Result assertions surface MCP misconfiguration with the underlying error message.

## What is NOT in this PR

- Additional scenario fixtures (\`webhook-receiver\`, \`vertical-slice-refactor\`, \`2d-kart-racer\`) — fixtures are maintainer-curated example data, not substrate PRs.
- Probe runner wiring for the canonical workdir.
- Recorded-replay layer — deferred per #1170 minimal-substrate audit.

## Test plan

- [x] \`uv run pytest tests/canonical/ -v\` (hermetic default) → 9 passed, 1 skipped (live-run gate).
- [x] \`uv run ruff check tests/canonical/test_canonical.py\` → clean.
- [x] \`uv run ruff format tests/canonical/test_canonical.py\` → clean.
- [x] \`uv run mypy tests/canonical/test_canonical.py\` → clean.

Refs #1157 (L0 lane), #1170 (L0 design), #1174 (L0-a skeleton), #1173 (L1-a catalog).

PR #1146, opened 2026-05-20
feat(auto): surface defaulted_sections in AutoPipelineResult

shaun0927:
## Summary
- Add ``defaulted_sections: tuple[str, ...] = ()`` field to ``AutoPipelineResult``.
- Populate it from ``ledger.summary()["defaulted_sections"]`` in ``_result()``.
- Zero behavior change. No ledger or signature changes.

## Why this scope
Live ``ooo auto`` runs against the #821 acceptance matrix observed that callers could not tell from the result meta whether the safe-default policy fired or which sections were auto-filled. The ledger already tracks them; the auto pipeline just was not propagating the field. This PR closes #821 acceptance criterion #3 for the ``defaulted_sections[]`` half.

## What is NOT done here
- No ``[from-auto][source]`` provenance promotion on ``assumptions[]`` — deferred to a follow-up because it requires a ``ledger.assumptions()`` return-type change with broader surface impact.
- No closure-mode field (PR-B1 adds ``interview_closure_mode``).
- No taxonomy/stop_reason_code change (PR-E).

## Test Plan
- [x] ``uv run pytest tests/unit/auto/test_pipeline_defaulted_sections_surface.py -v`` (3 new tests pass)
- [x] ``uv run pytest tests/unit/auto tests/integration/auto -q`` (no regression)
- [x] ``uv run ruff check`` on touched files
- [x] ``uv run mypy`` on touched files

Refs #821, #763, #1138.

PR #1134, opened 2026-05-19
feat(orchestrator): record durable workflow lifecycle events

shaun0927:
## Summary

Adds the durable workflow lifecycle event family for the #956 Workflow IR. Lifecycle events are persisted through the existing `EventStore` under a new `workflow_ir` aggregate type and can be replayed and projected deterministically without any change to live `parallel_executor` dispatch or any existing event family.

Refs #1131, #956.

### Scope

- `WorkflowLifecycleEvent` now serializes through a registered family (`WORKFLOW_LIFECYCLE_AGGREGATE_TYPE = "workflow_ir"`, `WORKFLOW_LIFECYCLE_EVENT_TYPES`) and exposes `from_base_event` for replay.
- `WorkflowLifecycleProjection` + `project_workflow_lifecycle` rebuild run/node/edge/checkpoint state from a lifecycle slice deterministically.
- `EventStore.append_workflow_lifecycle_event` / `replay_workflow_lifecycle` register the new family additively. No other event family is created or modified.
- Payloads are size-bounded and replay-unsafe keys (`stdout`, `stderr`, `prompt`, credentials, tokens) are rejected at the model boundary — proven by fixture.
- Lifecycle events carry only opaque `ControlContract` / `CheckpointStore` / `IOJournal` reference ids via `refs`.

### Out of scope (intentionally untouched)

- `parallel_executor` dispatch, node execution, live workflow behavior.
- Plugin lifecycle dispatch (#939), HITL authority (#960), evidence schema (#830/#978), workflow IR schema.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/unit/orchestrator/ -q` — 1636 passed, 1 skipped.
- [x] `PYTHONPATH=src python -m pytest tests/unit/orchestrator/test_workflow_lifecycle_events.py -q` — 21 passed.
- [x] `PYTHONPATH=src python -m pytest tests/unit/persistence/ -q` — 154 passed.
- [x] `ruff check` + `ruff format --check` clean on all touched files.
- [x] Replay fixture distinguishes `completed` / `failed` / `retried` / `cancelled` runs.
- [x] Same event slice rebuilds the same `WorkflowLifecycleProjection` regardless of input ordering.
- [x] Persisted-payload fixture asserts no raw `stdout` / `stderr` / `prompt` / credential substring leaks at either the `BaseEvent.to_db_dict()` layer or the round-trip SQLite layer.

PR #1205, opened 2026-05-24
fix: resolve post-merge review blocker foundations

Q00:
## Summary
- clear stale auto stop reason codes on recovery transitions
- expose missing auto result metadata: interview_closure_mode, active_task_class, runtime probe evidence, and external Ralph product status
- render defaulted sections in CLI ledger output
- force plugin runtime env path contracts instead of inheriting stale overrides
- make runtime timeout/deferred evidence safe for consumers
- fix task-class inference overlap and unmatched no-op behavior
- classify invalid status-run limits as malformed input
- make workflow lifecycle same-timestamp restarts replay to the fresh run
- lock mechanical-eval verdict evidence event linkage

## Original PR blocker coverage
Addresses blocker foundations for #1151, #1148, #1146, #1193, #1191, #1188, #1181, #1177, #1133, #1134, and #1132.

## Verification
- uv run ruff check src tests
- uv run pytest tests/unit/auto/test_stop_reason_code.py tests/unit/auto/test_pipeline_task_class_envelope.py tests/unit/auto/test_domain_inference.py tests/unit/auto/test_surface.py tests/unit/cli/test_auto_command.py tests/unit/plugin/test_firewall.py tests/unit/orchestrator/test_runtime_evidence.py tests/unit/cli/test_status_run.py tests/unit/orchestrator/test_workflow_lifecycle_events.py tests/integration/test_mechanical_eval_projection.py -q

### DIFF
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -648,15 +648,11 @@ class AutoPipeline:
                     state.seed_origin = SeedOrigin.AUTO_PIPELINE
                     # L1-d / #1171: derive the task class from the already-
                     # standardized ledger and prepend the catalog's default
-                    # AC template to the Seed. Single match → apply that class;
-                    # unmatched → apply the safe LIBRARY fallback from L1-b;
-                    # ambiguous → leave the user's AC untouched until the
-                    # L1-c interview-driver disambiguation hook resolves it.
-                    # ``state.active_task_class`` is set only when a concrete
-                    # class/fallback actually fired, so the envelope does not
-                    # pretend an unresolved ambiguous class was active.
+                    # AC template only for a concrete single match. Ambiguous
+                    # or unmatched ledgers leave the user's AC untouched until
+                    # a disambiguation hook resolves the class.
                     _inference = derive_domain_from_ledger(ledger)
-                    _task_class = _inference.single
+                    _task_class = next(iter(_inference.classes)) if _inference.is_single else None
                     if _task_class is not None:
                         _applied = apply_default_ac_template(seed, _task_class)
                         if _applied.injected_ac:
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -648,11 +648,15 @@ class AutoPipeline:
                     state.seed_origin = SeedOrigin.AUTO_PIPELINE
                     # L1-d / #1171: derive the task class from the already-
                     # standardized ledger and prepend the catalog's default
-                    # AC template only for a concrete single match. Ambiguous
-                    # or unmatched ledgers leave the user's AC untouched until
-                    # a disambiguation hook resolves the class.
+                    # AC template to the Seed. Single match → apply that class;
+                    # unmatched → apply the safe LIBRARY fallback from L1-b;
+                    # ambiguous → leave the user's AC untouched until the
+                    # L1-c interview-driver disambiguation hook resolves it.
+                    # ``state.active_task_class`` is set only when a concrete
+                    # class/fallback actually fired, so the envelope does not
+                    # pretend an unresolved ambiguous class was active.
                     _inference = derive_domain_from_ledger(ledger)
-                    _task_class = next(iter(_inference.classes)) if _inference.is_single else None
+                    _task_class = _inference.single
                     if _task_class is not None:
                         _applied = apply_default_ac_template(seed, _task_class)
                         if _applied.injected_ac:
--- a/src/ouroboros/auto/state.py
+++ b/src/ouroboros/auto/state.py
@@ -598,6 +598,8 @@ class AutoPipelineState:
         self.updated_at = now
         self.last_progress_message = message
         self.last_error = error
+        if error is None:
+            self.last_error_code = None
         # Authoring-backend attribution is scoped to the most recent
         # authoring failure; reset on every transition so a later
         # non-authoring blocker (grade_gate, seed_saver, run_starter)
--- a/src/ouroboros/auto/state.py
+++ b/src/ouroboros/auto/state.py
@@ -598,8 +598,6 @@ class AutoPipelineState:
         self.updated_at = now
         self.last_progress_message = message
         self.last_error = error
-        if error is None:
-            self.last_error_code = None
         # Authoring-backend attribution is scoped to the most recent
         # authoring failure; reset on every transition so a later
         # non-authoring blocker (grade_gate, seed_saver, run_starter)
--- a/src/ouroboros/mcp/tools/auto_handler.py
+++ b/src/ouroboros/mcp/tools/auto_handler.py
@@ -1219,8 +1219,6 @@ def _result_meta(result: AutoPipelineResult) -> dict[str, Any]:
         meta["stop_reason_code"] = result.stop_reason_code
     if result.interview_closure_mode is not None:
         meta["interview_closure_mode"] = result.interview_closure_mode
-    if result.active_task_class is not None:
-        meta["active_task_class"] = result.active_task_class
     if result.execution_job_status:
         meta["execution_job_status"] = result.execution_job_status
     if result.execution_job_error:
--- a/src/ouroboros/mcp/tools/auto_handler.py
+++ b/src/ouroboros/mcp/tools/auto_handler.py
@@ -1217,6 +1217,10 @@ def _result_meta(result: AutoPipelineResult) -> dict[str, Any]:
         meta["product_status"] = "not_verified_complete"
     if result.stop_reason_code is not None:
         meta["stop_reason_code"] = result.stop_reason_code
+    if result.interview_closure_mode is not None:
+        meta["interview_closure_mode"] = result.interview_closure_mode
+    if result.active_task_class is not None:
+        meta["active_task_class"] = result.active_task_class
     if result.execution_job_status:
         meta["execution_job_status"] = result.execution_job_status
     if result.execution_job_error:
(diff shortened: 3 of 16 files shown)


src/ouroboros/orchestrator/runtime_evidence.py
```diff
@@ -237,7 +247,7 @@ def run(
     ) -> RuntimeEvidence:
         return RuntimeEvidence(
             probe_kind=self.kind,
-            passed=True,
+            passed=False,
```
shaun0927: Reverts the #1181 contract. The class docstring three lines up still says *"returns a probe-PASS ... without failing the grade"*, so on `main` consumers depend on PASS. Flipping to False combined with #1190's `passed=False → BLOCKED` rule will block every PRODUCT_COMPLETE that touches a deferred probe kind. Please keep the PASS semantics here and open a follow-up issue if there's evidence the deferred placeholder should fail-closed.

```suggestion
            passed=True,
```

src/ouroboros/auto/pipeline.py
```diff
@@ -648,15 +648,11 @@ async def run(self, state: AutoPipelineState) -> AutoPipelineResult:
...
-                    # L1-c interview-driver disambiguation hook resolves it.
-                    # ``state.active_task_class`` is set only when a concrete
-                    # class/fallback actually fired, so the envelope does not
-                    # pretend an unresolved ambiguous class was active.
+                    # AC template only for a concrete single match. Ambiguous
+                    # or unmatched ledgers leave the user's AC untouched until
+                    # a disambiguation hook resolves the class.
                     _inference = derive_domain_from_ledger(ledger)
-                    _task_class = _inference.single
+                    _task_class = next(iter(_inference.classes)) if _inference.is_single else None
```
shaun0927: `DomainInference.single` on `main` is documented as *"Return the single matched class, or `fallback` for unmatched"*, and `derive_domain_from_ledger` returns `fallback=LIBRARY` with `reason='unmatched'` precisely so the consumer here applies the safe completion gate. Switching to `next(iter(_inference.classes)) if _inference.is_single else None` drops the unmatched→LIBRARY path that #1177 / #1188 deliberately wired up — and the test edit later in this PR flips `test_envelope_uses_library_fallback_when_unmatched` to assert no fallback, which is a behavioral reversal. Please revert to the property accessor (or, if the fallback should be retired, raise that as its own issue against `DomainInference.single`).

```suggestion
                    _task_class = _inference.single
```

src/ouroboros/auto/state.py
```diff
@@ -598,6 +598,8 @@ def transition(self, next_phase: AutoPhase, message: str, *, error: str | None =
         self.updated_at = now
         self.last_progress_message = message
         self.last_error = error
+        if error is None:
+            self.last_error_code = None
```
shaun0927: #1157 names **#1194** as the canonical landing spot for stop-reason-code clearing, and #1194's hunk uses `if next_phase is not AutoPhase.BLOCKED`. The two conditions diverge on `transition(BLOCKED, msg, error=None)` (this PR clears; #1194 preserves) and on non-BLOCKED → non-BLOCKED with an explicit error string (this PR preserves; #1194 clears). Two open PRs writing the same hunk with different conditions will conflict on merge. Suggest dropping this clearing here and letting #1194 land first (the test additions in `test_stop_reason_code.py` can stay if they pass against #1194's condition).

```suggestion
        self.last_error = error
```

src/ouroboros/plugin/firewall.py
```diff
@@ -571,13 +571,12 @@ def _plugin_runtime_env() -> dict[str, str]:
...
-        env.setdefault(
-            "OUROBOROS_PLUGIN_OUTPUT_DIR",
-            str(workdir / ".ouroboros" / "plugin-artifacts" / manifest.name),
+        env["OUROBOROS_PLUGIN_WORKDIR"] = str(workdir)
+        env["OUROBOROS_PLUGIN_OUTPUT_DIR"] = str(
+            workdir / ".ouroboros" / "plugin-artifacts" / manifest.name
         )
         if plugin_home is not None:
-            env.setdefault("OUROBOROS_PLUGIN_HOME", str(plugin_home))
+            env["OUROBOROS_PLUGIN_HOME"] = str(plugin_home)
```
shaun0927: #1193 just landed these three env vars with `env.setdefault(...)` so downstream entrypoints could pre-seed values (e.g. for an `.omx`-style overlay or test harness). Forcing direct assignment removes that affordance and quietly reverses #1193's API choice. If the goal is to prevent stale override leaks across invocations, that needs its own issue and a documented contract change — it shouldn't ride in as an unannounced revert.

```suggestion
        env.setdefault("OUROBOROS_PLUGIN_WORKDIR", str(workdir))
        env.setdefault(
            "OUROBOROS_PLUGIN_OUTPUT_DIR",
            str(workdir / ".ouroboros" / "plugin-artifacts" / manifest.name),
        )
        if plugin_home is not None:
            env.setdefault("OUROBOROS_PLUGIN_HOME", str(plugin_home))
```

src/ouroboros/mcp/tools/auto_handler.py
```diff
@@ -1217,6 +1217,10 @@ def _result_meta(result: AutoPipelineResult) -> dict[str, Any]:
         meta["product_status"] = "not_verified_complete"
     if result.stop_reason_code is not None:
         meta["stop_reason_code"] = result.stop_reason_code
+    if result.interview_closure_mode is not None:
+        meta["interview_closure_mode"] = result.interview_closure_mode
+    if result.active_task_class is not None:
+        meta["active_task_class"] = result.active_task_class
```
shaun0927: Collides with open PR **#1196**, which #1157 names as the canonical MCP-metadata cleanup for `active_task_class`. #1196's explicit decision is that the field is **always present, null when unavailable**, so clients can tell ambiguous inference apart from a missing protocol field. The conditional `if ... is not None` here silently downgrades to the looser contract. Please drop this two-line block and let #1196 own the surfacing (its insertion is at the unconditional `_result_meta` body section).

```suggestion
    if result.interview_closure_mode is not None:
        meta["interview_closure_mode"] = result.interview_closure_mode
```

shaun0927 (changes requested):
Thanks for the omnibus — most of the included fixes are clean (`_matches_web_service` independence, `defaulted_sections` CLI render, `status --limit > 0`, mechanical-eval evidence linkage, `TimeoutExpired.stdout` bytes handling, workflow lifecycle same-ts restart). I'd like to land it, but five hunks need to come out first because they either reverse a merged design decision in the `ooo auto` SSOT (#1157) or collide with an actively-open sibling PR explicitly named by that SSOT as the canonical landing spot.

### Reversals of merged decisions

**1. `DeferredProbe.passed` flipped True→False (reverses #1181)**
The class docstring on `main` is explicit: *"returns a probe-PASS RuntimeEvidence with outcome='deferred' so the verifier can flag the gap **without failing the grade**."* Combined with the #1190 contract (*"explicit probe passed=False ... downgrade the terminal to BLOCKED"*), flipping to False makes every canonical scenario whose `TaskClassProfile` includes a deferred probe kind block PRODUCT_COMPLETE. That's a global regression of the L3 minimal-substrate behavior, not a safety improvement.

**2. L1-b LIBRARY fallback removed (reverses #1177 + #1188)**
`DomainInference.single` is documented on `main` as *"Return the single matched class, or `fallback` for unmatched"* and `derive_domain_from_ledger` returns `fallback=LIBRARY, reason='unmatched'`. #1188's body explicitly states *"unmatched → apply the safe LIBRARY fallback from L1-b"*. Switching to `next(iter(...)) if is_single else None` silently drops the safest-completion-gate fallback; the test rewrite that asserts `active_task_class is None` for unmatched cements the reversal. If we want to retire LIBRARY-as-fallback that should be an own issue + a `DomainInference.single` change at the source, not a quiet flip at the consumer.

### Collisions with sibling open PRs (#1157 explicitly names them)

**3. `state.transition()` last_error_code clearing collides with #1194**
#1157 names #1194 as the remaining L2 follow-up for stop_reason_code clearing. #1194's condition is `if next_phase is not AutoPhase.BLOCKED` (clears only when leaving BLOCKED); this PR uses `if error is None` (clears on any error-less transition). The semantics diverge on `transition(BLOCKED, msg, error=None)` and on non-BLOCKED→non-BLOCKED with an explicit error string. Two PRs writing the same hunk with different conditions will conflict on merge. Suggest dropping this hunk and letting #1194 land first.

**4. `active_task_class` MCP meta collides with #1196**
#1157 names #1196 as the remaining L1 MCP-metadata cleanup. #1196's whole point is *"Preserve the field as null when inference was ambiguous or unavailable, so clients can distinguish an empty computed value from a missing protocol field."* That requires unconditional `meta["active_task_class"] = result.active_task_class`. The conditional `if result.active_task_class is not None` here silently opts out of that contract.

**5. firewall `setdefault` → direct assignment reverses #1193**
#1193 just landed these three env vars with `env.setdefault(...)` so downstream entrypoints could pre-seed values. Forcing assignment removes that affordance without a follow-up issue documenting the new contract. If "stale override leaks" is the concern, it should be tracked separately — #1193 was a deliberate setdefault choice.

### Suggested path to merge-ready

Apply the five suggestions below (one per hunk), then this PR carries only the legitimate scope:
- `_matches_web_service` independence (good — restores #1177 pattern-independence intent)
- `defaulted_sections` CLI rendering
- `status --limit > 0` validation
- mechanical-eval `evidence_event_ids` lock
- `TimeoutExpired.stdout/stderr` bytes-safe decode
- workflow lifecycle same-timestamp restart ordering + regression test
- MCP meta surface for `interview_closure_mode` + `runtime_probe_evidence` (these aren't covered by #1194/#1196)
- `product_status='not_verified_complete'` downgrade for unfinished Ralph plugin runs — please confirm there's a documented decision for this; if so a one-line comment citing the issue would help future readers.

Suggested merge order to avoid conflicts: #1195 → #1196 → #1194 → #1206 → #1213 → (this PR rebased).

shaun0927 (approved):
Force push 70e2ef7 applies all five revert suggestions from the previous review:

1. `DeferredProbe.passed=True` restored (#1181 contract)
2. `_inference.single` restored (#1177 + #1188 LIBRARY fallback)
3. `state.transition()` last_error_code hunk dropped (let #1194 own it)
4. `firewall.py setdefault` restored (#1193 contract)
5. `auto_handler.py` conditional `active_task_class` dropped (let #1196 own it)

Test envelope assertion and DeferredProbe test reverted to match. CI is green across the 8 checks. The PR now carries only the legitimate scope I called out as keepers in the previous review.

LGTM. Ready to merge after the smaller sibling PRs (#1195 → #1196 → #1194) land — those don't conflict with this branch but landing them first keeps the SSOT history readable.

PR #1189, opened 2026-05-22
feat(auto): wall-clock watchdog integration in AutoPipeline (L2-2)

shaun0927:
## Summary

L2-2 slice of #1172 — auto pipeline consumption of the L2-1 \`Watchdog\` substrate (#1178).

Wires the optional \`Watchdog\` into \`AutoPipeline\`:

- New optional \`watchdog: Watchdog | None = None\` field. Default \`None\` → zero behavior change.
- New private \`_check_watchdog(state, ledger)\` helper runs at \`run()\` entry. Parses \`state.created_at\` → calls watchdog → on a returned \`WatchdogDecision\` calls \`state.mark_blocked(..., error_code=WATCHDOG_STOP_REASON_CODE)\`, saves state, returns BLOCKED \`AutoPipelineResult\` — never running interview / seed / downstream phases.
- The watchdog itself appends the \`runtime.watchdog.cancel\` event; the pipeline only translates the decision into state transitions.

Single-point check at \`run()\` entry is the minimal v1 contract per the #1172 minimal-substrate redesign — catches the high-value case (resumed session with elapsed budget) without spreading boilerplate across every phase boundary.

## What lands

- \`src/ouroboros/auto/pipeline.py\`: \`Watchdog\` import + \`watchdog\` field + \`_check_watchdog\` private helper + \`run()\` entry call.
- \`tests/unit/auto/test_pipeline_watchdog_integration.py\` (new): 4 integration tests — no-watchdog backwards compat, under-budget no-op, over-budget BLOCKED + event + envelope, disabled-watchdog opt-out.

## What is NOT in this PR

- Per-phase watchdog checks — entry-only is enough for v1.
- MCP transport-timeout deprecation.
- AgentProcess cooperative cancellation handshake (#518).
- SKILL.md taxonomy update for the 9th stop_reason_code — bookkeeping follow-up.

## Test plan

- [x] \`uv run pytest tests/unit/auto/test_pipeline_watchdog_integration.py -v\` → 4 passed.
- [x] \`uv run pytest tests/unit/auto -q\` → 909 passed.
- [x] \`uv run ruff check\` on touched files → clean.
- [x] \`uv run ruff format\` on touched files → clean.
- [x] \`uv run mypy src/ouroboros/auto/pipeline.py\` → clean.

Refs #1157, #1172, #1178 (L2-1 substrate), #578 (RFC).

Issue #1172, opened 2026-05-22
Meta SSOT slice: L2 — Runtime watchdog v1 (minimal, lifts #578)


shaun0927:
# Meta SSOT slice: L2 — Runtime watchdog v1 (minimal)

> **Implementation cleanup (2026-05-23 KST).** L2 is partially implemented: #1178 merged `RuntimeControls` + wall-clock `Watchdog`; #1189 merged CLI/MCP/AutoPipeline consumption through `runtime.watchdog.cancel`. Treat old `runtime.watchdog.decision`, directive vocabulary, idle/no-progress timers, and subscriber-framework wording as superseded v2 ideas. Remaining known cleanup: #1194.

This issue is the **L2 implementation slice** of #1157. It is the smallest possible watchdog substrate that satisfies the SSOT invariant *"long-running run never stalls without a recorded reason"*.

## Why this is small on purpose

The earlier draft of L2 proposed a three-timer config (`idle_timeout` / `no_progress_timeout` / `safety_timeout`), a four-directive vocabulary (`WAIT` / `RETRY` / `UNSTUCK` / `CANCEL`), a subscriber pattern for cooperative cancellation, an `activity_events` vs `material_progress_events` event-set distinction, and a deprecation roadmap for ad-hoc timeouts spread across MCP / auto / Ralph layers. **That was 6× the substrate the actual problem requires.**

The actual problem is small: `ooo auto` sessions sometimes hang silently, and operators cannot tell *why*. The minimal substrate that fixes this:

- One timer per session (wall-clock).
- One outcome: when it fires, the session is cancelled and the EventStore records *why*.
- That's it.

Once v1 lands and we have evidence of *which kinds of stalls slip through wall-clock-only* (e.g. busy-loop without actual progress), the more refined timer types and directive vocabulary can be added as their own slices.

## Substrate honesty

L2 v1 introduces **one** new EventStore event family: `runtime.watchdog.cancel`. This is the only new substrate in #1157. Per the *Substrate honesty* note at the top of #1157, every other lane lights up existing substrate; L2 adds exactly this.

The earlier draft also proposed introducing a `runtime_control` aggregate type on the projection. **Confirmed additive** via code reading of `src/ouroboros/persistence/schema.py` — `aggregate_type` is a free-form `VARCHAR(100)`, no enum constraint, no Python allowlist. See the informational comment posted on #946.

### One config knob

```yaml
runtime_controls:
[...]

PR #1207, opened 2026-05-24
fix: make watchdog controls replay safe

Q00:
## Summary
- suppress duplicate runtime.watchdog.cancel events after process restart by consulting EventStore replay when available
- allow production CLI/MCP composition using load_runtime_controls(None) to honor OUROBOROS_SESSION_WALL_CLOCK_SECONDS
- keep load_runtime_controls compatible with existing runtime_controls config keys owned by the main config loader

## Original PR blocker coverage
Addresses watchdog/runtime-control blockers reported on #1178 and #1189.

## Verification
- uv run pytest tests/unit/runtime/test_controls.py tests/unit/runtime/test_watchdog.py tests/unit/auto/test_pipeline_watchdog_integration.py -q
- uv run ruff check src/ouroboros/runtime/controls.py src/ouroboros/runtime/watchdog.py tests/unit/runtime/test_controls.py tests/unit/runtime/test_watchdog.py

### DIFF
--- a/src/ouroboros/runtime/controls.py
+++ b/src/ouroboros/runtime/controls.py
@@ -14,6 +14,7 @@ deferral rationale.
 from __future__ import annotations
 
 from dataclasses import dataclass
+import os
 from pathlib import Path
 from typing import Any
 
@@ -75,14 +76,23 @@ def load_runtime_controls(path: Path | str | None = None) -> RuntimeControls:
 
     Behaviour:
 
-    - ``path is None`` → return ``RuntimeControls()`` with defaults.
+    - ``path is None`` → return defaults, optionally overridden by
+      ``OUROBOROS_SESSION_WALL_CLOCK_SECONDS``.
     - ``path`` exists but no ``runtime_controls`` key → defaults.
     - YAML parse error → ``ValueError`` so callers fail loudly.
     - Unrecognized keys under ``runtime_controls`` → ``ValueError``;
       the v2 expansion path adds new keys only through code review.
     """
     if path is None:
-        return RuntimeControls()
+        env_budget = os.environ.get("OUROBOROS_SESSION_WALL_CLOCK_SECONDS")
+        if env_budget is None:
+            return RuntimeControls()
+        try:
+            budget = int(env_budget)
+        except ValueError as exc:
+            msg = "OUROBOROS_SESSION_WALL_CLOCK_SECONDS must be an integer"
+            raise ValueError(msg) from exc
+        return RuntimeControls(session_wall_clock_seconds=budget)
     resolved = Path(path)
     if not resolved.is_file():
         msg = f"runtime_controls config not found at {resolved}"
@@ -104,7 +114,17 @@ def load_runtime_controls(path: Path | str | None = None) -> RuntimeControls:
     if not isinstance(block, dict):
         msg = f"runtime_controls block at {resolved} must be a mapping; got {type(block).__name__}"
         raise ValueError(msg)
-    allowed = {"session_wall_clock_seconds"}
+    # Keep compatibility with the existing runtime_controls config block.
+    # This loader consumes only the Auto session wall-clock knob and ignores
+    # the older generation/MCP controls owned by ouroboros.config.loader.
+    allowed = {
+        "session_wall_clock_seconds",
+        "mcp_tool_timeout_seconds",
+        "generation_idle_timeout_seconds",
+        "generation_no_progress_timeout_seconds",
+        "generation_safety_timeout_seconds",
+        "watchdog_poll_seconds",
+    }
     unknown = set(block.keys()) - allowed
     if unknown:
         msg = (
--- a/src/ouroboros/runtime/watchdog.py
+++ b/src/ouroboros/runtime/watchdog.py
@@ -36,7 +36,7 @@ from __future__ import annotations
 from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import UTC, datetime
-from typing import Protocol
+from typing import Protocol, runtime_checkable
 
 from ouroboros.events.base import BaseEvent
 from ouroboros.runtime.controls import RuntimeControls
@@ -64,6 +64,7 @@ WATCHDOG_STOP_REASON_CODE: str = "watchdog_wall_clock_exceeded"
 cancel event is observed."""
 
 
+@runtime_checkable
 class _EventAppender(Protocol):
     """Minimal protocol the watchdog needs from the EventStore.
 
@@ -73,6 +74,19 @@ class _EventAppender(Protocol):
     """
 
     async def append(self, event: BaseEvent) -> None: ...
+
+
+@runtime_checkable
+class _WatchdogEventReader(Protocol):
+    """Optional EventStore replay hook used for restart idempotency."""
+
+    async def query_events(
+        self,
+        aggregate_id: str | None = None,
+        event_type: str | None = None,
+        limit: int = 50,
+        offset: int = 0,
+    ) -> list[BaseEvent]: ...
 
 
 @dataclass(frozen=True, slots=True)
@@ -180,6 +194,15 @@ class Watchdog:
         budget = self._controls.session_wall_clock_seconds
         if elapsed_seconds <= budget:
             return None
+        if isinstance(self._appender, _WatchdogEventReader):
+            existing = await self._appender.query_events(
+                aggregate_id=session_id,
+                event_type=WATCHDOG_CANCEL_EVENT_TYPE,
+                limit=1,
+            )
+            if existing:
+                self._fired_sessions.add(session_id)
+                return None
         decision = WatchdogDecision(
             session_id=session_id,
             fired_at=now,
--- a/src/ouroboros/runtime/watchdog.py
+++ b/src/ouroboros/runtime/watchdog.py
@@ -202,7 +202,20 @@ class Watchdog:
             )
             if existing:
                 self._fired_sessions.add(session_id)
-                return None
+                event = existing[0]
+                fired_at = _coerce_datetime(event.data.get("fired_at")) or now
+                return WatchdogDecision(
+                    session_id=session_id,
+                    fired_at=fired_at,
+                    elapsed_seconds=_coerce_int(
+                        event.data.get("elapsed_seconds"),
+                        default=elapsed_seconds,
+                    ),
+                    configured_budget_seconds=_coerce_int(
+                        event.data.get("configured_budget_seconds"),
+                        default=budget,
+                    ),
+                )
         decision = WatchdogDecision(
             session_id=session_id,
             fired_at=now,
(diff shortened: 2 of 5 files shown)


shaun0927:
## Merge-readiness rationale (English)

This PR is ready to merge. It is the watchdog/runtime-controls hardening blocker called out on #1178 and #1189, and it took two REQUEST_CHANGES iterations to land the right semantics.

### What it does
Three coordinated changes inside the Track B L2 watchdog/runtime-controls surface:

1. **EventStore is the durable source of truth for terminal cancellation.** `Watchdog.check()` now consults `EventStore.query_events()` for an existing `runtime.watchdog.cancel` for the session *before* evaluating any current control knob. If a prior process already persisted a cancel event, the resumed pipeline reconstructs the original `WatchdogDecision` from the persisted payload and blocks — without appending a duplicate event.
2. **`load_runtime_controls(None)` honours `OUROBOROS_SESSION_WALL_CLOCK_SECONDS`.** This unblocks production CLI/MCP composition paths that don't pass a controls path, while keeping `int()` parsing strict (non-integer values raise `ValueError`).
3. **Compatibility with the existing `runtime_controls` config block.** The loader now accepts (and ignores) the older keys owned by `ouroboros.config.loader` (`mcp_tool_timeout_seconds`, `generation_*_timeout_seconds`, `watchdog_poll_seconds`) instead of rejecting them, so callers don't have to maintain two parallel YAML schemas.

A pair of `runtime_checkable` Protocols (`_EventAppender`, `_WatchdogEventReader`) keeps the dependency on the EventStore minimal and testable — the watchdog only requires the replay reader when the appender happens to expose `query_events`.

### Why it aligns with the SSOT direction
- Issue #961 Track B L2 (#1172 watchdog v1) records #1178/#1189 as merged L2 follow-ups, and SSOT explicitly lists #1207 as the "approved/clean/green" follow-up that closes the post-merge blockers on those slices.
- The change keeps the documented boundary: `Watchdog.check()` is the durable-cancellation source, and `AutoPipeline._check_watchdog()` is the state-transition consumer. The non-atomic gap between "append cancel event" and "save BLOCKED state" is now handled deterministically on resume.
- It does not change the watchdog's fire-once semantics, the cancel event schema, or the `BLOCKED` state contract.

### Why it is not over-engineered
- 5 files changed, +291/-12. Of those 291 lines, 161 are tests across three modules covering (a) env-override parsing, (b) duplicate-suppression, (c) same-budget replay, (d) restart with budget raised to 10000, and (e) restart with watchdog disabled (`budget=0`).
- Two new small helpers (`_coerce_datetime`, `_coerce_int`) reconstruct the persisted `WatchdogDecision` defensively without trusting the on-disk payload type — these are eight lines each and have explicit happy-path / fallback tests.
- The two `runtime_checkable` Protocols replace any need for adapter classes; the existing `EventStore` already implements both shapes.
- No new public API surface; no new env vars beyond the one stated; no new config keys.

### Why the bot's review trajectory matters
The first review caught the right bug: an over-budget process could append `runtime.watchdog.cancel` and then crash *before* the pipeline saved `BLOCKED`, so replaying with the current code would silently proceed. The fix moved replay into the "before any current-knob gate" path. The second review caught a tighter version of the same bug: even with replay in place, an operator who restarted with the budget *raised* (or the watchdog disabled) would bypass replay because the current-knob gate ran first. The fix at `src/ouroboros/runtime/watchdog.py:193` consults the persisted cancel event before `watchdog_enabled` or `elapsed_seconds <= budget`. Both regressions are now pinned with explicit tests.

### Why it is mergeable
- `reviewDecision: APPROVED` on commit `807bdd9`.
- Latest `ouroboros-agent` verdict: **APPROVE** — both prior blockers (replay too late; replay bypassed by raised/disabled budget) are explicitly withdrawn against current HEAD.
- The only non-blocking remark is a discoverability suggestion to document the env var in the CLI/config reference. That's a docs-only follow-up, not a merge gate.
- All 8 required checks green (Ruff, MyPy, Bridge TS, enforce-envelope, enforce-boundary, Tests 3.12/3.13/3.14).
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### Risk assessment
- **Boundary**: durable cancel-event ↔ in-memory `_fired_sessions` set ↔ in-process control knobs. Current HEAD orders them correctly: persisted event wins, then in-memory guard, then current knobs.
- **Replay**: explicitly tested. Existing sessions with prior cancel events transition to `BLOCKED` on resume regardless of the resumed process's budget configuration.
- **Wire format**: cancel event payload schema is unchanged; replay reads existing fields with defaults.
- **Backwards compatibility**: appenders that do not implement `query_events` (the `runtime_checkable` Protocol check fails) skip the replay branch entirely and behave exactly as before. No test infrastructure has to change.

Recommending merge.

shaun0927:
# PR Review Summary

_Posted via `/oh-my-claudecode:pr-review` skill — evidence-backed verdict._

## Verdict
**Approve**

## Scope Reviewed
- PR intent: make the auto-session watchdog restart-safe by treating the persisted `runtime.watchdog.cancel` event as the durable source of truth for terminal cancellation, regardless of how the resumed process's runtime-control knobs are configured; expose `OUROBOROS_SESSION_WALL_CLOCK_SECONDS` for `load_runtime_controls(None)`; keep the runtime-controls YAML loader compatible with older keys owned by `ouroboros.config.loader`.
- Main changed areas:
  - `src/ouroboros/runtime/controls.py` (+23/-3) — env override parsing + key-allow-list expansion.
  - `src/ouroboros/runtime/watchdog.py` (+78/-3) — `_WatchdogEventReader` Protocol, `_replay_existing_decision()` helper, and reordered `check()` so replay runs before any current-knob gate.
  - `tests/unit/runtime/test_controls.py` (+29/-6), `tests/unit/runtime/test_watchdog.py` (+82), `tests/unit/auto/test_pipeline_watchdog_integration.py` (+79).
- Tests reviewed: env override valid/invalid parsing; controls loader compatibility for additional keys; watchdog duplicate suppression; same-budget replay; replay with budget raised to 10000; replay with `OUROBOROS_SESSION_WALL_CLOCK_SECONDS=0`; pipeline integration covering both watchdog-disabled and raised-budget restart paths.
- Checks considered: Ruff Lint, MyPy, Bridge TypeScript, enforce-envelope, enforce-boundary, Tests Python 3.12 / 3.13 / 3.14, `ouroboros-agent` verdict on `807bdd9`.

## Blocking Issues
None.

## Warnings
None.

## Mutation-Test Thinking
- **Likely mutants the tests would kill**:
  - Moving the `_replay_existing_decision(...)` call *after* the `watchdog_enabled` check or after the `elapsed_seconds <= budget` check — would re-introduce both prior blockers and be killed by the disabled-budget and raised-budget restart tests.
  - Removing the `isinstance(self._appender, _WatchdogEventReader)` guard — would `AttributeError` for legacy in-memory appenders that don't implement `query_events`; the existing-appender integration test would fail.
  - Returning `None` from `_replay_existing_decision()` when `existing` is non-empty — same restart tests fail.
  - Forgetting to add the session to `_fired_sessions` after replay — duplicate-suppression tests would fail because a second `check()` call in the same process would re-emit a cancel event.
  - Strict-typing `event.data.get("fired_at")` so a non-ISO string blows up — `_coerce_datetime`'s fallback test covers this.
  - Changing the env-var name or removing the integer cast — `test_load_runtime_controls_applies_env_session_budget` and the `ValueError` test kill these.
  - Removing one of the new `allowed` keys from `controls.py` — the compatibility test that uses an existing config block fails.
- **Mutants the tests may not catch**:
  - A regression where `query_events(limit=1)` is changed to a larger limit but only the latest event is read — semantically identical, low concern.
  - A persistent cancel event whose `event.data` is `None` or missing — `event.data.get(...)` on a `None` would raise, but the `BaseEvent` contract guarantees a dict; the helpers default-coerce on string/int parse failures. Acceptable.
- **Additional tests recommended**: none for merge gating. Optional: a test where `event.data["fired_at"]` is a malformed ISO string to pin `_coerce_datetime`'s `None` branch explicitly (the production code already falls back to `fired_at_default`, but the explicit assertion would be cheap).

- **High-risk functions/modules**: `Watchdog.check()` gained one helper call and reorders the guard sequence. `_replay_existing_decision()` is straight-line: guard on `isinstance`, await query, coerce-and-return. No nested loops, no shared mutable state beyond `_fired_sessions` (already in-class).
- **Complexity increase**: bounded. The reorder is the most semantically loaded part of the change; it is two lines of moved code and is the *whole point* of the fix.
- **Test coverage concern**: branch coverage on the new replay path is comprehensive — the three "should-block" replay shapes (same budget, raised budget, disabled budget) are all asserted at both the watchdog unit boundary and the pipeline integration boundary.
- **Refactoring recommendation**: none. `_coerce_datetime` and `_coerce_int` are intentionally tiny defensive helpers; collapsing them would obscure their fallback semantics.

## Test Quality Assessment
- **Strong tests**: the pipeline-level integration tests (`test_pipeline_watchdog_integration.py:+79`) assert end-to-end that a session restarts into `BLOCKED` without appending a duplicate `runtime.watchdog.cancel` event. This is the contract the original bug violated, and it is now pinned.
- **Weak tests**: none observed.
- **Missing edge cases**: a multi-session appender where session A is cancelled and session B is over budget in the same process — not directly asserted, but the `_fired_sessions` set is keyed per session_id and the replay query filters by `aggregate_id=session_id`, so this is structurally safe.
- **Mocking concerns**: tests use small in-test appender stubs (`append` + `query_events`); they match the production EventStore's signature exactly. The bot's review confirmed the replay protocol matches `EventStore.query_events()`.

- **Auth/authz**: untouched.
- **Data**: read-only against the EventStore for replay; no schema migration.
- **Persistence**: cancel event payload schema unchanged. Replay reads `fired_at`, `elapsed_seconds`, `configured_budget_seconds` defensively with fallbacks.
- **Replay**: now correct across restart, env drift, and partial-failure. This is the central improvement.
- **Failure mode**: if the EventStore itself fails on `query_events()`, the exception propagates and the watchdog does not silently bypass — fail-closed behavior is preserved at the pipeline level.
- **Observability**: cancel events stay singular per session; operators do not see spurious duplicates on resume.

## Looks Good
- The fix correctly identifies the EventStore-vs-in-process-state split as the root cause and chooses durable persistence as the source of truth, which is the right boundary.
- The `runtime_checkable` Protocol pair lets the watchdog cooperate with appenders that don't expose `query_events` (production has it; some tests don't), without runtime type-check explosions.
- Test coverage spans unit *and* integration boundaries; both prior bot blockers have explicit regression tests.
- Env-var integer parsing is strict (raises on non-int), so silent misconfiguration is impossible.
- The YAML-key compatibility list documents itself in a comment about the dual-loader boundary.
- All CI green; clean against main.

## Final Recommendation
APPROVE. The PR closes both prior REQUEST_CHANGES blockers with explicit regression tests, and the resulting semantics (persisted cancel event is durable, env override is strict, dual loaders coexist) match the SSOT Track B L2 watchdog contract. No blocking findings, no warnings.

Review-Metadata:
verdict: APPROVE
skill: oh-my-claudecode:pr-review
head_sha: 807bdd96057a8878fb85b78301a9e5913cc3b0da

PR #1190, opened 2026-05-22
feat(auto): runtime-probe envelope + advisory probe_runner (L3-2)

shaun0927:
## Summary

L3-2 slice of #1176 — surfaces runtime acceptance evidence (L3-1 substrate, #1181) onto `AutoPipelineResult` via an optional `probe_runner` callback on `AutoPipeline`, and treats configured runtime-probe failures as PRODUCT_COMPLETE blockers.

Adds the minimal envelope + completion-gate plumbing:

- `AutoPipelineResult.runtime_probe_evidence: tuple[RuntimeEvidence, ...] = ()` — frozen evidence tuple; empty when no runner is wired (default/backwards compatible).
- `AutoPipeline.probe_runner` optional async callback. The caller decides which probes to invoke (often via `probes_for_task_class` with the active task class from L1-d / future binding work).
- `AutoPipeline._complete_runtime_probe_blocker` runs the callback once per `run()` before COMPLETE is returned on PRODUCT_COMPLETE paths, caches evidence, and surfaces it via every `_result()` call in the same run.
- Runtime evidence is no longer advisory-only when a runner is configured: explicit probe `passed=False` and runner infrastructure exceptions downgrade the terminal to `BLOCKED` instead of allowing a false PRODUCT_COMPLETE.
- Both Ralph-success direct completion and EVALUATE-pass completion paths invoke the same runtime-probe completion gate.

This remains minimal-substrate: no new EventStore event family, projection, CLI flag, probe kind, or command-source policy is introduced here.

## What lands

- `src/ouroboros/auto/pipeline.py`: `runtime_probe_evidence` field + typed `probe_runner` callback + completion-gate helper + cache + `_result()` plumbing.
- `tests/unit/auto/test_pipeline_runtime_probe_envelope.py`: integration tests for empty-by-default, successful evidence surface, runner exception blocking, explicit probe failure blocking, and EVALUATE-pass completion coverage.

## What is NOT in this PR

- New probe kinds (`sim_trace`, `render_hash`, `api_smoke`) or real probe implementations beyond the L3-1 `headless_run` substrate.
- EventStore/projection/audit-event persistence for runtime probes.
- Automatic command-source selection / task-class binding inside `AutoPipeline`; this slice consumes a wired runner rather than owning binding policy.
- L0 canonical scenario fixture consumption of runtime evidence.

## Test plan

- [x] `uv run pytest tests/unit/auto/test_pipeline_runtime_probe_envelope.py -q` → 5 passed.
[...]

PR #1213, opened 2026-05-24
fix: persist runtime probe evidence

Q00:
## Summary
- fixes #1190 runtime probe durability by persisting runtime_probe_evidence into AutoPipelineState
- preserves persisted probe evidence when completed sessions are replayed/resumed
- adds an opt-in public CLI/MCP runtime probe runner via OUROBOROS_RUNTIME_PROBE_COMMAND for PRODUCT_COMPLETE validation

## Verification
- uv run ruff check src/ouroboros/auto/pipeline.py src/ouroboros/auto/state.py src/ouroboros/auto/adapters.py src/ouroboros/orchestrator/runtime_evidence.py src/ouroboros/cli/commands/auto.py src/ouroboros/mcp/tools/auto_handler.py tests/unit/auto/test_pipeline_runtime_probe_envelope.py
- uv run pytest tests/unit/auto/test_pipeline_runtime_probe_envelope.py tests/unit/orchestrator/test_runtime_evidence.py -q

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -6,6 +6,7 @@ import asyncio
 from collections.abc import Callable
 from dataclasses import dataclass
 import hashlib
+import os
 from pathlib import Path
 from typing import Any
 
@@ -26,6 +27,7 @@ from ouroboros.mcp.tools.qa import QAHandler
 from ouroboros.mcp.tools.ralph_handlers import RalphHandler
 from ouroboros.mcp.tools.subagent import should_dispatch_via_plugin
 from ouroboros.mcp.types import MCPToolResult
+from ouroboros.orchestrator.runtime_evidence import HeadlessRunProbe, RuntimeEvidence
 from ouroboros.resilience.lateral import ThinkingPersona
 
 _SAFE_SEED_ID_FILENAME_CHARS = frozenset(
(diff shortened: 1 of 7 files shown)


shaun0927 (approved):
Cross-checked against #1190 (the L3-2 envelope + `probe_runner` contract) and #1181 (`RuntimeEvidence` substrate). Three things I verified:

1. **Persistence scope is correct** — `state.runtime_probe_evidence` is per-session, hydrated at the start of `run()` and replaces the explicit `self._last_probe_evidence = ()` clear. This is *not* a "leak across AutoPipeline instances" regression because a fresh `AutoPipelineState` still starts with `runtime_probe_evidence=[]`; the persistence only kicks in for resume/re-attach of the same logical session, which is exactly the gap #1190 acknowledged ("caches evidence... in the same run").
2. **`EnvRuntimeProbeRunner` is strictly opt-in** — returns `()` when `OUROBOROS_RUNTIME_PROBE_COMMAND` is unset, so wiring it whenever `complete_product=True` is behaviorally identical to the previous `probe_runner=None` default for any caller that doesn't set the env var. Aligns with #1190's "the caller decides".
3. **`RuntimeEvidence.to_dict/from_dict` round-trip is safe** — explicit type validation for each field, `_json_safe_payload` handles non-JSON-safe values via `default=str` fallback, and `_validate_loaded` rejects malformed entries. `from_dict` defaulting `payload.setdefault("runtime_probe_evidence", [])` covers older state files cleanly.

No conflicts with the other open auto PRs since this touches `auto/{pipeline,state,adapters}.py`, `mcp/tools/auto_handler.py`, and `orchestrator/runtime_evidence.py` in additive places — verified against #1196 / #1194 line ranges. Safe to merge.

PR #1345, opened 2026-06-07
fix(installer): prefer freshly installed ouroboros for setup

Q00:
## Summary
- Prefer the freshly installed uv tool-bin `ouroboros` before any stale executable already on PATH.
- Invoke the remembered fresh uv executable path directly for setup, so later fallback PATH probing cannot reorder it behind stale legacy bins.
- Preserve pipx/pip behavior by only scanning fallback home bins when no `ouroboros` command is visible.
- Add regression tests for stale PATH commands, stale `$HOME/.local/bin/ouroboros` under uv, and stale `$HOME/.local/bin/ouroboros` under pipx.

## Why
While doing the release smoke after `v0.40.1`, an isolated Pi install completed with exit 0 but ran setup through an older global `ouroboros`, producing `Unsupported runtime: pi`. That can make installer output look successful while runtime-specific setup was skipped or misconfigured.

## Review Response
- Addressed first requested change: `$HOME/.local/bin` fallback probing can no longer overtake the fresh uv tool bin, because step 4 executes the remembered fresh uv executable path directly.
- Addressed second requested change: the home-bin fallback is no longer unconditional for pipx/pip. It only runs when no `ouroboros` command is already visible, preserving the existing PATH command for non-uv install paths.
- Added focused regressions for both stale-home-bin cases.

## Verification
- `bash -n scripts/install.sh`
- `uv run --python 3.12 ruff check .`
- `uv run --python 3.12 python -m pytest tests/unit/scripts/test_install_runtime_selection.py tests/unit/cli/test_setup.py -q` -> 165 passed
- Isolated uv installer smoke with stale `$HOME/.local/bin/ouroboros`, custom `HOME`, `UV_TOOL_DIR`, `UV_TOOL_BIN_DIR`, and `OUROBOROS_INSTALL_RUNTIME=pi`: setup runs the fresh `UV_TOOL_BIN_DIR/ouroboros`, reaches `Configured Pi runtime`, and writes `.pi/agent/extensions/ouroboros-ooo-bridge.ts`.
- Isolated pipx installer smoke with fresh PATH `ouroboros` and stale `$HOME/.local/bin/ouroboros`: setup runs the PATH command and does not execute the stale home-bin command.

## Note
The real Pi native response smoke still fails in this local environment because the configured OpenAI key is rejected with 401; that is credential/auth state, not this installer path bug.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -488,21 +488,44 @@ else
   fi
 fi
 
-# Ensure ouroboros binary is in PATH (uv tool install may add to ~/.local/bin)
-if ! command -v ouroboros &>/dev/null; then
-  for p in "$HOME/.local/bin" "$HOME/.cargo/bin" "$HOME/bin"; do
-    if [ -x "$p/ouroboros" ]; then
-      export PATH="$p:$PATH"
-      break
+# Ensure setup runs the freshly-installed uv tool binary rather than a stale
+# command already on PATH. uv can install into UV_TOOL_BIN_DIR or its configured
+# tool bin directory without updating the current shell's PATH, so remember the
+# fresh executable and invoke that exact path below. For pipx/pip installs,
+# preserve the existing PATH command unless no ouroboros command is visible.
+OUROBOROS_BIN=""
+_prepend_path_if_ouroboros() {
+  local candidate="$1"
+  if [ -n "$candidate" ] && [ -x "$candidate/ouroboros" ]; then
+    export PATH="$candidate:$PATH"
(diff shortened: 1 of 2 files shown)


Q00:
Ready for re-review. Addressed the latest blocker by scoping fallback home-bin probing so pipx/pip keep the existing PATH command unless no ouroboros command is visible. Added regression coverage for the exact stale $HOME/.local/bin + pipx case, and all CI checks are passing on head 909538d3.

PR #1359, opened 2026-06-07
fix(mcp): stop orphaned servers and reclaim WAL; speed up session list

Q00:
## Problem

The TUI/MCP became very slow and the on-disk store ballooned — on one machine **13GB `ouroboros.db` + 17GB `-wal`**, of which only ~53MB was live data. Three compounding causes:

1. **`get_all_sessions()` took ~33s.** SQLite's planner satisfied `ORDER BY timestamp` by scanning all ~850k event rows via `ix_events_timestamp` and filtering inline, instead of using the selective `ix_events_event_type` (session events are a tiny subset). This is the first query the TUI session list runs, so the UI hung for ~33s on open.
2. **The `-wal` file grew unbounded.** With many concurrent MCP/TUI connections a reader was always pinning the WAL, so passive autocheckpoints could never truncate it, and nothing ran an explicit `TRUNCATE` checkpoint.
3. **`ouroboros mcp serve` had no shutdown hooks.** An unhandled `SIGTERM` bypassed the cleanup `finally`; a dead client left the server (especially `streamable-http`, which has no stdin EOF) orphaned and holding the DB open → process + connection accumulation.

## Changes

**`persistence/event_store.py`**
- `get_all_sessions()`: pin `INDEXED BY ix_events_event_type` via raw `text()` + `.columns(*events_table.c)` (to keep ORM result processing / JSON payload deserialization). SQLAlchemy/SQLite ignores `with_hint()`. **33.7s → 1.3s**, identical 453-row result.
- `close()`: explicit `wal_checkpoint(TRUNCATE)`, bounded by a short 2s `busy_timeout` connection so a peer write-lock cannot stall shutdown for the 30s default.

**`cli/commands/mcp.py`** (`_run_mcp_server`)
- Graceful `SIGTERM`/`SIGINT`/`SIGHUP` handlers → the cleanup `finally` (WAL collapse + PID cleanup) always runs.
- Parent-death watchdog: self-terminates when reparented away from its original ppid (covers child-subreaper hosts like `systemd --user`, not just pid 1).
- Re-raises a genuine serve-loop crash after cleanup instead of exiting 0 silently.

## Adversarial review

This change went through a multi-agent adversarial review (4 dimensions × verify), which surfaced and fixed 4 issues in the first draft:
- `signal.SIGHUP` crashed `mcp serve` on Windows at startup (tuple built before the `suppress`) → resolved via `getattr`.
- A serve-loop crash was swallowed by the `finally` `suppress(Exception)` (silent exit 0, a regression) → the serve task's exception is now re-raised.
- `close()` WAL `TRUNCATE` could block shutdown up to 30s under write-lock contention → short `busy_timeout`.
- Orphan watchdog `getppid()==1` never fired under `systemd --user` (subreaper) → compare `!= orig_ppid`.

## Verification
- `ruff` / `ruff format` / `mypy` clean
- 325 targeted tests pass
- Integration tested on a live server: SIGTERM graceful shutdown, orphan self-termination (parent SIGKILL → reparent → exit), and WAL truncate-on-close (4MB → 0)

### DIFF
--- a/src/ouroboros/cli/commands/mcp.py
+++ b/src/ouroboros/cli/commands/mcp.py
@@ -6,10 +6,12 @@ Start and manage the MCP (Model Context Protocol) server.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 from enum import Enum
 import json
 import os
 from pathlib import Path
+import signal
 import subprocess
 import sys
 from typing import Annotated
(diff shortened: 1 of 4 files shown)


Q00:
Addressed the review in `497fff05`:

**🚫 Blocker — teardown bypassed the adapter's owned-resource contract**
`_run_mcp_server` now routes teardown through `await server.shutdown()` instead of closing the stores directly. This closes owned resources in the documented order — **ControlBus (drain subscriber tasks) → EventStore (WAL collapse) → BrownfieldStore → mcp_bridge** — so control-bus tasks are drained and the bridge is closed before the EventStore is disposed. Re-verified on a live server: `mcp.server.shutdown` (adapter.py) fires on **both** the SIGTERM and orphan-exit paths.

**🟡 Follow-up — `INDEXED BY` brittleness**
`get_all_sessions()` now catches `OperationalError` and falls back to the planner's choice (plain `select`) on a fresh transaction, so a store created outside the bundled schema/migrations (missing `ix_events_event_type`) degrades gracefully instead of erroring. Documented inline.

**🟡 Test coverage**
Added:
- `test_falls_back_when_forced_index_missing` — drops the index and asserts correct, ordered results.
- `TestEventStoreClose` — WAL TRUNCATE checkpoint collapses `-wal`, `close()` idempotency, read-only stores skip the write checkpoint.
- `test_shutdown_routes_through_server_shutdown` — teardown awaits `server.shutdown()`.

**Verification:** ruff/format/mypy clean; 1031 targeted tests pass; live SIGTERM + orphan shutdown re-confirmed.

PR #1372, opened 2026-06-07
feat(orchestrator): pace non-Claude delivery within a configurable rate budget

deepakdgupta1:
## Summary

Third and final remediation from the June-2026 RCA (after #1360 P0 and #1361 P1, both now merged). It extends Ouroboros-side rate governance — which previously existed **only** on the native Claude adapter — to every runtime, so that raising fan-out concurrency on a CLI backend (hermes, codex, gemini, opencode, …) is *safe* rather than a stampede risk.

This PR is rebased on current `main` and contains a single commit.

## What changed

- **Reusable `RateLimitGate`** (`rate_limit.py`) wraps the existing acquire/heartbeat/force-reserve loop around `SharedRateLimitBucket`. **Dormant** when no limits are set — wiring it onto a path with no configured budget is a no-op.
- **Executor pacing** (`parallel_executor.py`): one shared gate per executor, consulted before both `execute_task` dispatch sites. Because all concurrent workers share the executor's single gate instance, it coordinates the in-level fan-out — exactly the R3 stampede (14 ACs at once).
- **No double-limiting Claude**: `ClaudeAgentAdapter` advertises `self_governs_rate_limit = True` (it already runs its own bucket), so the executor gate stays dormant for it.
- **Limits are fully data-driven** (`backend_limits.py`) — *every* concurrency/RPM/TPM value is now configurable without source edits, resolved highest-precedence-first:
  1. env (`OUROBOROS_<BACKEND>_RPM` / `_TPM`, `OUROBOROS_MAX_CONCURRENCY`)
  2. `~/.ouroboros/backend_limits.yaml` (mtime-cached, fault-tolerant — mirrors the existing `tool_capabilities.yaml` loader)
  3. built-in registry, then the serialize-by-default cap of 1

## Behavior change

**None by default.** With no env/config set, CLI runtimes behave exactly as after #1361 (serialized to 1, no pacing). Declaring an RPM/TPM is the opt-in that activates pacing and makes lifting `OUROBOROS_MAX_CONCURRENCY` safe.

## Design note / scope boundary

The gate lives on the `ParallelACExecutor` instance, so it coordinates that executor's parallel workers (the documented in-level stampede). It does **not** coordinate across separate executor instances for the same backend — a run-/adapter-scoped shared bucket would be a larger change and is intentionally out of scope here. Claude's per-adapter bucket is left untouched; consolidating it onto `RateLimitGate` is a possible low-risk follow-up.

## Test plan

- [x] `backend_limits` — config-file loader (missing/malformed/non-regular → ignored; mtime cache; path override), precedence (env > file > registry > default), per-backend RPM/TPM env, canonicalization, Claude ceilings + default cap overridable via file
- [x] `rate_limit` — `RateLimitGate` dormant when unconfigured; paces when budget set; force-reserves after max wait; backoff callbacks fire
- [x] `parallel_executor` — gate dormant for self-governing (Claude) adapter; dormant for CLI backend without config; activates with configured RPM; dispatch hook is a no-op when dormant and paces through an active gate
- [x] Full suite: **10,524 passed, 5 skipped**; `ruff` + `mypy` (v2.1.0) clean
- [x] Docs: new "Backend concurrency & rate limits" section + env-var reference rows

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -600,6 +600,9 @@ All environment variables have higher priority than the corresponding `config.ya
 | `OUROBOROS_OPENCODE_PERMISSION_MODE` | `orchestrator.opencode_permission_mode` | Permission mode when using OpenCode runtime. |
 | `OUROBOROS_MAX_PARALLEL_WORKERS` | `orchestrator.max_parallel_workers` | Requested maximum concurrent Acceptance Criteria workers for parallel execution. Must be a positive integer. |
 | `OUROBOROS_MAX_CONCURRENCY` | _(fan-out cap)_ | Caps the effective delivery fan-out for the connected backend, overriding Ouroboros' backend-aware default. Use it to raise CLI runtimes (`hermes`, `codex`, ...) above their serialized-by-default ceiling of 1. Must be a positive integer; blank/invalid values are ignored so the safety cap is never silently disabled. |
+| `OUROBOROS_<BACKEND>_RPM` | _(rate budget)_ | Per-backend requests-per-minute ceiling for the shared dispatch rate bucket. `<BACKEND>` is the runtime backend upper-cased with non-alphanumerics collapsed to `_` (e.g. `OUROBOROS_HERMES_CLI_RPM`, `OUROBOROS_OPENCODE_RPM`). Dormant unless set — declaring it is how you safely raise `OUROBOROS_MAX_CONCURRENCY` on a CLI runtime. Must be a positive integer; blank/invalid values are ignored. |
+| `OUROBOROS_<BACKEND>_TPM` | _(rate budget)_ | Per-backend tokens-per-minute ceiling for the shared dispatch rate bucket (same naming as `_RPM`). Dormant unless set. Must be a positive integer; blank/invalid values are ignored. |
+| `OUROBOROS_BACKEND_LIMITS` | _(config path)_ | Path to the backend-limits YAML file (default `~/.ouroboros/backend_limits.yaml`). See [Backend concurrency & rate limits](#backend-concurrency--rate-limits). |
 | `OUROBOROS_CLI_PATH` | `orchestrator.cli_path` | Path to the Claude CLI binary. |
 | `OUROBOROS_CODEX_CLI_PATH` | `orchestrator.codex_cli_path` | Path to the Codex CLI binary. |
 | `OUROBOROS_OPENCODE_CLI_PATH` | `orchestrator.opencode_cli_path` | Path to the OpenCode CLI binary. |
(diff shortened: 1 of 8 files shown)


deepakdgupta1:
Thanks for the review — addressed the blocker in `f7c1ae5`.

**Blocker (backend identity alignment):** Confirmed and fixed. I verified the actual `runtime_backend` values across all eight adapters: `hermes_cli`, `codex_cli`, `opencode`, `gemini_cli`, `goose`, `pi`, `copilot_cli`, `kiro` (the `runtime_backend` *property* returns the `_runtime_handle_backend` `*_cli` form). Operators configure runtimes by the user-facing selector name (`hermes`/`codex`/`gemini`/`copilot`, matching `orchestrator.runtime_backend` / `OUROBOROS_AGENT_RUNTIME`), so `OUROBOROS_HERMES_RPM` / `backends: hermes:` would not have matched the `hermes_cli` value the executor passes in. Fix: added `*_cli → <user-facing>` canonical aliases (`hermes_cli→hermes`, `codex_cli→codex`, `gemini_cli→gemini`, `copilot_cli→copilot`) so config, env keys, and the executor's `adapter.runtime_backend` all resolve to one identity. Docs now use the user-facing names with a canonicalization note, and I added regression tests proving the `hermes_cli` handle resolves the `OUROBOROS_HERMES_RPM` / `backends: hermes:` budget (`test_cli_handle_name_resolves_to_user_facing_env_key`, `test_user_facing_config_key_applies_to_cli_handle`).

**Follow-up (coordinator conflict-resolution sessions bypass the gate, `coordinator.py:383`):** Accurate — only decomposition and atomic AC dispatches are paced today. Routing `LevelCoordinator`'s own adapter calls through the shared gate is a worthwhile improvement, but it requires threading the gate into the coordinator and is broader than this PR's scope (and the gate is dormant by default). I'd prefer to land it as a focused follow-up rather than widen this change; noted for that.

Full suite green (10,539 passed, 5 skipped); `ruff check` + `ruff format --check` + `mypy` clean.

shaun0927:
## Hermes readiness + PR review

### Readiness comment
I re-reviewed this PR against Issue #961 as the SSOT for the AgentOS / `ooo auto` direction. In that frame, this PR is directionally appropriate because it is focused on making runtime/tooling contracts more explicit and truthful rather than adding speculative orchestration surface area.

**PR:** #1372 — feat(orchestrator): pace non-Claude delivery within a configurable rate budget  
**Current head:** `f7c1ae525`  
**Scope assessment:** This PR paces non-Claude delivery with configurable rate budgeting rather than unbounded dispatch.  
**Remediation/audit result:** No additional remediation was required beyond the final SSOT/design audit; the current head already matches the direction and has clean gates.  
**Gate status:** ouroboros-agent[bot] review is APPROVED on f7c1ae525. All visible status checks are green (9/9). GitHub merge state is `CLEAN` and review decision is `APPROVED`.  


Based on the current bot/design-note state and visible checks, I consider this PR merge-ready subject to normal repository policy and, where applicable, stack ordering.

### `pr-review` analysis
**Intent traced:** The change was reviewed for alignment with the #961 SSOT: preserve AgentOS/ooo-auto as a contract-first orchestration layer, expose degradation and ownership boundaries truthfully, and avoid broad speculative rewrites.  
**Scope traced:** The implementation remains within its PR-local concern; I did not find scope expansion that would undermine the roadmap direction.  
**Risk review:** The latest bot/design review is approved, merge state is clean, and status checks are green or not required. Remaining risk is ordinary integration/merge-order risk rather than a known design blocker.  
**Over-engineering review:** I did not find unnecessary new architecture in the final head. The accepted shape is minimal and corrective: it clarifies existing contracts, fixes drift, or adds targeted tests instead of creating a parallel system.  
**Merge recommendation:** APPROVE.

**Approve verdict:** APPROVE — this PR is acceptable to merge from the Issue #961 / AgentOS / `ooo auto` readiness perspective.

PR #1402, opened 2026-06-12
docs(rfc): configuration coherence — surface config, make reload honest (#1376)

deepakdgupta1:
RFC for the **next slice** of the usability/transparency theme in discussion #1376, after the merged breadcrumb + TUI slice (#1392) deliberately deferred `ooo config` to its own RFC.

## What it proposes
1. **Detect stale reconnect-bound edits and warn** — when a structural config field changes on disk but a live adapter still holds the old value, never let the no-op edit pass silently; name the field and say a reconnect is required.
2. **Surface the effective configuration** — `ooo config` shows each key as *active value · default · what it controls · source* (Ouroboros / runtime / env); a change shows a diff with a per-key *applies live vs. needs reconnect* flag.

Detect-and-warn ships first (cheap, removes the whole "I fixed it but it's still broken" loop); full live-reload is deferred. The owner corroborated the silent-MCP-reconnect pain directly and said this slice "deserves its own RFC rather than riding along."

Refs: discussion #1376 · builds on #1392

### DIFF
--- a/docs/rfc/config-coherence.md
+++ b/docs/rfc/config-coherence.md
@@ -28,55 +28,75 @@ ships first. Full live-reload is more invasive and is deferred.
 
 ## Context
 
-### Some config is read live, some is construction-bound
+### Config binds at different points, and nothing says which
 
-The inconsistency is real and unsignalled:
+The same edit can take effect immediately, on the next tool call, or only after a
+reconnect — depending on *where* the field binds. Nothing signals which:
 
-- `orchestrator/adapter.py` captures `self._permission_mode` at construction and
-  exposes it as a read-only property. **Construction-bound** — editing
-  `~/.ouroboros/config.yaml` does nothing to a live adapter until reconstruction
-  (an `/mcp` reconnect).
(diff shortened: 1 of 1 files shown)


deepakdgupta1:
Thanks for the review — addressed in the latest push:

- **Blocker (permission_mode):** you're right that a fresh MCP tool run re-reads `permission_mode` (`create_agent_runtime` → `get_agent_permission_mode` → `config/loader.py`), so it is *not* reconnect-required for a fresh run. Reframed the whole mechanism around a **binding-point taxonomy** — `live` / `tool_entry` / `handle_bound` — and narrowed detect-and-warn to the **`handle_bound` / inherited-handle** case on an in-flight session, which is the one that actually goes stale. `permission_mode` is now classified `tool_entry`.
- **Follow-up:** dropped the approximate "~13 sections / ~50 env overrides" counts (drift-prone).
- **Non-blocking:** clarified that `ooo config` is the skill/MCP shorthand and `ouroboros config` is the CLI surface.

PR #1403, opened 2026-06-12
docs(rfc): token frugality as one control loop — attribution + advisory guardrails (#1377)

deepakdgupta1:
RFC for the greenlit first slice of the token-frugality theme in discussion #1377.

## Frame
Per the owner, the four #1377 themes are **one control loop**, not four features: spend attribution = the *sensor*; the reflective guardrail loop = the *learning layer*; the user-held assurance dial = the *policy input*; completion-feasibility = the controller's *initial estimate*; the spend decision itself = the *controller* (its own RFCs, #1384).

## First slice (this RFC)
- **Spend attribution (the sensor)** — per-run `total_cost` + itemized *avoidable* portion (rework / dead-escalation / stagnation) + per-stage attribution, from already-event-sourced signals. Floor-preserving: only flags motion that didn't advance a *verified* AC.
- **Advisory reflective guardrail loop (the learning layer)** — emits one generalizable, owner-tagged, reversible, project-scoped guardrail per session; advisory in v1.

Records the two hard invariants (never reduce the outcome; never increase rework risk), the layered ownership, and that fan-out discipline already shipped (`backend_limits` + #1372). Adaptive concurrency is the named second slice.

Refs: discussion #1377 · composes with #1384 RFCs

### DIFF
--- a/docs/rfc/frugality-control-loop.md
+++ b/docs/rfc/frugality-control-loop.md
@@ -1,0 +1,144 @@
+# RFC — Token frugality as one control loop (attribution + advisory guardrails)
+
+> Status: **Draft**
+> Relates to [discussion #1377](https://github.com/Q00/ouroboros/discussions/1377)
+> (token frugality). Composes with the spend-decision mechanism RFCs —
+> [spend estimator](./spend-estimator.md) and
+> [spend actuator (effort dial)](./spend-actuator-effort-dial.md) (#1384) — and the
+> [decomposition reliability](./decomposition-reliability.md) RFC (#1385). Sibling
+> usability thread: [#1376](https://github.com/Q00/ouroboros/discussions/1376).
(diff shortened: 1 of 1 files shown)


deepakdgupta1:
Thanks — addressed in the latest push:

- **Blocker (broken sibling links):** the estimator / actuator / decomposition RFCs live in separate per-slice PRs (per the owner's "split per slice" guidance), so their `./*.md` targets aren't in this branch. Repointed the cross-links at their PRs — #1404, #1405, #1406 — as you suggested, so the doc is navigable now; they can become relative links once those merge.
- **Follow-up (stale TUI paths):** corrected to `src/ouroboros/tui/widgets/cost_tracker.py` and `src/ouroboros/tui/components/token_tracker.py`.

PR #1404, opened 2026-06-12
docs(rfc): the spend estimator — difficulty + stakes from measured inputs (#1384)

deepakdgupta1:
**Estimator half** of the split the owner requested in discussion #1384 (the **actuator half** is a sibling PR).

## What it proposes
A per-unit estimate of **difficulty** and **stakes** with explicit **confidence**, from **measured inputs only** — difficulty decoupled from length (a short, conceptually hard unit must rate hard), stakes = cost-of-being-wrong (reversibility / blast radius / sensitivity). It does *not* decide what to do with the estimate (that's the actuator RFC).

## Properties (restructured per owner)
- **v1 invariants:** two-axis, fail-safe-under-uncertainty (escalate, don't cheapen), monotonic/stable, faithful auditable rationale, and the owner-added **runtime-scoped applicability** (routable vs. advisory per backend).
- **v2 maturity goals:** calibration and cross-run learning (they collide with cold-start; v1 only *records* their inputs).

Grounded in current `main`: the estimate exists but its consumers are unprincipled or absent (the tier router has no non-test call site; inputs are fabricated; length is treated as difficulty).

Refs: discussion #1384 · sibling: spend actuator RFC

### DIFF
--- a/docs/rfc/spend-estimator.md
+++ b/docs/rfc/spend-estimator.md
@@ -1,0 +1,106 @@
+# RFC — The spend estimator: difficulty + stakes from measured inputs
+
+> Status: **Draft**
+> Relates to [discussion #1384](https://github.com/Q00/ouroboros/discussions/1384)
+> (complexity→investment mechanism). This is the **estimator half** of the split the
+> owner requested; the **actuator half** is
+> [spend actuator (effort dial)](./spend-actuator-effort-dial.md). Frugality stance
+> both serve: [#1377](https://github.com/Q00/ouroboros/discussions/1377) /
+> [frugality control loop](./frugality-control-loop.md).
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


deepakdgupta1:
Thanks — addressed in the latest push:

- **Blocker (live consumer):** good catch. Corrected the context — the *tier router* (`PALRouter.route`/`ModelRouter.route`) has no call site, but `estimate_complexity` itself *is* called at `execution/atomicity.py:271`. That module is off the live path and is slated for deletion in #1406, so there's no *live-path* consumer of the estimate today; the RFC now says exactly that rather than "no live consumer."
- **Blocker (links):** repointed the sibling RFC links at their PRs (#1405 actuator, #1403 frugality) until they land.
- **Nit:** reworded "Frugality stance both serve".

PR #1405, opened 2026-06-12
docs(rfc): the spend actuator — a reasoning-effort dial, not model switching (#1384)

deepakdgupta1:
**Actuator / wiring half** of the split the owner requested in discussion #1384 (the **estimator half** is a sibling PR).

## What it proposes (the owner's reshape)
The lever is **reasoning effort, not model switching**. Routing on the effort dial (Anthropic thinking budget, OpenAI reasoning effort, Codex `model_reasoning_effort`) is monotonic and cliff-free, makes escalation cheap (so fail-safe-on-uncertainty is practical), and keeps a mis-estimate from changing the executor's capability class (shrinking cold-start risk).

- Wire an **effort-first investment decision** into the live executor (`orchestrator/parallel_executor.py`).
- Escalation ladder: *effort low → effort high → bigger model* (model-tier switching is only the coarse outer rung + the no-effort-dial fallback, e.g. GLM).
- **Capability matrix + advise-fallback:** route where Ouroboros calls the LLM directly; *advise* (guardrail in the spec) where it delegates to a CLI runtime.
- **Remove** the unwired `PALRouter` / tier machinery rather than wiring it — it encodes the wrong actuator.

Refs: discussion #1384 · sibling: spend estimator RFC

### DIFF
--- a/docs/rfc/spend-actuator-effort-dial.md
+++ b/docs/rfc/spend-actuator-effort-dial.md
@@ -1,0 +1,109 @@
+# RFC — The spend actuator: a reasoning-effort dial, not model switching
+
+> Status: **Draft**
+> Relates to [discussion #1384](https://github.com/Q00/ouroboros/discussions/1384)
+> (complexity→investment mechanism). This is the **actuator / wiring half** of the
+> split the owner requested; the **estimator half** is
+> [spend estimator](./spend-estimator.md). Frugality stance:
+> [#1377](https://github.com/Q00/ouroboros/discussions/1377) /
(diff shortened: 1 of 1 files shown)


deepakdgupta1:
Thanks for the review (and the approve). Addressed the flagged documentation-integrity warning: the sibling RFCs (estimator #1404, frugality #1403, decomposition #1406) live in separate per-slice PRs per the owner's "split per slice" guidance, so I repointed the cross-links at those PRs rather than dangling `./*.md` targets. They can become relative links once the siblings merge.

PR #1406, opened 2026-06-12
docs(rfc): atomicity & decomposition reliability, re-grounded on the live executor (#1385)

deepakdgupta1:
RFC for discussion #1385, **re-grounded** on the owner's verification finding.

## The re-grounding
The module the original analysis dissected — `execution/atomicity.py` + `execution/decomposition.py`, via `DoubleDiamondExecutor` — has **no non-test caller in `src/`** (115 `execution.ac.completed` events, **zero** atomicity-check events). The live executor is **`orchestrator/parallel_executor.py`**.

## What it proposes
1. **Delete** the dead modules (repairing fabricated fields in code nothing calls is motion, not progress).
2. **Re-aim the discipline at the live split path:** a verified decomposition step (collectively-exhaustive / mutually-exclusive / each child measurably simpler, one repair retry then escalate) plus bounce-cause classification (too-big / bad-spec / environment) on the existing attempt-then-bounce loop, seeding splits from the bounce *trace*.

Properties restructured per owner (closed-loop → v2; boundary-robustness fires only mid-confidence) + the owner-added **live-path-verification-first** property. Lineage note: this transplants the methodology-first philosophy onto the executor that actually runs.

Refs: discussion #1385 · sibling mechanism: #1384 RFCs

### DIFF
--- a/docs/rfc/decomposition-reliability.md
+++ b/docs/rfc/decomposition-reliability.md
@@ -3,10 +3,10 @@
 > Status: **Draft**
 > Relates to [discussion #1385](https://github.com/Q00/ouroboros/discussions/1385)
 > (atomicity & AC-decomposition reliability). Sibling mechanism:
-> [spend estimator](./spend-estimator.md) (#1384) — both pre-execution judgments,
+> [spend estimator](https://github.com/Q00/ouroboros/pull/1404) (#1384) — both pre-execution judgments,
 > both shared the fabricated-output failure pattern. Frugality stance:
 > [#1377](https://github.com/Q00/ouroboros/discussions/1377) /
-> [frugality control loop](./frugality-control-loop.md).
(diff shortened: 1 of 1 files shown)


deepakdgupta1:
Thanks for the review (and the approve). Addressed all three notes:

- **Link warning:** repointed the sibling RFC links at their PRs (#1404 estimator, #1403 frugality, #1405 actuator) until they land.
- **Class name:** corrected `DoubleDiamondExecutor` → `DoubleDiamond`, and noted the modules are reachable only via `DoubleDiamond.run_cycle_with_decomposition`.
- **Criterion 1 contradiction:** clarified the removal order — drop the `run_cycle_with_decomposition` wiring (the only non-test importer) first, then remove the modules; the no-importer assertion runs *after* the wiring is gone.

PR #1408, opened 2026-06-12
refactor(mcp): consolidate tool-layer duplication, fix durable cancel + lost audit rows

Q00:
## Summary

Adversarially-verified duplication consolidation inside `src/ouroboros/mcp/tools/`, in four commits. Two of the consolidations fix real behavior bugs the copy-paste divergence had created.

1. **`e3780e19` — remove orphaned background-path subagent build (DUP-04).** The fall-through branch in `StartExecuteSeedHandler._handle_inner` built a `build_execute_subagent` payload that was never referenced (orphaned by 3c393c98). Deliberately NOT hoisted above the branch — that would reintroduce the session-id bug 3c393c98 fixed.

2. **`866b8dcd` — delete 12 zero-caller factory functions (DUP-03 dead portion).** Each verified to have zero callers across src/ and tests/ (same-named modules/local variables disambiguated). `get_ouroboros_tools` still yields 29 handlers; production-live and test-used factories kept.

3. **`02296daa` — extract the 5×-copied background-job pipeline (DUP P4) + durable-cancel fix.** New `mcp/tools/background.start_background_tool_job(...)` owns allocate_job_id → should_cancel guard → `run_with_agent_process(process_id, cancel_key=mcp_job:{job_id})` → start_job → receipt envelope. **Behavior fix:** `StartEvaluate` and `StartAuto` previously passed no cancel guard and no `cancel_key`, so `JobManager.cancel_job`'s durable `mcp_job:{job_id}` marker (whose docstring promises restart-visible cancellation) was written but never honored for them. Both now get the same guard + key, with tests proving the durable marker binds. Preserved: auto's lease lifecycle composes around the helper via hooks; execute_seed's post-receipt idempotency write untouched.

4. **`b4080a2f` — extract the 13×-copied plugin-dispatch terminal sequence (DUP P5) + audit-row fix.** New `subagent.dispatch_plugin_terminal(...)` guards `event_store.initialize()` (losing the audit row must never fail the dispatch), emits `subagent.dispatched`, returns the result envelope. **Fix:** several sync sites emitted on an uninitialized store and silently lost the audit row. The `delegated_to_subagent` / `delegated_to_plugin` status strings became constants WITHOUT unifying values (intentional public-contract distinction). Left alone by design: the two lateral_think no-emit sites and auto's lease-interleaved variant (it needs initialize-failure to be observable).

## Test plan

- `tests/unit/mcp`: 1157 passed (+6 new: durable-cancel proofs for start_evaluate/start_auto, `TestDispatchPluginTerminal` init-before-emit / swallowed-failure / envelope shape).
- `tests/integration/mcp`: 88 passed. `tests/unit/orchestrator -k agent_process`: 75 passed. ruff + format clean.


Q00:
Addressed the review (bdb2178d):

- **Blocker — public factory exports**: reverted the DUP-03 factory deletion outright (`d09af048`). `src/ouroboros/mcp/tools/__init__.py` and `definitions.py` are now byte-identical to `main` — `from ouroboros.mcp.tools import job_status_handler` etc. all import again. The zero-internal-caller cleanup will be re-proposed separately behind a deprecation cycle.
- **Follow-up (background.py)**: `on_enqueue_failure` is now try/except-guarded so a failing cleanup hook can never mask the original `start_job` exception (logged with `exc_info`).
- **Test-coverage suggestion**: already covered — `TestDispatchPluginTerminal::test_initialize_failure_is_swallowed_dispatch_still_succeeds` pins the fail-open envelope when `initialize()` raises.

`tests/unit/mcp`: 1157 passed.

PR #1418, opened 2026-06-13
fix(plugin): reload deferred MCP schema before each ouroboros_* call

### LINKED ISSUE #1417: [Bug] Multi-turn skills flash "Invalid tool parameters" on every ouroboros_* call — deferred MCP-tool schema loaded once at skill entry, not per call
## Summary

On Claude Code builds where the `ouroboros_*` MCP tools are registered as **deferred tools**, every multi-turn skill (`interview`, `pm`, `seed`, `evaluate`, `run`, `evolve`, plus the deferred-tool calls inside `setup`) shows a recurring **"Invalid tool parameters"** error. Root cause: the skills load the deferred tool schema only once (e.g. `interview` Step 0.5, `seed`/`run`/`evaluate`/`evolve` "Load MCP Tools", `pm` Step 1), but each subsequent turn runs in a fresh tool context where that deferred schema is no longer loaded. The first call on a later turn fails client-side input validation → "Invalid tool parameters". In most cases the runtime then re-discovers/loads the schema and the retry succeeds, so the operation completes — but the red error flashes on nearly every step, which reads as "it keeps erroring / is it broken?".

## Impact

- Primary symptom: the error appears repeatedly across the whole pipeline (interview → seed → run/evaluate/evolve, and setup), making the tool look broken and eroding user trust even when work completes.
- Verified non-fatal on the `setup` path: in a real session, `ouroboros_brownfield {action:"scan"}` ("13 repositories registered") and CLAUDE.md integration both completed successfully despite the error flashing — the failed first attempt self-corrects on retry.
- Risk: at user-decision gates (interview "confirm/decide", seed User-Adoption-Gate / `Balance: N expand / N sharpen / N remove`, setup "Step 3 CLAUDE.md integration") a first-attempt validation failure that does NOT auto-retry would stall the flow. Severity therefore ranges from cosmetic-but-noisy to blocking depending on whether the runtime retries that specific call.
- Related: #126 (deferred tools detected as unavailable → always Path B) is the inverse symptom of the same deferred-tool-loading weakness.

## Environment

- Ouroboros version: 0.41.0 (plugin + `ouroboros-ai` via `uvx --from ouroboros-ai[mcp,claude]`)
- OS: Linux 6.17.0-23-generic x86_64
- Python: 3.12.3
- Installation method: Claude Code plugin (marketplace `Q00/ouroboros`); MCP server launched by `uvx`
- Runtime/provider: Claude Code 2.1.177 (Anthropic). The runtime registers plugin MCP tools as **deferred tools** (schema not loaded until an explicit tool-discovery / ToolSearch query). This is common when many MCP servers/plugins are installed.

## Steps to reproduce

1. On a Claude Code build that registers plugin MCP tools as deferred tools (typically when many tools are installed), run any multi-turn ouroboros skill, e.g. `/ouroboros:interview <idea>` or `/ouroboros:setup`.
2. The skill's one-time load step loads the schema; the first `ouroboros_*` call succeeds.
3. On a later turn (next question / next iteration / next step), the skill issues another `ouroboros_*` call without re-loading the deferred schema.
4. Observe **"Invalid tool parameters"** on that call. The runtime usually re-loads + retries and the call then succeeds, so the error flashes on most steps without necessarily blocking.

## Expected behavior

Multi-turn skills complete without spurious "Invalid tool parameters" errors: the deferred tool schema is guaranteed loaded immediately before every `ouroboros_*` call, so no call is rejected by client-side validation.

## Actual behavior

The deferred schema is loaded once at skill entry. On later turns the first `ouroboros_*` call is rejected with "Invalid tool parameters" because the schema is not loaded in that turn's context. The error recurs throughout the pipeline; work often still completes via retry, but the experience looks broken.

## Logs or error output

```
<assistant narration of the next step, e.g. "Step 3 — CLAUDE.md integration" / interview gate / seed Balance gate>
  ⎿  Invalid tool parameters
```

Diagnostic confirmation (same environment):
- `ouroboros mcp serve` registers all `ouroboros_*` tools; `ouroboros --version` → 0.41.0; hooks are healthy.
- After explicitly loading a deferred tool schema, calling it returns a NORMAL application response, not "Invalid tool parameters". Example: `ouroboros_session_status {session_id:"__probe__"}` returned `Session not found: ...` (i.e. the call dispatched fine). This proves the failure is purely schema-load timing, not a genuine parameter-shape error (the relevant tool params are all optional/simple).
- Real `setup` session transcript: `ToolSearch "+ouroboros brownfield"` immediately preceding `ouroboros_brownfield {action:"scan"}` → success; CLAUDE.md integration via Write → success. Where the skill reloads right before the call, no error occurs.

## Minimal reproduction

Skill files that load the deferred schema once, then call `ouroboros_*` on later turns without re-loading:
- `skills/interview/SKILL.md` — Step 0.5 loads once; flow calls `ouroboros_interview` every turn.
- `skills/pm/SKILL.md` — Step 1 loads once; Step 2+ call `ouroboros_pm_interview` every turn.
- `skills/seed/SKILL.md` — "Load MCP Tools" once; Reflect/Refine loop calls `ouroboros_qa` / `ouroboros_lateral_think` per iteration (error surfaces at the `Balance:` User-Adoption-Gate).
- `skills/run/SKILL.md`, `skills/evaluate/SKILL.md`, `skills/evolve/SKILL.md` — same one-time-load pattern with multiple `ouroboros_*` calls.

No per-call reload between turns ⇒ deferred schema absent on turn N>1 ⇒ "Invalid tool parameters".

## Suggested fix

Make the deferred-schema load **idempotent and per-call**: immediately before every `ouroboros_*` MCP call inside a multi-turn skill, re-issue the tool-discovery/ToolSearch load query (a no-op when already loaded). If the load returns no matching tool, fall through to the skill's documented Path B / unavailable fallback instead of retrying the failing call.

Verified-working prototype patch (applied locally to the 6 SKILL.md files above): add a "deferred-schema guard" note to each skill's load step plus a per-call reminder at the call sites. Happy to open a PR.

- [ ] Multi-turn skills (`interview`, `pm`, `seed`, `evaluate`, `run`, `evolve`) complete without spurious "Invalid tool parameters" when `ouroboros_*` tools are deferred.
- [ ] The deferred-schema load is guaranteed before every `ouroboros_*` MCP call in a loop, not only once at entry.
- [ ] User-decision gates (interview confirm/decide, seed User-Adoption-Gate, setup Step 3) no longer flash validation errors.
- [ ] When a deferred tool is genuinely unavailable, the skill cleanly routes to Path B / documented fallback rather than retrying a failing call.

seongjin-lim:
## Summary

Multi-turn skills (`interview`, `pm`, `seed`, `run`, `evaluate`, `evolve`) load the deferred `ouroboros_*` MCP tool schema **only once at skill entry**. On Claude Code builds that register plugin MCP tools as *deferred tools*, a schema loaded on one turn is not guaranteed to remain loaded on the next turn's fresh tool context. The first `ouroboros_*` call on a later turn is then rejected client-side with **"Invalid tool parameters"** before it ever reaches the server.

This fixes #1417.

## Symptom

The error flashes on nearly every step of the pipeline (interview → seed → run/evaluate/evolve, and the deferred calls in `setup`). In many cases the runtime re-discovers the schema and the retry succeeds, so work still completes — but it reads as a broken tool and erodes trust. At user-decision gates (interview confirm/decide, seed User-Adoption-Gate / `Balance:` line, setup Step 3) a first-attempt failure that does not auto-retry can stall the flow.

## Root cause (verified)

- After explicitly loading a deferred tool's schema, calling it returns a normal application response — e.g. `ouroboros_session_status {session_id:"__probe__"}` → `Session not found: ...` — i.e. the call dispatches fine. So the failure is purely **schema-load timing**, not a genuine parameter-shape error (the relevant params are all optional/simple).
- A real `setup` session showed `ToolSearch "+ouroboros brownfield"` immediately preceding `ouroboros_brownfield {action:"scan"}` → success; where the skill reloads right before the call, no error occurs.

## Fix

Add a **deferred-schema guard** to each affected skill instructing the agent to re-run the *idempotent* tool-discovery load immediately before **every** `ouroboros_*` call (a no-op when already loaded), and to fall through to the documented Path B / not-found fallback when the load returns no tool.

Applied to both trees:
- `skills/` (the tree referenced by `.claude-plugin/plugin.json` → `"skills": "./skills/"`)
- `.claude-plugin/skills/` mirror (kept in sync)

Docs-only change (skill instructions); no code paths touched.

## Files

`interview`, `pm`, `seed`, `run`, `evaluate`, `evolve` — `SKILL.md` in each of the two skill trees (12 files, +140 lines).

## Testing

- Confirmed the guard's discovery queries (`+ouroboros interview`, `+ouroboros pm_interview`, etc.) load the schemas and that a subsequently-issued call validates and dispatches (returns app-level result, not "Invalid tool parameters").
- Verified anchor uniqueness and that the guard text inserts cleanly in all 12 files.

## Notes

- `skills/` and `.claude-plugin/skills/` are largely duplicated in the repo (only `pm` currently differs between them). If one is intended to be generated from the other, happy to drop the mirror edits and wire up generation instead — let me know.

### DIFF
--- a/.claude-plugin/skills/brownfield/SKILL.md
+++ b/.claude-plugin/skills/brownfield/SKILL.md
@@ -41,6 +41,14 @@ This may take a moment...
 
 **Implementation — use MCP tools only, do NOT use CLI or Python scripts:**
 
+**CRITICAL — deferred-schema guard (prevents "Invalid tool parameters"):**
+This skill can call `ouroboros_brownfield` across multiple turns (`scan`,
+`set_defaults`, `defaults`, and `set`). A deferred schema loaded for one turn is
+NOT guaranteed to remain loaded for the next. Immediately before EVERY
+`ouroboros_brownfield` call, re-run `ToolSearch query: "+ouroboros brownfield"`
+(idempotent — a no-op when already loaded). If the load returns no matching tool,
+stop with the MCP-not-available message instead of retrying the failing call.
+
 1. Load the brownfield MCP tool: `ToolSearch query: "+ouroboros brownfield"`
 2. Call scan+register:
    ```
@@ -117,7 +125,7 @@ Then stop.
 
 The user can select the recommended defaults (if any), choose "None", or type custom numbers.
 
-After the user responds, use ONE MCP call to update all defaults at once:
+After the user responds, re-run `ToolSearch query: "+ouroboros brownfield"`, then use ONE MCP call to update all defaults at once:
 
 ```
 Tool: ouroboros_brownfield
@@ -155,7 +163,7 @@ Scan only, no default selection prompt. Show the numbered list and stop.
 
 ### Subcommand: `defaults`
 
-Load the brownfield MCP tool and call:
+Re-run `ToolSearch query: "+ouroboros brownfield"`, then call:
 ```
 Tool: ouroboros_brownfield
 Arguments: { "action": "scan" }
@@ -170,7 +178,7 @@ No default repos set. Run 'ooo brownfield' to configure.
 
 ### Subcommand: `set <indices>`
 
-Directly set defaults without scanning. Parse the comma-separated indices from the user's input and call:
+Directly set defaults without scanning. Parse the comma-separated indices from the user's input, re-run `ToolSearch query: "+ouroboros brownfield"`, and call:
 
 ```
 Tool: ouroboros_brownfield
--- a/.claude-plugin/skills/evaluate/SKILL.md
+++ b/.claude-plugin/skills/evaluate/SKILL.md
@@ -52,6 +52,18 @@ The Ouroboros MCP tools are often registered as **deferred tools** that must be 
 
 **IMPORTANT**: Do NOT skip this step. Do NOT assume MCP tools are unavailable just because they don't appear in your immediate tool list. They are almost always available as deferred tools that need to be loaded first.
 
+**CRITICAL — deferred-schema guard (prevents "Invalid tool parameters"):**
+This skill makes `ouroboros_*` MCP calls across multiple turns, and each turn runs
+in a fresh tool context. A deferred tool's schema loaded on one turn is NOT
+guaranteed to still be loaded on the next. If you call any `ouroboros_*` MCP tool
+while its schema is not loaded in the **current** turn, the runtime rejects the
+call with **"Invalid tool parameters"** before it ever reaches the server.
+Therefore: **immediately before EVERY `ouroboros_*` MCP call in this skill, re-run
+the same tool-discovery load query you used above** (idempotent — a no-op when the
+schema is already loaded) so the schema is guaranteed present for that call. If a
+load ever returns no matching tool, switch to the documented fallback / Path B
+instead of retrying the failing call.
+
 ### Evaluation Steps
 
 1. Determine what to evaluate:
--- a/.claude-plugin/skills/evaluate/SKILL.md
+++ b/.claude-plugin/skills/evaluate/SKILL.md
@@ -53,16 +53,14 @@ The Ouroboros MCP tools are often registered as **deferred tools** that must be 
 **IMPORTANT**: Do NOT skip this step. Do NOT assume MCP tools are unavailable just because they don't appear in your immediate tool list. They are almost always available as deferred tools that need to be loaded first.
 
 **CRITICAL — deferred-schema guard (prevents "Invalid tool parameters"):**
-This skill makes `ouroboros_*` MCP calls across multiple turns, and each turn runs
-in a fresh tool context. A deferred tool's schema loaded on one turn is NOT
-guaranteed to still be loaded on the next. If you call any `ouroboros_*` MCP tool
-while its schema is not loaded in the **current** turn, the runtime rejects the
-call with **"Invalid tool parameters"** before it ever reaches the server.
-Therefore: **immediately before EVERY `ouroboros_*` MCP call in this skill, re-run
-the same tool-discovery load query you used above** (idempotent — a no-op when the
-schema is already loaded) so the schema is guaranteed present for that call. If a
-load ever returns no matching tool, switch to the documented fallback / Path B
-instead of retrying the failing call.
+This skill can call `ouroboros_evaluate` after a fresh turn. A deferred tool's
+schema loaded on one turn is NOT guaranteed to still be loaded on the next. If
+you call it while its schema is not loaded in the **current** turn, the runtime
+rejects the call with **"Invalid tool parameters"** before it reaches the server.
+Therefore: **immediately before EVERY `ouroboros_evaluate` call in this skill,
+re-run `ToolSearch query: "+ouroboros evaluate"`** (idempotent — a no-op when
+already loaded). If the load returns no matching tool, switch to the documented
+fallback instead of retrying the failing call.
 
 ### Evaluation Steps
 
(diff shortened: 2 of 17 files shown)


shaun0927:
Ready for re-review on current HEAD `ee0ceb4dcafcd9862430c0def898c5c8fe65e4aa`.

This addresses the latest `ouroboros-agent` blocker:
- `skills/seed/SKILL.md` and `.claude-plugin/skills/seed/SKILL.md` now instruct the agent to reload the discovery query for the specific MCP tool about to be called;
- the seed skill now explicitly maps `+ouroboros seed` → `ouroboros_generate_seed`, `+ouroboros qa` → `ouroboros_qa`, and `+ouroboros lateral` → `ouroboros_lateral_think`;
- this preserves the deferred-schema guard without routing QA/lateral calls through the seed schema query.

Verification:
- `uv run pytest tests/unit/skills/test_skill_artifacts.py tests/unit/test_runtime_skill_capability_docs.py -q` — 3 passed;
- `git diff --check` — passed;
- GitHub CI is green on current head.

shaun0927:
Updated the PR to address the deferred-schema review blockers.\n\nChanges pushed:\n- replaced the fragile "same tool-discovery query" guidance with per-called-tool / documented-tool-family reload guidance;\n- kept seed + interview lateral paths on `+ouroboros lateral`;\n- covered `setup` and `brownfield` `ouroboros_brownfield` calls before/after user turns;\n- added a regression assertion that packaged skill docs no longer contain the stale "same tool-discovery load query you used above" guidance and that multi-tool guards name the relevant discovery queries.\n\nVerification:\n- local: `ruff format --check .` PASS\n- local: `ruff check .` PASS\n- local: `pytest tests/unit/skills/test_skill_artifacts.py tests/unit/test_runtime_skill_capability_docs.py -q` PASS (4 passed)\n- GitHub checks on `ef953196ee0349d75ff73ee0e820a9a1a5113432`: all SUCCESS.\n\nReady for re-review.

shaun0927:
ouroboros-agent re-review requested for latest head `fe54fd286d86accf25e5aeba5196c5b2123279bf`.

Current evidence:
- GitHub checks are all green and mergeStateStatus is CLEAN on this head.
- The visible CHANGES_REQUESTED review is stale and references older head `8674d02f50b3aa8e4ea0b1a9967ee34b31fd9c93`.
- The documented blockers were addressed by explicit seed/interview per-tool schema reload guidance plus regression coverage.

Please review the latest head so the stale change-request state can be cleared.

Q00:
Closed out the remaining non-blocking / follow-up item (`a7cb6065`):

**Follow-up + Non-blocking (both pointed at `test_skill_artifacts.py:38`):** The deferred-schema guard regression test only pinned brownfield/setup/seed/interview, so drift in the guards added to **evaluate / evolve / pm / run** wouldn't be caught. Extended the expected query→tool map to all four:
- `evaluate`: `"+ouroboros evaluate"` → `ouroboros_evaluate`
- `evolve`: `"+ouroboros evolve"`→`ouroboros_evolve_step`, `"+ouroboros interview"`→`ouroboros_interview`, `"+ouroboros seed"`→`ouroboros_generate_seed`, `"+ouroboros lateral"`→`ouroboros_lateral_think`
- `pm`: `"+ouroboros pm_interview"` → `ouroboros_pm_interview`
- `run`: `"+ouroboros execute"` → `ouroboros_start_execute_seed` / `ouroboros_job_wait` / `ouroboros_job_result`

The assertions run against **both** `skills/` and `.claude-plugin/skills/` copies. `test_skill_artifacts.py` + `test_runtime_skill_capability_docs.py`: 4 passed. This leaves no outstanding blocking, follow-up, or non-blocking findings.

Q00:
Closed the remaining Low follow-up/non-blocking (`94b3be07`): added `('"+ouroboros execute"', 'ouroboros_ac_tree_hud')` to the run guard's expected map, so the direct drill-down path the run docs guard is pinned against drift too. `test_skill_artifacts.py`: 2 passed.

PR #1436, opened 2026-06-14
feat(cli): add read-only job event polling

### LINKED ISSUE #1424: No read-only observability into `ouroboros.db` (WAL + lock blocks external `sqlite3` reads)
## Summary

`~/.ouroboros/ouroboros.db` (the run/event store) cannot be inspected read-only with the `sqlite3` CLI while the MCP server is active — `.tables` returns empty (WAL + lock). There is no lightweight, external way to observe AC/job progress without going through the MCP tools, and those are chatty (see the `job_wait` long-poll issue). External schedulers/dashboards have no cheap polling path.

## Environment

- Ouroboros **0.41.0** (plugin; uv tool `ouroboros-ai`), MCP `ouroboros mcp serve`
- `persistence.database_path: data/ouroboros.db` (resolves to `~/.ouroboros/ouroboros.db`, ~95 MB, WAL mode with `-shm`/`-wal` sidecars)

## Steps to reproduce

While a job runs:

```
sqlite3 ~/.ouroboros/ouroboros.db ".tables"      # -> empty
sqlite3 ~/.ouroboros/ouroboros.db "SELECT name FROM sqlite_master LIMIT 5;"   # -> empty
```

## Expected

A documented, read-only way to observe progress for external monitors — without taking the writer's lock and without MCP token cost.

## Suggested fix

- Provide a tiny read-only CLI, e.g. `ouroboros events --job <id> --since <cursor>` or `ouroboros status --json <session_id>`, that opens the store read-only and prints AC/phase progress. External schedulers could then poll cheaply.
- Or document the correct read-only DSN for the WAL database (e.g. `file:...ouroboros.db?mode=ro&immutable=1` semantics / proper WAL reader settings) so `sqlite3` and scripts can tail progress without disturbing the writer.

## Why it matters

Long-running `start_execute_seed` jobs are the case where you most want a cheap external "how far along is it?" check — for a status line, a dashboard, or a scheduler deciding when to wake. Today the only progress source is MCP, which couples observability to the (expensive) conversational client.

Q00:
## Summary
- Add `ouroboros job events JOB_ID --since <cursor> --limit <n>` for low-cost external job polling.
- Open `ouroboros.db` with `EventStore(read_only=True)` so the command does not create schema, write WAL/checkpoint state, or require MCP.
- Return cursor-paged JSON containing job aggregate events and the next cursor for dashboards/schedulers.

## RCA
The EventStore already had a true read-only SQLite mode, but the public job CLI only exposed MCP-style status/wait/result handlers that open the default store path normally. External monitors therefore had no documented, scriptable read-only path and were pushed toward either direct `sqlite3` reads or token-heavy MCP polling.

## Tests
- `uv run pytest tests/unit/cli/test_job_command.py -q`
- `uv run ruff check src/ouroboros/cli/commands/job.py tests/unit/cli/test_job_command.py docs/cli-reference.md`
- `uv run ruff format --check src/ouroboros/cli/commands/job.py tests/unit/cli/test_job_command.py`

Closes #1424

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -91,7 +91,14 @@ CLI users wait and retrieve with the standard job surfaces:
 ouroboros job status JOB_ID
 ouroboros job wait JOB_ID
 ouroboros job result JOB_ID
+ouroboros job events JOB_ID --since 0 --limit 100
 ```
+
+`ouroboros job events` is the low-cost external observability surface for
+dashboards and schedulers. It opens `~/.ouroboros/ouroboros.db` read-only,
+does not create schema or write WAL/checkpoint state, and prints cursor-paged
+JSON for the job aggregate. Pass the returned `cursor` back as `--since` on
+the next poll.
 
 MCP clients use the matching job tools:
 
--- a/src/ouroboros/cli/commands/job.py
+++ b/src/ouroboros/cli/commands/job.py
@@ -3,13 +3,18 @@
 from __future__ import annotations
 
 import asyncio
+import json
+import os
+from pathlib import Path
 from typing import Annotated
 
 import typer
 
 from ouroboros.cli.formatters.panels import print_error
+from ouroboros.events.base import BaseEvent
 from ouroboros.mcp.tools.job_handlers import JobResultHandler, JobStatusHandler, JobWaitHandler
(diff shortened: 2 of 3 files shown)


Q00:
CI is green and local verification is documented in the PR body. Ready for review.

Issue #1468, opened 2026-06-15
[Task] Capability-matrix per-call effort actuator across all runtime adapters (#1405/#1404)


Q00:
Parent: #1465 · RFC: [#1405 spend-actuator](https://github.com/Q00/ouroboros/pull/1405) + [#1404 spend-estimator](https://github.com/Q00/ouroboros/pull/1404) · Discussion #1384

## Scope
Wire the **actuator** uniformly across **all** runtime adapters via a capability matrix. The lever is **reasoning effort, not model switching**. Apply **one notch lower** effort to verified decomposed children (hard floor = low).

- **Routable** (per-call knob exists) → ENFORCE: `anthropic_adapter` (`output_config.effort`) + `litellm_adapter` (`reasoning_effort` kwarg) already plumbed; **codex newly wired** via `codex exec -c model_reasoning_effort=<level>`; extend the same pattern to gjc / goose / opencode / copilot / pi as each knob allows.
- **Advise-fallback** (no per-call knob, e.g. ourocode ACP) → hand effort down in the spec, record `effort_mode = advised`, exclude from the strict proof.
- Behind a flag; **parity tests required** before merge.

## Acceptance Criteria
- [ ] Capability matrix records per-adapter effort-knob availability
- [ ] Enforced where a knob exists (codex newly wired), advised-and-recorded where not
- [ ] One-notch-lower applied only to verified-MECE children; floor = low

## Test Checklist
- [ ] A high-stakes/uncertain estimate escalates effort (never cheapens)
- [ ] CLI runtime with no knob falls back to advise and records it
- [ ] Parity test: behavior change is byte/event-order safe behind the flag
- [ ] Monotonic: higher difficulty/stakes never maps to lower effort

## Pass Criteria
Effort is enforced per-AC on routable adapters and cleanly advised+excluded elsewhere, feeding the triad's `effort_level` / `effort_mode`.

---
*Part of [Epic] #1465 | Generated by [Ouroboros](https://github.com/Q00/ouroboros) via `ooo publish`*

Q00:
## Design gate resolved — accepted implementation slice (#1473–#1479)

Per the SSOT #961 escape hatch (*"design-gated unless maintainers split an accepted implementation slice"*), I'm accepting the **#1473–#1479 effort stack** as the concrete implementation slice of this task. The capability-matrix per-call effort actuator described here is now realized on the live path:

| This task's scope | Delivered by |
|---|---|
[...]

Issue #1465, opened 2026-06-15
[Epic] Frugality you can prove — decompose → lower effort → still grounded (RLM control loop)


Q00:
> [!IMPORTANT]
> This Q00-authored epic is back in design/protocol reconciliation. Do not schedule or spend a paid real-runtime cohort from the unmerged protocol-v2 claims recorded in historical comments.

## Current authoritative state

- Current `main` still declares frugality proof protocol version 1 and has no `contract_partition_v1` implementation or accepted `frugality-proof` experiment command.
- PR #1763 shipped verified-MECE bounce-only decomposition, while #1600 supplied the current model-tier routing/proof substrate.
- Q00's reconciliation comment notes that the epic's framing is stale relative to the live routing code and requires a design decision before remaining work is picked up.
- Q00-authored PRs #1846 and #1847 measure focused evolve versus full-graph behavior. That is a different hypothesis from lower-model-tier child execution and cannot silently substitute for this epic's cohort.

## Ordered next steps

- [ ] Reconcile this issue, the frugality RFCs, current `main`, #1763, routing observations, and the outcome of #1846/#1847 into one maintainer-approved protocol.
- [ ] Decide explicitly whether the original lower-model-tier hypothesis is retained, redesigned, or rejected.
- [ ] If retained, merge a versioned admission/accounting/isolation implementation and publish the exact executable command, runtime, budget, artifacts, and owner.
- [ ] Only then activate a paid cohort: at least three real runs and the approved sample threshold, with immutable predeclared PASS/FAIL/INSUFFICIENT gates.
- [ ] Archive the evidence and close by explicit accept/reject/redesign decision rather than an unbounded rerun queue.

No implementation or empirical claim is being marked complete by this reconciliation.

---

## Goal

Build a **falsifiable, minimal closed-loop control system** that *proves* token-frugality of mixed reasoning-effort execution instead of asserting it.

**The RLM-inspired hypothesis:** if work is decomposed well (verified-MECE), each child AC becomes easier for the model, so it can run at **one notch lower reasoning-effort** and stay **frugal without losing grounding**. We prove this **deterministically — no LLM self-judgment** — via a per-AC triad:

> **triad = (token_spend × effort_level × TraceGuard verdict)**, measured against a shadow-replay parent-effort baseline.

This Epic **fuses three RFC themes into one control loop** (they are not three separate features):

| Control-loop role | RFC | Discussion |
|---|---|---|
[...]

PR #1472, opened 2026-06-17
feat(orchestrator): Agent-OS reasoning-effort contract + deterministic frugality-proof gate

Q00:
Implements the effort-enforcement foundation **and** the deterministic proof gate for Epic #1465 ("Frugality you can prove") / Task #1468, per [RFC #1405](https://github.com/Q00/ouroboros/pull/1405).

## Problem
Ouroboros decides an abstract reasoning-effort level per unit of work, but it reached **no agent runtime** — it was wired only into LLM-only provider adapters. So "enforced effort" was unrealizable for code tasks and the frugality proof's *enforced rows* were always empty.

## 1. The capability contract
- `RuntimeCapabilities.reasoning_effort_support` (defaults `IGNORED`; opt-in to `NATIVE`).
- **Enforcing runtimes (verified via each CLI/SDK):** Claude (Agent SDK `effort`), Codex (`-c model_reasoning_effort`), **Copilot** (`--reasoning-effort`). Allow-listed.
- **Advised (declared IGNORED with the verified reason in code):** Gemini, Goose, Hermes (global config only, no per-call override).
- `effort_routing.decide_effort` / `resolve_execute_effort` — the single pure policy: a decomposed child runs one notch lower (floored at "low"); mode is "enforced" only when the runtime declared NATIVE.

## 2. Routed at every live call site
Effort is routed wherever `execute_task` is called: `ParallelACExecutor` (multi-AC), the **runner's own single-AC and resume paths**, and the **coordinator** review pass — a live run proved single-AC executions previously skipped effort entirely. Each AC path emits `execution.ac.effort_routed` (effort_level × effort_mode × backend × ac_id).

## 3. Activation
`OrchestratorConfig.reasoning_effort` + `get_agent_reasoning_effort()` (env `OUROBOROS_AGENT_REASONING_EFFORT` → config → None). Default None ⇒ **dormant, zero behavior change**.

## 4. Deterministic frugality-proof gate
`frugality_proof.py` — the seed's FrugalityProofTriad as a pure, LLM-free judge. `assemble_triads()` joins the per-axis events by `ac_id`; `evaluate_proof()` computes PASS/FAIL in the seed's order: grounding regression is a **per-AC veto**, then sample sufficiency (≥20 triads / ≥3 runs), then aggregate token reduction (≥10%). A row counts only when effort was enforced, decomposition trustworthy, and all axes present. The effort axis is produced today; the token / grounding / shadow-replay event-type contract is fixed so those producers (seed AC2/AC4/AC5) have a precise target. Until wired, the gate returns `INSUFFICIENT_DATA` rather than asserting an unproven hypothesis.

## Regression fixed
`CopilotCliRuntime`/`GeminiCLIRuntime`/`GooseCliRuntime` subclass `CodexCliRuntime` and override `_build_command`; once Codex forwarded `reasoning_effort` they would have raised `TypeError`. Each override now accepts it.

## Verified
- Unit: effort policy, per-runtime enforcement, queryable event, helper, the proof gate (all branches).
- **Live (codex):** `reasoning_effort=high` → `execution.ac.effort_routed level=high mode=enforced` on both the parallel and runner paths; the proof machine reads those real events and honestly returns `INSUFFICIENT_DATA` (only the effort axis is produced).
- 2115+ orchestrator tests green.


Q00:
Split into a 6-PR stack, one per commit, for easier review: #1473 → #1474 → #1475 → #1476 → #1477 → #1478. Closing this combined PR.

PR #1476, opened 2026-06-17
feat(orchestrator): enforce effort on Copilot; declare rest advised (4/6)

Q00:
Copilot CLI `--reasoning-effort` joins Claude/Codex as NATIVE; Gemini/Goose/Hermes declared IGNORED (advised) with the verified reason. Fixes a TypeError regression: the three CodexCliRuntime subclasses' `_build_command` overrides now accept reasoning_effort.

---
**Stacked PR 4/6** for Epic #1465 / Task #1468 (RFC #1405). Base: `effort/03-event`. Review/merge in order.
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1473, opened 2026-06-17
feat(orchestrator): reasoning-effort capability contract (1/6)

Q00:
Adds `RuntimeCapabilities.reasoning_effort_support` (defaults IGNORED; opt-in NATIVE) and wires the Claude (Agent SDK `effort`) and Codex (`-c model_reasoning_effort`) runtimes to enforce it, plus the pure `effort_routing.decide_effort` policy and the ParallelACExecutor consumer. Dormant by default.

---
**Stacked PR 1/6** for Epic #1465 / Task #1468 (RFC #1405). Base: `main`. Review/merge in order.
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1478, opened 2026-06-17
feat(orchestrator): deterministic frugality-proof gate (6/6)

Q00:
`frugality_proof.py`: pure LLM-free triad assembler + PASS/FAIL gate (grounding per-AC veto → sample sufficiency → ≥10% reduction). Fixes the token/grounding/shadow-replay event contract; returns INSUFFICIENT_DATA until those producers are wired.

---
**Stacked PR 6/6** for Epic #1465 / Task #1468 (RFC #1405). Base: `effort/05-callsites`. Review/merge in order.

### DIFF
--- a/src/ouroboros/orchestrator/frugality_proof.py
+++ b/src/ouroboros/orchestrator/frugality_proof.py
@@ -126,19 +126,29 @@ def _event_data(event: object) -> Mapping:
 
 
 def assemble_triads(events: Iterable[object]) -> list[FrugalityTriadRow]:
-    """Join the per-axis events into one triad row per ``ac_id``.
+    """Join the per-axis events into one triad row per ``(run, ac_id)``.
 
     Accepts events as mappings or objects exposing ``type``/``event_type`` and
-    ``data``/``payload``. Unknown event types are ignored. Rows are keyed by
-    ``ac_id``; an event without an ``ac_id`` cannot be correlated and is skipped.
+    ``data``/``payload``. Unknown event types are ignored. An event without an
+    ``ac_id`` cannot be correlated and is skipped.
+
+    Rows are keyed by ``(run, ac_id)`` — **not** ``ac_id`` alone — because the proof
+    spans runs (``min_runs``) and the same logical AC id recurs every run. Keying by
+    ``ac_id`` only would let a later run's events overwrite an earlier run's in the
+    same slot, collapsing valid cross-run evidence to the last run. The run anchor is
+    each event's ``seed_run_id`` (falling back to ``execution_id``); a logical AC
+    therefore yields one row per run it ran in. Events carrying no run anchor share a
+    single implicit run — the original single-run behavior, preserved.
     """
-    acc: dict[str, dict] = {}
+    acc: dict[tuple[str | None, str], dict] = {}
 
     def slot(data: Mapping) -> dict | None:
         ac_id = data.get("ac_id")
         if not ac_id:
             return None
-        return acc.setdefault(str(ac_id), {"ac_id": str(ac_id)})
+        run = data.get("seed_run_id") or data.get("execution_id")
+        run_key = str(run) if run is not None else None
+        return acc.setdefault((run_key, str(ac_id)), {"ac_id": str(ac_id), "seed_run_id": run_key})
 
     for event in events:
         etype = _event_type(event)
@@ -152,8 +162,8 @@ def assemble_triads(events: Iterable[object]) -> list[FrugalityTriadRow]:
             row["is_decomposed_child"] = bool(data.get("is_decomposed_child", False))
             if data.get("parent_effort") is not None:
                 row["parent_effort"] = data.get("parent_effort")
-            if data.get("seed_run_id") is not None:
-                row["seed_run_id"] = data.get("seed_run_id")
+            # seed_run_id is established as part of the row key (see ``slot``), so it
+            # is not re-read here.
         elif etype == EVENT_TOKEN_ATTRIBUTION:
             row = slot(data)
             if row is None:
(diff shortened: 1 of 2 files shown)


Q00:
Roadmap gate: this PR references the frugality/RLM lane (#1465 / #1468 / #1405), whose canonical issues are still `needs-design` in the #961 AgentOS SSOT. Because this run is authenticated as the PR author, GitHub does not allow a `CHANGES_REQUESTED` review; applying `blocked-by-tier` and leaving this gate comment instead. Please do not land this implementation stack until the design gate is resolved or #961 is updated by @Q00 to explicitly allow it.

_Posted by [agentos-roadmap-warden](https://github.com/Q00/ouroboros/blob/main/agents/agentos-roadmap-warden.md) — bot. Reply with `/warden ignore` to suppress further comments on this thread._

Q00:
## Pushed: BLOCKING run-identity fix resolved

`assemble_triads()` now keys the accumulator by **`(run, ac_id)`** instead of `ac_id` alone, where the run anchor is each event's `seed_run_id` (falling back to `execution_id`). Cross-run evidence is no longer overwritten by the last run.

Your exact repro now holds: 2 ACs × 3 runs → **6 rows**, `runs=3`, not 2 rows / `runs=1`. Added a regression test (`test_same_ac_id_across_runs_stays_distinct`) plus an `execution_id`-anchor test. Events with no run anchor still share one implicit run (single-run behavior preserved).

CI red was a single stale test (`test_claude_declares_full_capabilities`), fixed at the base of the stack (#1473). Tier gate lifted (accepted slice of #1465/#1468).

Q00:
Addressed the last review (`a08a35d0`):

**Blocker — `counts_in_proof` ignores `is_decomposed_child`:** Fixed. `counts_in_proof` now requires `is_decomposed_child`, so a fully-measured **top-level** AC row (and the whole-seed direct-runner path) can no longer satisfy the gate. This matches the hypothesis exactly — only *decomposed children* running at lower effort than their parent are admissible; a top-level unit has no parent effort to lower from and no meaningful shadow-replay baseline.
- Row-level regression: `test_top_level_non_decomposed_row_never_counts` (fully measured, but `is_decomposed_child=False` → does not count).
- `evaluate_proof` regression reproducing your exact repro: `test_top_level_rows_alone_never_pass` — 21 fully-measured `is_decomposed_child=False` rows now return `INSUFFICIENT_DATA` with `counted_rows == 0`, not `PASS`.
- Updated the module + `counts_in_proof` docstrings to state the decomposed-child requirement.

The earlier run-identity `(run, ac_id)` keying fix is unchanged and still in place.

**Design/roadmap gate:** Resolved at the #961 SSOT (see #1476) — #1473–#1479 is the accepted #1468 implementation slice and is no longer design-gated.

Q00:
Addressed the latest review (`8791adae`):

**Blocker — `has_all_axes` accepts malformed `token_spend`:** Fixed. A row now needs a `token_spend` that is a **finite, non-negative** number and a `baseline_token_spend` that is **finite and strictly positive**, validated through a shared `_finite_number` guard that rejects `None`, `bool`, non-numeric types, and NaN/inf. A negative spend (your repro: `token_spend=-1.0` → 101% reduction → PASS) is now treated as a missing measurement, so the row is excluded. Zero spend stays valid (a child that cost nothing is maximally frugal).
- Regressions added: `test_negative_token_spend_row_does_not_count`, `test_non_finite_token_spend_row_does_not_count` (NaN/±inf), `test_zero_token_spend_is_valid`, and your exact repro `test_negative_token_spend_never_passes` — 21 decomposed enforced rows with `token_spend=-1.0` now return `INSUFFICIENT_DATA` with `counted_rows == 0`, not `PASS`.

**Follow-up (Medium) — validate all measurement fields at admission:** This `_finite_number` guard covers the numeric axes (token + baseline). `grounding_regression` is already coerced to `bool` in `assemble_triads`, and `effort_mode`/`is_decomposed_child` are checked by `is_enforced`/`counts_in_proof`, so the malformed-payload surface the follow-up describes is now closed for the axes that gate a PASS.

Q00:
Addressed the latest review (`dd43c8a7`):

**Blocker — `has_all_axes` admits a row with no real grounding measurement:** Fixed. The grounding axis now requires the actual TraceGuard output the contract defines, not just a defaulted flag: a non-empty `traceguard_verdict` AND a finite `unsupported_claim_rate` in `[0, 1]`, in addition to `grounding_regression is not None`. Your repro (21 enforced decomposed rows with `traceguard_verdict=None` / `unsupported_claim_rate=None` but `grounding_regression=False`) now yields `traceguard_verdict` absent → `has_all_axes=False` → excluded → `INSUFFICIENT_DATA`, not PASS.
- Regressions added: `test_missing_grounding_measurement_does_not_count`, `test_out_of_range_claim_rate_does_not_count` (rate < 0, > 1, NaN, inf), and `test_grounding_flag_without_measurement_never_passes` (your exact 21-row repro → `counted_rows == 0`).

This closes the grounding admission gap so a defaulted/malformed future producer cannot assert "no grounding loss" without the measurement, consistent with the token/baseline finite-value hardening in the prior commit.

Q00:
Addressed the latest review (`10ab6f97`):

**Blocker — truthy admission-boolean parsing:** Fixed. `assemble_triads` now parses `is_decomposed_child`, `decomposition_trustworthy`, and `grounding_regression` through a `_strict_bool` guard that accepts only real booleans. A malformed string like `"false"` no longer coerces to `True` via `bool("false")`:
- `is_decomposed_child` / `decomposition_trustworthy` → fail safe to `False` (excludes the row).
- `grounding_regression` → stays unset (`None`) so `has_all_axes` excludes the unmeasured row.

Your repro (21 rows with `is_decomposed_child="false"`) now returns `INSUFFICIENT_DATA` with `counted_rows == 0`.
- Regressions added through the public event-mapping path: `test_string_is_decomposed_child_does_not_truthy_admit`, `test_string_decomposition_trustworthy_does_not_truthy_admit`, `test_string_grounding_regression_stays_unmeasured`, and `test_string_boolean_payloads_never_pass`.

This completes the malformed-telemetry hardening of the admission boundary — numeric (token/baseline), grounding (verdict + rate), and now boolean flags all reject malformed payloads as missing rather than admitting them.

PR #1485, opened 2026-06-19
feat(auto): AI answer refiner drives interview ambiguity to convergence

Q00:
## Problem

`ooo auto` reliably **blocked at the seed QA gate**. Manual testing (all stages on codex) showed it was not a hang — the seed QA gate is `asyncio.wait_for`-bounded — but a structural block:

- **Root cause**: `AutoAnswerer` is a deterministic **template** engine. Its generic answers ("Stable text output suitable for verification", …) score low on `constraint_clarity`/`success_criteria_clarity`, so the LLM ambiguity scorer plateaus around **~0.5** and never reaches the `0.20` interview-closure threshold. The seed QA gate then honestly blocks the under-specified seed.

## Primary fix — AI answer refiner (f)

`AutoInterviewDriver._refine_answer`: when an `answer_refiner` is wired, a **generic** `CONSERVATIVE_DEFAULT`/`ASSUMPTION` answer (no blocker) is replaced with a concrete, goal-specific answer from the interview LLM, written into **both** `answer.text` (→ the transcript the backend re-scores) and the ledger entry value (→ Seed content). The deterministic answerer still owns **all** routing/safety (blocker detection, intent, source tagging) — only the *text* of already-safe generic answers is upgraded.

- `adapters.LLMAnswerRefiner(create_llm_adapter(use_case="interview"))` via `build_answer_refiner()`; wired in CLI `auto.py` + MCP `auto_handler.py` (best-effort; `None` falls back to template-only).

### E2E proof (was `blocked`, now `complete`)
```
answer_refined ×3  →  ambiguity 0.5 → 0.10  (below the 0.20 threshold)
→ interview closes → seed grade A → seed QA passes → Status: complete
```

## Supporting changes

- **(a)** Disable the unsafe-context veto on safe-default closure (empty `_UNSAFE_CONTEXT_PATTERNS`). The `legal/medical judgment` bank literally listed `contract`, so a CSV→JSON tool was blocked because an answer said "data contract". Owner chose global removal; the matcher + lateral-escalation machinery is kept live and tests re-inject the historical bank via `conftest._legacy_unsafe_bank`.
- **(b)** Wire `lateral_thinker` into the CLI auto path (parity with MCP) — the Issue #1248 safe-default escalation now also fires from the CLI.
[...]

PR #1487, opened 2026-06-19
fix(auto): refine every interview answer section so auto mode converges

Q00:
## Problem

`ooo auto "아주 간단한 todo CLI 만들어줘"` stalled in the interview at the
first product-decision question:

> 기존 관례가 없을 때는 할 일을 `~/.todo-cli.json` 같은 사용자별 로컬 JSON 파일에 저장하고,
> `done`/`delete`는 `list`에 표시되는 안정적인 숫자 ID로 지정하는 방식으로 확정해도 될까요?

In **auto mode there is no human** to answer, so the pipeline must auto-decide
and converge. Instead it hit `interview_phase_deadline` and produced a
`degraded: true` Seed (`recovery_reason: interview_phase_deadline`) — and never
executed, so no product was built.

## Root cause

The AI answer refiner (#1485) — the component that replaces the deterministic
answerer's generic placeholder with a concrete, goal-specific decision and
thereby drives ambiguity down — **never fired**. `_refine_answer` bailed out on:

```python
or len(answer.ledger_updates) > 1
```

But **every** real `AutoAnswerer` route emits ≥2 ledger updates:

| route | sections |
|---|---|
| `_product_behavior_answer` | constraints + acceptance_criteria |
| `_verification_answer` | verification_plan + acceptance_criteria |
| `_io_actor_answer` | actors + inputs + outputs |
| `_default_answer` | constraints + failure_modes |

So the guard skipped all of them: the backend only ever saw generic templates,
ambiguity never reached the 0.20 convergence threshold, and the phase deadline
fired. As a side effect, the generic placeholder
(`"Preserve the product behavior requested by the interview question: <raw question>"`)
was never replaced, leaking the raw question text into the Seed's
`constraints` / `acceptance_criteria`.

## Fix

Refine **every** ledger-update section. The refiner is already section-aware
(`refiner(goal, question, section, generic_value) -> concrete`), so each entry
gets its own concrete value — the transcript/ledger desync the old single-update
guard was avoiding **cannot occur** (we replace each entry with its own refined
value, not one shared string, and rebuild `answer.text` from those same values).

- Per-section calls run **concurrently** (`asyncio.gather`), so wall-clock stays
  ~one refiner call → phase deadline respected.
- **Partial failure is safe**: a section whose refiner call fails or returns
  blank keeps its original (in-sync) value; the rest are upgraded. If nothing
  refines, the original answer is returned unchanged.
- Fixing the refiner also removes the raw-question pollution, since the generic
  placeholder values are now replaced with concrete decisions.

## Tests

- Rewrote `test_multi_section_answer_is_not_refined` →
  `test_multi_section_answer_is_refined_per_section` (asserts each section gets
  its own concrete value + transcript rebuilt from them).
- Added `test_partial_section_refine_keeps_failed_section_in_sync`.
- `uv run pytest tests/unit/auto/test_interview_answer_refiner.py` → 8 passed.
- `uv run pytest tests/unit/auto/` → 1175 passed; 4 pre-existing `test_surface`
  failures (codex-config-leak env issue, identical on `main` — unrelated).
- ruff + mypy clean on changed files.

## Follow-ups (not in this PR)

- The interview-deadline recovery path (`partial_seed_from_evidence`) still
  injects unresolved-slot text into constraints; worth a separate hardening pass.
- Stagnation-lateral can consume the full phase timeout; consider a smaller
  per-intervention budget.

## R-run comparison

| Metric | Baseline | This PR | Ratio |
|---|---|---|---|
| Rounds completed in 600s | N/A | N/A | n/a |

**Why N/A (honest disclosure, not a substrate-only claim):** a canonical R-run
(`OUROBOROS_RUN_CANONICAL=1 uv run pytest tests/canonical/ -k cli-todo`) needs an
LLM-backed interview/refiner backend that is not available in this authoring
environment, so I could not capture real per-round numbers and will not fabricate
them. Expected per-round impact is neutral-to-positive: the per-section refiner
calls run **concurrently** (`asyncio.gather`) so they add ~one refiner call of
wall-clock per answered round, and prior to this change the refiner never fired at
all — the interview burned rounds on un-converging generic answers until it hit the
phase deadline. Driving ambiguity down should **reduce total rounds to terminal**,
not add latency. A maintainer with canonical-harness access should capture a real
R-run before merge if §I5 numbers are required.

### DIFF
--- a/src/ouroboros/auto/interview_driver.py
+++ b/src/ouroboros/auto/interview_driver.py
@@ -1473,23 +1473,31 @@ class AutoInterviewDriver:
         )
 
         refined_updates: list[tuple[str, LedgerEntry]] = []
-        concrete_parts: list[str] = []
+        transcript_parts: list[str] = []
+        refined_count = 0
         for (section, entry), concrete in zip(answer.ledger_updates, results, strict=True):
+            # Rebuild the transcript from the FINAL value of every section —
+            # refined where the call succeeded, original otherwise — so the text
+            # the backend acknowledges always describes the same committed content
+            # as the persisted ledger entries. A partial failure must never leave
+            # a ledger entry that is absent from the acknowledged transcript.
             if concrete:
                 refined_updates.append((section, replace(entry, value=concrete)))
-                concrete_parts.append(concrete)
+                transcript_parts.append(concrete)
+                refined_count += 1
             else:
                 refined_updates.append((section, entry))
-        if not concrete_parts:
+                transcript_parts.append(entry.value)
+        if not refined_count:
             return answer
         log.info(
             "auto.interview.answer_refined",
             auto_session_id=state.auto_session_id,
             sections=[section for section, _ in answer.ledger_updates],
-            refined_count=len(concrete_parts),
+            refined_count=refined_count,
             source=answer.source.value,
         )
-        return replace(answer, text=" ".join(concrete_parts), ledger_updates=refined_updates)
+        return replace(answer, text=" ".join(transcript_parts), ledger_updates=refined_updates)
 
     async def _maybe_concretize_on_stagnation(
         self,
--- a/src/ouroboros/auto/interview_driver.py
+++ b/src/ouroboros/auto/interview_driver.py
@@ -31,7 +31,7 @@ from ouroboros.auto.lateral_routing import (
     select_persona_for_qa_failure,
     select_persona_for_safe_default_block,
 )
-from ouroboros.auto.ledger import LedgerSource, LedgerStatus, SeedDraftLedger
+from ouroboros.auto.ledger import LedgerEntry, LedgerSource, LedgerStatus, SeedDraftLedger
 from ouroboros.auto.progress import AutoProgressCallback, AutoProgressEvent
 from ouroboros.auto.repo_context import repo_auto_answer_context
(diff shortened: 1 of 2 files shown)


Q00:
@ouroboros-agent Addressed the BLOCKING finding (partial-refinement transcript/ledger desync) in `6e7d0815`.

**Fix**: `_refine_answer` now rebuilds `answer.text` from the **final value of every section** — the refined value where the per-section refiner call succeeded, the original entry value where it failed/blanked — so the acknowledged transcript (`prefixed_text`, sent verbatim to `backend.answer` and recorded by `apply`) always covers every persisted ledger entry. A partial failure can no longer leave a ledger entry absent from the transcript.

**Tests**:
- Rewrote the partial-refine test to assert the failed section's value remains in the transcript (`test_partial_section_refine_keeps_transcript_and_ledger_in_sync`).
- Added the boundary test you asked for: `test_refined_transcript_covers_every_applied_ledger_entry` drives the refine → `apply` seam and asserts `prefixed_text` (what `backend.answer` receives) contains every applied ledger entry value, including the fallback section.

All CI green (perf-budget now satisfied; 3.12/3.13/3.14 tests pass). Please re-review at HEAD `6e7d0815`.

PR #1503, opened 2026-06-22
fix(providers): avoid FrozenInstanceError when normalizing JSON in ClaudeCodeAdapter

hanhosung:
## Summary

`ClaudeCodeAdapter._normalize_json_content` assigned `result.value.content` in place, but `CompletionResponse` is a **frozen** dataclass (`@dataclass(frozen=True, slots=True)`). So whenever a JSON `response_format` response needs extraction from surrounding prose/fences, the assignment raises:

```
File ".../providers/claude_code_adapter.py", line 317, in _normalize_json_content
    result.value.content = extracted
dataclasses.FrozenInstanceError: cannot assign to field 'content'
```

This makes **every semantic evaluation (Stage 2 of the evaluation pipeline) crash** when using the `claude_code` backend — `ouroboros_evaluate` / `ouroboros_qa` fail with `cannot assign to field 'content'`, and the pipeline only ever reaches Stage 1.

## Fix

Build a new instance via `dataclasses.replace(...)` instead of mutating the frozen dataclass:

```python
return Result.ok(replace(result.value, content=extracted))
```

## Why existing tests missed it

The JSON tests in `test_claude_code_adapter.py` used `MagicMock` responses, whose attributes are freely assignable — so the in-place mutation never raised and the bug was invisible. This PR:

- converts those tests to use **real `CompletionResponse`** objects via a small `_ok_response` helper, and
- adds a regression test (`test_json_normalization_on_real_frozen_response`) that drives the frozen dataclass directly. It **fails before this fix** with `FrozenInstanceError` and **passes after**.

## Verification

```
uv run pytest tests/unit/providers/test_claude_code_adapter.py -q   # 18 passed
uv run pytest tests/unit/providers/ -q                              # 119 passed, 4 skipped
uv run ruff check src/ouroboros/providers/claude_code_adapter.py tests/unit/providers/test_claude_code_adapter.py  # clean
```

Reproduced originally while running the full three-stage evaluation pipeline against an external project via the `claude_code` backend; after this fix the pipeline completes Stage 2 (semantic) successfully.

### DIFF
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -314,8 +314,11 @@ class ClaudeCodeAdapter:
             return result
         extracted = extract_json_payload(result.value.content)
         if extracted:
-            result.value.content = extracted
-            return result
+            # CompletionResponse is a frozen dataclass — build a new instance
+            # instead of mutating (mutation raises FrozenInstanceError).
+            from dataclasses import replace
+
+            return Result.ok(replace(result.value, content=extracted))
         return None
 
     async def _enforce_json(
--- a/tests/unit/providers/test_claude_code_adapter.py
+++ b/tests/unit/providers/test_claude_code_adapter.py
@@ -11,12 +11,31 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
+from ouroboros.core.types import Result
 from ouroboros.providers.base import (
     CompletionConfig,
+    CompletionResponse,
     Message,
     MessageRole,
+    UsageInfo,
 )
 from ouroboros.providers.claude_code_adapter import ClaudeCodeAdapter
+
+
+def _ok_response(content: str) -> Result[CompletionResponse, object]:
+    """A successful Result wrapping a real (frozen) CompletionResponse.
+
+    Use this instead of a MagicMock so tests exercise the real dataclass — a
+    MagicMock's attributes are freely assignable and would hide bugs like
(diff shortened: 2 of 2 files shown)


Q00:
Warden triage: closing this PR as **superseded by current `main`**, not because the report/fix was invalid.

What I inspected:
- PR #1503 body and latest review state: the FrozenInstanceError diagnosis is correct and the bot review is approved.
- The live `main` branch already contains the same durable fix in `src/ouroboros/providers/claude_code_adapter.py`: `_normalize_json_content()` rebuilds the frozen `CompletionResponse` with `dataclasses.replace(...)` instead of mutating `result.value.content`, and deep-copies `raw_response` while doing so.
- `main` also already has stronger regression coverage in `tests/unit/providers/test_claude_code_adapter.py`, including:
  - `test_normalize_json_content_rebuilds_frozen_completion_response`
  - `test_json_normalization_rebuilds_response_without_aliasing_raw_response`

Why not land this PR now:
- The PR is currently merge-conflicted against `main` in the exact same implementation/test areas.
- Resolving the conflicts by preserving current `main` truth leaves no useful additional behavior beyond what is already present.
- Keeping the PR open would create duplicate/superseded queue noise for Ouroboros.

Future acceptable work:
- If there is a remaining `claude_code` JSON-normalization case that `main` does not cover, please open a fresh narrow PR against current `main` with a failing regression test for that distinct case.

Thanks for catching the frozen dataclass mutation bug — the fix direction was right; it has simply already landed through the current mainline.

PR #1532, opened 2026-06-30
fix(capabilities): deliver tool-discovery convention via MCP server instructions

Q00:
## Why

The ouroboros capability matrix should be the **ubiquitous language** (SSOT) that every runtime understands. But Claude had **no delivery path** for it — only codex/opencode/gemini receive instruction artifacts (AGENTS.md etc), while Claude relies purely on per-skill `SKILL.md` prose. As a result the "how to load a deferred MCP tool" convention leaked into ~10 shared `SKILL.md` files as a **Claude-specific `ToolSearch`** tool name, which every *non-Claude* runtime then needed a band-aid to ignore (`capabilities.py` codex `call_mcp`: "Do not rely on Claude-specific `ToolSearch` names").

## What

Make the capability matrix the SSOT and deliver the conventions to **every MCP client (Claude included)** via the standard MCP server `instructions` field — the one channel all MCP hosts read at session start.

- **L1 — abstract vocabulary**: `ToolDiscoveryMechanism` enum (`DEFERRED_TOOL_SEARCH` / `NATIVE_RUNTIME_DISCOVERY` / `DIRECT_EXPOSURE`) + `BackendCapability.tool_discovery_mechanism`/`tool_discovery_tool_name` + `resolve_tool_discovery()`. Mirrors the existing `SubagentDispatchMode` abstract/concrete split. Mapping: claude→ToolSearch, codex→native, opencode→direct.
- **L2 — delivery channel**: `render_mcp_server_instructions()` — provider-neutral, **1312 bytes (<2KB host cap)**, front-loaded — wired through `MCPServerAdapter.serve()` → `FastMCP(instructions=...)` and defaulted in `create_ouroboros_server`. Now Claude reads the matrix at session start like every other provider.
- **L3 — de-leak**: `ToolSearch` → "runtime tool discovery" across **10 `SKILL.md` (both trees)**, matching the already-migrated seed/interview template. **Preserved verbatim**: every `+ouroboros X` discovery query, every `ouroboros_*` tool name, the `deferred-schema guard` strings, and the unstuck **fail-loud contract** meaning.
- **L4 — cleanup**: drop the Claude-specific cross-reference from the codex `call_mcp` guidance (now provider-neutral).

## Verification

- New: `resolve_tool_discovery` per-runtime mapping; `render_mcp_server_instructions` neutrality + 2KB budget + key conventions; adapter stores/wires `instructions`; end-to-end `create_ouroboros_server` default-injection (1312 bytes).
- Regression: backends / skills / mcp.server / codex+runtime+hermes artifact tests — **216 passed**; targeted+new — **124 passed**.
- Guard tests `test_skill_artifacts` + `test_codex_artifacts` pass; **zero `ToolSearch` remaining** in shared prose; unstuck contract diff reviewed.
- `ruff format` + `ruff check` clean.

## Notes

- `scripts/deleak_toolsearch.py` is the one-shot, deterministic migration used for L3 (kept as documentation of the canonical replacement vocabulary; safe to drop later).
- The MCP `instructions` field is a standard protocol field, so codex and any MCP-aware host read the **same** guidance → true cross-provider symmetry.


Q00:
## Addressed the blocker — `a1ba1c4c`

**Finding (valid):** `render_mcp_server_instructions()` declared an empty discovery result for an already-exposed tool to be a normal no-op, but the migrated skill bodies still treated "discovery found nothing" as MCP-unavailable → fell back / reported the dispatch surface broken. On direct-exposure or already-loaded runtimes that would skip a callable MCP tool or raise a false "surface broken". Agreed — the skill↔server contract was left half-reconciled.

**Fix:** availability now means **"the tool is callable by its `ouroboros_*` name"** (already exposed **or** loaded via discovery), not "discovery returned a match". Fallback fires only on **genuine absence** (no Ouroboros MCP server).

- `evaluate` / `run` / `status` / `evolve`: `If discovery finds the tool → MCP; if not → Fallback/Path B` → `If the tool is callable (already exposed, or loaded by discovery) → MCP; empty result for an already-exposed tool is expected; Fallback only if genuinely absent`.
- `qa`: MCP-mode/Fallback-mode gates reworded to callability, not discovery-hit.
- `unstuck`: **fail-loud contract preserved** — still stops and reports a broken dispatch surface, but only when the tool is truly not callable (neither exposed nor loadable), no longer on an expected empty discovery result. Both trees.

**Verification:** `test_skill_artifacts` + `test_codex_artifacts` pass (36); guard strings (queries / `ouroboros_*` names / `deferred-schema guard`) intact; zero `ToolSearch` remaining; `test_capabilities` + `test_adapter` green.

The registry direction you flagged as sound (backend-owned discovery capability + MCP server instructions) is unchanged; this only makes the shared skill prose agree with that boundary.

Q00:
## Second-round blocker addressed — `36b0e716`

**Finding (valid, narrowed):** the top-level Path A/B decision was reconciled, but the **deferred-schema guard tail** (“if the load returns no matching tool, switch to fallback”) still routed on discovery emptiness, and the already-migrated **seed/interview** skills — now governed by the new server-instruction contract — kept the same pattern. Agreed: on direct-exposure / already-loaded runtimes an empty discovery is an expected no-op, so these still risked skipping a callable tool.

**Fix:**
- Qualified **every** deferred-schema-guard `no matching tool` fallback with a callability check across `brownfield/evaluate/evolve/interview/pm/run/seed/setup` (both trees): *an empty load for an already-exposed tool is an expected no-op, not absence.*
- Converted the **seed** and **interview** top-level Path A/B decisions to the same callability test used by evaluate/run/status/evolve.
- **Closed the coverage gap** you named: new `test_packaged_skills_gate_fallback_on_callability_not_empty_discovery` asserts no packaged skill (either tree) routes to fallback/Path B on empty discovery alone.

**Non-blocking addressed:** added `ToolDiscoveryMechanism` / `render_mcp_server_instructions` / `resolve_tool_discovery` to the `capabilities.py` module-local `__all__` (verified `from ... import *` now exports them).

**Verification:** `test_skill_artifacts` + `test_codex_artifacts` + `test_capabilities` + `test_adapter` — 125 passed; grep confirms every fallback/Path-B routing sentence is now callability-gated; guard strings, discovery queries, and `ouroboros_*` names intact; zero `ToolSearch` remaining.

PR #1483, opened 2026-06-19
feat(orchestrator): add Grok Build CLI (grok) runtime backend

project820:
## Summary

Adds **Grok Build** (xAI's `grok` CLI) as a first-class **runtime-only** backend. Auth is subscription OAuth (`grok login`, SuperGrok / X Premium+) or `XAI_API_KEY`.

`GrokCliRuntime` extends `CodexCliRuntime` with `grok`'s headless contract: `grok -p <prompt> --output-format streaming-json --permission-mode <mode> [-m …]`, native permission mapping (`acceptEdits`/`bypassPermissions`; `default` coerced for headless), sentinel-model passthrough, and the **verified `thought`/`text`/`end` event schema**. The model catalog is **dynamically listed** via `grok models` (a grok-specific parser handles its bulleted output) so the picker reflects exactly what the user's Grok CLI can call, with `grok-build`/`grok-composer-2.5-fast` as the static fallback. Wired the same way as Antigravity (capability registry, config + env-denylisted path, loader, model catalog, runtime factory, `--runtime` enums). Runtime-only (`supports_llm=False`) for v1.

> 🔀 **Stacked PR — review after the parent merges.** This is the second of three stacked PRs (see #1481). It is branched on `feat/antigravity-backend` (PR #1482), so the diff shown here **includes that PR's commits** until it merges; I'll rebase to a clean grok-only diff afterward. Opened as a **draft** to signal that.
>
> **This PR's own changes (grok only):** `grok_cli_runtime.py`, `model_catalog.py` (grok entry + dynamic listing), `capabilities.py`, `config/{models,loader,__init__}.py` (grok), `backend_limits.py`, `runtime_factory.py` (grok dispatch), `cli/commands/{init,mcp,run}.py` (GROK enum), `docs/runtime-guides/grok.md`, `docs/{cli-reference,runtime-capability-matrix,runtime-guides/skill-capability-guides}.md`, and `tests/unit/{backends/test_model_catalog,config/test_loader,config/test_models,orchestrator/test_grok_cli_runtime}.py`.

## Test plan

- [x] `uv run pytest tests/ -q` passes (offline `uv build` wheel test excepted, as in PR #1482)
- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check src/ tests/` clean
- [x] `uv run mypy src/ouroboros` clean
- [x] New tests: grok runtime (command/permission/event schema), `grok models` parser + dynamic refresh, config/loader/factory/CLI-enum coverage
[...]

Issue #1481, opened 2026-06-19
Add Antigravity (agy) and Grok Build (grok) runtime backends + multi-LLM routing presets


project820:
## Summary

Add two new agentic CLI **runtime backends** to the pluggable `AgentRuntime` layer — **Antigravity** (the `agy` binary) and **Grok Build** (the `grok` binary) — plus one-click **multi-LLM stage-routing presets** in the `ouroboros config` settings GUI.

## Motivation

- **Antigravity is now required for Gemini consumer subscriptions.** Per Google's [transition notice](https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/), on **2026-06-18** the Gemini CLI stopped serving the AI Pro / Ultra / free consumer tiers and migrated those users to the new Antigravity CLI (`agy`); enterprise / API-key Gemini access is unaffected. The existing `gemini` backend shells out to the `gemini` binary, so subscription users currently have no working Google runtime. Antigravity is a separate Go binary (`agy`), not a flag rename, so it needs its own backend.
- **Grok Build CLI is unsupported.** xAI's official `grok` CLI (subscription OAuth via `grok login`, or `XAI_API_KEY`) is a capable terminal coding agent with a structured `--output-format streaming-json` stream, but there is no `grok` backend today.
- **Per-stage multi-LLM routing.** Ouroboros already supports `orchestrator.runtime_profile.stages` (interview / execute / evaluate / reflect) and surfaces it in the settings GUI. Recommended routing presets — built around the *generate ≠ verify* vendor-diversity principle the engine already encodes in `consensus.diversity_required` — make that power one click away.

## Proposed approach

Mirror the existing `gemini` backend end-to-end (capability registry → config schema → loader → model catalog → runtime factory → `--runtime` enums → runtime class → docs → tests). Both backends are **runtime-only** for v1 (`supports_llm=False`): `agy -p` prints plain text and `grok -p` drives an agentic run, so they back the orchestrator runtime but not structured LLM-completion calls (those can still route through `litellm`/OpenRouter for consensus diversity).

Commands are modeled against the **real installed binaries** (`agy` 1.0.10, `grok` 0.2.56), e.g.:

- `agy -p <prompt> --dangerously-skip-permissions [--model …]` — plain-text stdout (no `--output-format`).
[...]

PR #504, opened 2026-04-29
feat(orchestrator): Gemini CLI runtime backend (follow-up to #312)

shaun0927:
## Summary
- Lands the Gemini CLI **runtime** layer that PR #312 attempted but never merged. The Gemini *adapter* is already on `main` via #315; this PR fills the runtime/factory/CLI gap so Gemini works end-to-end as an orchestrator backend.
- Fixes the unresolved blocker the bot reviewer flagged on #312 at HEAD `880658d`: `_convert_event` now surfaces Gemini's terminal `result` event as the final assistant message instead of silently dropping it.
- Wires Gemini through every public CLI surface: `setup --runtime gemini`, `config backend gemini` (with `OUROBOROS_GEMINI_CLI_PATH` / persisted-path fallback), `mcp serve --llm-backend gemini`, `init --llm-backend gemini`.
- Decouples Gemini setup from the `[claude]` MCP extra — Gemini-only installs no longer pull the Claude SDK.

## Why a fresh PR instead of finishing #312
- #312 has 25 days of base drift; GitHub marks it `CONFLICTING`.
- It was opened from @yevheniikravchuk's fork, so we cannot push the rebase + fixes onto its branch directly.
- The branch's last bot review (after a brief APPROVE) re-flagged the dropped `result` event as a blocker. That fix needed to land before merge regardless.

This PR cherry-picks the structurally sound parts of #312 (runtime + normalizer) onto current upstream/main, applies every follow-up the bot reviewer flagged, and credits @yevheniikravchuk via `Co-Authored-By` so attribution stays intact. Once this lands we can close #312 with a thank-you and pointer here.

## What's in
- `src/ouroboros/orchestrator/gemini_cli_runtime.py` — extends `CodexCliRuntime`. `_OUROBOROS_DEPTH` recursion guard + `InputValidator` truncation (matches the #315 adapter pattern). No session resumption (Gemini CLI doesn't expose one).
- `src/ouroboros/providers/gemini_event_normalizer.py` — stateless NDJSON / plain-text normalizer.
- `runtime_factory.py` — `gemini` / `gemini_cli` aliases registered through `get_gemini_cli_path()`; rejection message now lists every supported backend.
- CLI wiring in `mcp.py`, `init.py`, `setup.py`, `config.py` (env-var fallback in `config backend gemini`).
- `docs/runtime-guides/gemini.md` — documents the contract the code actually implements.
- `_setup_gemini` is intentionally lightweight — does **not** require `[claude]` extras.

[...]

PR #1496, opened 2026-06-21
feat(orchestrator): leader-driven external worker pool + native-visibility opt-in

Q00:
## What

A **provider-neutral leader-driven worker pool**: ouroboros is the LEADER and drives an addressable, resumable worker session directly (`codex mcp-server`, `claude -p --resume`) below the `AgentRuntime` seam. A provider becomes a worker pool member by supplying a thin **spawn/resume transport** — the orchestration brain (ParallelExecutor / EventStore / validate_evidence) is untouched.

## Highlights

- **`LeaderDrivenWorkerRuntime`** + `claude_mcp` / `codex_mcp` transports (one runtime, swap the transport)
- **codex_mcp** persistent multi-turn session pool (anyio actor); `codex_session_index`
- **claude fork-from-parent**: worker inherits the host conversation context via `--resume <parent> --fork-session` with a fresh session id → **zero parent-transcript pollution** (live-verified)
- **Worker host-visibility is OPT-IN** (`OUROBOROS_NATIVE_SESSION_INDEX`, default **off**): by default neither floods the native app — codex skips `session_index`, claude runs with `--no-session-persistence`. The web dashboard (stacked PR) is the default worker view. Opt in to persist + fork + `--name "ooo:…"` (codex `ooo:` index entry) for native open/resume.
- **Provider tagging** in events (`orchestrator.session.started.runtime_backend`)
- Unified transient retry (`providers/retry`); `AssignmentMessage` builder (TASK/DELIVERABLE/SCOPE/VERIFY)
- Recursion hardening (depth guard + ouroboros tool/MCP denial) on workers

## Tests

New unit suites: worker_runtime, claude_worker_runtime, codex_mcp_runtime, codex_mcp_session_pool, codex_session_index, subagent_orchestration, assignment, retry + config flag. All green locally. (Pre-existing auto plugin-mode/ralph/interview failures are unrelated — reproduce on clean `main`.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1539, opened 2026-07-02
fix(orchestrator): register runtime-only backends in RuntimeHandle contract

project820:
## What & why

Post-merge review on #1483 (ouroboros-agent, CHANGES_REQUESTED) surfaced a real crash that landed on `main`: the `grok` runtime backend is unusable through generic resume/replay/control paths.

`GrokCliRuntime` subclasses `CodexCliRuntime` and inherits `_build_runtime_handle`, which constructs `RuntimeHandle(backend="grok_cli")` on every execute/resume/control iteration. But `grok_cli` was never added to `_RUNTIME_HANDLE_BACKEND_ALIASES` in `orchestrator/adapter.py`, so `RuntimeHandle.__post_init__` raised:

```
ValueError: Unsupported RuntimeHandle backend selector: grok_cli
```

Reproduced directly:

```python
>>> from ouroboros.orchestrator.adapter import RuntimeHandle
>>> RuntimeHandle(backend="grok_cli")
ValueError: Unsupported RuntimeHandle backend selector: grok_cli
```

## Scope

The same latent gap affects the other Codex/Gemini-derived runtimes that declare a `_cli` handle backend but were never registered, so this fixes all of them to keep the contract complete:

| runtime | `_runtime_handle_backend` | canonical |
|---|---|---|
| `GrokCliRuntime` | `grok_cli` | `grok_cli` |
| `AntigravityCLIRuntime` → `GeminiCLIRuntime` | `antigravity_cli` | `antigravity_cli` |
| `GeminiCLIRuntime` (pre-existing since #504) | `gemini_cli` | `gemini_cli` |

Canonicalizing to the `_cli` form matches each runtime's declared `_runtime_handle_backend` and the existing `codex`/`hermes`/`copilot` precedent. Config-layer spellings (`agy`, `grok_build`, etc.) are resolved to runtime classes before any handle exists and never flow into `RuntimeHandle`, so they are intentionally out of scope for the boundary map.

## Tests

`tests/unit/orchestrator/test_adapter.py`:
- extended the alias-normalization parametrization (construction) for gemini/grok/antigravity;
- added `from_dict` deserialization cases (backend-keyed and legacy provider-keyed);
- added `test_runtime_only_backends_register_in_handle_contract`, which exercises Grok's inherited `_build_runtime_handle` against the real method.

Local: `ruff` / `ruff format --check` / `mypy` clean; `pytest tests/unit/orchestrator/test_adapter.py tests/unit/orchestrator/test_grok_cli_runtime.py` → 102 passed.

Relates to #1481. Addresses the post-merge review on #1483.

### DIFF
--- a/src/ouroboros/orchestrator/adapter.py
+++ b/src/ouroboros/orchestrator/adapter.py
@@ -266,6 +266,12 @@ _RUNTIME_HANDLE_BACKEND_ALIASES = {
     "pi_cli": "pi",
     "gjc": "gjc",
     "gjc_cli": "gjc",
+    "gemini": "gemini_cli",
+    "gemini_cli": "gemini_cli",
+    "grok": "grok_cli",
+    "grok_cli": "grok_cli",
+    "antigravity": "antigravity_cli",
+    "antigravity_cli": "antigravity_cli",
 }
 
 
--- a/tests/unit/orchestrator/test_adapter.py
+++ b/tests/unit/orchestrator/test_adapter.py
@@ -262,6 +262,12 @@ class TestRuntimeHandle:
             ("claude_code", "claude"),
             ("codex", "codex_cli"),
             ("opencode_cli", "opencode"),
+            ("gemini", "gemini_cli"),
+            ("gemini_cli", "gemini_cli"),
+            ("grok", "grok_cli"),
+            ("grok_cli", "grok_cli"),
+            ("antigravity", "antigravity_cli"),
+            ("antigravity_cli", "antigravity_cli"),
         ],
     )
     def test_init_normalizes_legacy_backend_aliases(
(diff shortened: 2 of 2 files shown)


Q00:
Reviewed in depth — **verdict: MERGE** (no blocking findings).

Verified against current main:
- **Root cause, not symptom**: `GrokCliRuntime`/`AntigravityCLIRuntime`/`GeminiCLIRuntime` all declare `_runtime_handle_backend` values that were absent from `_RUNTIME_HANDLE_BACKEND_ALIASES`, so `RuntimeHandle.__post_init__` raised `ValueError: Unsupported RuntimeHandle backend selector` on every generic execute/resume/control iteration. Registering both the short names and the canonical `_cli` forms at the contract itself is the right fix, and matches the existing codex/hermes/copilot precedent (including canonical→self mapping for persisted-handle idempotency).
- **Complete**: every runtime's `_runtime_handle_backend` on main now resolves; #1496's `codex_mcp`/`claude_mcp` were already present and are untouched. Leaving config-layer aliases (`grok_build`, `agy`) out is correct — they never flow into handles.
- **Tests bite**: the parametrized cases and `test_runtime_only_backends_register_in_handle_contract` fail on main without this change (they drive the real inherited `_build_runtime_handle`).
- CI 10/10 green, no conflicts.

Two non-blocking follow-ups worth an issue:
1. This is the second hand-maintained alias table alongside `backends/capabilities.py` (which already declares per-backend name+aliases, with the opposite canonical direction) — that duplication is why grok/gemini were missed in the first place. Deriving the handle map from the capabilities registry would prevent recurrence.
2. A sweep test iterating every runtime subclass's `_runtime_handle_backend` and asserting it resolves would guard the next backend addition.

PR #1540, opened 2026-07-02
chore: delete confirmed dead code

Q00:
## Summary
- Delete the confirmed-dead standalone AC tree renderer and stale workflow display formatter.
- Remove orphaned tests and the example that only exercised the deleted workflow display surface.
- Preserve the live AC-tree HUD failed-status glyph coverage in `test_ac_tree_hud_handler.py`.

## Roadmap Checklist (PR-A)
- [x] All three precondition greps returned clean before deletion.
- [x] Deleted stale bytecode directories, if present.
- [x] Removed tracked dead-code surfaces and orphan tests/examples discovered by reverse-reference review.
- [x] `src/ouroboros/cli/formatters/__init__.py` does not re-export `workflow_display`.

## Validation
- [x] `uv run ruff check .`
- [x] `uv run ruff format --check .`
- [x] `uv run mypy src/`
- [x] `uv run pytest tests/unit/mcp/tools/test_ac_tree_hud_handler.py tests/unit/cli/formatters -q` (`96 passed`)
- [x] Exact-reference scan found no remaining tracked references to deleted modules.
- [x] 5 senior subagent reviews approved before upload.

Full non-e2e pytest was run and is not green due to unrelated existing failures in auto runtime dispatch/start_auto tests plus a transient wheel-build DNS failure. The PR-A diff does not touch those areas; focused and static validation pass.


Q00:
Reviewed with per-deletion reference evidence against current main — **verdict: MERGE** (no blocking findings).

- `cli/formatters/workflow_display.py`: zero live references to `render_workflow_state`/`WorkflowDisplay` — the only consumers are `examples/test_display.py` and its test, both deleted here. `cli/formatters/__init__.py` re-exports only `console`/`OUROBOROS_THEME`, so nothing dangles.
- `mcp/tools/ac_tree_hud_render.py`: confirmed dead twin — the live `ac_tree_hud_handler.py` never imports it and carries its own `_render_tree_lines` + `_MAX_RENDER_DEPTH=3`. Zero references to `render_ac_tree_markdown` anywhere (code, docs, entry points, dynamic imports).
- **Coverage preserved**: the deleted `test_ac_tree_hud_status_icons.py` had one live-renderer test (failed ✖ glyph) — the +17 in `test_ac_tree_hud_handler.py` migrates exactly that; done/executing/pending were already covered by the existing depth-1 test. `test_ac_tree_hud_render_depth3.py` only exercised the dead twin, so no replacement needed.
- The audit's stale `src/ouroboros/{execution,openclaw,secondary}/` dirs are `.pyc`-only with zero tracked files — outside this PR's reach (local cleanup, not a gap).
- merge-tree clean vs main, CI 10/10 green.

One non-blocking follow-up (pre-existing, not introduced here): the live `_MAX_RENDER_DEPTH=3` collapse path in `ac_tree_hud_handler.py` has never had a direct test — worth one when convenient.

PR #1545, opened 2026-07-02
feat(dashboard): merge provider-tagged web Kanban + delete dead TUI layers

Q00:
## Summary

**Phase D of `docs/run-metaharness-plan-2026-07.md`** (Option B of the dashboard-consolidation review): bring the provider-agnostic live web Kanban from the unmerged `feat/dashboard-web` branch into main, and delete the confirmed-dead legacy TUI layers. The web board is the only surface showing per-worker **provider identity** (claude/codex_cli/hermes_cli badges) — the control board for cross-harness runs.

### Part 1 — Web Kanban port (cherry-pick `fbe4832a`, re-wired against current main)
- `src/ouroboros/dashboard_web/` (~1.3k LOC, stdlib-only — no FastAPI/uvicorn per supply-chain policy): pure event reducer with terminal-state-sticky status (late `workflow.progress` snapshots cannot downgrade a completed card), read-only sqlite tail, SSE server (`/`, `/events`, `/healthz`, `/snapshot`, `/api/runs`), singleton daemon (O_EXCL lock, auto free-port, 600s idle self-exit — no zombies).
- Wiring: `execute_seed`/`start_execute_seed`/`start_auto` responses include `Live Dashboard: <url>` + `meta.dashboard_url`; default ON, opt-out `OUROBOROS_DASHBOARD=0`; resolved via `asyncio.to_thread`, strictly best-effort (never blocks/fails a run). Conflicts vs v0.44.0's changed wiring sites resolved and verified.
- `tests/conftest.py` pins `OUROBOROS_DASHBOARD=0` globally so unit tests never spawn a daemon.

### Part 2 — Dead TUI deletion (−4k LOC)
Every candidate grep-verified to have no live importer before deletion:
`screens/dashboard.py`, `screens/dashboard_v2.py`, `screens/hud_dashboard.py`, `widgets/drift_meter.py`, `widgets/parallel_graph.py`, `widgets/cost_tracker.py`, `components/token_tracker.py`, `components/agents_panel.py` + their dead tests (surgical removal from mixed test files). `dashboard_v3` (pause/resume, debug inspector, logs) is untouched — it has no web equivalent yet. `tests/unit/test_dashboard.py` was a filename false-positive (tests the MCP AC formatter) and is untouched.

Net: **+1811 / −4063** across 29 files.

## Testing
- 35 dashboard_web tests + 176 TUI tests pass; `import ouroboros.tui.app` clean.
- `ruff check` ✅ / `ruff format` ✅ / `mypy src/` ✅ (400 files, clean).
- Full `pytest tests/ --ignore=tests/e2e`: 11170 passed; the 17 failures are the pre-existing config/codex-leak set (verified identical on a clean checkout of the base, and re-verified post-rebase on clean `origin/main`).


Q00:
Both blockers addressed in `da4704fd`:

1. **Daemon reuse ignores DB identity**: daemon state now records the RESOLVED db path; reuse (fast path, in-lock double-check, and lock-loser wait) requires liveness AND exact db match. Pre-`db_path` state files never match → replaced, not trusted. On mismatch a fresh daemon spawns on a new free port and overwrites the state; the orphan idle-exits within 600s (deliberately not killed — commented). Regression test mirrors the repro: two sqlite DBs → second ensure returns `reused=False`, new port, `/api/runs` = exactly the second DB's runs.
2. **/snapshot label XSS**: the run id is HTML-escaped (innerHTML sink) and the whole status line embedded via `json.dumps` with `</` escaped — closes both the JS-string breakout and the script-element escape. Audited remaining interpolation paths (board JSON, /api/runs, live page runId) — all already encoded. Regression tests cover single/double quotes, backslash, and `</script>` payloads.

Full non-e2e gate: 11185 passed; only the 17 pre-existing config-leak failures. CI green.

Q00:
Blocker addressed in `5a10fa76`: the live dashboard page no longer dead-ends when it is opened before a run exists.

**The contract change** (`src/ouroboros/dashboard_web/page.py`): the auto flow publishes the daemon *base* URL while interview/seed are still running (`auto_handler.py` `_resolve_dashboard_base_url()`), so at open time `/api/runs` is legitimately empty. The old bootstrap fetched `/api/runs` exactly once and, on an empty list, permanently rendered "no active run" — never connecting when the run later materialized.

The page now:
- shows a `waiting for run…` state and **re-polls `/api/runs` in a loop** (`while (!runId)`) until a run appears, then attaches to the SSE stream via `connect()`;
- gates the poll with `setTimeout(..., WAIT_POLL_MS=3000)` so an idle page never spins hot against the shared SQLite file;
- still honors an explicit `?run=` immediately (returned first from `pickRun()`), and handles a run that appears *after* the page loaded — the poll loop covers exactly that case.
- stays dependency-free: vanilla JS, stdlib server, no new assets.

**Coverage:**
- `tests/unit/dashboard_web/test_page.py::TestLivePage::test_index_html_polls_for_pending_run` — asserts the rendered page contains the polling loop (`while (!runId)`, `pickRun()`, `setTimeout`, `WAIT_POLL_MS`, `waiting for run`) and no longer contains the old `no active run` dead end.
- `tests/unit/dashboard_web/test_daemon.py::TestPendingRun::test_empty_runs_and_polling_page_are_served` — spins the real server against an empty EventStore and proves `/api/runs` serves `200 {"runs": []}` and `/` serves the polling index page, i.e. the pending-run contract holds end-to-end.

Validation: `uv run ruff format . && uv run ruff check .` clean; `uv run pytest tests/unit/dashboard_web tests/unit/tui/test_screens.py tests/unit/tui/test_widgets.py tests/unit/tui/test_cancelled_display.py -q` → 112 passed (110 prior + 2 new).

Q00:
Re-requesting bot review: pending-run blocker addressed in `5a10fa76` (see comment above) — page now polls /api/runs with a waiting state until the run appears. `b73ede7d` is a no-op re-enqueue commit.

Q00:
Blocker addressed in `81e9cc25`: the live dashboard was published for the wrong database whenever a custom store was in play.

**Root cause.** Both dashboard-URL resolvers called `dashboard_url_for_run()` / `dashboard_base_url()` with no `db_path`, so the DB-scoped daemon always defaulted to `~/.ouroboros/ouroboros.db`. When the MCP server is started with `ooo mcp --db-path` it injects `EventStore("sqlite+aiosqlite:///<db>")` into the execute/auto handlers, so the published dashboard tailed the home DB while the execution wrote to the custom DB — `/api/runs` and `/events` were empty or showed unrelated runs.

**Fix — thread the store’s real DB path through at BOTH call sites:**
- `EventStore.sqlite_path()` (new) derives the filesystem path from the store’s own aiosqlite URL, understanding both the plain `sqlite+aiosqlite:///<path>` form and the read-only `…///file:<path>?mode=ro&uri=true` URI form; returns `None` for in-memory / non-SQLite backends (no local file to scope to).
- `execution_handlers._resolve_dashboard_url(...)` (both the sync `ExecuteSeedHandler.handle` path and the background `StartExecuteSeedHandler` path) and `auto_handler._resolve_dashboard_base_url(...)` now take `db_path` and forward it to the daemon, derived via `_event_store_db_path(self.event_store)` from each handler’s injected store. A repo sweep confirms these are the only two dashboard-URL call sites.

**Also (low-priority follow-up):** `reader._connect_readonly` now percent-encodes the DB path (keeping `/`) so a path containing `?`/`#` can’t be misparsed as the SQLite URI’s query/fragment.

**Regression coverage:**
- `tests/unit/persistence/test_event_store_sqlite_path.py` — `sqlite_path()` across custom / default-home / `:memory:` / non-sqlite / read-only-URI forms.
- `tests/unit/mcp/tools/test_dashboard_db_path.py` — `ExecuteSeedHandler`, `StartExecuteSeedHandler` and `StartAutoHandler` built with a custom-path `EventStore` surface that path via the exact attribute each call site reads, and the resolvers forward it to the daemon (incl. an end-to-end path from an `ExecuteSeedHandler`’s store).
- `tests/unit/dashboard_web/test_reader_readonly_uri.py` — read-only connect succeeds for a path containing `?`/`#` and still rejects writes.

Validation: `uv run ruff format . && uv run ruff check .` clean. `uv run pytest tests/unit/persistence/test_event_store_sqlite_path.py tests/unit/mcp/tools/test_dashboard_db_path.py tests/unit/dashboard_web tests/unit/tui/test_screens.py tests/unit/tui/test_widgets.py tests/unit/tui/test_cancelled_display.py -q` → 126 passed. `tests/unit/persistence/test_event_store.py` passes. The 11 failures in `test_start_auto.py` / `test_auto_handler_ralph_runtime.py` are pre-existing codex-config env leakage (identical on the parent commit, `opencode` vs `codex` assertion) — unrelated to this change.

Q00:
Blocker addressed in `f5c88a54`: the no-local-file fallback is now a tri-state, and the duplicated helper is gone.

**Contract (`src/ouroboros/mcp/tools/_dashboard.py::_daemon_db_target`)** — the resolvers branch on the injected store:
- **no store injected** → `(False, None)`: pass `db_path=None`, daemon uses its default (home DB) exactly as before;
- **store with a real SQLite file** → `(False, path)`: scope the daemon to that file (the custom-DB threading from the previous round);
- **store present but not dashboard-addressable** (`sqlite+aiosqlite:///:memory:` or a non-SQLite backend, where `sqlite_path()` is `None`) → `(True, None)`: **SUPPRESS** the dashboard URL entirely — the resolver returns `None` and never contacts the daemon, so it can no longer publish a dashboard for an unrelated home DB.

The prior bug was that `_event_store_db_path()` collapsed cases (a) and (c) to the same `None`, which `daemon._resolve_db(None)` turned into `~/.ouroboros/ouroboros.db`.

**Dedupe:** the helper + both resolvers were copy-pasted in `execution_handlers.py` and `auto_handler.py`. They now live in one shared `mcp/tools/_dashboard.py` (`resolve_dashboard_run_url` / `resolve_dashboard_base_url`); both handlers import from it and the call sites pass the store directly, so the tri-state has a single owner. (Removed the now-unused `asyncio` import from `auto_handler.py`.)

**Regression coverage** (`tests/unit/mcp/tools/test_dashboard_db_path.py`, rewritten):
- `_daemon_db_target` across all four cases (no-store / custom-file / `:memory:` / non-sqlite);
- `ExecuteSeedHandler`, `StartExecuteSeedHandler`, `StartAutoHandler` built with a custom-path store resolve to that path via the exact attribute each call site reads;
- an injected **in-memory** store yields **NO** dashboard URL and the daemon is asserted **never contacted** — for both the execute run-url path (incl. straight from an `ExecuteSeedHandler`) and the auto base-url path (incl. from a `StartAutoHandler`);
- no-store default still forwards `db_path=None` (behavior unchanged).

Validation: `uv run ruff format . && uv run ruff check .` clean. `uv run pytest tests/unit/mcp/tools/test_dashboard_db_path.py tests/unit/persistence/test_event_store_sqlite_path.py tests/unit/dashboard_web -q` → 59 passed. (Avoided the leaking `start_auto`/`ralph` files this round; the auto path is asserted via the targeted tests above.)

PR #1546, opened 2026-07-02
feat(orchestrator): cross-harness recovery & review — REDISPATCH_ALT_HARNESS, executor≠reviewer, outcome flywheel

Q00:
## Summary

**Phase X of `docs/run-metaharness-plan-2026-07.md`** — the meta-harness moves no single-vendor harness can make. Research found cross-vendor verification is unclaimed territory (parallel multi-vendor execution is mature — Vibe Kanban, Conductor — but nobody cross-verifies); this PR makes it first-class.

All logic lives in **new modules**; `parallel_executor.py` gains only a narrow hook (~3 insertion sites) to minimize conflicts with the sibling verify-by-default PR.

### X0 — Shared runtime picker (`orchestrator/runtime_picker.py`)
Deterministic "healthy (runtime×backend) that is NOT the one that just failed": availability via the existing `model_catalog.installed_backends()` detection, capability-gated, outcome weights as tie-break only.

### X1 — `REDISPATCH_ALT_HARNESS` recovery (fully wired, default ON)
New `RecoveryAction` + separate `alt_harness_policy()` layer (existing `_POLICY_TABLE` pins untouched). When an AC fails terminally with **fabrication-suspected**, **stall-retries-exhausted**, or **transient-exhaustion (sustained 429/529)**, the same AC is re-run ONCE on a different installed runtime (fresh worker session, recursion-guarded throwaway executor, cap consumed up-front). Emits an observable `execution.ac.alt_harness_redispatched` from→to event. Any negative decision or error → today's failure path untouched. Config: `execution.cross_harness_redispatch: true`.

### X2 — Executor ≠ reviewer binding (`evaluation/reviewer_independence.py`)
Consensus juries drop voters sharing the executor's vendor (backend→vendor + model→vendor mapping), but **never below a viable quorum (≥2)** — votes beat purity. Honest `reviewer_independence` label on `ConsensusResult`: `independent` / `same_vendor` / `unavailable` (single-vendor setups: label only, zero behavior change). Executor backend resolved best-effort from `execution.session.*` lifecycle events.

### X3 — N-version tournament (scaffolded, deliberately)
Worktree manager (real-git isolation-tested) + pure `plan_tournament`/`select_tournament_winner`/diff export-apply primitives + `execution.n_version_tournament: false` flag. The live per-AC trigger is intentionally NOT wired yet — it is entangled with the concurrently-edited retry area and cannot be exercised safely without real runtimes. Follow-up once the verify-by-default PR settles.

### X4 — Outcome flywheel (`orchestrator/backend_outcomes.py`)
Read-only, bounded EventStore aggregation per runtime_backend → `outcome_weights()` feeding X0's tie-break. Every failure path collapses to no-weights; never blocks dispatch. This is the moat: only a meta-harness ever sees comparative cross-harness outcomes.

## Testing
- 54 new unit tests across 6 files (picker determinism/exclusion, decision precedence, hook fires-once + falls back cleanly, reviewer quorum guard, tournament worktree isolation, outcomes aggregation).
- `ruff check` ✅ / `ruff format` ✅ / `mypy src/` ✅ (405 files, clean).
- Full `pytest tests/ --ignore=tests/e2e`: 11219 passed; the 17 failures are the pre-existing config/codex-leak set (verified identical on the stashed clean tree). Post-rebase targeted re-run: 494 passed.
- Honest caveat: the live alt-runtime spawn is stubbed in tests (CI can't spawn real vendor CLIs); the hook is fail-safe by construction — any spawn error logs a warning and returns to today's failure path.

Note: expect a small mechanical conflict with the verify-by-default PR in `parallel_executor.py`'s retry area — whichever merges second gets rebased.


Q00:
Blocker addressed in `ef10e62c`: unknown model vendors now provide **no** independence evidence. `independent` requires positive proof — at least one voter whose vendor is KNOWN and different from the executor's. The reported repro (executor `codex` + `["default"]*3` + backends `[codex, claude]`) now stamps the new honest label `unverified` (also used when the executor's own vendor is unmappable — previously another false-`independent` path). `same_vendor` is reserved for all-known-matching rosters; `filter_voter_models` keeps unknown-vendor voters (behavior pinned by test). 4 labels documented in the module docstring; regression tests cover the repro, a mixed roster (`default` + `gemini-2.5-pro` → `independent`), and the unmappable-executor case.

Full non-e2e gate: 11235 passed; only the 17 pre-existing config-leak failures. CI green.

Q00:
Rebased onto `origin/main` (past #1548 verify-by-default, #1551 structured AC, #1555/#1556 extractions, #1553 capabilities split) and addressed the blocker at head `49340f6e`.

**Blocker — failed alternate-harness attempts no longer hide workspace side-effects.** The alternate backend does run in the *same* `cwd` as the original executor, so a failed redispatch can leave real edits behind. Previously `_maybe_redispatch_alt_harness` returned the alternate result only on success and `None` otherwise, so the caller kept the original same-runtime failure and the alt's edits went unreported. Fixed in the redispatch hook (`parallel_executor.py`): the hook now returns the alternate's result **whether it succeeds or fails**. On failure it routes through `_annotate_alt_harness_result`, which surfaces the alternate attempt as the authoritative failure — the returned `ACExecutionResult.error` names the `from -> to` backends and explicitly states "the alternate backend ran in the shared workspace and may have modified it", preserving the alternate's own failure detail. A negative decision or an infra error to *spawn* the alternate still returns `None` so the original path is untouched (never make a failure worse). The one-per-AC redispatch cap is unchanged. The observable `execution.ac.alt_harness_redispatched` from->to event is still emitted before the alternate runs, so events and evidence match the surfaced result.

I chose "surface as authoritative failure" over a separate isolated worktree because it fits the existing single-workspace executor cleanly and needs no new infra, exactly as the merge recommendation allows.

**Regression test** (`test_alt_harness_hook.py`): `test_alt_failure_is_surfaced_as_authoritative` stubs an alternate that fails *after* editing the workspace and asserts the caller receives the alt's failed, annotated result (not `None`, not the original error verbatim) — carrying the alt backend name, the workspace-mutation flag, and the alt's own failure detail. Added `test_alt_success_result_returned_unchanged` for the winning path. The pre-existing `test_stall_retry_exhaustion_...` now disables `cross_harness_redispatch` to isolate the pure stall->failure conversion — note it previously passed only *because* alt failures were silently discarded, i.e. the exact bug this blocker flagged.

**Follow-up — `reviewer_independence` now exposed** (`evaluation_handlers.py`): the honest label resolved on `ConsensusResult` is now surfaced in both the structured MCP metadata (`stage3_reviewer_independence`) and the formatted Stage 3 output ("Reviewer Independence: ..."). Covered by `TestEvaluateHandler::test_format_result_surfaces_reviewer_independence`.

Validation: `ruff format`/`ruff check` clean; 746 passed across `test_cross_harness_redispatch`, `test_alt_harness_hook`, `test_parallel_executor`, `test_parallel_executor_retry_resume`, `tests/unit/evaluation/`, `test_runtime_picker`, `test_backend_outcomes`, `test_n_version_tournament`, and `TestEvaluateHandler`.

Q00:
Blocker addressed in `ae21c06c`: cross-harness redispatch now waits until the same-runtime recovery budget is spent.

**Root cause.** The alt-harness hook lived inside `_execute_single_ac` and fired on the first non-stall atomic failure (and on stall-abandon). But `_execute_single_ac` is only one same-runtime attempt: the batch-level V3 retry loop (`_ac_retry_attempts`) in `_run_batch_with_verify_and_retry` re-dispatches non-stall failures afterward. So the alternate harness pre-empted the configured same-runtime retries — violating the "AC has spent its same-runtime recovery budget" policy documented in `cross_harness_redispatch.py` / `failure_taxonomy.py`.

**Fix (gate on the final same-runtime attempt — keeps the redispatch decision in one layer).** I chose the gate over moving the hook because the redispatch mechanism needs the per-AC re-run bundle that only exists inside `_execute_single_ac`. A new `same_runtime_budget_exhausted` flag is threaded from the retry layer through `_execute_ac_batch` into `_execute_single_ac`, and **both** hooks (non-stall failure and stall-abandon) are gated on it. It is `True` only on the AC's final same-runtime attempt:
- initial dispatch when `ac_retry_attempts <= 0` (no retries configured), else `False`;
- the last V3 retry (`ac_retry_attempts[idx] >= _ac_retry_attempts`), computed at the batch layer.

An abandoned stall is also re-dispatched by the V3 loop (its error is rewritten off the stall sentinel, so `_is_retryable_failure` is true), so its hook is gated identically. Sub-AC/direct callers default to `True` (behavior preserved); the throwaway alternate executor still disables redispatch (recursion guard). The failed-alt surfacing you already accepted (authoritative failure naming the from->to backends + workspace-mutation note) is untouched.

**Regression test** (`test_alt_harness_defers_until_same_runtime_retry_budget_spent`): with `ac_retry_attempts=2` and always-failing dispatches, the `same_runtime_budget_exhausted` value observed by the worker batch is `[False, False, True]` across the initial dispatch + two retries — i.e. no alt-harness eligibility until the same-runtime budget is spent. The pre-existing `test_stall_retry_exhaustion_...` (direct `_execute_single_ac` call) keeps `cross_harness_redispatch=False` to isolate the pure stall->failure conversion.

Validation: `ruff format`/`ruff check` clean; 722 passed across `test_cross_harness_redispatch`, `test_alt_harness_hook`, `test_parallel_executor`, `test_parallel_executor_retry_resume`, `tests/unit/evaluation/`, and `TestEvaluateHandler`.

Q00:
Blocker addressed in `be01e7c5`: cross-harness redispatch now opens on the retry early-stop path.

**Root cause.** The budget gate opened alt-harness only when the retry counter reached `_ac_retry_attempts`. But `_run_batch_with_verify_and_retry`'s kill criterion marks an AC terminal earlier when the failure class repeats (`pending.discard` before the cap). Those dispatches ran with `same_runtime_budget_exhausted=False`, so the in-worker hook never fired — a repeated `FABRICATION_SUSPECTED` (explicitly `REDISPATCH_ALT_HARNESS`-eligible) reached terminal without ever consulting the alternate harness.

**Fix (treat repeated-class early-stop as budget exhausted).** This matches the existing gate semantically: the early-stop *is* the same-runtime recovery giving up, so the budget is spent. At the early-stop branch, when the dispatch was not already the final attempt (`not retry_batch_final`, i.e. its workers never got the hook), the batch layer now consults the alternate harness for the eligible AC via a new `_maybe_redispatch_alt_harness_for_batch_ac` helper. It rebuilds the top-level re-run bundle and defers to the same `_maybe_redispatch_alt_harness` the workers use, so the alternate-runtime decision, the one-per-AC cap, and the failed-alt surfacing (authoritative failure naming from->to backends + workspace-mutation note) all stay in one place. Final-attempt early-stops already fired the hook in-worker and are skipped, so there is no double redispatch (the per-AC cap backstops this regardless).

**Regression test** (`test_alt_harness_opens_on_retry_early_stop`): `ac_retry_attempts=2` with the same `FABRICATION_SUSPECTED` verdict on the initial attempt and retry 1 early-stops before the counter cap, yet the alternate harness is consulted exactly once (`alt_backends == ["codex"]`) and its failed result is surfaced as the authoritative outcome (error carries the `alt-harness` annotation). The prior ordering test (`test_alt_harness_defers_until_same_runtime_retry_budget_spent`) still holds: `[False, False, True]` across initial + two retries.

Noted and deferred per your review: the `n_version_tournament.py:165` `git diff HEAD` untracked-file concern is pre-wiring scaffolding — left untouched.

Validation: `ruff format`/`ruff check` clean; 723 passed across `test_cross_harness_redispatch`, `test_alt_harness_hook`, `test_parallel_executor`, `test_parallel_executor_retry_resume`, `tests/unit/evaluation/`, and `TestEvaluateHandler`.

Q00:
Blocker addressed in `8d7ffc0d`: the early-stop alt-harness success result is now verify-gated.

**Root cause.** My batch-level early-stop hook replaced the stored AC result with the alternate's without re-applying `_apply_verify_gate`. The alternate runs via `_run_single_ac_on_backend` -> `_execute_single_ac`, which has no seed-level success contract, so an alternate `success=True` with a failing `verify_command` or missing `expected_artifact` could be accepted as success — bypassing the verify-by-default contract.

**Fix.** In the early-stop branch, `_apply_verify_gate` is now applied to the alternate result before it is stored — a byte-for-byte mirror of the retry loop's gate on same-runtime results (`session_id`/`execution_id` and all). An alternate success against a failing contract is flipped to FAILED just like a same-runtime success.

**Worker-level alt path — already gated, no matching hole (as you asked me to verify).** The in-worker hook (final-attempt and the `ac_retry_attempts<=0` initial dispatch) returns its result up through `_execute_ac_batch`, and `_run_batch_with_verify_and_retry` gates every returned result: the initial dispatch at the `# V1 gate on freshly-successful ACs` loop, and each retry dispatch's `gated = await self._apply_verify_gate(...)` before the early-stop branch runs. So the worker alt result is verify-gated downstream on both paths; only the new batch-level early-stop replacement (which happens *after* that gate) needed closing, and it is now closed in the same place.

**Regression test** (`test_early_stop_alt_success_still_verify_gated`, in test_parallel_executor_verify_by_default.py): a repeated-FABRICATION AC early-stops before the cap, its alternate reports `success=True`, and `verify_command: exit 1` still gates the final result to FAILED with "Verify gate failed" — no contract bypass.

Validation: `ruff format`/`ruff check` clean; 747 passed across `test_cross_harness_redispatch`, `test_alt_harness_hook`, `test_parallel_executor`, `test_parallel_executor_retry_resume`, `test_parallel_executor_verify_by_default`, `tests/unit/evaluation/`, and `TestEvaluateHandler`.

PR #1548, opened 2026-07-02
feat(orchestrator): verify-by-default — AC success contracts, orchestrator-run verify gate, retry, fat-harness ON

Q00:
## Summary

**Phase V of `docs/run-metaharness-plan-2026-07.md`** — the highest-impact fix from the run failure-mode audit: today the default (parallel) run path marks an AC complete on the worker's own word (`success = not message.is_error`), never retries a failed AC (`ac_retry_attempts` was a dead counter), and fences all stagnation/lateral recovery to the sequential runner. On hard domains and brownfield, workers "believe" they finished and nothing catches the false positive.

### 1 — AC success contract (`AcceptanceCriterionSpec`)
`acceptance_criteria` items become structured specs — `description` (mandatory) + optional `verify_command` / `expected_artifacts` / `output_assertion`. Bare-string ACs coerce forever; **description-only specs serialize back to bare strings so existing seed files round-trip byte-identically** (tested against a real fixture). `ac_text()` helper rippled across ~16 consumers with identical behavior for legacy data. seed-architect now emits per-AC contracts (`AC: <desc> | verify: … | artifacts: … | expect: …`, 2 few-shot examples); unparseable lines fall back to description-only — parsing never fails on missing contract fields.

### 2 — V1: orchestrator-run verify gate (default ON)
When an AC carries `verify_command`, the ORCHESTRATOR runs it (asyncio subprocess, workspace cwd, configurable 600s timeout) and requires exit 0 + `output_assertion` substring. Worker self-report becomes irrelevant. Gate failure → existing failure taxonomy (`EVIDENCE_MISSING`, no new codes) + `execution.verify.failed` event + command output tail in the failure reason. Already-failed ACs skip the gate — one failure event per root cause, no double-fail. Contract-less ACs byte-identical to today.

### 3 — V2/V3: per-AC retry with kill criteria
Non-stall failures re-dispatch up to `execution.ac_retry_attempts` (default 2) with an enriched prompt (failure class + error tail + verify output tail). Identical failure class on consecutive attempts → stop early. Final attempt injects a lateral change-of-approach directive (shared `resilience/lateral.py` helper, loop-independent). Retries complete within the level, so recovered ACs naturally unblock their dependents.

### 4 — V4: trust leaks closed
- Sibling-evidence flips: a FAILED AC with a `verify_command` can only be flipped to satisfied if its own gate passes now.
- `--skip-completed`: runs the gate when a contract exists (fail → execute normally); contract-less skips are stamped `verification_status=assumed`.

### 5 — fat-harness ON by default (user mandate), both paths
- CLI: fresh runs enforce typed evidence + verifier PASS unless the seed opts out with `execution_mode: legacy`.
- MCP (`execute_seed`/`start_execute_seed` — the primary `ooo run` surface): same resolution, same vocabulary. **Deliberate behavior change:** `_validate_fresh_execution_mode` re-admits `legacy` as the explicit opt-out (#978 P5 removed it when the default runner was observe-mode and the token was redundant; verify-by-default makes it meaningful again).
- Plugin dispatch (structurally cannot enforce): explicit `fat_harness` requests still hard-reject; the default case downgrades with a visible `fat_harness_downgraded: "plugin_dispatch_cannot_enforce"` meta note. Resume: persisted contract always wins; sessions without one resume ON unless the seed says `legacy`.
- `ParallelACExecutor` constructor defaults unchanged (False/0) — ~100 direct-construction tests rely on them and every production path passes flags explicitly from the resolvers; the mandated OUTCOME (fat-harness active on every real run) is delivered at the resolver layer.

Config: `execution.run_verify_commands: true`, `verify_command_timeout_seconds: 600`, `ac_retry_attempts: 2`.

## Testing
- 16 new executor tests (gate pass / exit-fail / assertion-mismatch, retry increment/exhaust/early-stop, lateral on final attempt, sibling gate, skip-completed stamp) + schema round-trip/coercion tests + 5 new MCP resolution tests (fresh default ON, legacy opt-out, plugin downgrade note, plugin explicit reject, resume persisted-False preserved).
- `ruff check` ✅ / `ruff format` ✅ / `mypy src/` ✅ (400 files, clean).
- Full `pytest tests/ --ignore=tests/e2e`: 11191 passed; the 17 failures are the pre-existing opencode/codex/plugin set, verified byte-identical on the base commit. Existing test edits are representation-only (`ac_text()` wrapping forced by the element-type change) or deliberate default-assertion updates for the mandated flips — documented per file in the commits.

Note: expect a small mechanical conflict with #1546 (cross-harness) in `parallel_executor.py`'s retry area — whichever merges second gets rebased.

## R-run comparison

Substrate-only auto/ change: parameter plumbing / representation ripple only — the interview round loop, refiner, and grading mechanics are untouched, so there is no per-round comparison to capture (template-sanctioned explicit N/A).

| Rounds completed in 600 s      | N/A | N/A | n/a |
| Terminal reason                | N/A | N/A | n/a |
| EventStore event count         | N/A | N/A | n/a |

Budget compliance: [x] within 1.5× (no per-round cost path touched)

### DIFF
--- a/src/ouroboros/cli/commands/run.py
+++ b/src/ouroboros/cli/commands/run.py
@@ -421,33 +421,28 @@ def _load_skip_completed_markers(
 def _resolve_fat_harness_mode(seed_data: dict[str, Any]) -> bool:
     """Resolve the fresh-run fat-harness selector.
 
-    Fat-harness acceptance is opt-in until the shipped authoring/QA pipeline can
-    reliably produce profile-compatible typed evidence for every AC. Seeds that
-    request ``seed.orchestrator.execution_mode: fat_harness`` keep the stricter
-    verifier-gated path; missing/blank selectors use the legacy runner.
+    Verify-by-default: ``ooo run`` enforces typed evidence plus a verifier PASS
+    at AC acceptance unless the seed explicitly opts out with
+    ``seed.orchestrator.execution_mode: legacy``. Missing/blank selectors and an
+    explicit ``fat_harness`` selector both enable the stricter path.
     """
     orchestrator_config = seed_data.get("orchestrator")
     if not isinstance(orchestrator_config, dict):
-        return False
+        return True
 
     execution_mode = orchestrator_config.get("execution_mode")
+    if execution_mode in (None, "", "fat_harness"):
+        return True
     if execution_mode == "legacy":
-        print_error(
-            "seed.orchestrator.execution_mode='legacy' was removed after #978 P5; "
-            "omit the selector for the default runner or set execution_mode='fat_harness' "
-            "to opt in to typed evidence plus verifier PASS acceptance."
-        )
-        raise typer.Exit(1)
-    if execution_mode in (None, ""):
+        # Explicit opt-out of verify-by-default: fall back to the self-report
+        # runner. This is the supported escape hatch, not an error.
         return False
-    if execution_mode != "fat_harness":
-        print_error(
-            "seed.orchestrator.execution_mode must be 'fat_harness' when set "
-            f"(got {execution_mode!r})."
-        )
-        raise typer.Exit(1)
 
-    return True
+    print_error(
+        "seed.orchestrator.execution_mode must be 'fat_harness' or 'legacy' when set "
+        f"(got {execution_mode!r})."
+    )
+    raise typer.Exit(1)
 
 
 def _resolve_resume_fat_harness_mode(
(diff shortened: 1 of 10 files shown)


Q00:
Blocker addressed in `578f1446`: `expected_artifacts` is now enforced. The verify gate generalized to a success-**contract** gate applying when a spec has `verify_command` OR non-empty `expected_artifacts`; the artifact leg runs first (cheap), requires each entry to exist under the run workspace (absolute paths and `..` escapes rejected via `resolve()` + `is_relative_to` and reported as `escapes workspace`), and reports ALL missing entries in one failure reason. Same `EVIDENCE_MISSING` taxonomy class and `execution.verify.failed` event (now carrying `expected_artifacts` + `missing_artifacts`), same no-double-fail semantics. The two other gate consumers — sibling-evidence flips and `--skip-completed` — honor artifacts-only contracts too. 7 new tests (artifacts-only pass/fail, both escape forms, combined-leg specs, sibling-flip, skip-completed).

Also fixed in `e918cd5e`: one e2e test (`test_task_prompt_includes_acceptance_criteria`) missed the spec-migration ripple because the local gate excluded tests/e2e — representation-only `ac_text()` update; CI is now fully green including enforce-perf-budget (R-run N/A section added — substrate-only auto/ change).

PR #1576, opened 2026-07-06
fix(mcp): close the remaining JobManager terminal-event race (zombie job without terminal state)

Q00:
## Status

Three commits, each driven by CI evidence from the previous one:

1. `90bf1738` — swallowed-append hardening + inline backstop. **Refuted as the primary mechanism** by this PR's own CI (job 85408070770): the flake reproduced with the backstop in the branch, no log line.
2. `f72cf8e4` — cancellation-proof detached backstop + in-process stranded-job reconciler in `get_snapshot`. **Changed the failure mode** (CI job 85415178221): the job now terminalizes, but as INTERRUPTED while the runner returned SUCCESS — proving both terminal appends *silently vanished* while a later append succeeded.
3. `8f96188d` — the actual root cause, found and **reproduced deterministically**: a silent append-void in `EventStore` for `:memory:` databases.

## Root cause: the `:memory:` append-void (src/ouroboros/persistence/event_store.py)

`EventStore` used **StaticPool** for `sqlite+aiosqlite:///:memory:` (the chaining test fixture's store). StaticPool hands the **same aiosqlite connection to concurrent `engine.begin()` blocks with no mutual exclusion**. SQLite transactions are connection-scoped, so:

```
task A (append terminal event):   INSERT (implicit BEGIN) ........... COMMIT → no-op!
task B (monitor cancelled          |                          |
        mid-SELECT poll):          |   exception → ROLLBACK ──┘  ← discards A's INSERT
result: append() returns SUCCESS; no exception; no log; event GONE
```

sqlite3 `commit()` with no active transaction is a **no-op**, so task A's append reports success. In this test two jobs (run + chained evaluate), two monitors, and agent-process lifecycle appends all share one connection; `_run_job`'s finally cancels monitors mid-poll, so ROLLBACKs land inside other tasks' INSERT→COMMIT windows. On a loaded CI runner the congested aiosqlite worker-thread queue stretches those windows to many ms.

Evidence:
- **Deterministic repro**: slow the commit 100ms, run one `append()` concurrently with one failing `engine.begin()` block → `append returned SUCCESS, persisted rows: 0`. Committed as `test_memory_store_concurrent_rollback_cannot_void_append` (fails on StaticPool, passes with the fix).
- **Stress repro of the actual CI failures**: with random 2-40ms aiosqlite latency injected (simulating a loaded runner), the chaining test fails **2/15** locally on StaticPool with the exact CI signatures — including round 3's `INTERRUPTED != COMPLETED`.
- **Why only this test**: it is the only job-manager test on a `:memory:` store (`test_run_evaluate_chaining.py:34`); the rest use file-backed stores with per-connection pools. **Why round 2/3's raw CI log shows no captured log section**: nothing ever raised.
- Full trace of the earlier theories is preserved in the commit messages; both were real defense-in-depth gaps but not the void itself.

## Why not just serialize the one connection

A one-connection exclusive pool closes the interleaving but stress testing exposed the second `:memory:` hazard: **asyncio cancellation invalidates the pooled connection**, and its replacement is a fresh empty database (`no such table: events` — 3/100 under stress). A plain `:memory:` database cannot survive the loss of its only connection, period.

## Fix (commit 3)

Back `:memory:` stores with SQLite's **process-shared in-memory VFS (`memdb`, SQLite ≥ 3.36)**, unique name per store instance, plus a keepalive anchor connection:
- pooled connections join ONE shared database with normal connection-scoped transactions → interleaving impossible;
- the database **survives connection invalidation**;
- `:memory:`-store semantics preserved (one store instance == one private DB; `sqlite_path()` still `None`);
- ancient-SQLite fallback: serialized one-connection queue pool.

**Stress result: 100/100 green** with injected latency (vs 3/100 serialized pool, 2/15 StaticPool).

## Defense-in-depth kept from commits 1-2 (all still tested)

- Detached, cancellation-proof release-point backstop in `_run_job`'s finally — now **replays the body's real intended terminal event** (result_text / result_meta incl. `chained_evaluate_job_id`) before any guard-status fallback.
- In-process stranded-job reconciler in `get_snapshot` (scoped to jobs this manager instance started; defers to drain/live holders) — the second net that converted round 2's 60s zombie into round 3's visible wrong-status, and remains the safety net for any future silent loss.
- Swallowed-append last-resort guard (commit 1).

## Diagnostics insurance

The chaining test's failure paths now dump **every persisted job stream** (run + chained evaluate, with result_meta keys), engine pool status, and manager task/backstop state — any residual flake will name its own mechanism.

## Done evidence

- `ruff check` / `ruff format --check`: clean; `mypy src/`: 423 files, no issues
- persistence + job manager + drain + chaining suites: **293 passed**
- chaining test plain 30× loop: 0 failures; **jittered 100× loop: 0 failures**
- new regression tests verified to fail on the code they target (git stash)
- `pytest tests/ --ignore e2e --ignore unit/mcp --ignore integration/mcp`: **10101 passed**, only the documented pre-existing codex-config-leak failures

### DIFF
--- a/src/ouroboros/mcp/job_manager.py
+++ b/src/ouroboros/mcp/job_manager.py
@@ -449,6 +449,11 @@ class JobManager:
         coroutine ``close()`` cleanup.
         """
         cancel_context = False
+        # The terminal event the body intends to persist, captured before each
+        # append so the release-point backstop can replay the REAL payload
+        # (result_text / result_meta — e.g. chained_evaluate_job_id — matter to
+        # consumers) instead of a guard-status fallback if the append is lost.
+        intended_terminal: tuple[str, dict[str, Any]] | None = None
         try:
             await self.update_status(job_id, JobStatus.RUNNING, f"Running {job_type}")
 
@@ -484,28 +489,32 @@ class JobManager:
                             event_id=f"{_DRAIN_INTERRUPTED_EVENT_ID_PREFIX}{job_id}",
                         )
                     raise
+                cancelled_data = {
+                    "status": JobStatus.CANCELLED.value,
+                    "message": "Job cancelled",
+                }
+                intended_terminal = ("mcp.job.cancelled", cancelled_data)
                 await self._append_terminal_event_with_fallback(
                     "mcp.job.cancelled",
                     job_id,
-                    {
-                        "status": JobStatus.CANCELLED.value,
-                        "message": "Job cancelled",
-                    },
+                    cancelled_data,
                 )
                 raise
             except Exception as exc:
                 snapshot = await self.get_snapshot(job_id)
                 if snapshot.is_terminal:
                     return
+                failed_data = {
+                    "status": JobStatus.FAILED.value,
+                    "message": f"Job failed: {exc}",
+                    "error": str(exc),
+                    "is_error": True,
+                }
+                intended_terminal = ("mcp.job.failed", failed_data)
                 await self._append_terminal_event_with_fallback(
                     "mcp.job.failed",
                     job_id,
-                    {
-                        "status": JobStatus.FAILED.value,
-                        "message": f"Job failed: {exc}",
-                        "error": str(exc),
-                        "is_error": True,
-                    },
+                    failed_data,
                 )
             else:
                 snapshot = await self.get_snapshot(job_id)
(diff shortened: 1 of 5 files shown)


Q00:
**Post-merge note on the CHANGES_REQUESTED review** (head_sha `f72cf8e4`, the round-2 state):

The flagged blocker — a successful run intermittently terminalizing as `INTERRUPTED` instead of `COMPLETED` — was exactly the residual the final commit `8f96188d` (round 3) fixed:

1. **Root cause found at the store layer**: StaticPool hands the same `:memory:` aiosqlite connection to concurrent `engine.begin()` blocks; a concurrent rollback silently discards another task's uncommitted terminal INSERT and its COMMIT no-ops. Fixed by backing `:memory:` stores with SQLite's process-shared **memdb VFS** (unique name per store + keepalive anchor; serialized-pool fallback for sqlite <3.36).
2. **Backstop now replays the real intended terminal event** (full `result_text`/`result_meta` incl. `chained_evaluate_job_id`) before any guard-status fallback — a successful run can no longer be downgraded by the recovery nets.
3. **Evidence**: jittered 100× loop on the chaining test: 0/100 failures (vs 2/15 on StaticPool); deterministic repro test `test_memory_store_concurrent_rollback_cannot_void_append` fails on the old pool and passes with memdb. Post-merge, the previously ~40%-flaky `test_successful_run_enqueues_chained_evaluate_job` has passed on every subsequent PR CI run (#1580–#1583, 3 Python versions each).

No further change needed for this finding; flagging here so the review record reflects that the reviewed sha predates the fix.

PR #1581, opened 2026-07-06
feat(auto): A2 interview trace artifact — greppable per-run projection over EventStore + ledger

Q00:
## What & why

**A2 — Interview trace artifact** (run-metaharness plan, Phase A). Per interview run, emit a grep-able export of what actually happened: questions, answers, ambiguity-score trajectory, lateral results, timeout/fallback/degraded flags, promoted-vs-rejected decisions (with their A1 provenance), and the evaluate/QA outcome.

This is a **projection, not a store**. It reads the two durable sources that already exist by the time an `ooo auto` run finalizes — the auto `SeedDraftLedger` (question history + every decided contract field with `source`/`status`/`provenance`, A1 #1579) and the `EventStore` (interview event stream: ambiguity trajectory, lateral advisories, lifecycle flags) — and writes plain files under the project working dir. No new store, no event-schema changes.

## Module + export API

`src/ouroboros/auto/trace_export.py`:

- `async export_trace_from_state(state, ledger, *, event_store, out_root=None) -> Path` — core projection.
- `async export_interview_trace(run_id, *, auto_store, event_store, out_root=None) -> Path | None` — **manual / A3-CLI entry**: loads a past run from `AutoStore` + EventStore and projects it. Importable and cleanly callable; A3 wraps this. (A3 CLI itself is intentionally NOT built here.)
- `async best_effort_export_trace(state, ledger, *, event_store, out_root=None) -> Path | None` — the pipeline finalize hook; logs and swallows any failure so trace generation has **zero impact on the run**.

## File schema (`<cwd>/.ouroboros/traces/<run_id>/`, run_id == auto_session_id)

One JSONL file per stream (each line self-describing via `type`) plus a human digest. Empty streams omit their file.

| File | Lines |
|------|-------|
| `questions.jsonl` | `question` (round, question, answer from ledger history) + `response_event` (event-store round records) |
| `ambiguity.jsonl` | per-event `ambiguity_score` trajectory (round, from/to milestone), time-ascending |
| `lateral.jsonl` | `lateral` events (persona, directive, outcome) + `lateral_final` from state |
| `decisions.jsonl` | `decision` per ledger entry: section, key, value, source, **provenance**, status, **promoted** (active) vs rejected, **gated** (model_inferred/timeout_default) |
| `flags.jsonl` | `event_flag` (timeout/fallback/degraded/safe_default/…) + `state_flag` (terminal_phase, closure_mode, degraded_seed, blocker, open_gaps) |
[...]

PR #1582, opened 2026-07-06
feat(cli): A3 experience-store CLI — ooo harness list/show/trace/diff/frontier over A2 traces

Q00:
## A3 — Experience-store CLI (Phase A, final piece)

Thin, read-only `ouroboros harness` command group over A2's greppable interview traces (`<cwd>/.ouroboros/traces/<run_id>/`, merged #1581). Filesystem-first; when a run has persisted auto state but no trace on disk, it projects on-demand via A2's `export_interview_trace`. **No new MCP tools; no writes to `auto/` state** — read-only use of `trace_export.py` + `AutoStore`/`EventStore`. All output is plain deterministic text; the only timestamp shown is data-derived from persisted state or stored events (never wall-clock-at-render).

### Subcommand walkthrough

| Command | What it does | Example |
|---|---|---|
| `harness list` | Enumerate exported traces **plus** persisted-but-unexported runs (marked `not exported`). Columns: run_id, data-derived timestamp, status, phase, grade, final ambiguity, provenance-histogram digest. | `ooo harness list` |
| `harness show <run_id>` | Render `summary.md` + key `outcome.json` fields (grade, final ambiguity, counts, provenance histogram, unverified-provenance count, blocker). Exports on-demand when missing. | `ooo harness show auto_ab12cd34` |
| `harness trace <run_id> --grep <pat> [--stream ...]` | Grep across the JSONL streams; prints `file:line:` matches in a fixed, deterministic order. `--stream` restricts to one of `questions\|ambiguity\|lateral\|decisions\|flags`. Regex, with literal-substring fallback on invalid pattern. | `ooo harness trace auto_ab12cd34 --grep timeout --stream flags` |
| `harness diff <a> <b>` | Side-by-side of key outcome metrics (grade, final ambiguity, per-stream counts) with `delta (B-A)`, plus a per-stream line-count table straight from disk. | `ooo harness diff auto_ab12 auto_cd34` |
| `harness frontier --metric <name>` | Rank all traced runs by a metric from `outcome.json`. Unknown metric → clean error listing available ones. | `ooo harness frontier --metric final_ambiguity` |

### Metrics supported by `frontier`
`final_ambiguity` (asc) · `rounds` (asc) · `decisions` (desc) · `promoted` (desc) · `rejected` (asc) · `gated` (asc) · `ambiguity_points` (desc) · `lateral` (asc) · `flags` (asc) · `unverified_provenance` (asc). Each direction is metric-defined; `None`-valued runs sort last; ties break on run_id for determinism. These are exactly the fields A2's `outcome.json` (+ A1 provenance findings) makes available — no fabricated metrics.

### Design notes
- On-demand export opens the `EventStore` **read-only** (never mutates the event log) and tolerates an absent DB (projects from auto state alone).
- `list` reads persisted auto-state files raw (no `from_dict` validation) so a pruned/partial state never aborts the listing; timestamp priority is persisted `created_at` → earliest stream `at` → `n/a`.
- `frontier` ranks only runs already exported on disk (it does **not** bulk-project every persisted run — that would be an unbounded side effect on a read command).

### Validation
- `uv run ruff check .` — clean · `uv run ruff format --check .` — clean (1035 files)
- `uv run mypy src/` — Success, no issues in 429 source files
- `uv run pytest tests/unit --ignore=tests/unit/mcp` — **9896 passed**, incl. 12 new `tests/unit/cli/test_harness.py` (list / show / trace-grep / stream-filter / diff / frontier + on-demand-export fallback + unknown-metric + unknown-run + unknown-stream error paths).
- Pre-existing failures (verified identical on clean `origin/main` via `git stash -u`, unrelated to this PR): 6 codex-config-leak tests (`test_surface` ×4, `test_auto_runtime_dispatch` ×2); `test_wheel_contains_profile_yamls` is a full-suite build-artifact flake (passes in isolation on this branch); `test_runner.py::test_execute_seed_success` hangs in this sandbox (spawns a real runtime) in isolation on clean `origin/main` too.

Scope: `src/ouroboros/cli/` (new `commands/harness.py` + registration in `main.py`) + `tests/unit/cli/test_harness.py`. No `src/ouroboros/auto/` changes → no R-run section applies.

### DIFF
--- a/src/ouroboros/cli/commands/harness.py
+++ b/src/ouroboros/cli/commands/harness.py
@@ -1,0 +1,705 @@
+"""Experience-store CLI (A3 / run-metaharness plan).
+
+A thin, read-only surface over the greppable interview traces that A2
+(:mod:`ouroboros.auto.trace_export`) projects under
+``<cwd>/.ouroboros/traces/<run_id>/``. Each run directory holds one JSONL
+file per stream (``questions``/``ambiguity``/``lateral``/``decisions``/
+``flags``), an ``outcome.json`` of key metrics, and a human ``summary.md``.
+
+This module never *writes* run state. Its only mutation is a best-effort,
+on-demand **projection**: when a requested run has persisted auto state but
+no trace on disk yet, it calls
+:func:`ouroboros.auto.trace_export.export_interview_trace` to materialize the
+same deterministic files the pipeline finalize hook would have written. All
+rendered output is plain deterministic text (no wall-clock stamps — any
+timestamp shown is derived from persisted state or stored events).
+
+Subcommands::
+
+    ouroboros harness list
+    ouroboros harness show <run_id>
+    ouroboros harness trace <run_id> --grep <pattern> [--stream <name>]
+    ouroboros harness diff <a> <b>
+    ouroboros harness frontier --metric <name>
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Callable
+import json
+from pathlib import Path
+import re
+from typing import Annotated, Any
+
+import typer
+
+from ouroboros.auto.state import AutoStore
+from ouroboros.auto.trace_export import (
+    AMBIGUITY_FILE,
+    DECISIONS_FILE,
+    FLAGS_FILE,
+    LATERAL_FILE,
+    OUTCOME_FILE,
+    QUESTIONS_FILE,
+    SUMMARY_FILE,
+    export_interview_trace,
+)
+from ouroboros.cli.formatters.panels import print_error
+
+app = typer.Typer(
+    name="harness",
+    help="Inspect A2 interview traces (list/show/trace/diff/frontier).",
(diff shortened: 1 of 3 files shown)


Q00:
**Resolution note for review 4638593562 (CHANGES_REQUESTED):** the path-traversal blocker is fixed in follow-up PR #1587 — `run_id` is validated before any filesystem lookup (allowlist `^[A-Za-z0-9._-]+$` matching real `auto_<hex12>` session ids, explicit `..`/separator/absolute rejection) plus a symlink-safe `is_relative_to(traces_root)` backstop, all via the shared `_ensure_trace()` resolver used by show/trace/diff; 9 parametrized attack-case regression tests added; the low-priority `expanduser` note is also done. #1587 was **approved by ouroboros-agent** and merged. (Merged-PR reviews cannot be dismissed; this comment is the resolution record.)

PR #1590, opened 2026-07-07
fix(cli): auto waits for its run-handoff job instead of cancelling it on exit

Q00:
## Summary

Live smoke testing on `main` found that a direct CLI `ouroboros auto "<goal>"` drove interview→seed→grade successfully, then started the execute phase as a **detached** in-process `JobManager` job and exited — printing `run_handoff_started` / "not verified complete; execution is still external/pending". Because the `JobManager` lives in the dying process, `asyncio.run` teardown cancelled the job almost immediately (observed: `job_3ce3660f6503` created → CANCELLED in ~200ms). Net effect: a single CLI `ouroboros auto` invocation **never actually executed the seed**.

## Confirmed cancel mechanism

- `src/ouroboros/cli/commands/auto.py:294` — `asyncio.run(_run_auto(...))`. When `_run_auto` returns, the event loop is torn down.
- `src/ouroboros/mcp/job_manager.py:1963-1971` (`drain` docstring, authoritative): *"Called by the serve shutdown path... Without an explicit drain, job tasks are killed by `asyncio.run` teardown after `EventStore.close()`..."*. The CLI path has **no** `drain()` call (that is serve-only).
- `src/ouroboros/mcp/job_manager.py:496-499` — the `_run_job` `CancelledError` handler persists `mcp.job.cancelled` when not draining, so the teardown cancel lands as a plain `CANCELLED` terminal.

## Fix

Scope: `src/ouroboros/cli/commands/auto.py` + tests only (no `auto/` or `job_manager.py` semantic changes).

1. **Default: wait for the run job.** After the handoff starts (`_is_run_handoff_only_completion` + a `job_id`), `_run_auto` keeps the event loop alive and awaits the execute job to a terminal state using the **live** `StartExecuteSeedHandler._job_manager` (the one that owns the running task). Progress is streamed via the existing `JobWaitHandler` (`ouroboros_job_wait`) renderer and the terminal receipt via `JobResultHandler` (`ouroboros_job_result`) — no new renderer invented. The terminal snapshot is reconciled onto `AutoPipelineResult` (mirrors `mcp/tools/auto_handler._reconcile_execution_job_snapshot`), so the final output shows a real completed/failed verdict and the exit code reflects the run result.
2. **`--no-wait`** restores fire-and-forget. No pre-existing detach flag existed for this path (`--skip-run` stops before the run; `--complete-product` already runs synchronously), so `--no-wait` is new. It prints an honest notice that CLI-mode execution does **not** survive process exit unless a persistent owner (e.g. a running `ouroboros mcp serve`) holds it, and points at `ouroboros job wait <id>`.
3. **Bounded wait.** No artificial cap (a real run takes minutes); each poll long-polls up to 5s then re-checks. `Ctrl-C`/`CancelledError` during the wait cancels the job cleanly via the existing `JobManager.cancel_job` path and prints an interrupt notice.
4. **Zero change to the MCP/server auto path.** The wait is gated inside the CLI `_run_auto` only.

## MCP-path note

The MCP `ouroboros_start_auto` path was never affected: its server process is long-lived, so the background execute job keeps running and is reconciled to a terminal status on resume via `_reconcile_execution_job_snapshot`. That long-lived owner is exactly why this bug was invisible outside direct CLI mode.

## Tests

New `tests/unit/cli/test_auto_run_handoff_wait.py` (6 tests):
- wait drives a fast job to `COMPLETED` and reconciles `status=complete` — asserts **no `mcp.job.cancelled` event** is written (the regression).
- a failing run job flips the result to `failed` (exit code reflects run result).
- no-op when the job handle is unknown / absent (plugin dispatch).
- default CLI invocation requests `wait=True`; `--no-wait` passes `wait=False` and prints the honest detach notice.

## Validation

- `uv run ruff check .` + `ruff format --check .` — clean.
- `uv run mypy src/` — clean (429 files).
- `uv run pytest tests/unit/cli tests/unit/auto` (excl. mcp) — **2346 passed**; the only 6 failures are the pre-existing codex-config-leak cases in `test_surface` / `test_auto_runtime_dispatch` (confirmed failing on clean `origin/main`, unrelated to this change).
- New auto CLI wait tests: **6 passed**.

---

This change is a resume-path correctness fix (durable state reconciliation on
run handoff); it does not alter the auto convergence loop's per-cycle cost.
No per-cycle comparison applies, so every cell is filled explicitly as N/A.

### DIFF
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -1497,6 +1497,33 @@ class AutoPipeline:
                     return await self._handoff_to_ralph(
                         state, ledger, seed, review, run_subagent=None
                     )
+                # Non-complete-product RUN resume with a persisted run handle.
+                # A persisted handle proves the execute job was *dispatched*,
+                # not that it reached terminal success (Q00/ouroboros#1590):
+                # the owning process may have exited (deadline/Ctrl-C/kill)
+                # leaving the job cancelled, or the run may have paused
+                # (usage-limit) or failed. Reconcile the owned job's terminal
+                # state before declaring COMPLETE so ``--resume`` cannot return
+                # a stale product-complete for an incomplete run. When no poll
+                # channel is available (plain-function run starter, pruned job,
+                # etc.) fall back to the historical "trust the handle" behavior
+                # so genuinely-complete sessions and legacy handles are
+                # unaffected.
+                run_verdict = (
+                    await _wait_owned_run_job_terminal(
+                        self.run_starter,
+                        state.job_id,
+                        timeout_seconds=self._deadline_capped_timeout(
+                            state, state.phase_timeout_seconds(AutoPhase.RUN)
+                        ),
+                    )
+                    if state.job_id
+                    else None
+                )
+                blocked = self._block_resume_if_run_not_successful(state, run_verdict)
+                if blocked is not None:
+                    self._save(state)
+                    return self._result(state, ledger, review=review, blocker=state.last_error)
                 state.transition(
                     AutoPhase.COMPLETE, "execution already started; using persisted run handle"
                 )
@@ -1850,6 +1877,49 @@ class AutoPipeline:
         )
         self._save(state)
         return self._result(state, ledger, blocker=state.last_error)
+
+    def _block_resume_if_run_not_successful(
+        self, state: AutoPipelineState, run_verdict: dict[str, Any] | None
+    ) -> bool | None:
+        """Block a non-complete-product RUN resume unless the owned job succeeded.
+
+        ``run_verdict`` is the owned execute job's terminal ``result_meta``
+        (from :func:`_wait_owned_run_job_terminal`), or ``None`` when no poll
+        channel was available. Returns ``True`` when the state was marked
+        BLOCKED (the run did not reach terminal success), or ``None`` to let the
+        caller proceed to COMPLETE.
+
+        Conservative on ambiguity: ``None`` verdict (unpollable / pruned /
+        plain-function starter) or a non-terminal ``running``/``queued`` status
+        preserves the historical "trust the persisted handle" behavior, so
+        genuinely-complete and legacy sessions are unaffected. Only an
+        *observed* non-success terminal (paused / failed / cancelled /
+        interrupted / unknown) blocks — resumably, via ``run_starter`` — so a
+        later ``--resume`` re-reconciles instead of returning stale COMPLETE.
+        """
+        if run_verdict is None:
+            return None
+        status = _optional_str(run_verdict.get("status"))
+        success = run_verdict.get("success")
+        if success is True or status == "completed":
+            return None
+        if status in {"running", "queued", "cancel_requested"}:
+            # Still in flight elsewhere — cannot prove failure; do not cancel or
+            # block a run another owner may still complete.
+            return None
+        if status == "paused":
+            state.mark_blocked(
+                "resumed run execution paused before completion; resume the "
+                "paused run before continuing",
+                tool_name="run_starter",
+            )
+            return True
+        detail = status or "unknown"
+        state.mark_blocked(
+            f"resumed run execution did not reach terminal success: {detail}",
+            tool_name="run_starter",
+        )
+        return True
 
     def _deadline_capped_timeout(self, state: AutoPipelineState, phase_timeout: float) -> float:
         """Return ``phase_timeout`` capped by the remaining pipeline deadline.
--- a/src/ouroboros/auto/state.py
+++ b/src/ouroboros/auto/state.py
@@ -782,6 +782,41 @@ class AutoPipelineState:
         self.last_authoring_backend = None
         return True
 
+    def close_completed_run_handoff_as_failed(self, message: str) -> bool:
+        """Correct a prematurely-COMPLETE run handoff into a FAILED terminal.
+
+        Companion to :meth:`reopen_completed_run_handoff_to_run` for the
+        *genuine failure* case (the owned execute job reached a FAILED terminal
+        or reported ``success is False``). Unlike a paused / cancelled /
+        deadline outcome — which is resumable and re-opens to ``RUN`` — a
+        genuine failure must preserve a non-resumable failed terminal contract:
+        ``--resume`` must NOT retry it and must NOT report the run as complete.
+
+        Sets the phase directly to ``FAILED`` (``COMPLETE`` has no forward
+        transition edge, matching :meth:`reopen_completed_run_handoff_to_run`)
+        and sets ``last_tool_name = None`` so :meth:`resume_capability`
+        classifies the state as :attr:`AutoResumeCapability.NONE`
+        (``_recoverable_phase_for_tool(None)`` is ``None``). The run handles are
+        left intact for diagnostics. No-op (returns ``False``) unless the state
+        is currently ``COMPLETE``.
+        """
+        if self.phase is not AutoPhase.COMPLETE:
+            return False
+        now = utc_now_iso()
+        self.phase = AutoPhase.FAILED
+        self.phase_started_at = now
+        self.last_progress_at = now
+        self.updated_at = now
+        self.last_progress_message = message
+        self.last_error = message
+        self.last_error_code = None
+        # Intentionally not "run_starter": a genuine run failure is a terminal
+        # dead end, not a RUN-recoverable blocker, so resume_capability() must
+        # resolve to NONE rather than routing back into the RUN phase.
+        self.last_tool_name = None
+        self.last_authoring_backend = None
+        return True
+
     def is_terminal(self) -> bool:
         """Return True when the state cannot continue automatically."""
         return self.phase in TERMINAL_PHASES
--- a/src/ouroboros/auto/state.py
+++ b/src/ouroboros/auto/state.py
@@ -747,6 +747,41 @@ class AutoPipelineState:
         self.last_tool_name = tool_name
         self.transition(AutoPhase.FAILED, message, error=message)
 
+    def reopen_completed_run_handoff_to_run(self, message: str) -> bool:
+        """Reopen a prematurely-COMPLETE run handoff back to the RUN phase.
+
+        ``AutoPipeline.run`` persists ``COMPLETE`` immediately after a run
+        handoff is *dispatched* (to avoid duplicate run starts on resume), but
+        that COMPLETE is not verified product completion. When a caller later
+        observes the execute job finish paused / unsuccessfully / cancelled (for
+        example the CLI in-process wait), the durable state must be corrected so
+        a later ``--resume`` reconciles the owned run job instead of returning
+        the stale COMPLETE (the ``pipeline.run`` COMPLETE fast-path returns
+        immediately for ``AutoPhase.COMPLETE``).
+
+        ``_ALLOWED_TRANSITIONS`` has no ``COMPLETE -> *`` edge by design — auto
+        never *advances* out of COMPLETE — so this corrective re-open sets the
+        phase directly rather than via :meth:`transition`. It is a no-op (returns
+        ``False``) unless the state is currently ``COMPLETE``; the run handles
+        (``job_id`` / ``execution_id`` / ``run_session_id``) and
+        ``run_handoff_status`` are left intact so the RUN-phase resume path can
+        reconcile the owned job. ``last_tool_name`` is set to ``"run_starter"``
+        so the state classifies as RUN-recoverable (resumable).
+        """
+        if self.phase is not AutoPhase.COMPLETE:
+            return False
+        now = utc_now_iso()
+        self.phase = AutoPhase.RUN
+        self.phase_started_at = now
+        self.last_progress_at = now
+        self.updated_at = now
+        self.last_progress_message = message
+        self.last_error = None
+        self.last_error_code = None
+        self.last_tool_name = "run_starter"
+        self.last_authoring_backend = None
+        return True
+
     def is_terminal(self) -> bool:
         """Return True when the state cannot continue automatically."""
         return self.phase in TERMINAL_PHASES
--- a/src/ouroboros/auto/state.py
+++ b/src/ouroboros/auto/state.py
@@ -767,6 +767,16 @@ class AutoPipelineState:
         ``run_handoff_status`` are left intact so the RUN-phase resume path can
         reconcile the owned job. ``last_tool_name`` is set to ``"run_starter"``
         so the state classifies as RUN-recoverable (resumable).
+
+        The absolute pipeline deadline (``deadline_at`` / ``deadline_at_epoch``)
+        is CLEARED. The premature COMPLETE was often reached because the wait's
+        own deadline expired, so the persisted absolute deadline is already
+        past; leaving it would make ``--resume`` hit ``AutoPipeline.run``'s
+        deadline gate (which fires on an expired deadline BEFORE the RUN
+        reconciliation) and immediately re-block with ``pipeline_timeout`` —
+        turning the advertised resume into a dead end. Clearing both fields lets
+        ``AutoPipelineState.from_dict`` re-arm a fresh budget on the next load
+        (its non-terminal re-arm path), so a fresh resume gets a fresh deadline.
         """
         if self.phase is not AutoPhase.COMPLETE:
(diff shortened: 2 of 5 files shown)


Q00:
Addressed the BLOCKING finding in a4fe9126 — the wait reconciliation no longer treats `JobStatus.COMPLETED` as run success by itself.

**What was wrong:** `_reconcile_run_handoff_result` mapped every COMPLETED execute job to `status="complete"` + `AutoResumeCapability.NONE`. But `ExecuteSeedHandler._classify_synchronous_execution_status` maps `SessionStatus.PAUSED` → `result_meta={"status": "paused", "success": None}, is_error=False` — the JOB completes while the EXECUTION is paused (usage-limit pause etc.), so a paused run would have been reported as complete and non-resumable.

**Fix:** the CLI wait path now reads the execution-level verdict from `snapshot.result_meta` (`_run_meta_verdict`), mirroring the complete-product resume gate in `auto/pipeline.py` (persisted-handle branch, `result_meta["status"]` / `["success"]`):

- `"paused"` → `status="blocked"` with resume guidance; run handle (`job_id`/`execution_id`/`run_session_id`) and `RESUME` capability preserved — never complete/non-resumable. CLI exits 1.
- `success is False` or meta status in `{failed, cancelled, interrupted}` → `failed` with detail.
- Allowlist parity: terminal meta without an explicit success signal (`success is not True and status != "completed"`) is NOT evidence of run success → `blocked`, resume preserved.
- Meta status falls back to the job's own terminal status when absent — same semantics as `auto/adapters._wait_for_job_terminal`'s `setdefault`, so a COMPLETED job with empty meta still passes (pipeline gate parity).
- Only genuine success meta → `complete` + `NONE`.

**Tests:** added the requested regression — waited snapshot `JobStatus.COMPLETED` + `result_meta={"status": "paused", "success": None}` reconciles to NOT successful (`blocked`), resume capability preserved, handles intact; plus the unknown-meta allowlist case and a CLI-surface test asserting exit code 1 for the paused/blocked verdict. `tests/unit/cli/test_auto_run_handoff_wait.py`: 9 passed.

**Validation:** `ruff check .` / `ruff format --check .` clean, `mypy src/` clean (429 files), `tests/unit/cli` + `tests/unit/auto`: 2400 passed with only the 6 known pre-existing codex-config-leak failures (confirmed failing on clean origin/main).

Q00:
Addressed the second BLOCKING finding in 7aa6520c — the default run-handoff wait is now bounded by the pipeline/CLI deadline.

**What was wrong:** `_await_run_handoff_terminal` looped `while not snapshot.is_terminal` on 5s polls with no total deadline, so a wedged execute job kept the process alive forever and bypassed the advertised top-level `--timeout` once the handoff started.

**Fix:**
- The wait now carries the pipeline's armed monotonic deadline (`state.deadline_at`) — the SAME budget the CLI `--timeout` contract enforces (`arm_deadline()` = `monotonic + pipeline_timeout_seconds`). The pipeline consumed part of it during interview/seed/grade; the wait gets the **remainder**. Each poll window is clipped to the remaining budget via `asyncio.wait_for(..., timeout=min(poll, remaining))`.
- When no deadline was armed (defensive), a fresh `pipeline_timeout_seconds` window applies via `fallback_timeout_seconds` — never unbounded, no new invented default.
- **On expiry:** cancel the run job cleanly via the existing `JobManager.cancel_job` path (same as the Ctrl-C semantics), wait a short grace for the terminal `mcp.job.cancelled` event, then return a bounded verdict — `status="blocked"`, NOT complete, resume handle + `ouroboros auto --resume` guidance preserved, `AutoResumeCapability.RESUME`, non-zero exit — mirroring the paused reconciliation shape.
- `--timeout --help` now documents that the deadline also bounds the run-handoff wait.

**Regression test (bot-requested):** `test_wait_deadline_cancels_wedged_job_and_reports_bounded_verdict` — a never-terminal execute job + a 0.5s deadline returns within the bound (asserts elapsed < 10s under an outer 15s guard), the job reaches `CANCELLED` with an `mcp.job.cancelled` event written, the result is `blocked` with `RESUME` capability + handle preserved; plus a CLI-surface test asserting exit code 1 for the deadline/blocked verdict.

**Non-blocking (your suggestion):** added a `TODO(Q00/ouroboros#1590)` to extract a shared job→auto-result reconciliation helper with `_reconcile_execution_job_snapshot`; the MCP path is intentionally not refactored in this PR.

**Validation (scoped — disk was tight so no full-tree run):** `ruff check` + `ruff format --check` clean, `mypy src/ouroboros/cli/commands/auto.py` clean, `tests/unit/cli/test_auto_run_handoff_wait.py` + `tests/unit/cli/test_auto_command.py`: **48 passed** (11 wait + 37 auto-command).

Q00:
Addressed the third BLOCKING finding in 8b98f5d3 — the "resume preserved" contract is now real, not masked by test setup.

**Root cause (bot was right):** production `AutoPipeline._result()` emits `resume_capability=state.resume_capability()`, and `AutoPipelineState.resume_capability()` maps `AutoPhase.COMPLETE → AutoResumeCapability.NONE`. My `_reconcile_run_handoff_result` / `_run_wait_deadline_result` initialized `resume_capability` FROM that incoming COMPLETE result, and the blocked branches never overrode it — so a real paused / unknown-success / deadline-cancelled run reported `blocked` with NO resume capability. The tests hid this by building `_handoff_only_result()` with RESUME, which the pipeline never emits.

**Fix:**
- Every blocked-but-resumable branch now sets `resume_capability = AutoResumeCapability.RESUME` **explicitly**, regardless of the incoming capability: paused execution, unknown-success allowlist gate, cancelled/interrupted mid-run, and the deadline-cancel verdict (`_run_wait_deadline_result`).
- Genuine success stays `complete`/NONE; genuine `JobStatus.FAILED` keeps its non-resumable NONE.

**Test realism (bot's non-blocking note, done):**
- `_handoff_only_result()` now emits `resume_capability=NONE` to match real `AutoPipeline._result()` output for a COMPLETE handoff.
- The paused / unknown-success / deadline tests now assert the input was NONE and the reconciliation **upgrades** to RESUME. Proof: with the fix reverted (old `auto.py`, NONE input), those 3 tests FAIL (`3 failed, 8 passed`); with the fix they pass — demonstrating the bug and the fix.

**Verified against production:** `pipeline.py` `_result()` → `resume_capability=state.resume_capability()`; `state.py` `resume_capability()` → `COMPLETE` returns `NONE`.

**Validation:** `ruff check` + `ruff format --check` clean, `mypy src/ouroboros/cli/commands/auto.py` clean, `tests/unit/cli/test_auto_run_handoff_wait.py` + `tests/unit/cli/test_auto_command.py`: **48 passed** (11 wait + 37 auto-command).

Q00:
Addressed the fourth BLOCKING finding in 0e683ac2 — the durable auto state is now corrected, not just the in-memory result.

**Root cause (bot was right, and deeper than the last round):** `pipeline.run()` persists the durable state as `AutoPhase.COMPLETE` at run handoff (to avoid duplicate dispatch), and the wait only replaced the returned in-memory result. So a later plain `ouroboros auto --resume <id>` hit the `pipeline.run` COMPLETE fast-path and returned stale product-complete.

I verified the "contained option" (reopen durable → BLOCKED) alone does **not** stick: the default (non-complete-product) RUN-resume branch (`pipeline.py`) transitions a persisted run handle straight to COMPLETE **without reconciling the owned job**, so a reopened state just re-completes on the next resume. Both places had to be fixed.

**Fix:**
- `pipeline.py`: the non-complete-product RUN-resume branch now reconciles the owned execute job via the existing `_wait_owned_run_job_terminal` (new `_block_resume_if_run_not_successful`) before declaring COMPLETE. Paused/failed/cancelled/interrupted/unknown → `mark_blocked` resumably (`run_starter`); observed success/`completed` → COMPLETE (unchanged). Conservative on ambiguity: no poll channel (plain-function starter, pruned job) or a still-running/queued status keeps today's trust-the-handle behavior, so genuinely-complete and legacy sessions are untouched (the existing plain-function-starter resume test still passes).
- `state.py`: `reopen_completed_run_handoff_to_run()` corrects a premature COMPLETE back to RUN (COMPLETE has no forward transition edge by design — documented corrective re-open), keeping the run handle so the RUN-resume reconcile can classify it.
- `cli/auto.py`: after a non-success wait verdict, the corrected durable state is persisted via the same `AutoStore` the pipeline uses, and the in-memory `resume_capability` is derived from the durable state (memory == durable). Genuine success leaves durable COMPLETE (fast-path intact).

**Durability regression tests (the ones that were missing):**
- CLI: a paused wait verdict, then **loaded fresh from the store** (new `AutoStore` instance, exactly as `--resume` would), asserts the durable phase is NOT COMPLETE and `resume_capability()` is RESUME with the handle intact; genuine success leaves durable COMPLETE.
- Pipeline: a default RUN resume with a persisted **paused** owned job reconciles to `blocked` (not complete); a **succeeded** owned job still completes; the pre-existing plain-function-starter resume still completes (backward-compat, requirement 3).

**Verified against production:** `pipeline.run` COMPLETE fast-path and the non-complete-product RUN-resume handle branch; `AutoPipelineState.resume_capability()` (RUN → RESUME, COMPLETE → NONE).

**Validation:** `ruff check .` clean, `ruff format --check` clean, `mypy src/` clean (429 files). `tests/unit/auto` + `tests/unit/cli`: only the 6 known pre-existing codex-config-leak failures remain (baseline-diffed — zero new regressions; note I caught and fixed a self-inflicted `mark_failed` drop before pushing). Required CLI files `test_auto_run_handoff_wait.py` + `test_auto_command.py`: 50 passed.

Added the `## R-run comparison` section to the PR body (all cells N/A — this is a resume-path correctness fix with no per-cycle cost change).

Q00:
Addressed the interrupt-path gap in e43bd482 — the persistence boundary is now consistent across every exit path.

**The hole (bot was right):** the Ctrl-C / CancelledError handler cancelled the execute job and re-raised, but never called `_persist_resumable_run_verdict`. So an operator interrupt cancelled the run yet left the durable auto session as COMPLETE — a later `ouroboros auto --resume <id>` would report success for a run that was actually cancelled. Same class of bug as the deadline/paused durability issue, on the one remaining exit path.

**Fix:**
- The interrupt handler now corrects the durable state via the **same** `_persist_resumable_run_verdict` used by the deadline / paused / terminal-non-success paths, **before** re-raising the interrupt. Ordering is deliberate: the synchronous persist (`reopen_completed_run_handoff_to_run` + `store.save`) runs FIRST so it always lands even if the subsequent cancel `await` is itself interrupted; then the job is cancelled; then the interrupt propagates. A cancelled run is now durably resumable (phase RUN), never durably COMPLETE.
- Fixed the misleading copy: was "Ctrl-C to cancel the run and detach" (it cancels, does not detach) → now "Ctrl-C cancels the run; resumable with ooo auto --resume". The interrupt warning also now prints the resume command.

**Test (same durability proof shape as deadline/paused):** `test_wait_interrupt_persists_resumable_durable_state` — start the wait against a wedged job, deliver a cancellation, then load state **fresh from the store** (new `AutoStore`, as `--resume` would) and assert it is NOT COMPLETE and classifies resumable (phase RUN, `resume_capability()` RESUME).

**Validation:** `ruff check .` + `ruff format --check .` clean, `mypy src/` clean (429 files). `tests/unit/cli`: only the 2 known pre-existing codex-config-leak failures (1133 passed). Required files `test_auto_run_handoff_wait.py` + `test_auto_command.py`: 51 passed.

Q00:
Addressed the failure-vs-resumable inconsistency in 87e8f615 — genuine failures now preserve the failed-terminal contract instead of being reopened to RUN.

**The bug (bot was right):** `_persist_resumable_run_verdict` reopened **every** non-`complete` verdict to RUN and overwrote capability with `state.resume_capability()` (RESUME for RUN). So a genuinely FAILED execute job — which `_reconcile_run_handoff_result` had correctly classified `status="failed"`, `NONE` — was flipped to durable RUN and returned resumable, making real failures look retryable. The existing failed test missed it because it didn't pass `state`/`store`, so the persist path never ran.

**Fix — distinguish resumable from genuinely-failed:**
- `_persist_resumable_run_verdict` now branches on the reconciled verdict:
  - `status == "blocked"` (paused / deadline / interrupt / cancelled / interrupted job / unknown-success) → reopen durable to RUN, capability from state (RESUME). Resumable outcomes unchanged.
  - `status == "failed"` (job `JobStatus.FAILED` / `success is False`) → persist a durable **FAILED** terminal via the new `state.close_completed_run_handoff_as_failed` (**NOT** reopened to RUN), and keep the reconciled **NONE** — the reconciled verdict is authoritative for failures; `--resume` must not retry an already-failed run.
  - `status == "complete"` → unchanged (durable COMPLETE, fast-path intact).
- `state.close_completed_run_handoff_as_failed` sets phase `FAILED` directly (COMPLETE has no forward transition edge, same pattern as the RUN reopen) with `last_tool_name=None`, so `resume_capability()` resolves to `NONE` (`_recoverable_phase_for_tool(None)` is `None`) — a genuine run failure is a terminal dead end, not a RUN-recoverable blocker.

**Test realism (bot-flagged):** new `test_wait_failed_run_preserves_failed_terminal_durably` passes **`state` AND `store`** into `_await_run_handoff_terminal` and asserts a genuine failure (a) stays `NONE`, (b) is persisted **FAILED** — loaded fresh from the store: not RUN, not COMPLETE, (c) `--resume` preserves failed (`resume_capability()` is `NONE`). The paused/deadline/interrupt durability tests still assert reopen-to-RUN (those SHOULD be resumable).

**Verified against production:** `AutoPipelineState.resume_capability()` (FAILED + unmapped `last_tool_name` → `NONE`); `_recoverable_phase_for_tool(None)` → `None`; `_ALLOWED_TRANSITIONS[COMPLETE]` is empty (hence the direct-set corrective re-open).

**Validation:** `ruff check .` + `ruff format --check .` clean, `mypy src/` clean (429 files). `tests/unit/auto` + `tests/unit/cli`: 2408 passed, only the 6 known pre-existing codex-config-leak failures (baseline-diffed, zero new regressions). Required files `test_auto_run_handoff_wait.py` + `test_auto_command.py`: 52 passed.

Q00:
Addressed the deadline dead-end in 1f321371 — the reopen-to-RUN resume contract is now honest.

**The bug (bot repro was exact):** the deadline-cancelled reopen re-opened durable state to RUN and advertised `--resume`, but left the already-expired absolute deadline (`deadline_at` / `deadline_at_epoch`) in place. A later `ouroboros auto --resume <id>` hits `AutoPipeline.run`'s deadline gate (pipeline.py:607-619) — which fires on an expired deadline **before** the RUN reconciliation — and is immediately rewritten to `pipeline_timeout` / `NONE`. The advertised resume was a dead end.

**Fix:** `reopen_completed_run_handoff_to_run()` now **clears** `deadline_at` and `deadline_at_epoch`. `AutoPipelineState.from_dict` already re-arms a fresh budget for a non-terminal state whose deadline fields are both `None` (its existing non-terminal re-arm path, lines ~1139-1152), so a fresh resume gets a fresh, non-expired deadline and the resume-time gate passes into the RUN reconciliation. I verified against the gate at pipeline.py:607-619 (`deadline_at is not None and is_deadline_expired()`) and `arm_deadline`'s idempotent-unless-None contract.

**Scope:** applies to every resumable reopen-to-RUN (paused / deadline / interrupt / cancelled). Genuine success (COMPLETE) and genuine failure (`close_completed_run_handoff_as_failed` → terminal FAILED; the gate is skipped for terminal phases and `from_dict` won't re-arm a terminal) are untouched.

**Tests (bot's shape):**
- CLI: after a deadline-cancelled wait, the durable state loaded fresh is RUN with a re-armed, **NOT-expired** deadline (`not is_deadline_expired()`), so the resume gate passes.
- Pipeline (end-to-end): a reopened-to-RUN state saved+loaded through the store drives `pipeline.run` and does **not** return `pipeline_timeout` — it reaches the owned-job reconciliation (paused → blocked). Existing paused/succeeded/backward-compat resume tests kept.

**Validation:** `ruff check .` + `ruff format --check .` clean, `mypy src/` clean (429 files). `tests/unit/auto` + `tests/unit/cli`: 2410 passed, only the 6 known pre-existing codex-config-leak failures (baseline-diffed, zero new regressions). Required files `test_auto_run_handoff_wait.py` + `test_auto_command.py`: pass (54 in the wait+command files).

PR #1591, opened 2026-07-07
fix(orchestrator): normalize paths and unwrap shell wrappers in fat-harness evidence matching (release blocker)

Q00:
## Release blocker

Since #1548 flipped fat-harness verification **ON by default**, *every* `ouroboros run workflow <seed>` failed — on codex **and** claude runtimes — even when the worker's output was completely correct. Reproduced twice (isolated sandbox + maintainer's real codex env) with the trivial seed `hello.py defines greet(name)`:

```
Fat-harness verifier failed (unsupported evidence claims:
  files_touched: hello.py;
  tests_passed: python3 -c "from hello import greet; assert greet('World') == 'Hello, World!'; print('OK')")
→ FABRICATION_SUSPECTED → alt-harness redispatch (#1546) → fails again → 0/1
```
Control: same seed with `orchestrator.execution_mode: legacy` → PASSES.

## Confirmed mechanism (verified end-to-end, not assumed)

I reproduced the maintainer's exact transcript through the matcher and captured the live codex evidence with temporary instrumentation. Two independent causes, in `src/ouroboros/orchestrator/evidence/`:

1. **`files_touched` (`claims.py:_runtime_messages_support_file_claim`)** — the worker claims the workspace-relative `hello.py`; the transcript `Edit` event records the absolute `/private/tmp/<run>/hello.py`. When no workspace cwd was threaded to the verifier (`task_cwd` None — the live case), `_workspace_relative_file_claim` returns `None` and the function short-circuits to `False` **before ever inspecting the Edit event**. (With a cwd, `/tmp`↔`/private/tmp` was already handled via `Path.resolve`; the None path had no fallback.)

2. **`tests_passed` (`test_detection.py:_runtime_messages_support_test_claim`)** — the AC verifies itself with an inline `python3 -c "...; assert ...; print('OK')"`, wrapped by codex as `/bin/zsh -lc "python3 -c \"...\""`. The command matcher already unwraps the shell wrapper (so `commands_run` matched), but the `tests_passed` path gates every candidate through `_looks_like_test_command`, which only recognizes formal runners (pytest/unittest/gradle/mvn/npm test). An inline `python3 -c` assertion — and the natural-language prose the worker sometimes emits instead (`greet('x') returned 'Hello, x' and command output was OK`) — could never be matched.

## Matching-tier design (matcher only — default stays ON; no worker-prompt / taxonomy / flag changes)

**Paths** (`_runtime_messages_support_file_claim`): keep the existing workspace-anchored tiers (exact transcript match → cwd-relative↔absolute resolve equivalence → basename-if-file-exists), then add a **most-permissive tier**: match the raw claim against a structured mutation event (`Edit`/`Write`/`NotebookEdit` path, or an explicit shell write) by **basename-with-parent-dir suffix**, resolving both sides so `/tmp`↔`/private/tmp` compare equal. Works even without `task_cwd`. Still requires a real mutation event, so a fabricated/near-miss filename with no matching event rejects.

**Commands** (`_runtime_messages_support_test_claim`): accept a `tests_passed` claim backed by an **inline-code verification command** (`python3 -c` / `node -e` — an interpreter whose non-zero exit means a real failed check) that matches a transcript Bash event via the wrapper-aware matcher **and** ran with a runtime success signal. The candidate is the claim itself (command-form) or an already-backed `commands_run` entry (prose-form, anchored to a command the worker recorded and that really ran). New helpers `_is_inline_verification_command` / `_is_test_runner_family_command` in `shell_parsing.py`. Tiering stays conservative: exact → normalized-equal (wrapper/quote/whitespace-insensitive) → the inline-verification gate.

## Fabrication detection survives (negative-test evidence)

The three pre-existing fat-harness guard tests still pass, plus new negatives:
- near-miss filename (`goodbye.py` vs transcript `hello.py`) → REJECT
- near-miss command payload (`...=='WRONG'`) / command with no transcript event → REJECT
- prose `tests_passed` with **no** successfully-run inline command → REJECT
- read-only/print commands (`cat fake_results.txt` emitting a fake "passed" line, bare `pytest` for an unmentioned stale node-id) → still REJECT (excluded from the inline-verification gate; stay on the strict node-id path)
- skip-test invocations (`mvn -DskipTests`) and no-tests-ran builds → still REJECT

## Live proof (release gate)

The maintainer's exact seed, real configured backend (codex), clean committed code:

```
Parallel Execution Complete
Success: 1/1
- Task 1: [COMPLETED] hello.py defines greet(name) returning the string Hello, <name>
Execution completed successfully!
Session ID: orch_91c1e7c3c5c7
Completeness 0.82
```
Single verifier attempt, **no redispatch** (previously 0/1 → FABRICATION_SUSPECTED → alt-harness → 0/1). The live evidence that now passes: `files_touched: ["hello.py"]` against Edit `/private/tmp/<run>/hello.py`; `commands_run` the `python3 -c` command against the `/bin/zsh -lc "..."` Bash event; `tests_passed` as prose anchored to that successful command.

## Validation
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — no issues (429 files)
- `tests/unit/orchestrator/` (excluding the pre-existing-hanging `test_runner.py::test_execute_seed_success`, confirmed to hang on unmodified code too) — **2307 passed**
- Evidence/verify focused files (`test_parallel_executor.py`, `test_evidence_schema.py`, `test_verifier.py`, `test_parallel_executor_atomic_judgment.py`, `test_parallel_executor_verify_by_default.py`) — **all green**, incl. new regression tests

Scope: `src/ouroboros/orchestrator/evidence/` + tests only. Existing tests unmodified.


Q00:
Both blocking findings addressed in `aa0bcdde` — thanks, both were legit.

**Blocker 2 (over-widened anti-fabrication boundary)** — fixed. `_is_inline_verification_command` no longer accepts any interpreter `-c`/`-e` command that merely exits 0. The inline payload must now contain a real check construct — `assert` / `raise` / `throw` / `die` / `unittest` / `pytest` / `sys.exit(` / `os._exit(` / `process.exit(` (regex `_INLINE_VERIFICATION_CONSTRUCT_RE`). So:
- `python3 -c "print('OK')"` + a prose `tests_passed` claim → **REJECT** (exit 0 is meaningless; no check performed) — this is your repro.
- `python3 -c "from hello import greet; assert greet('x')=='Hello, x'; print('OK')"` (the AC's declared verify_command, actually run, `assert` present) → **ACCEPT**.

Read-only/print commands (`cat`/`echo`), bare `pytest` stale node-ids, faked `cat` output, and skip-test invocations all remain rejected as before. New regression tests: `test_atomic_verifier_rejects_inline_command_without_a_real_check` (print-only → reject) contrasted with `test_atomic_verifier_accepts_prose_tests_passed_backed_by_run_command` (real assertion → accept).

I went with the assertion-construct gate rather than threading the declared `verify_command` into the matcher — it's self-contained, keeps the maintainer's real case passing, and closes the `print('OK')` hole. Happy to add the declared-verify_command contract match on top if you'd prefer that as a second acceptance path.

**Blocker 1 (portability)** — fixed. Replaced the macOS-specific `/tmp`→`/private/tmp` test with `test_atomic_verifier_accepts_symlinked_cwd_against_resolved_transcript_path`, which creates a **real** symlink inside `tmp_path` via `os.symlink`, runs with `task_cwd` set to the symlinked dir, and records the transcript Edit under the resolved real dir — proving the `Path.resolve` equivalence tier without depending on the host `/tmp` layout (skips if symlink creation isn't permitted). The normalization code is unchanged.

**Validation re-run:**
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — clean (429 files)
- `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`, confirmed to hang on unmodified code) — **2308 passed**
- **Live gate still 1/1** with the tightened rule: `Success: 1/1`, `[COMPLETED]`, session `orch_4c4555792fe7`.

Q00:
You're right — the assertion-keyword heuristic was a losing game (`python3 -c "assert True"` is a "real construct" that verifies nothing). Switched to the principled PR-F/G contract binding in `3421e4a0`.

**Plumbing (step 1):** `ac_spec` was already threaded to `_execute_atomic_ac` by PR-G but stopped there. It now flows on into `_run_atomic_verifier_pass` → `_verify_atomic_evidence_against_runtime_messages` → `_runtime_messages_support_test_claim` as `verify_command`.

**New `tests_passed` rule for inline commands (step 2):**
- **AC declares a `verify_command`** → accept a `tests_passed` claim ONLY when a transcript command **matches the declared `verify_command`** (reusing the wrapper-unwrap + normalization matcher) **AND** that run shows a success signal. The declared command is the oracle:
  - `python3 -c "from hello import greet; assert greet('x')=='Hello, x'; print('OK')"` (the AC's own contract, actually run) → **ACCEPT** (maintainer's real case)
  - `python3 -c "assert True"` + fabricated prose, AC declares the greeting command → **REJECT** (the trivial command doesn't match the declared oracle) — your repro
- **AC declares NO `verify_command` (legacy)** → unchanged: formal test runners only, no arbitrary inline command accepted. Byte-identical to pre-PR behavior.

**Dropped (step 3):** `_is_inline_verification_command` + the assertion-keyword regex, plus the now-dead `_is_test_runner_family_command` / `_command_executable_names` / `_TEST_RUNNER_EXECUTABLES`. The prose-claim "any backed inline command with a keyword + success chunk" path is gone.

**Tests (step 4):**
- `..._accepts_prose_tests_passed_backed_by_declared_verify_command` — declared command run + prose → accept
- `..._rejects_inline_command_that_does_not_match_declared_verify_command` — `assert True` vs a different declared verify_command → reject (your repro)
- `..._rejects_tests_passed_when_declared_verify_command_not_run` — declared command never ran → reject
- `..._rejects_inline_command_for_legacy_ac_without_verify_command` — legacy AC, inline command → reject (formal runner still required)
- All pre-existing fabrication guards still pass (bare-pytest stale node-id, `cat`-faked output, skip-test, no-tests-ran, masked-form).

**Validation (step 5):**
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — clean (429 files)
- `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`, confirmed to hang on unmodified code) — **2309 passed**
- **Live gate still 1/1** with the principled rule: `Success: 1/1`, `[COMPLETED]`, session `orch_ea9aaeb5944b` (the seed's declared `verify_command` was actually run, so it matches the oracle and passes).

Q00:
Good catch — that raw-claim file fallback did re-open a scope hole. Fixed in `b0d265bc`.

**Root of the hole:** with `task_cwd` set, an absolute outside-workspace claim (or a `../` escape) makes `_workspace_relative_file_claim` return `None`, and the unconditional raw fallback then trusted `touch <abspath>` command-text — so `files_touched:["/tmp/.../outside.py"]` + `touch /tmp/.../outside.py` + exit 0 passed. Key realization: when `task_cwd` is set, Tier A already resolves both sides (`_file_claim_matches_runtime_path`), so it handles the relative↔absolute (and `/tmp`↔`/private/tmp`) form mismatch on its own — **the raw fallback was only ever needed for `task_cwd=None`.**

**Fix, split by workspace-known:**
- **`task_cwd` SET** → resolve the claim against cwd. Inside workspace → existing tiers (form-mismatch already covered). Outside workspace or `..` escape (`relative_claim is None`) → **REJECT**, with no fall-through to command-text mutation. Closes the hole.
- **`task_cwd` None** (workspace genuinely unknown — the original live case) → scope to the transcript's own structured mutation events: accept only an exact basename-with-parent-dir match to an **Edit/Write/NotebookEdit** path; do **not** trust `touch <abspath>` command text, and reject absolute claims / `..` traversal outright (`_file_claim_matches_runtime_path` enforces this). Stated honestly: I chose "scope to the transcript's Edit/Write events" over inventing a workspace root, since those events are the only reliable in-run signal when no cwd was threaded.

The permissive tier still fixes the same-file relative↔absolute form mismatch that made the live run pass; it can no longer widen scope.

**Tests (step 4):**
- `test_files_touched_rejects_absolute_outside_workspace_claim_with_touch` — bot's repro → reject
- `test_files_touched_rejects_parent_traversal_claim_escaping_cwd` — `../` escape → reject
- `test_files_touched_accepts_in_workspace_relative_vs_absolute_edit` — in-workspace form mismatch → accept
- `test_files_touched_task_cwd_none_rejects_touch_command_text` — unknown cwd: `touch` text → reject, Edit event → accept, near-miss → reject
- All pre-existing file/evidence guards still pass.

**Validation:**
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — clean (429 files)
- `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`) — **2313 passed**
- **Live gate still 1/1**: `Success: 1/1`, `[COMPLETED]`, session `orch_e76947a9925c` (its files are in-workspace, so it passes Tier A cleanly).

Q00:
Real catch — the declared-verify_command success check could be satisfied by assistant narration. Fixed in `4be459f7`.

**Root of the hole:** the chunk loop appends following messages (assistant turns with `tool_name=None` included), and `_message_contains_test_success` reads `AgentMessage.content`. So a worker could invoke the declared command with **no** `exit_code` / **no** tool result and let a following assistant message ("the tests passed") self-report success.

**Fix — runtime-signal-only success:** new `_message_has_runtime_success_signal(message)` accepts success ONLY from runtime-produced evidence:
- a structured **non-error** `exit_code == 0`, or `subtype == "success"`, or `status ∈ {completed, success, succeeded}`;
- test-success text from **runtime output fields** via `_runtime_message_test_proof_text`, which includes `message.content` *solely* for tool-result messages the runtime emits — **never** assistant narration.

The verify_command chunk check now uses this instead of `_message_contains_test_success`. Assistant `content` from a non-result turn can no longer satisfy the declared-command success.

- `python3 -c "...assert..."` matches declared command + narration-only "passed", no `exit_code`/tool-result → **REJECT** (your repro)
- declared command + structured `exit_code 0` (or a tool-result with exit 0 / "1 passed" runtime text) → **ACCEPT**

Verified directly: narration-only → reject; bash `exit_code 0` → accept; tool-result exit 0 → accept; tool-result test-success text → accept; non-matching command + exit 0 → reject.

**Tests:**
- `test_atomic_verifier_accepts_declared_verify_command_with_structured_exit_code` — runtime exit_code 0 → accept
- `test_atomic_verifier_rejects_declared_verify_command_success_only_via_narration` — narration-only success → reject (your repro)
- existing declared/prose/legacy/guard cases unchanged.

**Validation:**
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — clean (429 files)
- `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`) — **2315 passed**
- **Live gate still 1/1**: `Success: 1/1`, `[COMPLETED]`, session `orch_1e4aa485ba27` — the seed's declared verify_command exits 0 with a real Bash tool-result (`exit_code 0` / stdout `OK`), so the runtime-signal check accepts it.

Q00:
Right — `completed != passed`. Fixed in `764bda64`.

**Root of the hole:** `_message_has_runtime_success_signal` treated `status ∈ {completed, success, succeeded}` as success, but for a Bash tool `status="completed"` only means the command **finished** (including a non-zero exit). A failing verify_command with a generic "completed" status + `AssertionError` in output + no `exit_code` satisfied fat-harness.

**Fix — unambiguous success, and reject failure output:**
- Dropped the generic `status` check entirely.
- Success now requires: structured non-error `exit_code == 0` (non-zero exit → explicit failure), **or** a trusted tool-result `subtype == "success"`, **or** test-success text from runtime output fields — **and** only when the runtime output carries **no failure marker** (`_RUNTIME_FAILURE_MARKER_RE`: `AssertionError`, `Traceback`, `Exception`, `panic`, `exit code/status [1-9]`, `non-zero exit`).
- `exit_code == 0` short-circuits (strongest signal); `exit_code` non-zero → reject even if output says "passed".
- The failure-marker regex deliberately omits bare `\berror\b` so it doesn't reject "0 errors" / "Errors: 0" zero-failure summaries — it targets the exception-shaped markers that `_text_contains_test_success`'s word-boundary scan misses (e.g. `AssertionError`).

Verified directly (9 cases): exit 0 → accept; `completed`+`AssertionError` no exit_code → **reject** (your repro); `completed` only, no proof → reject; tool-result `subtype=success` → accept; narration-only → reject; non-matching command → reject; `exit_code 1` + "1 passed" text → reject; tool-result "1 passed" runtime text → accept; live shape (`OK`, exit 0) → accept.

**Tests:**
- `test_atomic_verifier_rejects_declared_verify_command_completed_status_with_failure` — your repro → reject
- `test_atomic_verifier_accepts_declared_verify_command_with_tool_result_success_subtype` — success subtype → accept
- existing exit_code-0 accept / narration reject / non-match reject cases unchanged.

**Validation:**
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — clean (429 files)
- `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`) — **2317 passed**
- **Live gate still 1/1**: `Success: 1/1`, `[COMPLETED]`, session `orch_facea6a64ac2` — the declared verify_command exits 0 via a real Bash tool-result, so the runtime-signal check accepts it.

Q00:
Right — prefix match defeats the oracle. Fixed in `ba20011b` with a dedicated normalized-equality matcher.

**Root of the hole:** the oracle reused `_runtime_message_supports_command_claim`, whose test-command fallback (`claims.py:211-212`) accepts any transcript command that *starts with* the claimed invocation. So `verify_command: pytest` was satisfied by `pytest --collect-only` / `--help` / `-k nope` / `--fixtures` (all exit 0, run no tests).

**Fix — `_commands_are_equivalent` (normalized EQUALITY, not prefix):**
- Unwrap shell wrappers (`zsh/bash/sh -lc/-c`) on the transcript side; strip outer quotes; collapse whitespace; lowercase — via `_command_equality_forms`.
- Add an argv-normalized spelling so quote-only differences (`--tests "X"` vs `--tests X`) compare equal (still the same command).
- Require the declared and transcript form-sets to **intersect** — no `startswith`, no plumbing tolerance. A mode-changing flag (`--collect-only`, `-k`, `-n`, `--help`, `--fixtures`) or a neutering redirect makes it a **different** command → reject.
- A benign `cd`/env preamble is tolerated via the safe-preamble inner command (equality on the inner, not a prefix).
- The loose `_runtime_message_supports_command_claim` stays only on the strict formal-runner path; the oracle no longer touches it.

Verified directly: `pytest` vs `pytest` / wrapped `pytest` → accept; vs `--collect-only`/`--help`/`-k nope`/`--fixtures`/`-n 4` → reject; inline `python3 -c "..."` verbatim (wrapped) → accept; vs different-arg / `assert True` → reject; `--tests "X"` vs `--tests X` → accept; `> /dev/null` → reject.

**Tests:**
- `test_atomic_verifier_rejects_pytest_oracle_prefix_or_mode_variants` (parametrized: `--collect-only`/`--help`/`-k nope`/`--fixtures`/`-n 4 --collect-only`) → all reject
- `test_atomic_verifier_accepts_pytest_oracle_exact_and_wrapped` → exact + shell-wrapped accept
- prior exit_code / narration / completed-status / inline-verbatim cases unchanged.

**Validation:**
- `ruff check .` / `ruff format --check .` — clean
- `mypy src/` — clean (429 files)
- `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`) — **2323 passed**
- **Live gate still 1/1**: `Success: 1/1`, `[COMPLETED]`, session `orch_6e2e1edc28e9` — the declared verify_command was run verbatim (wrapped), unwrap+normalize equal → accept.

@coordinator this closes the match-precision class. Boundary status: fabrication (`tests_passed`) is now oracle = declared verify_command, matched by **equality** (not prefix), with success requiring a **runtime** signal (`exit_code==0` / success subtype / clean success text, never a bare completion status or assistant narration); scope (`files_touched`) is workspace-scoped. If the bot surfaces yet another oracle-match corner, I'll flag it — but I believe we're at the boundary's floor now.

Q00:
## Architecture pivot: transcript reconstruction → authoritative execution (`e6eee505`)

The repeated `tests_passed` findings were all symptoms of one wrong approach: reconstructing "did a test pass" from the worker's transcript. The decisive evidence was the live gate — a worker legitimately verifies with its **own sample** (`greet('World')`) rather than the declared `verify_command`'s sample (`greet('x')`), so an exact oracle is flaky and a fuzzy one is a fabrication vector. There is no stable floor in transcript matching.

**The authoritative, non-fabricatable check already existed:** `_run_ac_verify_gate` runs the **declared** `verify_command` itself (real subprocess, real exit code + `output_assertion`). The bug was ordering — it ran only *after* the fragile fat-harness `tests_passed` reconstruction, so that reconstruction's failure pre-empted the real check.

**Fix:**
- **Contract-carrying AC** (declares `verify_command`) → `tests_passed` is dropped from the fat-harness required-evidence set (`_effective_evidence_schema_for_ac` gains `has_success_contract`). With `files_touched` + `commands_run` backed, `result.success=True` → the AC reaches `_apply_verify_gate` → it **executes the declared command** → its exit code decides.
- **Legacy AC** (no `verify_command`) → `tests_passed` reverts to the original pre-PR strict behavior (formal test runners only). **The entire inline-verification / oracle-equality / argv / runtime-success machinery is removed** — `test_detection.py` and `shell_parsing.py` are back to `origin/main`. The `tests_passed` matcher surface is gone.
- **Kept:** `claims.py` `files_touched` path-normalization + workspace-scoping (the real maintainer fix: `hello.py` vs `/private/tmp/.../hello.py`), and the `commands_run` wrapper-unwrap.

**Ordering verified in code + live:** `_apply_verify_gate` runs per-result gated on `result.success` (parallel_executor.py:3372); dropping `tests_passed` lets a contract AC pass fat-harness and reach the gate.

**Mandatory live proofs (real codex backend):**
- POSITIVE **2×, not flaky**: `orch_ef65ea6e0910`, `orch_1d186b82b8c1` → **1/1**. The orchestrator runs the declared `greet('x')` command; worker sample variance (`'World'`) is now irrelevant. (Bonus: an intermediate run showed the worker *bending its code* to satisfy the declared command — direct evidence it's the oracle.)
- NEGATIVE (delegation **enforced**, not bypassed): declared `verify_command: python3 -c "import sys; sys.exit(1)"` → **0/1**, `execution.verify.failed` event with `reason="verify_command exited with status 1"` (`orch_8454aa979783`). A broken/unsatisfiable declared command fails the run via the gate regardless of the worker's self-reported `tests_passed`.

**Tests:** `_effective_evidence_schema_for_ac` drops `tests_passed` for contract ACs; contract-AC verifier delegates `tests_passed`; legacy-AC verifier keeps strict formal-runner semantics; `_apply_verify_gate` flips a contract AC to FAILED when the declared command exits non-zero (focused delegation-enforcement test); `files_touched` normalization/scoping retained.

**Validation:** `ruff check .` / `ruff format --check .` clean; `mypy src/` clean (429 files); `tests/unit/orchestrator/` (excl. pre-existing-hanging `test_runner.py::test_execute_seed_success`) — **2313 passed**.

PR #1592, opened 2026-07-08
fix(orchestrator): ground contract-AC evaluation in authoritative execution, not transcript reconstruction

Q00:
## Summary
Completes the #1591 architecture pivot: for a **contract-carrying AC** (its spec declares a `verify_command`), the fat-harness stops reconstructing evidence from the worker transcript and defers to the orchestrator's own authoritative checks. This closes the release blocker where a correct `ooo run` / `ooo auto` on the **codex** backend was rejected as `FABRICATION_SUSPECTED` even though the produced code was correct.

### What changed
1. **`files_touched` → filesystem oracle.** When a contract AC also declares `expected_artifacts`, `files_touched` is dropped from the fat-harness required-evidence set (mirroring #1591's `tests_passed` drop). The verify-gate's `_missing_expected_artifacts` (real `os.path.exists` check in the run workspace) is the oracle — so it no longer matters whether the worker wrote files via a structured `Edit` event or a bash heredoc / `apply_patch`. Contract ACs **without** `expected_artifacts` keep transcript-backed `files_touched` (no filesystem oracle to delegate to). Strictly stronger than transcript reconstruction.
2. **`output_assertion` normalization.** Condition / exit-code phrases (`exit code 0`, `exit 0`, `returns 0`, `success`, `passed`, `no errors`, …) are normalized to `None` via a `fullmatch` validator, so a redundant, impossible-to-satisfy assertion no longer strands a correct run. Only literal stdout (`OK`, `5 passed`) survives. Exit-code 0 is already gated by `verify_command`'s own exit status.
3. **Seed authoring.** `seed-architect` now emits single-line verify commands and literal-only `expect` values; seed generation rejects heredoc/multiline `verify_command` and retries the extraction.

### Anti-fabrication is preserved / strengthened
Dropping `files_touched` only removes a *self-reported* evidence check. The authoritative `_apply_verify_gate` → `_run_ac_verify_gate` runs on **all** completion paths (initial dispatch, same-runtime retry, alt-harness redispatch) and independently flips `result.success = False` when a declared artifact is missing on disk or the `verify_command` exits non-zero. A worker cannot pass by touching nothing.

### Validation
- `ruff check .` / `ruff format --check .` / `mypy src/` — clean (429 source files)
- Isolated PR tests: `test_seed.py` + `test_seed_generator.py` + `test_parallel_executor.py` + `test_parallel_executor_verify_by_default.py` → **383 passed**
- Live (real codex backend, full working tree): `ooo run workflow ...` **3/3**; fresh `ooo auto ... --timeout 1200` **4/4, Artifact state: complete_verified**; run/auto event DB shows **0** occurrences of `unsupported evidence claims` / `output_assertion ... not found` / `Fat-harness verifier failed`.

### Scope
Evidence-gate + seed-authoring correctness only. CLI aliases, the `ooo seed` command, the auto run-wait reconciliation, and the Seed-QA prompt refinements ship in a companion PR. Does not touch `src/ouroboros/auto/`.

### DIFF
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -22,6 +22,16 @@ Specific, measurable criteria for success.
 Format: one `AC:` line per criterion
 Example: "AC: Tasks can be created | verify: python -m pytest tests/test_tasks.py | artifacts: NONE | expect: created task"
 
+`verify` / `verify_command` semantics:
+- Use exactly one single-line shell command.
+- NEVER use heredoc or multiline shell syntax such as `<<`, `<<'PY'`, `cat <<EOF`, line-continuation scripts, or an unterminated command block. The AC contract format is one line, so multiline command bodies will be lost.
+- For Python snippets, use `python -c "..."` / `python3 -c "..."`; for longer checks, require a pytest-discoverable test artifact and use `python -m pytest -q`.
+
+`expect` / `output_assertion` semantics:
+- Use `expect` ONLY for a literal string that the verify command prints to stdout verbatim, such as `OK` or `5 passed`.
+- NEVER use a condition, status, or exit-code description such as `exit code 0`, `exit 0`, `returns 0`, `success`, `no errors`, `passed`, or `passes`.
+- If the command has no distinctive stdout literal to assert, write `expect: NONE`. Exit-code 0 is already verified separately by the runner.
+
 **Granularity contract (read carefully):**
 - Produce **3-7** acceptance criteria. Each criterion is **one independently valuable, user-visible outcome** — not an implementation step.
 - Do **NOT** pre-decompose criteria into executable sub-tasks. Splitting work into atomic units is the execution engine's job at runtime; doing it here multiplies token cost with no benefit.
@@ -82,6 +92,7 @@ Few-shot examples:
 
 ```
 ACCEPTANCE_CRITERIA:
-AC: `python -m pytest tests/test_tasks.py` passes for task create/list flows | verify: python -m pytest tests/test_tasks.py | artifacts: NONE | expect: passed
+AC: Task create/list flows pass automated verification | verify: python -m pytest tests/test_tasks.py -q && echo OK | artifacts: NONE | expect: OK
+AC: Greeting import check prints OK | verify: python -c "from hello import greet; assert greet('Alice') == 'Hello, Alice'; print('OK')" | artifacts: hello.py | expect: OK
 AC: README documents the CLI usage examples | verify: NONE | artifacts: README.md | expect: NONE
 ```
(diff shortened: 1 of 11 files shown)


Q00:
Fixed the reproduced BLOCKER (delegation with the verify gate disabled).

**Root cause**: dropping `files_touched`/`tests_passed` from the fat-harness required set delegates them to `_run_ac_verify_gate` — but `_apply_verify_gate` returns early when `run_verify_commands=False` (`parallel_executor.py:3393`). So with the gate off, a contract AC declaring `expected_artifacts=("hello.py",)` could complete on `commands_run` alone with no artifact on disk: the transcript check was dropped AND the replacement gate never ran.

**Fix**: a new `verify_gate_active` flag (threaded from `self._run_verify_commands`) now gates **both** delegation drops — the `files_touched` drop *and* the pre-existing `tests_passed` drop. Default is `False` (fail-safe: never delegate unless the gate is confirmed active). With `run_verify_commands=False` the transcript-backed evidence is retained (the pre-delegation behavior), so a missing / self-reported artifact is rejected.

Threaded `verify_gate_active` through every site that already forwards `has_expected_artifacts`:
- `evidence/ac_classification.py`: `_effective_evidence_schema_for_ac` (both drops now require the flag), `_out_of_scope_evidence_fields_for_ac`, `_out_of_scope_evidence_values_for_ac`, `_scoped_evidence_record_for_ac`
- `evidence/verification.py`: `_verify_atomic_evidence_against_runtime_messages`
- `parallel_executor.py`: computed at the atomic acceptance boundary (`verify_gate_active = self._run_verify_commands`) and passed through `_observe_atomic_typed_evidence`, `_run_atomic_verifier_pass`, `_emit_atomic_typed_evidence_event`, the `_verify_atomic_...` wrapper, and their internal schema/scoped/out-of-scope calls
- `atomic_prompt_builder.py`: `verify_gate_active = executor._run_verify_commands` so the worker's displayed required-fields stay correct when the gate is off

The content-based `_is_validation_only_ac` / `_is_documentation_only_ac` drops are unchanged (they are not gate delegations).

**Tests**: updated the direct schema/verifier delegation tests to pass `verify_gate_active=True`; added schema tests (`verify_gate_active=False` → `files_touched`+`tests_passed` retained; `True` → both dropped); added an integration test mirroring `test_contract_ac_with_artifacts_accepts_unbacked_files_touched_claim` but with `run_verify_commands=False` and no artifact written, asserting the AC is **not** accepted and both fields stay required.

Validation: ruff check/format, mypy `src/` (429 files), and `test_parallel_executor.py` + `test_parallel_executor_verify_by_default.py` + `test_seed.py` + `test_seed_generator.py` (386 passed) all green.

PR #1597, opened 2026-07-08
perf(brownfield): bound scan depth, drop worktree expansion, parallelize validation

### LINKED ISSUE #1596: brownfield scan: unbounded-depth walk, serial git subprocesses, and noisy worktree expansion
## Background

`ooo brownfield` scans the filesystem for existing git repositories and worktrees and registers them so users can pick "default repos" that become context for later workflows. The scan is implemented by `scan_home_for_repos` in `src/ouroboros/bigbang/brownfield.py`, which walks the user's home directory, and for every repo it finds it shells out to `git` to validate the path and to expand the repo into its linked-worktree family.

This registry is not cosmetic — its output is load-bearing for two consumers:
- **PM interview** loads the registered default/all repos as codebase-exploration context (primary consumer).
- **`evaluate`** falls back to the registered default repo as the working directory that gates Stage 1/Stage 2 (`_resolve_evaluate_working_dir`).

Regular interview, the auto pipeline, and seed execution use cwd-/seed-context-based brownfield detection and do **not** read this registry, so the blast radius of changing the scan's discovery rules is limited to the two consumers above.

### Why the scan does not need every worktree
The intended consumer of this registry is `ooo pm`. A PM interview only needs **one main repo as its reference** — it does not need the full linked-worktree family of a repo enumerated. So exhaustively registering every worktree (via `git worktree list` expansion) adds no value for the consumer, while it does add cost (a subprocess per repo) and noise (internal/hidden worktrees) to a registry whose consumer just wants the main repo. This is the core reason the expansion should go.

## Problem awareness

The scan is slower than it needs to be and, more importantly, it registers worktrees that are noise for those consumers. Inspecting a real machine (via `git worktree list` across the tracked repos) made the second point concrete: the only worktrees the current expansion logic *uniquely* contributes — beyond what the plain filesystem walk already finds — are Ouroboros's own internal orchestration worktrees under `~/.ouroboros/worktrees/**` (some already `prunable`) and worktrees parked under hidden `.worktrees/` dirs. Those are exactly the paths a user would never want as interview/evaluate context, yet the scan pays a per-repo subprocess cost to discover and inject them. The real work worktrees (`~/podo-backend-rfc-*`, etc.) already live inside the scan root and are found by the walk itself, so the expansion adds cost without adding value in the common case.

### 1. Unbounded-depth walk over the entire home directory
`os.walk(root)` traverses the full `~/` tree with no depth limit. It prunes at repo roots, skip dirs, and dot dirs, but still descends arbitrarily deep into non-repo subtrees. In practice registered repos live at `~/repo` (depth 1) or `~/group/repo` (depth 2); deeper nesting is rare and not worth the traversal cost.

### 2. Serial `git` subprocess calls per repo
For every discovered repo the scan runs `git` subprocesses **serially**:
- `_is_git_worktree` (`git rev-parse --is-inside-work-tree`) for validation, plus
- `_list_git_worktrees` (`git worktree list --porcelain`) for expansion.

Each spawn costs ~10–50ms and they run one after another, so wall-clock grows linearly with the number of repos. These calls are process-bound (the GIL is released during `subprocess`), so they are trivially parallelizable but currently are not.

### 3. Redundant double validation of the same path
`_add_repo_and_worktrees` validates `repo_path` with `_is_git_worktree` at its guard, then validates it **again** inside the candidate loop (the repo path is `candidates[0]`). Every found repo pays for two identical `git rev-parse` calls.

### 4. `git worktree list` expansion pulls in out-of-scope / internal worktrees
When a normal repo root is found, the scan expands it via `git worktree list --porcelain` and registers linked worktrees **even outside the scan root** — which, as noted above, only uniquely surfaces internal/hidden worktrees while charging a per-repo subprocess cost. Given that the consumer (`ooo pm`) only references a single main repo, enumerating the whole worktree family is unnecessary by design, not just inefficient.

## Proposed fix
1. Bound the walk to a shallow depth (default 2 levels below the scan root).
2. Drop the `git worktree list` expansion — register only repos/worktrees the walk reaches directly (self-only).
3. Split the scan into two phases: a cheap subprocess-free walk that collects candidate paths, then parallel validation of those candidates via a thread pool.
4. Validate each candidate exactly once.

This keeps all real work repos/worktrees (they live within the depth-bounded walk) while dropping internal/hidden worktrees, and removes the serial-subprocess and double-validation costs.

cohemm:
Closes #1596

## Purpose

The brownfield repo scan (`scan_home_for_repos`) is slower than necessary and pollutes its registry with worktrees that are noise for its consumers. It walked the entire home tree with no depth limit and ran serial `git` subprocesses (validate + `worktree list` expansion + a redundant second validate) per repo. The `git worktree list` expansion only uniquely surfaced internal/hidden worktrees (`~/.ouroboros/worktrees/**`, `.worktrees/`) that no user wants as interview/evaluate context.

Goal: make the scan fast and make its output contain only the repos/worktrees users actually work in. The intended consumer is `ooo pm`, and a PM interview only needs **one main repo as its reference** — it does not need a repo's full linked-worktree family enumerated. So exhaustively registering every worktree adds no value for the consumer while adding cost and noise; dropping that is by design, not just an optimization. The registry is load-bearing for **PM interview** (exploration context) and **`evaluate`** (working-dir fallback), so both correctness and cost matter. Full problem analysis and blast radius in #1596.

## What changed & why

- **Depth-bounded walk (default 2 levels).** Registered repos live at `~/repo` (depth 1) or `~/group/repo` (depth 2); deeper nesting is rare. Bounding the walk stops it from descending arbitrarily deep into non-repo subtrees. Exposed as `max_depth` for callers that need more reach.
- **Dropped `git worktree list` expansion.** Removed `_list_git_worktrees` / `_add_repo_and_worktrees`. Each repo/worktree is registered self-only when the walk reaches it directly. Intent: `ooo pm` only references a single main repo, so enumerating a repo's whole worktree family is unnecessary by design. Real work worktrees already live inside the bounded walk and are still found; the expansion only added internal/hidden worktrees (noise) at a per-repo subprocess cost, so removing it improves both signal and speed.
- **Two-phase, parallel scan.** Phase 1 is a subprocess-free `os.walk` that only collects candidate `.git` paths; Phase 2 validates them in parallel via a `ThreadPoolExecutor`. Intent: the `git` validation releases the GIL during `subprocess`, so overlapping the process-spawn cost turns a serial O(N) cost into near-constant wall-clock.
- **Single validation per candidate.** The old path validated the same repo twice (guard + `candidates[0]`); now each candidate runs `git rev-parse` exactly once.
- **Docs.** Updated `skills/brownfield/SKILL.md` scan-boundary section to match the new depth-bounded, no-expansion behavior.

Out of scope / unaffected: regular interview, the auto pipeline, and seed execution use cwd-/seed-context brownfield detection and do not read this registry.

## Success criteria

- All brownfield unit tests pass: `tests/unit/bigbang/test_brownfield.py` + `tests/unit/mcp/tools/test_brownfield_handler.py` → **135 passed**.
- New tests assert the behavior contract: depth-2 repo discovered, depth-3 repo excluded, `max_depth` configurable, and worktrees under hidden dirs are not expanded.
- A repo/worktree that lives within the depth-bounded walk (e.g. `~/podo-backend-rfc-*`) is still registered; internal worktrees under `~/.ouroboros/worktrees/**` are not.
- `ruff check` clean.
- Real `~/` smoke run completes in ~0.17s for 79 repos, including depth-2 repos (`Developer/browser-harness`, `grape-worktrees/*`).

Accepted trade-offs: repos deeper than 2 levels are no longer discovered by the default scan (raise `max_depth` if needed), and internal/hidden worktrees are intentionally excluded.

### DIFF
--- a/skills/brownfield/SKILL.md
+++ b/skills/brownfield/SKILL.md
@@ -33,8 +33,7 @@ Show scanning indicator:
   Scanning for Existing Projects...
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-Looking for git repositories and worktrees under the scan root directory.
-Linked worktrees reported by discovered normal repo roots may also be registered, even outside the scan root directory.
+Looking for git repositories and worktrees up to two directories below the scan root.
 Local repos and repos with any remote name are eligible.
 This may take a moment...
 ```
@@ -55,7 +54,7 @@ stop with the MCP-not-available message instead of retrying the failing call.
    Tool: ouroboros_brownfield
    Arguments: { "action": "scan" }
    ```
-   This walks `scan_root` for valid seed repos/worktrees and registers them in DB. For each discovered normal repo root with a `.git` directory, Git-reported linked worktrees are also considered, even when they live outside `scan_root`. A linked worktree found under `scan_root` with a `.git` file is registered itself, but it is not used to register its main worktree or sibling worktrees outside `scan_root`. Existing defaults are preserved.
+   This walks `scan_root` (up to two directory levels deep) for valid seed repos/worktrees and registers them in DB. Each repo or worktree found directly by the walk is registered self-only — Git worktree families are not expanded, so worktrees outside the depth-bounded walk (e.g. under `.ouroboros/worktrees`) are not pulled in. Existing defaults are preserved.
 
 The scan response `text` already contains a pre-formatted numbered list with `[default]` markers. **Do NOT make any additional MCP calls to list or query repos.**
 
@@ -81,10 +80,9 @@ Then stop.
 ### Scan boundaries
 
 - The filesystem walk starts at `scan_root`; when omitted, `scan_root` defaults to the current user's home directory.
-- Repositories are only discovered directly by walking directories inside `scan_root`.
+- Repositories are discovered by walking directories inside `scan_root`, at most two levels deep (so `~/repo` and `~/group/repo` are found; deeper nesting is not).
 - Dot-prefixed directories and known noisy directories such as `node_modules` are not walked as seed locations.
-- Git worktrees are different: once a normal repo root with a `.git` directory is discovered, Ouroboros runs `git worktree list --porcelain` and may register those linked worktrees even if their paths are outside `scan_root`.
-- A linked worktree found inside `scan_root` with a `.git` file is registered itself, but it is not used to register its main worktree or sibling worktrees outside `scan_root`.
+- Both normal repos (`.git` directory) and linked worktrees (`.git` file) are registered when the walk reaches them. Git worktree families are NOT expanded — a worktree is only registered if the walk finds it directly, not because its main repo's Git metadata reports it.
 - Local repos, repos without remotes, and repos whose remotes are not named `origin` are all eligible.
 
 **Step 2: Default Selection**
--- a/src/ouroboros/bigbang/brownfield.py
+++ b/src/ouroboros/bigbang/brownfield.py
@@ -4,8 +4,7 @@ Manages the global brownfield registry in ``~/.ouroboros/ouroboros.db``
 via :class:`~ouroboros.persistence.brownfield.BrownfieldStore`.
 
 Business-level operations:
-- Scan-root discovery for valid seed git repos/worktrees
-- Linked worktree discovery from normal repo roots via Git metadata
+- Scan-root discovery for valid seed git repos/worktrees (depth-bounded walk)
 - README/CLAUDE.md parsing for one-line description generation (Frugal model)
 - Async CRUD delegated to BrownfieldStore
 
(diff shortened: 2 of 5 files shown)


Q00:
@cohemm can you resolve the non blocking issue and follow-up?

cohemm:
@Q00 Resolved the review follow-up and the non-blocking suggestion in d40bd09d.

The scan behavior was already correct in code — these were stale docs/comments still describing the removed worktree expansion. Fixed all of them (docs/comments only, no behavior change):

- **Follow-up #1** — `src/ouroboros/bigbang/brownfield.py` module docstring: removed the "Linked worktree discovery from normal repo roots via Git metadata" line.
- **Non-blocking** — `src/ouroboros/mcp/tools/brownfield_handler.py`: updated the scan action docstring, `_handle_scan` docstring, and the inline workflow comment so none of them claim worktree expansion anymore.
- **"Not met" requirement (contract truthfulness across surfaces)** — `src/ouroboros/cli/commands/setup.py`: fixed `_scan_and_register_repos` / scan-command docstrings and the scan spinner label to describe the depth-bounded, self-only walk.

Verified: 135 brownfield unit tests pass, `ruff check` clean.

PR #1601, opened 2026-07-11
feat(orchestrator): progressive model-tier escalation per retry

Q00:
## Summary
- Model-tier escalation is now progressive: one notch per retry at-or-past `economics.escalation_threshold`, capped at frontier (was: single notch regardless of retry count)
- With `escalation_threshold: 1` a decomposed child ladders **frugal → standard → frontier** across its three attempts — the last attempt always runs the strongest tier before cross-harness redispatch takes over
- Shipped default (`threshold: 2`) keeps the first escalation step byte-identical to v0.50.3; only 4th+ attempts (non-default retry budgets) climb further
- Effort routing deliberately keeps its single-notch rule (documented asymmetry)

## Test plan
- 83 new/updated assertions in test_model_routing.py (ladder tables for thresholds 1/2, frontier ceiling cap, composition with child-drop and suggested_tier)
- Full routing/proof suite: 243 passed; mypy/ruff clean

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -270,7 +270,7 @@ Configures the PAL Router (Progressive Adaptive LLM): cost tiers, escalation on 
 ```yaml
 economics:
   default_tier: frugal          # "frugal" | "standard" | "frontier"
-  escalation_threshold: 2       # Consecutive failures before upgrading tier
+  escalation_threshold: 2       # Retry attempt escalation begins; then +1 tier per retry
   downgrade_success_streak: 5   # Consecutive successes before downgrading tier
   tiers:
     frugal:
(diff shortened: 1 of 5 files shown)


Q00:
Addressed the P2 finding — and it turned out to expose a real production gap, not just a missing test.

**What the finding surfaced:** the retry early-stop's pending-escalation predicate used the arithmetic proxy `retry < escalation_threshold`, which only defeats early-stop for the single threshold crossing. A decomposed child starts one tier below top-level, so its progressive ladder tops out one retry PAST the threshold: with `threshold=2` the child ladder is frugal → frugal → standard → frontier (retry 0..3), but early-stop fired after the retry-2 standard dispatch and the retry-3 frontier dispatch could never happen.

**Fix (ladder-truth instead of proxy):** the predicate now asks the router directly whether the NEXT scheduled retry resolves to a different enforced model than the one just dispatched (`decide_model` recomputed for both attempts, `is_decomposed_child` read from the dispatched result). Escalation stays pending until the resolved model stops climbing; once the ladder is exhausted, identical-class early-stop resumes exactly as before. Dormant router / non-NATIVE runtimes degrade to the old behavior.

**Tests (executor-level, as the non-blocking note requested — doc claim now cites them):**
- `test_retry_decomposed_child_reaches_retry3_frontier`: ac_retry_attempts=3, threshold=2, identical failure class every attempt → 4 dispatches, final at frontier (was a strict xfail documenting the gap; now a normal passing test)
- `test_retry_top_level_walks_whole_ladder_to_frontier`: 3 dispatches then early-stop resumes (no 4th)
- exhausted-ladder and dormant-router early-stop behavior pinned unchanged

376 passed on the touched suites; mypy/ruff clean.

PR #1602, opened 2026-07-11
feat(dashboard): surface model-tier routing and token telemetry in TUI and web

Q00:
## Summary
The frugality telemetry shipped in v0.50.3 was invisible: `execution.ac.model_routed`, `execution.ac.token_attribution.reported`, and `execution.frugality_proof.evaluated` were dropped at BOTH display gates (web SSE reader allowlist + TUI `create_message_from_event` fall-through).

**Web Kanban**
- Per-card model-tier/model badge (latest routing wins, so an escalated retry updates the badge) + compact token count
- Header: run-total tokens (`Σ`) + one-line frugality-proof verdict
- All folded in the shared `reduce_board` reducer with incremental-fold/full-reduce agreement preserved

**TUI (dashboard_v3)**
- AC tree: tier/model badge appended next to the existing provider badge (same order-safe stamp path via `_apply_provider_tags`)
- Node detail panel: Model (id + tier + enforced/advised) and Tokens rows
- Progress bar: run-total tokens; frugality verdict line at run end

## Test plan
- Web: 61 passed (kanban reducer cases per event, page render, shared board web/TUI agreement tests extended)
- TUI: 191 unit + 375 tui/tree-scoped passed; mypy clean on all touched files
- Defensive payload handling mirrored from the provider ledger (missing/malformed → omit, never render None)

### DIFF
--- a/src/ouroboros/dashboard/board.py
+++ b/src/ouroboros/dashboard/board.py
@@ -51,6 +51,12 @@ _TOOL_EVENTS = frozenset(
         "execution.coordinator.tool.started",
     }
 )
+_TELEMETRY_EVENTS = frozenset(
+    {
+        "execution.ac.model_routed",
+        "execution.ac.token_attribution.reported",
+    }
+)
 
 
 # Terminal statuses: once an AUTHORITATIVE event sets one, a node is finished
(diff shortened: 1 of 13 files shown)


Q00:
Addressed the blocker. The finding was accurate: the TUI polled only exact-match `(execution, id)`/`(session, id)` aggregates while the per-AC telemetry producers persist under worker-scoped aggregate ids (`exec_<id>_node_<NODEID>`) with the run id only in payload fields — so the new handlers were non-functional for real decomposed runs.

**Fix:** the subscription now uses the store's related-events query with the exact predicates the web Kanban reader applies (aggregate exact-match OR `json_extract(payload,'$.execution_id'/'$.session_id')`), advancing a single rowid cursor across all matched scopes — exactly-once delivery, still O(new_events) per poll. Bonus hardening: `_session_related_event_conditions` no longer emits a `payload.session_id == ''` predicate when the session id is empty (blank-field rows could match unrelated runs).

**Subscription-level regression added** (as requested): persists a `model_routed` event under a worker-scoped aggregate id with the run id only in the payload, drives the real poll boundary, and asserts the `ACModelRouted` message is produced.

Gates: tui+dashboard+dashboard_web 253 passed; broad `-k "tui or tree or event_store"` 479 passed; deliver-gate consumers of the touched store helper 38 passed; mypy/ruff clean.

Q00:
All three findings addressed.

**Blocker (fail-open blank scope):** confirmed — my prior hardening removed the blank-session predicates but left `or_(*conditions)` consuming a possibly-empty list, which emits no WHERE clause and returns the whole store. Both `query_session_related_events` and `query_session_related_events_after` now fail closed: zero scope predicates → zero rows, and the incremental variant returns its cursor unchanged so the poll stays idempotent. `query_execution_related_events` needs no guard (its three predicates are unconditional). Regressions cover both variants with unrelated rows seeded, plus the motivating TUI case (blank session + real execution_id still returns that execution's related rows).

**Follow-up + non-blocking (TUI token parser):** the parser now mirrors `fold_telemetry_event`'s finite-number guard — bool/NaN/±Inf/negative `token_spend` is rejected at parse time (no message produced) instead of relying on downstream negative filtering. Test cases added next to the existing malformed-payload cases.

Gates: persistence+tui+deliver-gate 292 passed; broad `-k "tui or tree or event_store"` 487 passed; mypy/ruff clean.

PR #1603, opened 2026-07-12
fix(dashboard): reject negative token_spend; refresh stale retry-ladder doc

Q00:
## Summary
Follow-up on PR review feedback from #1601 / #1602:

- **`board.py` `_coerce_spend()`**: reject negative `token_spend` (malformed for a cumulative counter) to mirror the TUI's parse-time guard in `tui/events.py`. Previously a stale/malformed `execution.ac.token_attribution.reported` event could make the web/shared board under-report runtime token spend while the TUI rejected the same event — a real divergence between the two surfaces. `token_reduction_pct` (a signed delta where negative legitimately means spend increased, per `test_frugality_proof.py::test_..._-5.0`) keeps its existing behavior via a new `reject_negative` kwarg, so this doesn't regress that case.
- **`docs/config-reference.md:321`**: refreshed stale wording — `test_retry_decomposed_child_reaches_retry3_frontier` has no `xfail` marker and currently asserts the decomposed child *reaches* the frontier tier (4th dispatch), not that it stops one rung short.

**Investigated, not changed**: the review also flagged that `board.py`'s telemetry fold lacks the `ac_index` fallback join key the TUI supports when `node_id` is absent. Traced both production emitters (`ParallelACExecutor`'s `execution_event_emitter` and `runner.py`'s direct-run path) — `node_id` is always populated whenever `ac_index` carries a meaningful value on the paths that emit these two event types, so the TUI's fallback is unreachable defensiveness rather than a real event contract. No change made to avoid an extraneous/speculative diff.

## Test plan
- [x] `uv run pytest tests/unit/dashboard tests/unit/dashboard_web tests/unit/tui -q` — 259 passed
- [x] `uv run pytest tests/unit/orchestrator/test_frugality_proof.py -q` — confirms negative `token_reduction_pct` still accepted (20 passed)
- [x] New regression test: `test_negative_spend_is_rejected_like_the_tui_parser`
- [x] `ruff check` / `ruff format --check` / `mypy` on changed files — clean


Q00:
Addressed the blocking finding in commit c1117985:

- `board.py:_coerce_spend()` now wraps `float(value)` in `try/except (OverflowError, ValueError)`, matching the existing project convention in `orchestrator/frugality_proof.py::_finite_number`. Verified the exact repro from the review no longer crashes: `reduce_board([{"event_type": "execution.ac.token_attribution.reported", "payload": {"node_id": "n1", "token_spend": 10**10000}}])` now returns `total_tokens: 0.0` instead of raising `OverflowError`.
- Also fixed the identical latent bug in the TUI's parse-time guard (`tui/events.py`), since `board.py`'s guard is documented to mirror it — leaving the TUI unfixed would have reintroduced the exact TUI/board divergence this PR exists to close.
- Added regression coverage for the oversized-int case in `test_board_shared.py` (new `test_oversized_int_spend_is_omitted_not_crashed`) and `test_events.py` (added `10**400` to the existing `test_token_attribution_rejects_non_finite_or_negative_spend` parametrize — used `10**400` rather than `10**10000` to stay under pytest's `sys.set_int_max_str_digits()` limit for test-id string conversion, while still exceeding float's max range).

`uv run pytest tests/unit/dashboard tests/unit/dashboard_web tests/unit/tui -q` — 261 passed. `ruff check`/`ruff format --check`/`mypy` on changed files — clean.

Q00:
Addressed the sibling blocker in commit b5506064:

- `tui/events.py`'s `execution.frugality_proof.evaluated` handler converted `token_reduction_pct` with a bare `float(pct)`, so `token_reduction_pct=10**400` still raised `OverflowError`. Verified the reviewer's exact repro no longer crashes: `create_message_from_event(...)` now returns `FrugalityProofEvaluated(token_reduction_pct=None, ...)`.
- Extracted the `token_spend` guard from the previous commit into a shared `_coerce_finite_spend()` helper (mirrors `dashboard.board._coerce_spend`) and reused it for both call sites — `token_spend` with `reject_negative=True`, `token_reduction_pct` with `reject_negative=False` (a negative pct legitimately means spend increased, so it's kept).
- This also closes a latent gap the review didn't flag directly: NaN/±Inf `token_reduction_pct` previously survived unrejected in the TUI (unlike the board, which already rejected them via `_coerce_spend`) — now consistent on both sides.
- Added regression tests: negative pct is kept; NaN/±Inf/oversized-int pct are omitted without crashing.

`uv run pytest tests/unit/dashboard tests/unit/dashboard_web tests/unit/tui -q` — 266 passed. `ruff check`/`ruff format --check`/`mypy` on changed files — clean.

PR #1604, opened 2026-07-12
fix(mcp): correct install hints and align [mcp,claude] extras across CLI, skills, and docs

ASak1104:
## Summary

A bare `uv tool install ouroboros-ai` succeeds silently, but the MCP server then fails to boot when Claude Code starts — the tool env has no `mcp` package, and every recovery path around the failure pointed somewhere wrong. Four aligned changes:

1. **Wrong recovery hint** (`src/ouroboros/cli/commands/mcp.py`) — the `ImportError` handler in `mcp serve` suggested `uv add mcp`, which targets *project* dependencies, not the environment actually running the server. Since the CLI cannot know the install mode, the hint now covers both: the uv tool fix (`uv tool install 'ouroboros-ai[mcp,claude]'`, the canonical spec `_ensure_claude_mcp_entry()` already uses at `src/ouroboros/cli/commands/setup.py:109`) and a pip/pipx fix. Brackets are escaped so rich does not parse `[mcp,claude]` as a markup tag (render verified).
2. **Doctor remediations shared the same flaw** — `ouroboros mcp doctor` emitted the same `uv add` form in four remediations. All now use `uv tool install`, and the uv-tool forms always name the full canonical extras: `uv tool install` replaces the tool env with exactly the requested extras, so advising a subset (e.g. `[mcp]` alone) would strip the user's other extras and re-introduce this very boot failure. Doctor tests now assert the `[mcp,claude]` substring so the remediations cannot silently regress.
3. **`[claude]`-only installs relied on a transitive dep** — the setup/update skills (and their `.claude-plugin/` copies) installed `ouroboros-ai[claude]` and only worked because `claude-agent-sdk` happens to depend on `mcp`: an implementation detail, not a contract, and not part of the `[claude]` extra in `pyproject.toml`. All install paths now use `[mcp,claude]`, and the setup skill's `DEPS_MISSING` checks verify the `mcp` package too (the pipx check chains two `pip show` calls because `pip show a b` exits 0 when only one package is present), and the "uvx: command not found" troubleshooting fallback no longer recommends a bare `pip install ouroboros-ai` before wiring `python -m ouroboros mcp serve` — that path booted straight into the ImportError this PR prevents. This also removes a config churn: the skill previously wrote a `[claude]`-only uvx entry into `~/.claude/mcp.json` that `ouroboros setup` would immediately rewrite as stale args on its next run. Docs that feed an MCP server registration (README Kiro quick start, `docs/runtime-guides/kiro.md`, `docs/runtime-guides/claude-code.md`, `docs/runtime-capability-matrix.md`, `docs/getting-started.md`) now recommend `[mcp,claude]` as well.
4. **Stale `"timeout": 600`** — `src/ouroboros/cli/commands/setup.py:123-127` deletes this key as legacy when repairing `~/.claude/mcp.json`, but the static plugin `.mcp.json` and the manual config examples in `docs/cli-reference.md` still carried it. Per the Claude Code MCP docs, `timeout` is a per-server *tool-call* timeout in milliseconds and values below 1000 are ignored — so `600` has been dead config all along. Removing it is a behavioral no-op that aligns the static artifacts with what setup already writes.

Follow-up to #335, which moved `setup.py`/`install.sh` to `[mcp,claude]` but left the skills, doctor remediations, plugin `.mcp.json`, and parts of the docs behind.

**Scope notes**: per-extra menu listings (`README.md:162`, `docs/getting-started.md:113`) keep `[claude]` since they document what each extra individually provides, but now carry a "pair with `[mcp]` for the MCP server" note. Deliberately out of scope: `anthropic_adapter.py`'s `uv add anthropic` hint belongs to the LLM-provider path, not the MCP install story — happy to sweep it in a follow-up. Possible further follow-ups: a preflight extras check in `_ensure_claude_mcp_entry()` that detects/repairs a broken tool install.

## Test plan

- [x] `uv run pytest tests/ -n 4` — 12130 passed, 5 skipped; 3 failures are pre-existing and environment-caused (verified identical on unmodified `main` in a clean worktree: 2× codex runtime-profile tests + 1× evolution-adapter test pick up host-local ouroboros config)
- [x] `uv run pytest tests/unit/cli -k "mcp or setup"` — 329 passed; `test_setup.py` + `test_mcp_doctor.py` — 213 passed with the new `[mcp,claude]` remediation assertions
- [x] `uv run mypy src/ouroboros` clean (434 files)
- [x] `python3 scripts/check-auto-boundary.py` and `python3 scripts/check-max-turns-envelope.py` — 0 findings
- [x] `pre-commit run` on all changed files — passed
- [x] New hints render correctly through rich markup (verified by rendering): `uv tool install 'ouroboros-ai[mcp,claude]'`
- [x] `.claude-plugin/.mcp.json` validated with `python3 -m json.tool`
- [x] `mcp==1.28.1` pin satisfies `claude-agent-sdk`'s `mcp>=1.23.0,<2.0.0` requirement (no resolution conflict)

## Related issues

Refs #335


ASak1104:
Addressed the review in 24385122 (current HEAD):

| Finding | Response |
|---|---|
| **Blocker** — `skills/setup/SKILL.md:623` no-uvx fallback installs bare `ouroboros-ai` then wires `python -m ouroboros mcp serve` | Fixed in both copies (`skills/` and `.claude-plugin/skills/`): the fallback now installs `'ouroboros-ai[mcp,claude]'`. |
| **Follow-up (Medium)** — doctor tests only assert `remediation != ""` | Added `"[mcp,claude]" in result.remediation` assertions to both fail-path tests (`test_fails_when_not_importable`, `test_fails_when_not_importable_on_claude_backend`). `test_setup.py` + `test_mcp_doctor.py`: 213 passed. |
| **Non-blocking** — `README.md:162` / `docs/getting-started.md:113` generic extras listings | Both menu lines now note "pair with `[mcp]` for the MCP server". Kept as `[claude]` entries since they document what each extra individually provides. |

Also re-ran ruff / ruff format / mypy / pre-commit on the changed files — all clean.

Q00:
Addressed the remaining non-blocking suggestions and follow-up findings from both review rounds in commit 73370092:

**Follow-up (medium, round 1)** — `tests/unit/cli/test_mcp_doctor.py`: two of the four `[mcp,claude]` remediation paths (`check_claude_agent_sdk_import`'s non-claude-backend warn, `check_litellm_import`'s warn) only asserted `remediation != ""`. Strengthened both to assert the canonical extras substring, matching the two paths already covered.

**Follow-up (low) + non-blocking (round 2)** — `skills/update/SKILL.md` and `skills/setup/SKILL.md` (plus their `.claude-plugin/` packaged copies): quoted the unquoted `ouroboros-ai[mcp,claude]` package specs in copy-pasteable shell snippets (6 occurrences in update's uv/pipx/pip commands, 2 in setup's `DEPS_MISSING` remediation lines) — unquoted brackets trigger zsh's default "no matches found" glob error.

**Left untouched (round 1 non-blocking)**: the per-extra menu listings in `README.md`/`README.ko.md`/`README.zh-CN.md`/`docs/getting-started.md` already carry the "pair with `[mcp]`" note and are explicitly marked out-of-scope in the PR description (they document each extra individually).

`uv run pytest tests/unit/cli/test_mcp_doctor.py tests/unit/cli -k "mcp or setup"` — 51 + 329 passed. `ruff check`/`ruff format --check` on changed files clean.

Q00:
Approved — thanks for the thorough re-review across all three rounds! Merging now. 🙏

Issue #1400, opened 2026-06-11
fix(orchestrator): bound provider attempts and clean up runtime resources on every exit path


Q00:
> [!NOTE]
> Last reconciled: 2026-08-03, `main@cb658aa`.

## Current state

#1763 delivered the live-decomposition contract: `bounce_only` is the default, `TOO_BIG` requires durable execution evidence, Verified-MECE trust is trace-informed and same-run only, the decision is replayable, one repair is bounded, and the dead module path is guarded.

The only remaining close blocker is bounded-attempt provider cleanup in PR #1854. That PR is changes-requested because async-iterator and child-process cleanup is not yet guaranteed on every exit path.

## Close rule

After that exact blocker is fixed, the revised PR head is verified, and #1854 merges, close this issue as completed. No additional decomposition split is warranted.

---

## Context

Discussion #1385 established acceptance properties for the atomicity/decomposition mechanism, but the analyzed module is dead code (see #1399). The live mechanism — `orchestrator/parallel_executor.py` — has better bones (decomposition profiles with `axis: testable_unit` and explicit `min_unit`, `atomic_verifier_verdict`, recorded depth-cap warnings) but suffers the same disease at larger scale:

- 6,300+ line module where load-bearing judgments ride on text-surface heuristics — e.g. AC classification via `re.finditer(r"\b(?:and|then|while|plus)\b|[,;:]", ...)`, the same shape as the keyword fallback #1385 showed to be anti-correlated on adversarial input.
- Much of the bulk reads as accreted per-failure regex patches (`_runtime_message_*` evidence matchers) — symptom treatment without stated principles.

## Direction

1. **Redo the #1385-style failure analysis against `parallel_executor.py`** — the acceptance properties from that discussion are the yardstick (executor-relative judgment, graded confidence, real measured outputs, reasoned asymmetric error bias, verified decomposition, guaranteed termination with recorded compromises, safe LLM-less degradation, faithful auditable rationale). @deepakdgupta1 offered exactly this kind of analysis; this issue is open to claim.
2. **Add the missing discipline to the de facto attempt-then-bounce loop** (empirically confirmed: ACs execute at seed granularity, evaluation bounces failures):
   - **Bounded attempts** — turn/time caps so an under-decomposed unit's cost is capped
   - **Bounce-cause classification** — too-big vs. bad-spec vs. environment vs. model, so the remedy (decompose vs. escalate effort vs. retry) matches the cause
[...]

PR #1613, opened 2026-07-13
Make decomposition recovery evidence-driven and fail-closed

shaun0927:
## Summary

- add a versioned `DecompositionDecisionRecord` with explicit `ATOMIC`, `SPLIT`, `UNKNOWN`, and `ESCALATED` outcomes plus a separate `BounceCause`
- persist finalized decisions by stable execution node ID and restore them fail-closed from parallel-execution checkpoints
- add `execution.decomposition_mode = preflight | bounce_only | off` while preserving `preflight` as the compatibility default
- classify failed atomic attempts before retry-budget exhaustion; only evidence-backed `TOO_BIG` classifications may enter bounce decomposition
- require structured proposals, bounded structural validation, an independent semantic attestation, and at most one repair before a bounce split is trusted
- gate child model-tier lowering and shadow-replay trust on the finalized decomposition decision
- emit bounded decision/bounce events, record depth-cap compromises, and redact trace, classifier, persisted rationale, and retry-failure data
- keep alternate-harness redispatch behind the existing same-runtime retry-budget gate

## Compatibility and deferred work

This PR intentionally keeps `preflight` as the default. Legacy preflight arrays remain executable for compatibility, but are recorded as untrusted and do not receive the child model-tier discount.

The following remain separate evidence-gated follow-ups:

- default flip to `bounce_only`
- confidence-gated extra sampling
- global wall-clock or turn caps
- closed-loop calibration
- final issue closure

## Validation

- `2904 passed, 2 skipped, 2 deselected` for `tests/unit/orchestrator` plus execution config tests, with `test_runtime_evidence.py` ignored and two baseline profile tests deselected
- the excluded 8 failures were reproduced unchanged on clean `upstream/main`: 2 local Codex-profile configuration failures and 6 missing-`python` executable failures
- focused fail-closed/decomposition regression set passed after final hardening
- Ruff check and format check passed for all 19 changed files
- mypy passed for all 8 changed source files
- `py_compile` and `git diff --check` passed
- read-only final code review found no blocking findings

Not tested: live-LLM end-to-end execution.

Analysis and scope confirmation: https://github.com/Q00/ouroboros/issues/1400#issuecomment-4954156302

Refs #1400


Q00 (changes requested):
## Request changes — decomposition attestation is not actually independent

`ParallelExecutor._dispatch_decomposition_prompt()` always uses the same `self._adapter` and `self._inherited_runtime_handle`. Both proposal generation and `_attest_decomposition_proposal()` call this path, so the same model in the same session context can approve the decomposition proposal it generated.

A successful attestation promotes the split to trusted status and enables the child model-tier discount. This means a non-independent self-approval can grant permission to use a cheaper model.

Required changes:

1. Separate the trust boundaries of the proposer and attester. Use a separate adapter/session or a deterministic verifier that does not share conversational history.
2. Keep the split untrusted and prohibit the child-tier discount whenever independence has not been established.
3. Add a regression test proving that attestation performed with the same adapter and inherited runtime handle cannot grant trust.
4. Resolve the current `main` conflicts and rerun the full decomposition and routing suites.

Q00 (approved):
현재 HEAD를 최신 main 기준으로 재검증했습니다. 독립 세션 attestation, bounce recovery, checkpoint persistence, trust-gated model routing 회귀 테스트가 통과했고 차단 결함이 남지 않았습니다.

Issue #1466, opened 2026-06-15
[Task] Verified-MECE decomposition on live executor + delete dead modules (#1406)


Q00:
Parent: #1465 · RFC: [#1406 decomposition-reliability](https://github.com/Q00/ouroboros/pull/1406) · Discussion #1385

## Scope
Harden the **live** split path and delete the off-path dead modules — the precondition that makes lowering a child's effort honest.

- Verified-MECE split in `orchestrator/parallel_executor.py`: children **collectively exhaustive, mutually exclusive, each measurably simpler**; one repair retry then escalate; record **forced-atomic compromises** as events.
- Decompose **only on a too-big bounce**, never preemptively (no difficulty estimator).
- Delete off-path dead modules: `execution/atomicity.py`, `execution/decomposition.py`, and the `DoubleDiamond.run_cycle_with_decomposition` wiring (zero live callers — 115 `execution.ac.completed` events, 0 atomicity-check events).

## Acceptance Criteria
- [ ] Live split performs verified-MECE decomposition with one repair retry then escalate, recording forced-atomic compromises
- [ ] Wiring deleted first, then both dead modules removed
- [ ] Test asserts no non-test importer of the two modules remains after wiring deletion

## Test Checklist
- [ ] Overlapping / scope-dropping split is caught and repaired or escalated (never silently accepted)
- [ ] Forced-atomic compromise emits an event
- [ ] Import-graph test green (no `src/` regression)

## Pass Criteria
Verified-MECE discipline lives solely on the live path; dead modules gone with no regression.

---
*Part of [Epic] #1465 | Generated by [Ouroboros](https://github.com/Q00/ouroboros) via `ooo publish`*

PR #1595, opened 2026-07-08
feat(zcode): add zcode LLM adapter for interview/run backend

lee3Q:
## What

Adds `ZcodeCliLLMAdapter` (subclass of `CodexCliLLMAdapter`) so ouroboros can use the local zcode CLI as an LLM backend for interview question generation and evaluate — removing the implicit `claude_code` dependency on those paths.

**Depends on #1568** (zcode runtime backend). Base `main` for visibility; will rebase after #1568 merges.

## Why

`ouroboros_interview` currently routes question generation through the `claude_code` backend (Claude Agent SDK → Claude Code CLI). When that CLI isn't authenticated/configured, interview fails (401 / permission-deny-rule mismatch) — observed in a zcode↔ouroboros MCP session. zcode should be a first-class LLM backend, not just a runtime executor.

## Measured surface

`zcode --help` on 0.14.5 and 0.15.0 (same facts `ZcodeCLIRuntime` already relies on):
- `--prompt --json --mode --cwd` one-shot
- Emits ONE buffered JSON summary (`response` / `usage` / `sessionId`) — not Codex's NDJSON event stream
- No `--model` flag (hard `Unknown option` rejection); model selected via `~/.zcode/cli/config.json` (`model.main`)

## Changes

- `src/ouroboros/providers/zcode_cli_adapter.py` — `ZcodeCliLLMAdapter(CodexCliLLMAdapter)`. Overrides only the zcode-specific surface: `_build_command` (zcode flags, no Codex artifacts like `exec`/`--output-last-message`/`--model`), `_complete_once` (single-JSON parse via `communicate()`, real usage tokens), `_build_child_env` (strips `OUROBOROS_LLM_BACKEND` as a recursion guard), permission mapping (`acceptEdits`→`edit`, `bypassPermissions`→`yolo`).
- `src/ouroboros/providers/factory.py` — `get_zcode_cli_path` import (the prior `_create_zcode_adapter` stub referenced it undefined) + import sort.
- `src/ouroboros/backends/{capabilities,factory_registry}.py` — zcode `supports_llm=True` + `llm_adapter_factory="_create_zcode_adapter"` (factory wiring).
- `tests/unit/providers/test_zcode_cli_adapter.py` — 19 tests (command shape, permission, model suppression, complete parsing, factory routing).

## Activation

`OUROBOROS_LLM_BACKEND=zcode` (env) or `config.llm.backend: zcode`.

## Verification

[...]

PR #1615, opened 2026-07-13
fix(orchestrator): fail closed when investment authority is uncertain

### LINKED ISSUE #1398: RFC: effort-first investment decision — difficulty+stakes estimator, reasoning-effort actuator, remove unwired tier router
## Context

From discussion #1384: there is effectively no complexity-based investment decision in the live path today — `estimate_complexity` → `PALRouter` → `ModelRouter` have no non-test call sites, and where exercised the estimate mistakes verbosity for difficulty (weighted sum of token count / fabricated tool list / depth, uncalibrated constants).

Split into two RFCs per the #1384 reply, tracked together here until the first is drafted.

## RFC A — the estimator

- Two-axis minimum: **difficulty AND stakes** (reversibility, blast radius, sensitivity). A trivial-to-state but high-stakes unit (auth, schema migration, payment path) must be investable on stakes alone.
- Inputs are **measured signals only** — no counts re-encoded as placeholder lists.
- Difficulty decoupled from length: a short hard unit rates hard, a long mechanical one easy.
- v1 invariants (safety): monotonicity (more difficult/higher stakes never maps cheaper; no threshold cliffs), escalate-under-uncertainty, faithful auditable rationale (recorded numbers are the ones actually used).
- Calibration / cross-run learning is a **v2 maturity goal** (cold-start collision) — v1 records honestly so v2 can close the loop.

## RFC B — the actuator and wiring

- **Primary lever = reasoning effort, not model switching**: route on the backend's thinking/effort dial (Anthropic thinking budget, OpenAI-style reasoning effort, Codex `model_reasoning_effort`). Monotonic and cliff-free by construction; escalation is cheap so fail-safe-on-uncertainty is practical; mis-estimates don't change the executor's capability class (shrinks cold-start risk).
- Escalation ladder: effort low → effort high → bigger model (model switching survives only as the coarse outer rung, and as fallback where no effort dial exists — e.g. GLM's near-binary thinking toggle).
- **Runtime-scoped applicability**: where Ouroboros calls the LLM directly it routes; where it delegates to a CLI runtime with no per-call control it can only advise. Needs a capability matrix + explicit advise-fallback (composes with the #1374 `ParamSupport` work).
- **Remove, don't wire**: the existing `PALRouter`/tier machinery encodes the wrong actuator and gets deleted.

Refs: discussion #1384 (acceptance properties), #1377 (frugality invariants), #1374 (capability declarations)

shaun0927:
## Summary

- add optional AC-level `InvestmentSpec` metadata with bounded difficulty, stakes, provenance, and confidence values while preserving legacy string serialization
- normalize that metadata into an auditable `InvestmentAssessment`; only complete low/low, declared-or-measured, high-confidence authority may lower reasoning effort by one notch, while any high axis imposes a high floor
- emit `execution.ac.investment_assessed`, embed the exact assessment in `execution.ac.effort_routed`, route every investment-bearing fresh run through the per-AC executor, and block whole-session resume until it can restore per-AC authority
- require `decomposition_trustworthy is True` before a decomposed child may use a lower model tier; current live decomposition has no trusted issuer, so live children remain at the base tier
- preserve investment authority across recursive children, retries, and alternate-harness reruns without copying a parent success contract into child prompts
- reconcile the estimator/actuator RFCs and related runtime/config documentation with current `main`

## Policy boundaries

- no natural-language difficulty scoring
- no token, tool, artifact, command, file, or keyword-count lowering proxies
- no live verified-MECE trust issuer in this PR; that remains follow-up work under #1466
- no shadow-replay enablement, frugality-proof redesign, or new provider capability work
- explicit model pins, retry escalation, backend capability enforcement, and the model-routing kill switch retain precedence

## Test plan

- [x] Hermetic full suite: `12164 passed, 5 skipped`
- [x] `ruff check .`
- [x] `ruff format --check .`
- [x] `mypy src/` (`434` source files)
- [x] Targeted direct, sequential, resume, recursion, alternate-harness, investment-event, effort-routing, model-routing, retry, and runner regressions
- [x] Independent code review and acceptance verification; the direct-runner authority gap found during review was fixed before re-review approval

## Related issues

Closes #1398
Refs #1466

Issue analysis and confirmed scope: https://github.com/Q00/ouroboros/issues/1398#issuecomment-4954157717


shaun0927:
Addressed the blocking direct-runner authority gap in `9b2bf4a9a`. Any Seed with AC investment metadata now routes through the per-AC executor; explicit sequential runs preserve `force_sequential_levels`, and whole-session resume fails closed until it can restore per-AC authority. Added single-AC, incomplete-metadata, sequential, and resume regressions. Local full suite: `12164 passed, 5 skipped`; all GitHub checks are green. Please re-review the latest HEAD.

shaun0927:
Addressed the review in `9b2bf4a9` (current HEAD):

- investment-bearing single-AC and explicit sequential runs now use the per-AC executor instead of the whole-seed direct path
- sequential callers retain one-AC-per-stage ordering
- whole-session resume now fails closed before tool setup or `execute_task` when per-AC investment authority cannot be restored
- complete, incomplete/low-confidence, sequential, and resume regressions were added

All required checks are green; local hermetic suite is `12164 passed, 5 skipped`.

Q00 (dismissed):
## Request changes — effort-lowering authority is not fail-closed

The current HEAD conflicts with `main`, so the integrated result cannot be validated. Independently of those conflicts, the `can_cheapen` condition in `assess_investment()` treats `provenance="declared"` as equivalent to `"measured"` lowering authority.

As a result, an AC author can reduce the baseline effort by one level merely by declaring `difficulty=low`, `stakes=low`, and `confidence=high`. A self-declaration therefore gains permission to reduce execution investment without independently observed or verified evidence, which conflicts with #1398's measured-only and uncertainty-never-cheapens principles.

Required changes:

1. Rebase onto the latest `main` and resolve the conflicts.
2. Do not allow a plain `declared` value to reduce effort. Lowering should require measured evidence or a separately established trusted authority.
3. Add a regression test proving that `declared + low/low + high confidence` results in `can_cheapen=False` and preserves the baseline effort.
4. Rerun the full suite and the effort-routing integration tests after resolving the conflicts.

Q00 (approved):
Approved at d5df84efc. The measured-only cheapening rule and declared-metadata regression remain intact after integrating current main; investment authority and decomposition trust stay separate across all execution paths. Targeted: 754 passed, 1 skipped. Full suite: 12,613 passed, 10 skipped. Ruff and MyPy pass.

Q00 (approved):
Approved at aa9c407ff. The final branch integrates #1595 and current main; legacy string acceptance criteria are safe under investment detection, and the prior Python 3.14 checkpoint timing failure is covered by the event-based retry test now on the integrated branch. CI-equivalent Python 3.14 parallel suite: 12,808 passed, 12 skipped; Ruff and MyPy pass.

PR #1617, opened 2026-07-14
fix(orchestrator): stabilize Codex runtime identity and GPT-5.6 routing

gronxb:
## Summary

Stabilize Codex-backed Ouroboros runs without weakening their execution identity:

- tolerate only Codex Desktop-managed project and hook state updates during an active runtime;
- continue failing closed when executable, security, model, provider, or profile settings change;
- route Codex worker tiers to the GPT-5.6 model family supported by ChatGPT-account Codex.

### Runtime identity

The Codex CLI runtime originally hashed `~/.codex/config.toml` byte-for-byte. Codex Desktop legitimately mutates `projects.*` trust state and `hooks.state.*` approval state, so those writes aborted active runs with `Codex configuration changed after runtime initialization`.

The global config also contains runtime-critical settings such as `[mcp_servers]`, `sandbox_mode`, and `approval_policy`. A safe canonicalization therefore cannot use a narrow routing-key allowlist: it must retain every setting except the two known Codex-managed state subtrees.

### Model routing

The shipped OpenAI tier ladder still selected `gpt-5.1-codex-mini`, `gpt-5-codex`, and `gpt-5.2`. ChatGPT-account Codex rejects those IDs with HTTP 400 before a worker starts, causing every retry to fail at model invocation.

## Changes

- Canonicalize the global Codex config by excluding only `projects.*` and `hooks.state.*` before hashing.
- Keep MCP server configuration, sandbox/approval policy, routing settings, all profile-v2 `*.config.toml` files, and the resolved `CODEX_HOME` in the runtime fingerprint.
- Add regression coverage proving MCP command, sandbox mode, and approval policy changes invalidate the fingerprint while Codex-managed state changes do not.
- Route Codex tiers as:
  - frugal → `gpt-5.6-luna`
  - standard → `gpt-5.6-terra`
  - frontier → `gpt-5.6-sol`
- Normalize previously shipped Codex tier defaults to the GPT-5.6 family while preserving explicit custom model IDs.
- Keep the backend model catalog and config TUI presets aligned with the runtime defaults.

## Verification

- [x] Full unit suite: `11812 passed, 2 skipped`
- [x] Review-focused suite: `287 passed`
- [x] `uv run ruff check src/ tests/`
- [x] `uv run ruff format --check src/ tests/`
- [x] `uv run mypy src/ouroboros` — 435 source files
- [x] `python3 scripts/check-auto-boundary.py`
- [x] No-excuse rules check for the changed files
- [x] Manual runtime QA: managed `projects`/`hooks.state` changes remain accepted; an MCP command change raises the restart-required runtime error
- [x] Routing red/green coverage: prior shipped defaults normalize to Luna/Terra/Sol; custom IDs remain unchanged
- [x] Real router driver resolves `codex_cli` to Luna/Terra/Sol
- [x] Authenticated Codex CLI calls to all three GPT-5.6 models return a non-empty response with exit code 0
- [x] GitHub Actions on the latest head: 10 checks passed

## Related issues

Refs #1600

### DIFF
--- a/src/ouroboros/backends/model_catalog.py
+++ b/src/ouroboros/backends/model_catalog.py
@@ -31,7 +31,7 @@ from ouroboros.backends.capabilities import (
     get_backend_capability,
     runtime_backend_choices,
 )
-from ouroboros.config._model_defaults import DEFAULT_OPUS_MODEL, DEFAULT_SONNET_MODEL
+import ouroboros.config._model_defaults as model_defaults
 
 # Backends whose runnable model is the CLI's own configured default rather
 # than a Claude model id. Mirrors the loader's sentinel frozensets
@@ -74,11 +74,11 @@ class BackendModelCatalog:
 
 
 # Hand-curated additions per backend, appended after the loader-mirroring
-# default entry. Keep entries verifiable: the codex ids below were confirmed
-# against a live `opencode models` listing of the OpenAI catalog.
+# default entry. Keep entries verifiable: the Codex ids below were confirmed
+# against live ChatGPT-account Codex CLI calls.
 _EXTRA_KNOWN_MODELS: dict[str, tuple[str, ...]] = {
     "claude": ("claude-haiku-4-5-20251001",),
-    "codex": ("gpt-5-codex", "gpt-5", "gpt-5-mini"),
+    "codex": model_defaults.DEFAULT_CODEX_TIER_MODELS,
     # Grok Build model slugs, after the CLI-owned "default" sentinel. Verified
     # against a live `grok models` listing (grok-build, grok-composer-2.5-fast).
     "grok": ("grok-build", "grok-composer-2.5-fast"),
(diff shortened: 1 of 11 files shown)


Q00:
Hi @gronxb  I only just went back through this one properly, and I'm sorry no human showed up while it was open.

The bug you were fixing is real. Codex Desktop writing `projects` / `hooks.state` mid-run shouldn't kill an active runtime, and today it does. I want that tracked regardless of what happens to this PR.

But I'd rather understand your call than second-guess it: what made you close it? If you hit something after marking it ready, I'd genuinely like to know what — that's more useful to me than the diff is.

No pressure to re-open anything.

gronxb:
I just wanted to run ooo run, but I noticed on the Ouroboros dashboard that all of the tasks were failing.

While debugging the issue, I realized that the run command doesn’t properly support the 5.6 model. I asked Codex to try fixing it and open a PR.

However, I didn’t review the diff carefully, so I closed the PR for now.

PR #1626, opened 2026-07-15
feat(pm): explore brownfield repos via persistent snapshot worktrees pinned to remote main

cohemm:
## Summary

PM brownfield exploration previously read the developer's live checkout, so a stale or dirty local main leaked outdated context into PRD generation. Each registered brownfield repo now gets one long-lived detached worktree under `~/.ouroboros/pm-snapshots/<name>-<path-hash>`, created once on first use and refreshed deterministically on every PM interview start (`git fetch origin` + hard reset to `origin/HEAD`, falling back to `origin/main` → `origin/master` → local `HEAD` when offline or remote-less). The refresh is wired into the MCP handler (`_handle_start`, which the `select_repos` 2-step flow also funnels into), not skill prose, so it runs unconditionally.

Key properties:
- **Reuse, not re-create**: the worktree persists across sessions; each PM start only fetches and hard-resets it, so there is no per-session checkout cost and no cleanup lifecycle to get wrong.
- **Durable artifacts stay durable**: the PM seed, `pm.md`, `InterviewState.codebase_paths`, and explore context headers all record the source checkout path (`source_path`), never the snapshot location. Only exploration and in-interview code reads target the snapshot.
- **Never breaks the interview**: any git failure (non-git path, fetch timeout, worktree error) logs a warning and falls back to the original repo path.
- Configurable via `orchestrator.pm_snapshot_worktrees` (default on) and `orchestrator.pm_snapshot_root`.

Scope note: the OpenCode bridge (plugin dispatch) path is unchanged — there the child session owns seed generation, so the source-path mapping doesn't apply; the in-process engine path used by Claude Code runtimes is fully covered.

## Test plan

- [x] `uv run pytest tests/unit/ -q` — 24 pre-existing failures identical on clean `main` (Python 3.14 rc env issue); this branch adds 0 new failures and +9 new passing tests
- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check` clean on touched files
- [x] `uv run mypy src/ouroboros/core/pm_snapshot.py src/ouroboros/mcp/tools/pm_handler.py src/ouroboros/bigbang/pm_interview.py` clean
- [x] New unit tests build real git remotes with stale clones: snapshot pins to `origin/main` while local stays behind, refresh reuses the same worktree and cleans dirty/untracked files, stale non-worktree dirs are re-created, non-git/remote-less repos fall back gracefully
- [x] Verified locally against real repos (`podo-backend` 38 commits behind, `podo-app` 17 behind): first run created detached worktrees at the exact `origin/main` SHA (~7s for both), second run reused and refreshed them in place, source checkouts untouched

## Related issues

N/A

### DIFF
--- a/src/ouroboros/bigbang/pm_interview.py
+++ b/src/ouroboros/bigbang/pm_interview.py
@@ -16,7 +16,7 @@ questions for classification and collects PM-specific metadata.
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 import json
 from pathlib import Path
 from typing import Any
@@ -286,6 +286,18 @@ class PMInterviewEngine:
                 model=self.model,
             )
             results = await explorer.explore(paths)
+
+            # Snapshot worktrees are scan locations only — present the
+            # durable source checkout in the injected context.
+            source_by_scan_path = {
+                r["path"]: r["source_path"] for r in repos if r.get("source_path")
+            }
+            if source_by_scan_path:
+                results = [
+                    replace(res, path=source_by_scan_path.get(res.path, res.path))
+                    for res in results
+                ]
+
             self.codebase_context = format_explore_results(results)
 
             # Share context with classifier
@@ -424,8 +436,12 @@ class PMInterviewEngine:
             if brownfield_repos and self.codebase_context:
                 state.is_brownfield = True
                 state.codebase_context = self.codebase_context
+                # Persist the durable source checkout, not an ephemeral
+                # snapshot worktree path, into interview state.
                 state.codebase_paths = [
-                    {"path": r["path"], "role": "primary"} for r in brownfield_repos if "path" in r
+                    {"path": r.get("source_path") or r["path"], "role": "primary"}
+                    for r in brownfield_repos
+                    if "path" in r
                 ]
                 state.explore_completed = True
             log.info(
@@ -1258,8 +1274,17 @@ Include them as original question text in "decide_later_items":
             if item not in all_decide_later:
                 all_decide_later.append(item)
 
-        # Include brownfield repos — use session-stored repos, not DB
-        brownfield_repos = tuple(dict(r) for r in self._selected_brownfield_repos)
+        # Include brownfield repos — use session-stored repos, not DB.
+        # Snapshot worktree paths are working locations only; the seed must
+        # record the durable source checkout (``source_path``) instead.
+        def _durable_repo(repo: dict[str, str]) -> dict[str, str]:
+            entry = dict(repo)
+            source = entry.pop("source_path", None)
+            if source:
+                entry["path"] = source
+            return entry
+
+        brownfield_repos = tuple(_durable_repo(r) for r in self._selected_brownfield_repos)
 
         return PMSeed(
             pm_id=f"pm_seed_{interview_id}",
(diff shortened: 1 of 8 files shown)


cohemm:
Addressed the review blocker in b592f49a:

**Blocker 1 — plugin-dispatch path bypassed snapshot refresh**: Fixed, plus the CLI gap flagged in the follow-up finding.
- The refresh hook moved from the MCP handler into `PMInterviewEngine.start_interview`, so every engine-driven entry point (MCP in-process start/select_repos, CLI `ooo pm`, direct `start_interview`) shares one hook instead of per-caller wiring.
- Plugin dispatch now redirects `selected_repos` to refreshed snapshot worktrees before they are persisted in pm_meta or embedded in the subagent payload — both the 1-step start and the select_repos step. The child session therefore reads remote-main state.
- The cwd-derived fallback (plugin start without `selected_repos`) is deliberately NOT redirected: cwd is the user's live working repo and may hold intentional WIP that the interview is about. This is documented in code and pinned by a test.

**Follow-up — test coverage for plugin/CLI paths**: Added.
- `tests/unit/mcp/tools/test_pm_handler_snapshot.py` pins the plugin-dispatch contract (persisted pm_meta paths, subagent payload paths, cwd-fallback exclusion).
- `tests/unit/bigbang/test_pm_interview_brownfield_db.py::test_start_redirects_repos_to_snapshot_worktrees` pins the engine hook, which covers the CLI path since `ooo pm` funnels through `ask_opening_and_start` → `start_interview`.

Verification: `pytest tests/unit/mcp tests/unit/bigbang tests/unit/core tests/unit/cli/test_pm*.py` — 2486 passed; ruff + mypy clean. The Python 3.12 CI failure on the previous run was two unrelated flaky tests (`test_job_manager` terminalization timing, `test_synapse_after_turn_integration` signal-expiry boundary); re-run requested.

cohemm:
Addressed the latest review (req_1784272436_224) and fixed the CI failure.

**Blocker — plugin generate loses durable source path (`pm_handler.py:552`):** resolved in a33884f1. Plugin start/select now persist repo *records* (`{path: <snapshot scan path>, source_path: <registered checkout>}`) in `pm_meta["brownfield_repos"]` instead of bare snapshot strings. Resume turns keep projecting the scan path, while `generate` projects `source_path` via `_plugin_repo_paths(..., durable=True)`, so PM seeds/handoffs never record `~/.ouroboros/pm-snapshots/...` locations. Covered by `test_generate_restores_durable_source_paths` (start → select_repos → generate, asserting the subagent prompt gets the durable source path back), plus record-shape and fallback tests — closing the coverage gap called out in the review.

**CI failure:** `test_round5_fixes.py::test_1step_start_persists_selected_repos` still asserted the pre-snapshot plain-string persistence format and broke on all Python versions after a33884f1. Fixed in 3b49f88e by asserting both scan and durable projections resolve back to the caller's paths.

Note: `test_job_manager.py::test_crash_inside_terminalization_still_persists_terminal_state` (Python 3.14 only) also fails on current `main` CI (run 29585704950) — pre-existing and unrelated to this PR.

Q00 (approved):
최신 HEAD 3b49f88e4를 검증했습니다. snapshot scan path와 durable source_path가 plugin start/select/generate 및 engine/CLI 경로에서 분리 보존되며 74개 관련 테스트와 전체 CI가 통과했습니다.

PR #1627, opened 2026-07-15
fix(brownfield): guarantee repo list visibility by ending turn before default selection

cohemm:
## Purpose

The default-repo selection flow in `ooo brownfield` (and setup Step 5) was unusable: the "Default Repos" dialog appeared without the numbered repo list the user needs to answer it. The grid was emitted as assistant text *between* two tool calls (scan → grid → `AskUserQuestion`), and the host UI does not guarantee rendering of inter-tool-call text. This PR restructures the flow so list visibility is guaranteed before any selection is requested.

## Changes (with intent)

- `skills/brownfield/SKILL.md` Step 2: replaced the `AskUserQuestion` dialog with a **turn-boundary pattern** — the repo grid plus a plain-text selection prompt ("numbers / keep / none") is now the final message of the turn. Intent: turn-final text is the only rendering-guaranteed surface in the host UI.
- Added explicit next-turn reply parsing rules (numbers → indices, "keep" → no-op, "none" → clear, anything else → re-ask, never guess) followed by the single `set_defaults` MCP call. Intent: keep exactly one MCP write and remove the dead-end "Select repos" option that collected no numbers and forced an extra round-trip.
- `skills/setup/SKILL.md` Step 5: applied the same pattern ("keep" proceeds to Step 6). Intent: the flaw was a copy of the same block; fixing only one file would leave the two flows divergent.
- Documented both rendering constraints inside the skill bodies (inter-tool-call text may not render; option `preview` boxes truncate). Intent: prevent the dialog pattern from being reintroduced by a future edit that only reads the skill file.
- Added a "keep" confirmation message case in Step 3. Intent: every reply branch now has a defined user-visible outcome.
- **Non-goal: embedding the repo list in `AskUserQuestion` option `preview` fields.** Verified live with a 68-repo scan — the preview box is fixed-height, non-scrollable, and silently truncates ("9 lines hidden"). The absence of any dialog here is a decision, not an omission.
- **Non-goal: changing `ouroboros_brownfield` server behavior.** The scan/set_defaults contracts and response format are untouched; the defect was purely in skill presentation sequencing.
- **Trade-off:** the structured option dialog (arrow-key selection) is given up in exchange for guaranteed list visibility. Free-text replies need parsing, which is why the parse rules are specified in the skill rather than left to the model.

## Success criteria

Running `ooo brownfield` shows the full numbered repo list on screen before any selection is requested, and one chat reply completes the default update.

- [ ] After scan, the repo grid + selection prompt is the final message of its turn, with no tool call after it
- [ ] Replying with numbers (e.g. "2, 3, 7") produces exactly one `set_defaults` call and a confirmation listing the chosen repos
- [ ] "none" clears all defaults; "keep" (when defaults exist) makes no MCP call and confirms defaults unchanged
- [ ] No `AskUserQuestion` dialog appears anywhere in the brownfield/setup default-repo selection
- [ ] `pytest tests/unit/orchestrator/test_skill_tool_mapping.py tests/unit/test_codex_artifacts.py` passes (47 tests)

### DIFF
--- a/.claude-plugin/skills/brownfield/SKILL.md
+++ b/.claude-plugin/skills/brownfield/SKILL.md
@@ -33,8 +33,7 @@ Show scanning indicator:
   Scanning for Existing Projects...
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-Looking for git repositories and worktrees under the scan root directory.
-Linked worktrees reported by discovered normal repo roots may also be registered, even outside the scan root directory.
+Looking for git repositories and worktrees up to two directories below the scan root.
 Local repos and repos with any remote name are eligible.
 This may take a moment...
 ```
@@ -55,7 +54,7 @@ stop with the MCP-not-available message instead of retrying the failing call.
    Tool: ouroboros_brownfield
    Arguments: { "action": "scan" }
    ```
-   This walks `scan_root` for valid seed repos/worktrees and registers them in DB. For each discovered normal repo root with a `.git` directory, Git-reported linked worktrees are also considered, even when they live outside `scan_root`. A linked worktree found under `scan_root` with a `.git` file is registered itself, but it is not used to register its main worktree or sibling worktrees outside `scan_root`. Existing defaults are preserved.
+   This walks `scan_root` (up to two directory levels deep) for valid seed repos/worktrees and registers them in DB. Each repo or worktree found directly by the walk is registered self-only — Git worktree families are not expanded, so worktrees outside the depth-bounded walk (e.g. under `.ouroboros/worktrees`) are not pulled in. Existing defaults are preserved.
 
 The scan response `text` already contains a pre-formatted numbered list with `[default]` markers. **Do NOT make any additional MCP calls to list or query repos.**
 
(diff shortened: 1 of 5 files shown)


cohemm:
Addressed the review feedback.

**Changes:**
- `.claude-plugin/skills/brownfield/SKILL.md`: replaced the stale outside-root worktree-expansion scan contract with the depth-bounded, self-only registration contract (synced with root `skills/brownfield/SKILL.md`)
- `.claude-plugin/skills/setup/SKILL.md`: same scan-contract sync for the Step 5.5 brownfield scan section
- `tests/unit/skills/test_skill_artifacts.py`: added `test_brownfield_scan_contract_matches_runtime_in_every_skill_bundle`, a drift guard that fails if stale worktree-expansion phrasing reappears (or the self-only/two-level contract disappears) in either `skills/` or `.claude-plugin/skills/`

Verified: `tests/unit/skills/test_skill_artifacts.py` (10 passed), `tests/unit/bigbang/test_brownfield.py` + `tests/unit/mcp/tools/test_brownfield_handler.py` (135 passed).

Q00 (approved):
최신 HEAD 083dd693b를 검증했습니다. 두 턴 저장소 선택 UX와 root/Claude-plugin scan 계약 동기화, drift guard를 확인했고 145개 관련 테스트와 전체 CI가 통과했습니다.

PR #1630, opened 2026-07-16
fix(orchestrator): recover artifact contracts and qa env

lee3Q:
## Summary

This PR fixes three upstream reliability gaps found while diagnosing repeated `ooo run` / `ooo qa` failures:

- add artifact-family task strategy/profile support for `artifact`, `document`, `documentation`, and `presentation`
- make `ooo qa` bootstrap the same cached shell environment as MCP/run paths before invoking QA
- let the seed-level verify gate recover runtime false-negatives when an AC's explicit `expected_artifacts` / `verify_command` contract passes

The local installed-tool mismatch that was also found during diagnosis is intentionally not part of this PR. This change only contains bug classes that are reproducible from source contracts and tests.

## Details

Artifact/document/presentation Seeds were previously easy to route through code-shaped execution evidence, which can reject valid file-artifact work or require irrelevant code-test evidence. This PR adds `ArtifactStrategy` and built-in artifact-family profiles so those task types use artifact-shaped tools, prompts, and verifier expectations.

`ooo qa` now calls the same shell-env bootstrap used by MCP/run entrypoints. Without this, QA could fail or hang in environments where provider keys are available through Ouroboros shell-env cache rather than the process environment.

For ACs with a declared success contract, the orchestrator now treats `verify_command` or `expected_artifacts` as the durable success oracle. If a runtime reports failure after producing the requested artifact, the verify gate can emit `execution.verify.recovered` and mark the AC succeeded. If the runtime reports success but the declared verify command fails, the existing failure flip remains intact.

## Validation

- `ruff check` on touched source/tests: clean
- `ruff format --check` on touched source/tests: clean
- focused pytest suite: `85 passed`
- QA shell-env focused pytest: `2 passed`
- `git diff --check`: clean

Live smoke in the clean PR worktree was attempted but blocked by local memory pressure before artifact creation. Separate installed-tool smoke after the same patch path completed successfully with `Success: 1/1` and exact output `OZO_RUN_SMOKE_OK\n`; `ooo qa` returned a real verdict (`REVISE 0.76`) instead of failing from missing provider environment.


Q00 (approved):
Approved at 9e60300fa. The artifact/QA contract is fail-closed at the typed-evidence boundary, cached verify outcomes avoid duplicate command execution, and the verify-only code regression is covered. Local targeted suite: 734 passed, 1 skipped; Ruff/MyPy and all current CI checks pass.

Issue #1636, opened 2026-07-17
benchmark(interview): capture redacted cold/warm OAuth latency evidence


sergiobuilds:
## Current state

PRs #1640 and #1645 completed the instrumentation and documentation:

- successful response events separate `total`, `ambiguity_scoring`, `question_generation`, and `advisory_build`;
- start and later turns report the correct phase set;
- completion, failure, collaborator exceptions, and parent handoff close active spans;
- the exporter and redaction boundary are documented and tested.

## Remaining task

Capture the live OAuth benchmark evidence required by the original acceptance criteria. No additional runtime implementation is currently justified.

## Acceptance criteria

- [ ] Follow `docs/guides/interview-latency-benchmark.md` on a recorded current-`main` SHA.
- [ ] Capture at least 20 cold and 20 warm samples in the defined environment.
- [ ] Attach a redacted report containing `n`, p50, and p95 for the total and phase timings.
- [ ] Link the report to #1635 and record provider/runtime/Python versions without secrets or raw prompt content.
- [ ] If the live-trace requirement is waived, record explicit maintainer approval and the reason instead.

Close after either the evidence is attached or the live-trace criterion is explicitly waived.

Last reconciled: 2026-08-03, `main@cb658aa`.

sergiobuilds:
## Maintainer-environment evidence captured before instrumentation

A redacted existing log window for `interview_20260717_152758` shows:

- `15:34:19.884552Z` — `mcp.tool.interview.response_recorded`
- `15:34:49.311212Z` — `mcp.tool.interview.question_asked`
- observed server-side interval: **29.427 s**

Within that interval, `codex_cli_runtime.initialized` appears at:

- `15:34:22.717283Z`
- `15:34:25.604754Z`
- `15:34:36.502755Z`
- `15:34:43.533001Z`
- `15:34:47.513188Z`

Another runtime initialization appears at `15:35:07.470948Z`, after the server logged `question_asked`.

Interpretation boundary: this proves a substantial engine-side interval and repeated runtime initializations in one maintainer environment. The old log does not label each initialization as scoring, question generation, or advisory, so it does **not** prove exact attribution or that every user sees the same two-minute wall clock. Draft PR #1640 adds the missing phase labels for subsequent reproduction.

sergiobuilds:
[...]

Issue #1635, opened 2026-07-17
RFC: make the first 10 minutes of interviews fast, truthful, and non-blocking


sergiobuilds:
> [!NOTE]
> Last reconciled: 2026-08-03, `main@cb658aa`. The former 11-PR proposal is superseded by the live work breakdown below.

## Current work breakdown

- [ ] #1636 - capture redacted cold/warm phase-latency evidence
- [ ] #1638 - establish the structured, novice-friendly one-decision presentation contract
- [ ] #1639 - make advisory delivery non-blocking across host runtimes
- [ ] #1637 - make fresh Codex configuration truthful and isolate nested LLM child execution from Ouroboros MCP
- [ ] #1650 - reduce user turns and effort through the dependency-ordered interview UX program

Closed or unmerged draft PRs are implementation references only.

## Product evidence budgets

- First question: warm p50 <= 3 seconds, p95 <= 8 seconds in the defined benchmark environment.
- Follow-up: p50 <= 4 seconds, p95 <= 10 seconds in the defined benchmark environment.

These are product evidence gates, not wall-clock CI assertions. OAuth live benchmarks are recorded separately; deterministic fake-provider/fake-CLI contracts belong in CI.

---

## Maintainer report

On Codex, an interactive `ooo interview` can take roughly two minutes per turn in a maintainer environment, while the generated questions are difficult enough that even a maintainer has to stop and interpret them. The latency was noticeable before question advisory subagents were added, so removing subagents alone would not address the underlying problem.

We do not yet have evidence that every user sees the same wall-clock time. However, several costs are structural on the current Codex path and therefore can affect more than one installation. This issue should establish a reproducible baseline before claiming a universal regression.

## Current-main evidence

As of `7068835c`:

1. **The persisted default can be misleading on Codex.** `ClarificationConfig.default_model` still defaults to `DEFAULT_OPUS_MODEL`, currently `claude-opus-4-8` (`src/ouroboros/config/models.py`, `src/ouroboros/config/_model_defaults.py`). Runtime loading can normalize shipped Claude defaults for Codex, but fresh/setup-facing config can still present an Opus name even when Codex is the actual backend.
2. **Question advisory fan-out is effectively mandatory when metadata is present.** `authoring_handlers.py` attaches `meta.question_advisory_subagents`; `skills/interview/SKILL.md` says hosts MUST fan out every payload. The intended UX benefit is valuable, but the question should not wait for that work.
[...]

PR #1640, opened 2026-07-17
feat(interview): emit per-phase response timings

sergiobuilds:
## Summary

Adds privacy-safe monotonic phase timings to successful question-bearing interview responses so #1635 can distinguish server-side work from host advisory time.

Part of #1636.

## What is measured

`interview.response.emitted.data.timings_ms` now reports:

- `total`
- `ambiguity_scoring`
- `question_generation`
- `advisory_build`

Start turns explicitly report no ambiguity-scoring phase. Resume-pending turns report no question-generation phase. Later answer turns keep scoring and question generation separate.

`advisory_build` covers server-side advisory preparation, including payload construction, dispatch metadata, and fan-out registration when configured. It does not measure host-side advisory execution or wait time.

## Privacy boundary

The new fields contain durations only. They do not add prompt text, answer text, model output, local paths, or environment values.

## Verification

- `PYTHONPATH=src python -m pytest tests/unit/mcp/tools/test_interview_response_diagnostics.py tests/unit/mcp/tools/test_interview_lateral_review_advisory.py tests/unit/mcp/tools/test_interview_length_guard_envelope.py -q`
  - 29 passed
- All interview-tool unit tests: 92 passed in current-head review
- Ruff check and format check passed
- MyPy passed
- `git diff --check` passed
- GitHub checks passed on Python 3.12, 3.13, and 3.14 plus coverage, Bridge TypeScript, and policy gates

## Known follow-up

This first slice measures successful question-bearing responses. #1636 remains open for:

- completion, failure, and parent-handoff timing spans;
- stable event-schema and null-semantics documentation;
- the opt-in OAuth live benchmark and redacted trace procedure.

## Local push-hook note

The implementation commit was pushed with `--no-verify` because the local global document hook rejects pre-existing upstream Markdown files without that workstation's private frontmatter convention. Repository-scoped tests, lint, type checking, diff checks, GitHub CI, and current-head review all passed.

### DIFF
--- a/src/ouroboros/events/interview.py
+++ b/src/ouroboros/events/interview.py
@@ -104,6 +104,10 @@ def interview_response_emitted(
     transcript_chars: int,
     ambiguity_prefix_present: bool,
     is_length_guard: bool,
+    total_duration_ms: float | None = None,
+    ambiguity_scoring_duration_ms: float | None = None,
+    question_generation_duration_ms: float | None = None,
+    advisory_build_duration_ms: float | None = None,
 ) -> BaseEvent:
     """Diagnostic event recording the shape of an MCP question-bearing response.
 
@@ -124,6 +128,12 @@ def interview_response_emitted(
             "transcript_chars": transcript_chars,
             "ambiguity_prefix_present": ambiguity_prefix_present,
             "is_length_guard": is_length_guard,
+            "timings_ms": {
+                "total": total_duration_ms,
+                "ambiguity_scoring": ambiguity_scoring_duration_ms,
+                "question_generation": question_generation_duration_ms,
+                "advisory_build": advisory_build_duration_ms,
+            },
         },
     )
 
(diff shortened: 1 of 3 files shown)


shaun0927:
Pushed tests-only follow-up `a025a6343` to make the timing diagnostics evidence mutation-resistant:

- phase-aware deterministic clock assertions now catch swapped `ambiguity_scoring`, `question_generation`, and `advisory_build` labels plus total undercounting;
- resume-without-pending now verifies question-generation timing;
- length-guard coverage now pins `advisory_build: null` when that phase does not execute.

Verification on current HEAD: targeted local suites **29 passed**, Ruff/format/diff checks passed, and all required GitHub checks are green across Python 3.12/3.13/3.14, MyPy, Ruff, Bridge TypeScript, and policy gates.

@ouroboros-agent please re-review current HEAD `a025a6343b3e0e22acf2cbb6af6ccf74dc7f2d99`.

PR #1648, opened 2026-07-18
feat(orchestrator): gate-anchored decomposition trust + lateral escalation ladder

Q00:
## Summary

- The RLM frugality thesis lowers a decomposed child AC's model tier unconditionally, assuming decomposition made it easy enough — but nothing verified that assumption, so bad/ambiguous decompositions silently got handed to weak models and repeatedly failed validation.
- A root AC that kept failing even at the strongest configuration had no recovery path beyond retrying the identical setup, and eventually surfaced as a final FAILED status with no operator-visible signal explaining why.
- Neither of the above was visible anywhere a human looks (TUI, Kanban, HUD, or the conductor's own decision reports), and the TUI additionally hard-truncated AC text to fixed character counts regardless of terminal width.

## Changes

**Decomposition trust (`decomposition_attestation.py`, new)**
- A decomposition round is judged trustworthy only when every sibling sub-AC passed its own verify gate AND the parent AC's own verify gate also passes when re-evaluated over the children's combined effect on the workspace — reusing the existing verify-gate machinery as the single source of truth instead of a separate heuristic.
- Fail-closed: an AC with no verify command, or missing verify-gate evidence on either axis, is never optimistically treated as trustworthy.
- Wired into `model_routing`'s child-tier-drop decision so an untrustworthy prior round withholds the cheap-tier drop on that root AC's next retry. Emits `execution.ac.decomposition_attested`.

**Lateral-persona escalation ladder (`lateral_escalation.py`, new)**
- A root AC that fails twice in a row at the terminal configuration (atomic, frontier model tier, max reasoning effort) now escalates through the existing lateral-thinking personas instead of blindly retrying the same setup. Persona selection reuses the same pattern-aware routing the UNSTUCK_LATERAL auto-pipeline already uses: for a generic/unclassified failure (the common case here) the actual order is hacker → contrarian → architect → researcher → simplifier — contrarian is that selector's universal fallback, tried second, not last.
- Once all personas are exhausted and the AC still fails, it's parked with a durable `execution.ac.parked_for_operator` event and continues retrying at a much longer, bounded cadence — never surfacing a final FAILED status, but now with a real operator-visible signal instead of a silent loop.
- Gated behind `EconomicsConfig.lateral_escalation_enabled` (default off), mirroring this codebase's existing opt-in posture for experimental execution-cost levers.
- Infra-fatal failures (adapter crash, auth error) are exempted and still surface as failed through the existing mechanism.

**Observability wiring**
- Kanban board (`dashboard/board.py`, web `page.py`) now folds both new events into card badges (trust verdict, escalation/parked state).
- The AC-tree HUD (`ac_tree_hud_handler.py`) surfaces trust/escalation state across all its verbosity modes.
- The conductor's decision-recording response (`conductor_handler.py`) now echoes back the submitted `verification_summary`/`selected_action` plus relevant trust/escalation state, instead of a bare "decision recorded" confirmation.

**TUI truncation fix**
- `ac_tree.py`/`ac_progress.py` labels now derive their truncation budget from the widget's actual available width instead of a fixed character constant (30/40/45/50 chars regardless of terminal size).
- The web Kanban's client-side title truncation now appends an ellipsis when it actually truncates.

## Notes

- Full test suite (`pytest -n 4`) diffed against the branch point: identical pre-existing failure set on both sides (unrelated opencode/codex backend-dispatch tests) — zero regressions introduced by this change.
- `lateral_escalation_enabled` ships default-off; a real infinite-sleep hang was caught in this branch's own full-suite run before the fix, which is why it's opt-in rather than on by default for now.


lee3Q:
@/private/tmp/ozo-pr1648-fallback-comment-round2.md

Q00:
Superseded by #1681.

The decomposition-attestation judge, lateral-escalation primitives, observability requirements, focused tests, and Textual display-cell fixes remain useful reference material. The integration is being re-sequenced so execution authority and attempt-vs-acceptance event semantics land before reusable trust, recovery, and routing depend on them.

Future work will be delivered as approval-gated stacked PRs, with only one unapproved implementation PR open at a time. The branch is intentionally preserved for implementation and test reuse.

PR #1662, opened 2026-07-19
Anchor PM interview success criteria to delivered behavior, not post-launch KPIs

### LINKED ISSUE #1663: PM interview success-criteria questions drift into post-launch KPI measurement
## Problem

When the PM interview (`ooo pm`) probes success criteria, it sometimes generates questions about **post-launch outcome measurement** instead of the specification itself. Observed while creating a PM project for an AlimTalk (notification template) change:

- "How much should the conversion rate rise after the template change?"
- "For how many days will you observe the metric to declare success?"

These are not PRD success criteria. A PRD is a **contract between the PM and the developers**: it communicates the PM's intent and the policies the product must embody, and its success criteria define what the delivered feature must satisfy for the PM to accept it as built — "this is what I had in mind", expressed in terms of behavior and policy. KPI and conversion tracking are what the PM does **after** the feature ships; they are the PM's own follow-up work, not obligations the developers can deliver on, and therefore have no place in this contract.

## Root cause

Two mechanisms combine to pull the interview toward business metrics:

1. **Scoring feedback loop.** The ambiguity scorer's rubric for `success_criteria_clarity` asks "Are success criteria measurable?" (`src/ouroboros/bigbang/ambiguity.py`). Its weakest-dimension justification is injected into every question-generation prompt as "Weakest area / Reason", and persona question candidates deterministically target the worst-scoring dimension (`src/ouroboros/bigbang/interview.py`). An LLM reading "measurable" with no counter-steering resolves this pressure toward quantitative business metrics.
2. **Classifier passthrough.** `QuestionClassifier`'s PLANNING category explicitly listed "Success metrics, KPIs, acceptance criteria" (`src/ouroboros/bigbang/question_classifier.py`), so KPI questions matched PLANNING and were forwarded to the PM unchanged — even though its DECIDE_LATER category already covers "questions requiring data that doesn't exist yet", which post-launch metrics are.

## Constraint

The inner InterviewEngine and its ambiguity rubric are the system's shared philosophy and must not be modified. The correction belongs in the PM steering layer (`_PM_SYSTEM_PROMPT_PREFIX`) and the PM-only `QuestionClassifier`: steer with a one-paragraph philosophy of what a PRD is (not per-question prohibition lists), and route post-launch agenda into the existing DECIDE_LATER deferral plumbing, which already exempts deferred items from clarity-score penalties.

cohemm:
## Summary

Steers the PM interview so success-criteria questions stay focused on **policy and delivered behavior** (what the PM must observe in the shipped feature to say "this is what I intended") instead of drifting into **post-launch outcome metrics** (conversion uplift targets, KPI observation windows). The fix lives entirely in the PM steering layer; the inner InterviewEngine and ambiguity scorer are untouched.

## Background

- **Observed problem.** In PM interviews (e.g. an AlimTalk template change project), the engine asked questions like "how much should conversion rise?" and "how many days will you observe success?". These are post-launch measurement agenda, not PRD success criteria — a PRD is a **contract between the PM and the developers**, communicating the PM's intent and the policies the product must embody; its success criteria define what the delivered feature must satisfy for the PM to accept it as built. KPI and conversion movement is what the PM tracks **after** the feature ships — it is not an obligation the developers can deliver on, so it does not belong in the PRD.
- **Root cause.** The ambiguity scorer's `success_criteria_clarity` rubric ("Are success criteria measurable?") feeds its weakest-dimension justification back into every question-generation prompt, and persona candidates deterministically target the worst-scoring dimension. Left unsteered, the generator resolves that pressure toward business-metric questions. The `QuestionClassifier` then let them through because its PLANNING category explicitly listed "Success metrics, KPIs".

## Design decisions

1. **Do not touch the interview engine** (`interview.py`, `ambiguity.py`). The engine and its scoring rubric are the system's shared philosophy; correction belongs in the PM-specific steering layer that wraps it.
2. **Steer with a philosophy statement, not prohibition lists.** `_PM_SYSTEM_PROMPT_PREFIX` gains one paragraph defining the PRD's boundary. Since the steering prefix is re-prepended above the full inner prompt every round (including the "Weakest area" scoring feedback), it governs *how* the generator resolves a weak success-criteria signal — toward acceptance/policy questions — rather than suppressing specific question types.
3. **Route post-launch agenda through the existing DECIDE_LATER plumbing.** The classifier (PM-layer only; used exclusively by `PMInterviewEngine`) now classifies post-launch outcome measurement as DECIDE_LATER. Deferred items already flow into scoring `additional_context` with an explicit "do NOT penalise" instruction, so deferring these questions also stops the scorer from dragging the interview back toward metrics — no engine change needed.
4. **Keep user-confirmed deferral (no auto-skip).** Auto-answering DECIDE_LATER questions was previously removed due to MCP 120 s timeouts on consecutive deferrals; this PR preserves the offer-and-confirm flow.

## Changes

| File | Change |
|------|--------|
| `src/ouroboros/bigbang/pm_interview.py` | Add PRD-boundary philosophy paragraph to `_PM_SYSTEM_PROMPT_PREFIX` |
| `src/ouroboros/bigbang/question_classifier.py` | PLANNING: replace "Success metrics, KPIs, acceptance criteria" with "Acceptance criteria — what the delivered behavior and policy must satisfy"; DECIDE_LATER: add post-launch outcome measurement (conversion uplift, KPI targets, adoption goals, observation windows) |

## Test plan

- [x] `uv run pytest tests/unit/bigbang/test_pm_interview.py tests/unit/bigbang/test_question_classifier.py tests/unit/bigbang/test_classifier_context.py -q` — 96 passed
- [x] `uv run ruff check` and `uv run ruff format --check` clean on both touched files
- [x] `uv run mypy` clean on both touched files

## Related issues

Fixes #1663

### DIFF
--- a/src/ouroboros/bigbang/pm_interview.py
+++ b/src/ouroboros/bigbang/pm_interview.py
@@ -18,8 +18,10 @@ from __future__ import annotations
 
 import asyncio
 from dataclasses import dataclass, field, replace
+import functools
 import json
 from pathlib import Path
+import re
 from typing import Any
 
 import structlog
@@ -71,13 +73,9 @@ You are a Product Requirements interviewer helping a PM define their product.
 Assume the resulting product requirements document will drive all downstream work through AI workflows, so elicit decisions precise enough for autonomous planning, implementation, and verification.
 If a product question is not settled, preserve that uncertainty explicitly instead of inventing certainty; capture assumptions and decide-later items as first-class PM output.
 
-A PRD is a contract between the PM and the developers: it communicates the PM's
-intent and the policies the product must embody. Success criteria define what
-the delivered feature must satisfy for the PM to accept it as built — "this is
-what I had in my mind" — expressed in terms of behavior and policy. Post-launch
-outcomes (adoption, conversion, KPI movement) are the PM's own follow-up work,
-not obligations the developers can deliver on, so they have no place in this
-contract.
+A PRD is a contract between the PM and the developers: success criteria are the
+behavior and policy the PM must observe in the delivered feature to accept it
+as built. Post-launch outcomes have no place in this contract.
 
 Focus on: goal, user stories, constraints, success criteria, assumptions.
 
@@ -134,6 +132,31 @@ _INNER_GUIDANCE_STATIC_MARKERS = (
 _PM_CONTRACT_MARKER = "contract between the PM and the developers"
 
 
+@functools.lru_cache(maxsize=1)
+def _inner_base_prompt_sections() -> tuple[str, ...]:
+    """Sections of the inner interviewer base prompts, for parity checks.
+
+    The inner builder truncates its base prompt from the end, so every
+    ``##`` section the unconstrained build retains completely must also
+    survive the steered build completely (a section head without its tail
+    means the boundary rules were cut mid-text). Sections are collected
+    from both base prompt variants; the parity filter against the actual
+    unconstrained build decides which apply at runtime.
+    """
+    from ouroboros.agents.loader import load_agent_prompt
+    from ouroboros.bigbang.interview import _TOOLLESS_INTERVIEW_BASE_PROMPT
+
+    texts = [_TOOLLESS_INTERVIEW_BASE_PROMPT]
+    try:
+        texts.append(load_agent_prompt("socratic-interviewer"))
+    except Exception:  # noqa: BLE001 - agent file is optional at runtime
+        pass
+    sections: list[str] = []
+    for text in texts:
+        sections.extend(s for s in re.split(r"(?=\n## )", text) if s.strip())
+    return tuple(sections)
+
+
 def _shed_one_paragraph(block: str) -> str:
     """Drop the lowest-priority paragraph from a fitted steering block.
 
@@ -155,28 +178,34 @@ def _shed_one_paragraph(block: str) -> str:
 def _fit_steering_paragraphs(steering: str, budget: int) -> str:
     """Fit whole steering paragraphs into ``budget`` characters.
 
-    Paragraphs are included in document order and atomically: one that does
-    not fit (with its trailing ``"\\n\\n"`` separator) is dropped along with
-    everything after it, so a reduced budget can only narrow the steering
-    policy — never cut a sentence in half or strip the exclusion clause off
-    a paragraph and invert its meaning.
+    Paragraphs are included atomically: one that does not fit (with its
+    trailing ``"\\n\\n"`` separator) is skipped whole, so a reduced budget
+    can only narrow the steering policy — never cut a sentence in half or
+    strip the exclusion clause off a paragraph and invert its meaning.
 
-    Returns the fitted block ending with ``"\\n\\n"``, or ``""`` when not
-    even the first paragraph fits.
+    Selection follows the shedding priority, not document order: the
+    PRD-contract paragraph is placed first, then the remaining paragraphs
+    in document order as budget allows. The fitted paragraphs are emitted
+    in their original document order.
+
+    Returns the fitted block ending with ``"\\n\\n"``, or ``""`` when no
+    paragraph fits.
     """
+    paragraphs = [p for p in steering.split("\n\n") if p.strip()]
+    by_priority = sorted(
+        range(len(paragraphs)),
+        key=lambda i: (0 if _PM_CONTRACT_MARKER in paragraphs[i] else 1, i),
+    )
     fitted_len = 0
-    fitted: list[str] = []
-    for paragraph in steering.split("\n\n"):
-        if not paragraph.strip():
-            continue
-        cost = len(paragraph) + 2  # paragraph + "\n\n" separator
-        if fitted_len + cost > budget:
-            break
-        fitted.append(paragraph)
-        fitted_len += cost
-    if not fitted:
+    chosen: set[int] = set()
+    for index in by_priority:
+        cost = len(paragraphs[index]) + 2  # paragraph + "\n\n" separator
+        if fitted_len + cost <= budget:
+            chosen.add(index)
+            fitted_len += cost
+    if not chosen:
         return ""
-    return "\n\n".join(fitted) + "\n\n"
+    return "\n\n".join(paragraphs[i] for i in sorted(chosen)) + "\n\n"
 
 
 @dataclass
@@ -586,10 +615,12 @@ class PMInterviewEngine:
 
             # Guidance the unconstrained inner build retains must also
             # survive the steered build; steering yields otherwise. The
-            # ambiguity snapshot and the perspective panel are required as
-            # their full text so the "Weakest area" lines and the panel's
-            # breadth/closure/seed-ready safeguards survive — not just a
-            # heading or a mid-sentence fragment.
+            # ambiguity snapshot, the perspective panel, and each base
+            # prompt section (role/context boundaries etc.) are required as
+            # their full text so the "Weakest area" lines, the panel's
+            # breadth/closure/seed-ready safeguards, and the interviewer
+            # boundary rules survive — not just a heading or a mid-sentence
+            # fragment.
             base_full = original_build(
                 state,
                 initial_context=initial_context,
@@ -599,7 +630,12 @@ class PMInterviewEngine:
             panel_text = self.inner._build_perspective_panel_prompt(state)
             required_markers = [
                 marker
-                for marker in (*_INNER_GUIDANCE_STATIC_MARKERS, snapshot_text, panel_text)
+                for marker in (
+                    *_INNER_GUIDANCE_STATIC_MARKERS,
+                    *_inner_base_prompt_sections(),
+                    snapshot_text,
+                    panel_text,
+                )
                 if marker and marker in base_full
             ]
 
--- a/src/ouroboros/bigbang/pm_interview.py
+++ b/src/ouroboros/bigbang/pm_interview.py
@@ -474,6 +474,13 @@ class PMInterviewEngine:
 
         Idempotent — if already installed, replaces previous wrapper to prevent
         stacking across multiple start/resume calls on the same engine instance.
+
+        The wrapper preserves the inner engine's prompt-budget contract: the
+        steering prefix is charged against the caller-supplied ``max_chars``
+        (or the engine's default system-prompt cap) instead of being prepended
+        on top of an already-capped prompt, so the combined prompt never
+        exceeds the budget the caller computed against the serialized-prompt
+        safety ceiling.
         """
         self._pm_steering = getattr(self, "_pm_steering", _PM_SYSTEM_PROMPT_PREFIX)
 
@@ -483,9 +490,23 @@ class PMInterviewEngine:
 
         original_build = self._original_build_system_prompt
 
-        def _pm_build_system_prompt(state: InterviewState, *args, **kwargs) -> str:
-            base = original_build(state, *args, **kwargs)
-            return self._pm_steering + "\n\n" + base
+        def _pm_build_system_prompt(
+            state: InterviewState,
+            initial_context: str | None = None,
+            max_chars: int | None = None,
+        ) -> str:
+            steering_block = self._pm_steering + "\n\n"
+            cap = max_chars or self.inner._MAX_SYSTEM_PROMPT_CHARS
+            # Never pass 0 down: the inner builder treats falsy max_chars as
+            # "use the default cap", which would blow the budget again.
+            inner_budget = max(1, cap - len(steering_block))
+            base = original_build(
+                state,
+                initial_context=initial_context,
+                max_chars=inner_budget,
+            )
+            # Hard cap as final safety net (degenerate tiny-cap case).
+            return (steering_block + base)[:cap]
 
         self.inner._build_system_prompt = _pm_build_system_prompt  # type: ignore[assignment]
 
--- a/src/ouroboros/bigbang/pm_interview.py
+++ b/src/ouroboros/bigbang/pm_interview.py
@@ -481,6 +481,13 @@ class PMInterviewEngine:
         on top of an already-capped prompt, so the combined prompt never
         exceeds the budget the caller computed against the serialized-prompt
         safety ceiling.
+
+        Under tight caps the steering yields, never the inner prompt: the
+        inner builder is guaranteed its minimum system-prompt allocation
+        first, because it carries the engine's operating instructions (role,
+        one-question rule, round, initial context, ambiguity snapshot).
+        Steering is philosophy — safe to truncate or drop when a saturated
+        history leaves only the minimum budget.
         """
         self._pm_steering = getattr(self, "_pm_steering", _PM_SYSTEM_PROMPT_PREFIX)
 
@@ -495,8 +502,12 @@ class PMInterviewEngine:
             initial_context: str | None = None,
             max_chars: int | None = None,
         ) -> str:
-            steering_block = self._pm_steering + "\n\n"
             cap = max_chars or self.inner._MAX_SYSTEM_PROMPT_CHARS
+            inner_reserve = min(cap, self.inner._MIN_SYSTEM_PROMPT_CHARS)
+            steering_block = self._pm_steering + "\n\n"
+            steering_budget = max(0, cap - inner_reserve)
+            if len(steering_block) > steering_budget:
+                steering_block = steering_block[:steering_budget]
             # Never pass 0 down: the inner builder treats falsy max_chars as
             # "use the default cap", which would blow the budget again.
             inner_budget = max(1, cap - len(steering_block))
--- a/src/ouroboros/bigbang/pm_interview.py
+++ b/src/ouroboros/bigbang/pm_interview.py
@@ -114,6 +114,33 @@ Respond ONLY with valid JSON in this exact format:
 """
 
 
+def _fit_steering_paragraphs(steering: str, budget: int) -> str:
+    """Fit whole steering paragraphs into ``budget`` characters.
+
+    Paragraphs are included in document order and atomically: one that does
+    not fit (with its trailing ``"\\n\\n"`` separator) is dropped along with
+    everything after it, so a reduced budget can only narrow the steering
+    policy — never cut a sentence in half or strip the exclusion clause off
+    a paragraph and invert its meaning.
+
+    Returns the fitted block ending with ``"\\n\\n"``, or ``""`` when not
+    even the first paragraph fits.
+    """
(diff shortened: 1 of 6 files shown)


cohemm:
Addressed the review in 8161ffd7 (note: the review ran against 5bd31318; 8d984394 had since reworded the steering paragraph to contract framing, which made the prefix even longer — the budget fix covers both).

**Blocker 1 — steering bypasses prompt-budget enforcement: fixed.**
Verified the finding: `ask_next_question` computes `system_prompt_budget` against the serialized-prompt safety ceiling and passes it as `max_chars`, and the old wrapper prepended the steering block *after* that cap was applied. `_pm_build_system_prompt` now charges the steering block against the caller's cap (default `_MAX_SYSTEM_PROMPT_CHARS` when none is supplied) before delegating to the original builder, clamps the inner budget to ≥1 so a falsy value is never interpreted as "use the default cap", and hard-caps the combined result as a final safety net. New `TestPMSteeringPromptBudget` covers the default cap, a caller-supplied cap, and the degenerate tiny-cap case with steering installed.

**Follow-up — no regression coverage for the prompt policies: added.**
- `TestPMSuccessCriteriaBoundary` asserts the steering prefix defines the PRD as a PM-developer contract ("accept it as built", behavior-and-policy success criteria) and excludes post-launch outcomes from that contract. Since the steered system prompt starts with this prefix (asserted in the budget tests), PM question generation demonstrably receives the boundary guidance.
- `TestClassificationPromptPostLaunchBoundary` slices the classifier prompt by section and asserts PLANNING lists delivered-behavior acceptance criteria and contains no "KPI"/"Success metrics", while DECIDE_LATER lists post-launch outcome measurement (conversion uplift, KPI targets, adoption goals, observation windows).

Both test classes fail if either policy paragraph is removed, closing the mutation gap noted in the review.

**Verification:** `tests/unit/bigbang` 607 passed; MCP PM handler tests 98 passed; ruff check/format and mypy clean on all touched files.

cohemm:
Addressed the second review's blocker in c196c033.

**Blocker — saturated histories evict the inner interviewer prompt: fixed.**
Verified the reproduction premise: `_MIN_SYSTEM_PROMPT_CHARS` is 1,200 and the steering block is 1,309 chars, so at the minimum runtime budget the previous wrapper gave the inner builder `max(1, 1200 - 1309) = 1` char and the hard cap left steering-only output. The wrapper now reserves `min(cap, inner._MIN_SYSTEM_PROMPT_CHARS)` for the inner prompt **first**, and steering only receives the remainder (`max(0, cap - reserve)`), truncated or dropped to fit. The precedence is deliberate: the inner prompt carries the engine's operating instructions (role, one-question rule, round, initial context, ambiguity snapshot) and must always survive; steering is philosophy and safe to shed under pressure. At comfortable budgets (default 3,500 cap) nothing changes — full steering plus a 2,191-char inner prompt.

**Regression coverage added** in `TestPMSteeringPromptBudget`:
- `test_saturated_history_minimum_budget_preserves_inner_prompt` — cap = `_MIN_SYSTEM_PROMPT_CHARS`: asserts the total cap holds and the inner dynamic-header markers ("conducting a Socratic interview", "reduce ambiguity") survive while steering yields entirely.
- `test_mid_cap_truncates_steering_before_inner_prompt` — cap = 2,000: steering is truncated to exactly the space above the inner reserve and the inner markers still survive.
- Default-cap and caller-cap tests now also assert the inner operating instructions are present alongside the full steering prefix.

**Verification:** `tests/unit/bigbang` 609 passed; MCP PM handler tests 98 passed; ruff check/format and mypy clean.

cohemm:
Addressed the third review's blocker in a2c57ce5.

**Blocker — intermediate budgets garble or invert the steering policy: fixed.**
Confirmed the failure: at a 2,000-char cap the previous wrapper sliced the steering block at character 800, ending mid-phrase ("Post-launch outcomes (adoption, conversion") and concatenating directly into the inner header — capable of inverting the exact policy this PR introduces. The wrapper now composes steering via `_fit_steering_paragraphs`: paragraphs are fitted **whole**, in document order, against the budget above the inner reserve; the first paragraph that does not fit is dropped along with everything after it, and the fitted block always ends with a clean paragraph separator. Verified composition across budgets: 0 paragraphs below 453 chars of headroom, then 1 → 2 → 4 complete paragraphs as headroom grows — never a partial sentence.

**Test replacement — semantic composition instead of raw truncation.**
The mid-cap test that explicitly accepted character-level truncation is gone. In its place, a parametrized sweep over caps 1,200–2,600 (step 100) asserts on the composed runtime prompt: (1) total cap holds, (2) inner operating instructions survive at every cap, (3) every steering paragraph is either fully present or fully absent (a paragraph head appearing without its tail fails the test), and (4) the post-launch exclusion clause is never separated from its subject — if "Post-launch outcomes" appears, "no place in this contract" must too.

**Verification:** `tests/unit/bigbang` 623 passed (15 new parametrized cases); MCP PM handler tests 98 passed; ruff check/format and mypy clean.

cohemm:
Addressed the fourth review's blocker in 16dfaf2a.

**Blocker — long initial context evicts the ambiguity snapshot: fixed.**
Reproduced exactly as described: with a 1,800-char (`_MAX_INITIAL_CONTEXT_SYSTEM_CHARS`) context, a stored ambiguity breakdown, and the default 3,500 cap, the wrapped prompt lost the answer-prefix legend and the entire ambiguity snapshot while the unwrapped build retained both.

The fixed 1,200-char reserve was the wrong tool — the inner prompt's essential size varies with state. The wrapper now derives the requirement from the state itself: it builds the unconstrained inner prompt at the full cap, collects the guidance that build retains (answer-prefix legend, brownfield intent hint, and the **full ambiguity snapshot text** — requiring only the heading would still let the "Weakest area" lines be cut), then sheds steering paragraphs one at a time until every retained marker also survives the steered build, or no steering remains. Since the inner header truncates from the end, a surviving later element guarantees everything before it is intact.

Shedding priority: supporting paragraphs go first; the PRD-contract paragraph — the policy #1663 exists to enforce — is kept until nothing else remains, so the policy stays operative on exactly the mid-interview rounds where the "Weakest area: Success Criteria" signal fires.

**Probed outcomes at the default cap:** short context → all 4 steering paragraphs, full snapshot, parity with unwrapped; long context + snapshot → contract paragraph survives alongside the full snapshot and legend; long brownfield context + snapshot → steering yields entirely, wrapped output keeps full parity with the unwrapped build.

**Regression coverage added:** the review-requested case (`_MAX_INITIAL_CONTEXT_SYSTEM_CHARS` context + ambiguity breakdown) asserting the full snapshot text and legend survive; a contract-outlives-supporting-paragraphs test; and a brownfield saturation test asserting steering yields before any inner guidance is lost.

**Verification:** `tests/unit/bigbang` 626 passed; MCP PM handler tests 98 passed; ruff check/format and mypy clean.

cohemm:
Addressed the fifth review's blocker in 440f0868.

**Blocker — default-cap prompts truncate the perspective panel: fixed.**
Reproduced: on a first-round PM prompt at the default 3,500 cap, the wrapped build ended mid-sentence inside the panel ("...written-output requirements w") and lost the breadth-recap and seed-ready safeguards, while the unwrapped build retained the full panel.

The perspective panel now joins the ambiguity snapshot as a **full-text** preservation requirement, derived per state from `_build_perspective_panel_prompt`: whatever guidance the unconstrained build retains completely must survive the steered build completely, or steering sheds another paragraph. Probed outcomes at the default cap: first round → steering sheds one paragraph (the uncertainty paragraph, whose content is already stated in the opening paragraph) and keeps the role + PRD-contract paragraphs alongside the complete panel with all safeguards; snapshot rounds → contract paragraph, full panel, and full snapshot coexist; long-context rounds where the unwrapped build itself truncates the panel are treated by parity and keep the contract + full snapshot as before; brownfield saturation still sheds steering entirely.

**Regression coverage:** the default-cap test now asserts the composed runtime prompt semantically — exact full panel text present, "breadth recap" and "seed-ready" safeguards intact, contract policy retained, steering paragraphs atomic; the reduced caller-cap test asserts the same coexistence. Existing long-context, brownfield, saturated-minimum, and 1,200–2,600 sweep tests all still pass unchanged.

**Verification:** `tests/unit/bigbang` 626 passed; MCP PM handler tests 98 passed; ruff check/format and mypy clean.

cohemm:
Addressed the sixth review (blocker + warning) in fe91a432.

**Blocker — base-prompt boundary sections displaced: fixed.**
Measured the actual composition: the unwrapped default-cap build retains 1,002 chars of the Socratic base prompt, which completely contains `# Socratic Interviewer` intro, `## CRITICAL ROLE BOUNDARIES` (cum. 493), and `## CONTEXT BOUNDARIES` (cum. 768) — `## RESPONSE FORMAT` onward is already truncated in the unwrapped build. Every `##` section of both inner base prompt variants (agent markdown and the toolless prompt) now joins the parity markers: any section the unconstrained build retains completely must survive the steered build completely, or steering sheds.

**Warning — document-order fitting contradicts contract-first priority: fixed.**
`_fit_steering_paragraphs` now selects by shedding priority (contract paragraph first, then document order), emits in document order, and skips whole paragraphs that don't fit instead of stopping at the first overflow. A regression test pins the single-paragraph budget case: the contract wins over the generic opening.

**Consequence and trade-off, stated explicitly.** With the boundary sections required, the true surplus above parity at the default first-round cap is ~232 chars. The contract paragraph is therefore compacted from 482 to 220 chars (all asserted phrases kept), so the first round now carries the contract policy alongside the complete boundary sections and full panel. On rounds where parity leaves no surplus at all — reduced caller caps and snapshot rounds, where the ambiguity snapshot plus ROLE BOUNDARIES consume the budget — steering sheds entirely and the wrapped build keeps full parity with the unwrapped one. On those rounds the feature's second mechanism (the classifier routing post-launch outcome questions to DECIDE_LATER) remains the guard. This is the direct consequence of two sound constraints meeting a saturated 3,500-char budget: complete inner parity and any PM steering cannot both hold everywhere; where they conflict, steering yields, as this review series established.

**Regression coverage:** default-cap boundary-sections test (both sections complete alongside the contract), priority-fit unit test, and the caller-cap test now asserts parity semantics (whatever unwrapped retains completely survives) instead of assuming steering presence.

**Verification:** `tests/unit/bigbang` 628 passed; PM/classifier/interview slices 642 passed; ruff check/format and mypy clean.

Q00:
@cohemm can you resolve the follow up issue? I love your direction

src/ouroboros/bigbang/pm_interview.py
```diff
@@ -106,6 +112,102 @@
...
+# element implies everything before it is intact.
+_INNER_GUIDANCE_STATIC_MARKERS = (
+    "[from-research]",
+    "not on discovering what exists.",
+)
+
+# Identifies the PRD-contract paragraph — the policy #1663 exists to
+# enforce. It is shed last so tight budgets drop the supporting paragraphs
+# before the policy itself.
+_PM_CONTRACT_MARKER = "contract between the PM and the developers"
```
Q00, line 121: This one is good. but i think we can make these markers to first class object. what do you think?

cohemm: Done in b4577c85 — the markers are now first-class: `InnerGuidanceInvariant` (frozen dataclass with `name` / `why` / `resolve`) and the `INNER_GUIDANCE_INVARIANTS` registry declare everything PM steering must never evict, and the wrapper just consults the registry. Making `resolve` a per-build callable also fixed the approval review's follow-up finding for free: base-prompt sections are resolved through the agent loader on every build, so there is no stale cache after an operator prompt reload. Thanks for the nudge — the declarative form is much easier to reason about.

cohemm:
b4577c85 resolves the approval review's follow-up finding and the maintainer's requests in one restructure — @Q00 this covers "resolve the follow up issue" plus the first-class markers suggestion. `interview.py` remains untouched throughout.

**1. Guidance invariants are first-class — and wrapper-agnostic** (as of fc8e6f64). The scattered marker constants, the cached section splitter, and the inline snapshot/panel checks are replaced by `InnerGuidanceInvariant` + the `INNER_GUIDANCE_INVARIANTS` registry — six named, explained invariants (initial context, answer-prefix legend, brownfield intent hint, ambiguity snapshot, perspective panel, base-prompt sections), each resolving its full marker texts per build. The contract and its enforcement live in a neutral module, `bigbang/inner_guidance.py` (`reserve_steering_extension` + `compose_steered_prompt`), so the PM wrapper is just one consumer: any future steering wrapper reuses the same interview-layer-first guarantees with its own steering text and shed priority, demonstrated in `test_inner_guidance.py`.

**2. Follow-up finding (stale section cache) fixed.** The `lru_cache` is gone; the base-prompt-sections invariant resolves through the agent loader on every build, so `agents.loader.clear_cache()` reloads are honored immediately — an operator-supplied boundary section is protected from the first build after a reload. Regression test included (`test_base_prompt_sections_follow_agent_loader`).

**3. Steering now rides in a reserved budget extension instead of displacing inner content.** `_install_pm_steering` widens the PM-owned inner *instance's* `_MAX/_MIN_SYSTEM_PROMPT_CHARS` by exactly the steering length (class defaults and dev interviews untouched; idempotent across resume reinstalls). `ask_next_question` computes budgets and trims history against the widened numbers, so the wire ceiling (`_MAX_TOTAL_PROMPT_CHARS`) stays enforced by the inner engine itself. On the normal path the inner build keeps the engine's *designed* budget — `test_normal_path_inner_build_matches_designed_dev_build` asserts byte-identity with an unwrapped dev build — and the full steering lives in the extension. Probed across round types (first round, snapshot rounds, long-context, brownfield saturation, minimum-floor cap): full steering present in all of them, where the previous displacement design shed it on most. The cost is explicit: history capacity shrinks by the extension (~804 chars), trimming the oldest round roughly one round earlier in very long interviews. The fit-and-shed path from earlier review rounds survives as the interview-layer-first safety net for caller-supplied small caps and history-decline pressure, driven by the same invariant registry.

**4. Steering prefix deduplicated** (1,045 → 802 chars): the uncertainty rule was stated twice (opening paragraph + `PM_UNCERTAINTY_GUIDANCE`); the duplicate is removed and the downstream-AI sentence tightened. All prior phrase-level regression tests still pass.

**5. Classifier hardening ported from #1664, with credit:** the PRD-contract philosophy paragraph now precedes the category list (the classifier gets the "why", not just reworded bullets), and the DECIDE_LATER bullet adds the generalizing rationale — post-launch measurement "depends on real-world usage data" — alongside the enumerated examples.

---

## Design rationale (decision record)

**The concern that drove this change.** Under the previous displacement design, steering coverage was measured at roughly 3 of 30 rounds: the prefix survived only the pre-scoring rounds. From round 4 the ambiguity snapshot consumed the surplus, and once history saturated the serialized budget the fit budget was structurally zero — the policy silently dropped out on exactly the rounds where weakest-dimension pressure pushes generation toward KPI questions. Worse, whether steering applied flickered with snapshot length round to round. The maintainer's requirements were explicit: the PM policy must be effective on most rounds, and deterministically — "sometimes on, sometimes off" is not acceptable.

**Alternative considered and rejected: a generic cap raise (+500).** Measured, not assumed: raising `_MAX_SYSTEM_PROMPT_CHARS` to 4,000 also enlarges the unwrapped baseline, which then retains more of the base prompt (`## RESPONSE FORMAT` becomes complete), so the parity bar rises with the cap and absorbs the extra budget — the snapshot-round surplus actually *decreased* (159 → 124 chars). A generic raise can never fund steering; the parity staircase eats it.

**Decision: a reserved extension sized exactly to the prefix.** Widening the budgets by `len(prefix) + 2` and keeping the unwrapped comparison point at the *designed* cap makes the extension invisible to the parity bar: the inner build stays byte-identical to a dev interview, and steering presence becomes deterministic across every round type. The size is self-describing — it tracks the prefix, so nobody has to re-tune a magic number when the steering text changes.

**The trade-off was accepted deliberately.** The extension is paid for by history capacity (~804 chars, i.e. the oldest round is trimmed roughly one round earlier in very long interviews). That was judged the cheapest place to pay: marginally older history versus silently losing the PRD policy on most rounds.

**The priority rule is unchanged.** "The interview layer always wins" holds throughout — now stated declaratively by `INNER_GUIDANCE_INVARIANTS` — and when a caller-supplied cap genuinely cannot host both, the shed path still makes steering yield first.

**Verification:** `tests/unit/bigbang` 631 passed (incl. new budget-extension, dev-parity, reload, and classifier-philosophy tests); full bigbang+MCP+CLI sweep 3,198 passed; ruff check/format and mypy clean.

cohemm:
Addressed the seventh review (blocker + warning) in fc8e6f64, together with a structural change that made both fixes trivial.

**Structural: the steering contract is now a wrapper-agnostic module.** Following up on @Q00's first-class-objects suggestion, `bigbang/inner_guidance.py` now owns both sides of the contract: the `INNER_GUIDANCE_INVARIANTS` registry (declaration) and `reserve_steering_extension` + `compose_steered_prompt` (enforcement, with parameterized shed priority). `PMInterviewEngine`'s wrapper is reduced to a single `compose_steered_prompt(..., shed_last_marker=_PM_CONTRACT_MARKER)` call. Any future steering wrapper — QA steering, team conventions — reuses the same interview-layer-first guarantees instead of re-deriving them; `test_inner_guidance.py` demonstrates this with a hypothetical second wrapper using its own steering text and shed priority. `interview.py` remains untouched.

**Blocker — saturated path evicts baseline-retained initial context: fixed.** Verified the premise: the registry had no invariant for the dynamic header's own content, so at the reachable saturated cap the steered build could cut the user's initial context while all listed markers passed. A new `initial-context` invariant resolves the exact header-embedded context text (via `_initial_context_for_system_prompt`), and `compose_steered_prompt` additionally protects the effective context override (the prompt-safe summary path) that state-based invariants cannot see — both under the same baseline filter, so contexts the unwrapped build itself trims are not steering's debt. Regression test drives the **real `ask_next_question` history-budget path** with 24 saturated rounds and a 1,611-char context, asserting the full embedded context (including its tail marker) survives the steered system prompt.

**Warning — legend truncated mid-line behind a fragment marker: fixed.** The `answer-prefix-legend` invariant now requires the complete final legend line of either builder variant instead of the bare `[from-research]` fragment. Since the header truncates from the end, the full last line surviving implies the entire legend above it is intact. Covered by a second saturated run-path regression test.

Because the contract is declarative now, both fixes amount to registry entries — the extraction paid for itself in the same commit.

**Verification:** `tests/unit/bigbang` 638 passed (new: 2 saturated run-path tests, 4 generic-wrapper/registry tests); full bigbang+MCP+CLI sweep 3,204 passed; ruff check/format and mypy clean.

PR #1667, opened 2026-07-19
feat(orchestrator): add the Ouroboros AC runtime

Q00:
Depends on #1648

## Summary

- Introduces an Ouroboros-owned acceptance-criterion runtime above Claude, Codex, OpenCode, and other `AgentRuntime` drivers.
- Defines atomic work by one independently verifiable outcome plus bounded blast radius; token thresholds remain planning/continuation safeguards, not semantic boundaries.
- Combines grill-style decision closure with RLM-inspired external memory, bounded retrieval, fresh provider sessions, and gate-proven trust.

### AC capsule vertical slice

- Compiles every AC attempt into a versioned, provider-neutral capsule with semantic identity, workspace, dispatch authority, success contract, and bounded context references.
- Persists only redacted/hash-based manifests and authorizes provider effects only after the capsule is durable.
- Starts a fresh provider session per AC attempt and allows resume only for the exact same capsule, workspace, runtime authority, retry, and node identity.

### Trust and recovery boundaries

- Binds tools, prompt modes, retry/sibling context, runtime execution identity, routing policy, verification modes, and custom verifier authority into capsule identity.
- Requires custom verifiers to expose a bounded canonical `verification_identity_contract()` before their trust can survive a process restart; otherwise their identity is process-local and fails closed.
- Refuses fresh redispatch when an interrupted attempt may already have emitted non-idempotent tool effects, including terminal-session crash gaps.

### RLM-inspired frugality

- Keeps large context outside the provider prompt and selects a newest-first bounded reference frontier with omission count and rolling digest.
- Removes synthetic paging claims until a real resolver exists, and bounds references and success-contract materialization before hashing or prompt construction.
- Caches repeated authority projections and incrementally replays attempt history to keep the runtime economical without weakening trust.

### Architecture

- Adds the RFC: **Ouroboros Runtime — an Acceptance-Criterion VM for AI Engineering**.
- Establishes the loop `GRILL -> COMPILE -> PAGE -> RUN -> PROVE -> LEARN` and separates Decision Threads, Context Pages/Probes, AC Capsules, continuation segments, and verifiers.

## Validation

- Pre-PR gear: `R3` (`12 files`, `3,873 additions`, authority/session-sensitive boundary).
- Independent adversarial review: final 2-vote verdict `CLEAN`.
- `tests/unit/orchestrator`: `3,587 passed, 1 skipped`.
- Full suite: `13,479 passed, 12 skipped`; the same 17 existing Auto/OpenCode environment failures reproduce on the #1648 base and are unrelated to this diff.
- Ruff check/format, targeted MyPy, and mechanical pre-PR checks pass.

## Notes

- This is intentionally stacked on #1648 and implements the first complete vertical slice: capsule compilation plus session ownership.
- Durable context-page resolution, read-only Probe execution, speculative frugal promotion, and token-triggered continuation remain follow-up stack layers.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,6 +26,7 @@ replayable execution contract on your choice of runtime backend.
 ### Architecture
 
 - [System Architecture](./architecture.md) - Six-phase architecture, runtime abstraction layer, and core concepts
+- [Ouroboros AC Runtime RFC](./rfc/ouroboros-ac-runtime.md) - Acceptance-criterion VM: Grill, context paging, fresh AC capsules, gate-anchored trust, and RLM-inspired frugality
 - [Interview Milestone Lateral Contract](./rfc/interview-milestone-lateral-contract.md) - Proposed contract for bounded lateral review at ambiguity milestone transitions
 - [CLI Reference](./cli-reference.md) - Command-line interface flags and options
 - [Configuration Reference](./config-reference.md) - All `config.yaml` options and environment variables
--- a/docs/events.md
+++ b/docs/events.md
@@ -154,8 +154,12 @@ Lifecycle events carry the same `ac_dispatch_id`. Dispatch events form one
 validated predecessor chain, and recovery considers only its unique head; it
 never infers a successor from timestamps, replay-list position, or lifecycle
 event type priority across boundaries. Only a correlated reusable same-attempt
-handle at that head can resume. A failed head without such a handle remains
-ambiguous, because provider effects may precede the reported adapter failure. A
+handle at that head can resume. Terminal failure takes precedence over a
+resumable handle: a head with a durable `execution.session.failed` is NOT
+resumed even when it carries a reusable handle, because provider effects may
+precede the reported failure and resuming would send another provider turn. The
+only safe resume of a failed head is an explicit `execution.session.recovered`
+linkage to a live replacement session; without it, recovery fails closed. A
 completed head blocks same-attempt redispatch as already terminal and
 reconstructs its durable successful result for the caller. When the AC ran a
 verify gate, the completed lifecycle also carries a bounded exact-shape
--- a/docs/events.md
+++ b/docs/events.md
@@ -107,12 +107,13 @@ session freshness.
 | `session_id` | `string` | Orchestrator session id |
 | `semantic_ac_key` | `string` | Stable semantic AC identity |
 | `capsule_fingerprint` | `string` | SHA-256 identity of the complete capsule contract |
-| `capsule` | `object` | Strict versioned `ACExecutionCapsule` contract |
+| `capsule_manifest` | `object` | Strict versioned manifest whose free-form values and paths are represented only by SHA-256 digests |
 | `session_origin` | `string` | `"fresh"` or `"restored_same_attempt"` |
 
-The nested capsule contains compact facts and typed context references, not a
-provider transcript. A native runtime handle is bound to the capsule fingerprint
-and may resume only the same AC attempt.
+The nested manifest contains compact identities and typed context-reference
+digests, not a provider transcript or copies of free-form Seed/prompt/workspace
+text. Recovery strictly parses and re-fingerprints the manifest before a native
+runtime handle may resume the same AC attempt.
 
 ### mcp.job.cancelled
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -93,6 +93,27 @@ name further.
 | `ac_id` | `string` | Legacy source acceptance-criterion identifier for the execution unit |
 | `status` | `string` | Legacy worker completion status: `"passed"` means completed, `"failed"` means failed |
 
+### execution.ac.capsule.compiled
+
+Authority-bearing event emitted before a provider driver may start an AC
+attempt. Unlike observe-only telemetry, persistence failure blocks dispatch.
+The capsule makes Ouroboros — rather than Claude, Codex, or another provider —
+the owner of AC identity, context references, workspace authority, and native
+session freshness.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `execution_id` | `string` | Durable execution/run anchor |
+| `session_id` | `string` | Orchestrator session id |
+| `semantic_ac_key` | `string` | Stable semantic AC identity |
+| `capsule_fingerprint` | `string` | SHA-256 identity of the complete capsule contract |
+| `capsule` | `object` | Strict versioned `ACExecutionCapsule` contract |
+| `session_origin` | `string` | `"fresh"` or `"restored_same_attempt"` |
+
+The nested capsule contains compact facts and typed context references, not a
+provider transcript. A native runtime handle is bound to the capsule fingerprint
+and may resume only the same AC attempt.
+
 ### mcp.job.cancelled
 
 Emitted when a background MCP job is cancelled.
--- a/docs/events.md
+++ b/docs/events.md
@@ -221,7 +221,7 @@ and decomposed-parent verify gates.
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `verify_key` | `string` | Stable per-attempt gate identity. Namespaces: `atomic:…` (atomic gate), `decomposed_parent:…` (decomposed-parent gate), `apply_gate:…` (failed-runtime recovery gate), `flip_gate:…` (sibling-evidence flip gate), `skip_completed:…` (`--skip-completed` proof gate) |
+| `verify_key` | `string` | Stable per-attempt gate identity. `acgate:<execution>:<ac_index>:<retry>` is the single authoritative per-attempt key shared by the atomic gate, the failed-runtime recovery gate, and the sibling-flip gate, so a post-verify failure never reruns the command under a different key; `decomposed_parent:…` keys the decomposed-parent gate and `skip_completed:…` the pre-execution `--skip-completed` proof gate. A command-bearing completed-attempt recovery requires exactly one matching intent+outcome under this key before its cached PASS is trusted. |
 | `verify_command_digest` | `string` | Digest of the exact command, expected artifacts, and output assertion |
 | `execution_id` | `string` | Durable execution/run anchor |
 | `session_id` | `string` | Orchestrator session id |
--- a/docs/events.md
+++ b/docs/events.md
@@ -148,7 +148,11 @@ exact follow-up phase is not reconstructed from persisted metadata, so replaying
 the original AC prompt is refused rather than risking repeated acceptance work.
 Capsule authority also binds the actual subprocess `executable` (canonical path
 and realpath) and any launcher (e.g. a Zcode Electron/Node host), so a restart
-cannot resume the same capsule under a different binary.
+cannot resume the same capsule under a different binary. For a runtime that wraps
+a transport (e.g. `LeaderDrivenWorkerRuntime`), the transport's delegated
+`executable_identity_contract()` also contributes its command-affecting policy
+(e.g. Claude's `--add-dir` / `--disallowedTools`), so a change in launched-binary
+access changes authority too.
 
 Lifecycle events carry the same `ac_dispatch_id`. Dispatch events form one
 validated predecessor chain, and recovery considers only its unique head; it
@@ -205,7 +209,7 @@ and decomposed-parent verify gates.
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `verify_key` | `string` | Stable per-attempt gate identity (`atomic:…` or `decomposed_parent:…`) |
+| `verify_key` | `string` | Stable per-attempt gate identity. Namespaces: `atomic:…` (atomic gate), `decomposed_parent:…` (decomposed-parent gate), `apply_gate:…` (failed-runtime recovery gate), `flip_gate:…` (sibling-evidence flip gate), `skip_completed:…` (`--skip-completed` proof gate) |
 | `verify_command_digest` | `string` | Digest of the exact command, expected artifacts, and output assertion |
 | `execution_id` | `string` | Durable execution/run anchor |
 | `session_id` | `string` | Orchestrator session id |
--- a/docs/events.md
+++ b/docs/events.md
@@ -132,12 +132,23 @@ that provider invocation.
 |-------|------|-------------|
 | `ac_dispatch_id` | `string` | Unique correlation id for this exact provider entry |
 | `previous_ac_dispatch_id` | `string \| null` | Exact predecessor boundary in the same attempt; `null` only for the chain root |
+| `dispatch_kind` | `string` | Provider-entry phase: `"primary"` replays the AC prompt, `"session_signal_followup"` replays a Synapse signal turn. Absent on legacy records is treated as `"primary"`. |
+| `signal_id` | `string \| null` | SessionSignal id (follow-up dispatches only) |
+| `signal_mode` | `string \| null` | Effective signal mode, e.g. `"inform"`/`"after_turn"` (follow-up dispatches only) |
+| `follow_up_input_digest` | `string \| null` | `sha256:` digest of the exact rendered follow-up prompt (follow-up dispatches only) |
 | `execution_id` | `string` | Durable execution/run anchor |
 | `session_id` | `string` | Orchestrator session id |
 | `capsule_fingerprint` | `string` | Capsule authority used for this dispatch |
 | `session_origin` | `string` | `"fresh"` or `"restored_same_attempt"` |
 | `runtime_backend` | `string \| null` | Provider-neutral runtime selector |
 | `runtime` | `object \| null` | Minimal reconnect/authority handle; excludes cwd, transcript, tool catalog, capability graph, control plane, and arbitrary metadata |
+
+A non-primary (`session_signal_followup`) head fails closed on recovery: the
+exact follow-up phase is not reconstructed from persisted metadata, so replaying
+the original AC prompt is refused rather than risking repeated acceptance work.
+Capsule authority also binds the actual subprocess `executable` (canonical path
+and realpath) and any launcher (e.g. a Zcode Electron/Node host), so a restart
+cannot resume the same capsule under a different binary.
 
 Lifecycle events carry the same `ac_dispatch_id`. Dispatch events form one
 validated predecessor chain, and recovery considers only its unique head; it
@@ -158,8 +169,43 @@ such a linkage, conflicting session ids or equally authoritative response-chain
 cursors remain ambiguous and fail closed.
 
 The recovery handle contract is exact-shape, metadata-allowlisted, and bounded
-to 64 KiB before deserialization. Persisted payloads cannot smuggle excluded
-runtime context back into replay through unknown fields.
+to 64 KiB (measured in UTF-8 bytes) before deserialization. Persisted payloads
+cannot smuggle excluded runtime context back into replay through unknown fields.
+
+### execution.ac.dispatch.sealed
+
+Seals an already-completed provider turn immediately before a SessionSignal
+follow-up dispatch supersedes it. If the follow-up's own dispatch record fails
+to persist — or a crash lands before the follow-up reaches a terminal outcome —
+the predecessor becomes the chain head again. A sealed head with no durable
+terminal completion fails closed on recovery instead of resuming and replaying
+the original AC prompt over already-completed work.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `ac_dispatch_id` | `string` | The predecessor dispatch being sealed |
+| `execution_id` | `string` | Durable execution/run anchor |
+| `session_id` | `string` | Orchestrator session id |
+| `capsule_fingerprint` | `string` | Capsule authority of the sealed dispatch |
+
+### execution.ac.verify.intent / execution.ac.verify.outcome
+
+Make a non-idempotent `verify_command` single-shot across crash recovery. The
+intent is persisted immediately before the command runs and the outcome
+immediately after; both are keyed to a stable `verify_key` (the AC/node identity
+plus retry attempt) and a `verify_command_digest` binding the exact command and
+contract. On recovery the orchestrator consumes a recorded outcome instead of
+re-running the command; an intent with no matching outcome — the command may
+already have run — fails closed rather than replaying it. Applies to both atomic
+and decomposed-parent verify gates.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `verify_key` | `string` | Stable per-attempt gate identity (`atomic:…` or `decomposed_parent:…`) |
+| `verify_command_digest` | `string` | Digest of the exact command, expected artifacts, and output assertion |
+| `execution_id` | `string` | Durable execution/run anchor |
+| `session_id` | `string` | Orchestrator session id |
+| `verify_gate_outcome` | `object` | (outcome only) Bounded exact-shape gate result: `passed`, `reason`, `output_tail`, `missing_artifacts` |
 
 ### mcp.job.cancelled
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -182,12 +182,21 @@ cannot smuggle excluded runtime context back into replay through unknown fields.
 
 ### execution.ac.dispatch.sealed
 
-Seals an already-completed provider turn immediately before a SessionSignal
-follow-up dispatch supersedes it. If the follow-up's own dispatch record fails
-to persist — or a crash lands before the follow-up reaches a terminal outcome —
-the predecessor becomes the chain head again. A sealed head with no durable
-terminal completion fails closed on recovery instead of resuming and replaying
-the original AC prompt over already-completed work.
+General fail-closed marker: a sealed dispatch that ends up the active recovery
+head (with no durable terminal completion) is refused rather than resumed, so
+consumers must NOT read a seal as completion. It is written for every phase where
+a provider turn ran but its authoritative terminal is not yet durable, so a crash
+in that window cannot replay the AC prompt over already-done or ambiguous work:
+
+- a completed provider turn, immediately after its stream and BEFORE signal
+  handling, verification, evidence, and shadow replay (the later
+  `execution.session.completed` supersedes the seal on the normal path);
+- an effectful or opaque-driver stall (a stall after an observed tool effect, or
+  any stall on a driver that does not attest real-time tool-effect visibility);
+- a SessionSignal follow-up before it supersedes its (already-completed)
+  predecessor, and a stalled effectful follow-up.
+
+A sealed head fails closed on recovery instead of resuming and replaying the AC.
 
 | Field | Type | Description |
 |-------|------|-------------|
@@ -213,7 +222,11 @@ and decomposed-parent verify gates.
 | `verify_command_digest` | `string` | Digest of the exact command, expected artifacts, and output assertion |
 | `execution_id` | `string` | Durable execution/run anchor |
 | `session_id` | `string` | Orchestrator session id |
-| `verify_gate_outcome` | `object` | (outcome only) Bounded exact-shape gate result: `passed`, `reason`, `output_tail`, `missing_artifacts` |
+| `verify_gate_outcome` | `object` | (outcome only) Bounded exact-shape gate result: `passed`, `reason`, `output_tail`, `missing_artifacts` (count- and length-capped, with `missing_artifacts_omitted` when truncated). Enforced ≤ 64 KiB UTF-8 both at persistence and on replay. |
+
+An outcome is trusted only when preceded by exactly one matching intent for the
+same `verify_key` and `verify_command_digest`; orphan, duplicate, or out-of-order
+records fail closed rather than manufacture acceptance evidence.
 
 ### mcp.job.cancelled
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -120,6 +120,47 @@ also records `omitted_context_count` and a rolling `omitted_context_digest`.
 This is an auditable bounded-retrieval fact, not a claim that a page resolver
 already exists.
 
+### execution.ac.attempt.dispatched
+
+Authority-bearing transition persisted immediately before Ouroboros invokes a
+provider for an AC attempt, including resumed SessionSignal turns. A compiled
+capsule alone does not prove whether a particular provider boundary was crossed;
+this event makes every entry distinct and durable. Persistence failure blocks
+that provider invocation.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `ac_dispatch_id` | `string` | Unique correlation id for this exact provider entry |
+| `previous_ac_dispatch_id` | `string \| null` | Exact predecessor boundary in the same attempt; `null` only for the chain root |
+| `execution_id` | `string` | Durable execution/run anchor |
+| `session_id` | `string` | Orchestrator session id |
+| `capsule_fingerprint` | `string` | Capsule authority used for this dispatch |
+| `session_origin` | `string` | `"fresh"` or `"restored_same_attempt"` |
+| `runtime_backend` | `string \| null` | Provider-neutral runtime selector |
+| `runtime` | `object \| null` | Minimal reconnect/authority handle; excludes cwd, transcript, tool catalog, capability graph, control plane, and arbitrary metadata |
+
+Lifecycle events carry the same `ac_dispatch_id`. Dispatch events form one
+validated predecessor chain, and recovery considers only its unique head; it
+never infers a successor from timestamps, replay-list position, or lifecycle
+event type priority across boundaries. Only a correlated reusable same-attempt
+handle at that head can resume. A failed head without such a handle remains
+ambiguous, because provider effects may precede the reported adapter failure. A
+completed head blocks same-attempt redispatch as already terminal and
+reconstructs its durable successful result for the caller. When the AC ran a
+verify gate, the completed lifecycle also carries a bounded exact-shape
+`verify_gate_outcome`, so crash recovery does not replay a potentially
+non-idempotent `verify_command`. Recovery must not manufacture a fresh provider
+session merely because no first streamed message became durable.
+
+If `execution.session.recovered` explicitly links a failed resume id to a
+replacement resume id, that durable edge supersedes the failed handle. Absent
+such a linkage, conflicting session ids or equally authoritative response-chain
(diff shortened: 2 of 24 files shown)


Q00:
Addressed the prior review on HEAD `063c64865`:

- persist `execution.ac.attempt.dispatched` before every provider entry, including SessionSignal follow-up turns
- correlate lifecycle/tool events with a unique dispatch id and validated predecessor chain
- fail closed on ambiguous/corrupt dispatch history; recover completed dispatches without redispatch
- preserve OpenCode capsule/dispatch authority through compact persistence
- enforce manifest/reference and recovery-payload bounds
- persist cached verify outcomes so non-idempotent verify commands are not replayed after crash

Validation: focused recovery/capsule/runtime suite 480 passed; `tests/unit/orchestrator` 3625 passed, 1 skipped; full suite 13519 passed, 12 skipped with only the unchanged 17 Auto/OpenCode plugin-mode baseline failures; Ruff/MyPy clean. Fresh local R3 review (correctness, security, design, performance, contracts) is clean.

Please re-review the new HEAD.

Q00:
Addressed the three blockers from review `4731141469` on new HEAD `d5c47262b`:

**1. SessionSignal follow-up phase identity (`parallel_executor.py:10031`)** — dispatch events now persist an explicit `dispatch_kind`; follow-up dispatches additionally persist `signal_id`, `signal_mode`, and the exact rendered follow-up input digest (`sha256:`). The recovery loader fails closed (`AmbiguousACExecutionError`) on a non-primary dispatch head rather than resuming it and letting `_execute_atomic_ac` replay the original AC prompt. A legacy dispatch record with no `dispatch_kind` keeps its historical `primary` meaning (no inference); a follow-up missing its phase identity is rejected as corrupt.

**2. Alternate-harness verify policy (`parallel_executor.py:8628`)** — the cross-harness redispatch executor now inherits `run_verify_commands` and `verify_command_timeout_seconds`, so an operator-disabled or custom-timeout verify command is preserved on the alternate path instead of reverting to `True`/`600`.

**3. Effectful stall (`parallel_executor.py:9933`/`8437`)** — a stall observed *after* the dispatch emitted a durable tool effect (tracked on `LeafDispatchState.tool_effects_observed`) now fails closed as an ambiguous terminal outcome instead of taking the ordinary stall-retry path, which would bump `retry_attempt` and fresh-redispatch past the recovery ambiguity guard and could duplicate a non-idempotent effect. Effect-free stalls keep the existing retryable sentinel. The raise is re-propagated cleanly to the batch boundary (terminal FAILED, no bounce/redispatch), never reclassified as infra-fatal.

**Regression tests added:**
- follow-up dispatch head → fail-closed recovery; follow-up missing phase identity → rejected
- alternate redispatch inherits `run_verify_commands=False` + custom timeout
- effectful stall → provider entered exactly once, no redispatch; effect-free stall → still retryable

**Validation:** `tests/unit/orchestrator` — 3624 passed, 1 skipped; the only failures are 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (verified identical on untouched HEAD `063c64865`, only `python3` is installed). Focused capsule / runtime-scope / checkpoint-recovery / cross-harness / synapse suites all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `d5c47262b`.

Q00:
Addressed the three blockers from review `4731250656` (and the low follow-up) on new HEAD `47e6a6a3f`:

**1. Failed signal-follow-up write leaving a completed primary replayable (`parallel_executor.py:10043`)** — before persisting a SessionSignal follow-up dispatch, the already-completed predecessor is durably sealed via a new `execution.ac.dispatch.sealed` event. If the follow-up write fails (or a crash lands before its terminal outcome), recovery finds the predecessor as the chain head, sees it sealed, and fails closed instead of resuming it and replaying the original AC prompt. Regression: a real loader recovery where the follow-up write was lost fails closed.

**2. Verify command not single-shot across crash recovery (`parallel_executor.py:10279` + decomposed `7646`)** — a durable verify intent is persisted before a non-idempotent `verify_command` runs and its outcome after, keyed to a stable per-attempt `verify_key` + `verify_command_digest` (`execution.ac.verify.intent`/`.outcome`). Recovery consumes the outcome instead of re-running; an intent without an outcome fails closed; unreadable history fails closed. Applied to both the atomic gate and the decomposed-parent gate (which had no durable outcome before). Regressions: a recovered gate consumes the outcome (command runs once), and an intent-without-outcome fails closed without re-running.

**3. Capsule authority not binding the executable (`parallel_executor.py:2703`)** — the dispatch authority contract now includes a provider-neutral `executable` identity (canonical path + realpath) and any launcher (e.g. Zcode's Electron/Node host). Two otherwise-identical Codex adapters at `/bin/true` vs `/bin/false` now produce different authority/fingerprint. Regressions cover both the Codex path and the Zcode launcher.

**Follow-up (Low, `adapter.py:738`)** — the 64 KiB dispatch-recovery payload bound now measures `len(encoded.encode("utf-8"))`, not code points.

**Validation:** `tests/unit/orchestrator` — 3624 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (verified identical on untouched HEAD, only `python3` is installed). New restart-focused regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean. New events documented in `docs/events.md`.

Please re-review the new HEAD `47e6a6a3f`.

Q00:
Addressed the four blockers from review `4731331435` on new HEAD `ad98431c7`:

**1. Failed dispatch head resumed with a handle (`ac_runtime_handle_manager.py:1281`)** — terminal failure now takes precedence over a resumable handle. A head with a durable `execution.session.failed` is no longer resumed just because it has a reusable handle; the only safe resume is an explicit `execution.session.recovered` linkage to a live replacement, otherwise recovery fails closed before handle selection. Regression: a failed head with a correlated handle now fails closed; retries still occur via a fresh attempt.

**2. Effectful-stall ambiguity not persisted (`parallel_executor.py:10003`)** — the stall-after-effect path now seals the dispatch (`execution.ac.dispatch.sealed`) before raising, so a cold restart fails closed on the sealed head instead of re-entering the provider via the still-resumable session handle. Regression asserts the seal is persisted before the raise.

**3. `_apply_verify_gate` bypassed the single-shot oracle (`parallel_executor.py:11153`)** — the failed-runtime recovery gate now routes through `_run_verify_gate_single_shot` with a stable per-attempt key, so replaying a failed result consumes the durable outcome instead of running the `verify_command` twice. Regression: replaying a failed result runs the command once.

**4. Executable authority missed delegated transports (`parallel_executor.py:2597`)** — `_runtime_executable_identity` now honors an optional provider-neutral `executable_identity_contract()` and probes a wrapped `_transport`, so runtimes whose real CLI lives under a transport (e.g. `LeaderDrivenWorkerRuntime._transport._cli_path`) fingerprint their actual launcher. `LeaderDrivenWorkerRuntime` implements the contract delegating to its transport. Regressions cover both the explicit contract and the transport-probe fallback with `/bin/true` vs `/bin/false`.

**Validation:** `tests/unit/orchestrator` — 3632 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3` is installed; verified identical on untouched HEAD). New cold-restart/replay regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean. Recovery docs updated.

Please re-review the new HEAD `ad98431c7`.

Q00:
Addressed the four blockers from review `4731397433` on new HEAD `099b97183`:

**1. Provider completion not durable before post-processing (`parallel_executor.py:10007`)** — a completed provider turn is now sealed (`execution.ac.dispatch.sealed`) immediately after the provider stream, before signal/verify/evidence/shadow. A crash in that window fails closed on recovery instead of resuming the handle and replaying the AC prompt; the later `session.completed` supersedes the seal on the normal path. Regression asserts the seal precedes the terminal completion.

**2. Not every verify_command uses the oracle (`parallel_executor.py:6540` + `11492`)** — the `--skip-completed` proof gate (`skip_completed:…`) and the sibling-evidence flip gate (`flip_gate:…`) now route through `_run_verify_gate_single_shot`, joining the atomic, decomposed-parent, and `apply_gate` paths. Every command-bearing gate is single-shot. Regression: replaying the flip gate runs the command once.

**3. Delegated authority incompletely fingerprinted (`worker_runtime.py:153`)** — the Claude worker transport now publishes `executable_identity_contract()` including `--disallowedTools` / `--add-dir` / session persistence; `LeaderDrivenWorkerRuntime` delegates to it, and `_runtime_executable_identity` threads a bounded `command_policy` into capsule authority. Two workers with different `context_reference_dirs` now produce different authority. Regressions cover the executor pass-through and the real transport contract.

**4. SessionSignal follow-up stall ignored (`parallel_executor.py:10244`)** — per-dispatch stall/effect state is reset before the follow-up stream, and a follow-up that emits a tool effect then stalls is now sealed and fails closed (instead of being recorded as an acknowledged, successful signal turn). Regression is a full in-process after-turn run whose follow-up acknowledges, emits an Edit effect, then stalls: it fails closed and seals the follow-up dispatch, with no successful completion recorded.

Also documented the `apply_gate:`/`flip_gate:`/`skip_completed:` verify_key namespaces (the round-4 non-blocking note) and the delegated command policy.

**Validation:** `tests/unit/orchestrator` — 3637 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3` is installed; verified identical on untouched HEAD). New restart/replay regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `099b97183`.

Q00:
Addressed the six blockers from review `4731428073` on new HEAD `c1a5a374b`:

**1. Stall safety depended on observed tool messages (`parallel_executor.py:10054`)** — added `RuntimeCapabilities.realtime_tool_effect_visibility` (default False). The retryable stall path now requires the driver to attest real-time effect visibility; otherwise (buffered/opaque drivers like the Claude worker transport) the stall seals the dispatch and fails closed. Regressions cover both the attesting-retryable and opaque-fail-closed paths.

**2. Dispatch-without-capsule silently skipped (`ac_runtime_handle_manager.py:759`)** — a matching `attempt.dispatched` with a missing prerequisite `capsule.compiled` now fails closed instead of returning `None` and letting the caller re-enter as fresh. Regression added.

**3. Verify outcome trusted without an intent (`parallel_executor.py:11086`)** — outcomes are trusted only when preceded by exactly one matching intent (same `verify_key` + `verify_command_digest`); orphan, duplicate, or out-of-order records fail closed. Regression injects an orphan outcome and asserts the verifier is not trusted/run.

**4. Bare executable names fingerprinted via realpath not PATH (`parallel_executor.py:2600`)** — bare commands now resolve through the effective child PATH (`shutil.which`) before fingerprinting; an unresolved name is a distinct fail-visible identity. Regression: the same bare name under two PATHs (→ different binaries) yields different authority.

**5. Execution policy omitted signal-hub state (`parallel_executor.py:2880`)** — capsule execution policy now includes `session_signal_hub_enabled`, so an executor that registers/replays signals and runs follow-up turns no longer shares an authority hash with one that does not. Regression added.

**6. Verify outcome persistence unbounded (`execution_event_emitter.py:303`)** — outcomes are bounded (missing_artifacts count/length caps + a hard 64 KiB UTF-8 cap) at the source, at persistence, and on replay. Regression: 1,000 missing paths are capped with an omission count; a pathological payload over the byte cap is rejected.

Also broadened the `execution.ac.dispatch.sealed` docs (it is a general fail-closed marker, not only completion) and documented the verify intent/outcome ordering and bounded outcome.

**Validation:** `tests/unit/orchestrator` — 3644 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `c1a5a374b`.

Q00:
Addressed the four blockers from review `4731466527` (each a follow-up to a prior-round fix) on new HEAD `960976518`:

**1. Capabilities contract omitted `realtime_tool_effect_visibility` (`parallel_executor.py:2675`)** — the field is now serialized in `_runtime_capabilities_contract` and strict-parsed by the checkpoint dispatch-contract validator, so it participates in capsule/checkpoint authority. Regressions: a capsule-authority difference and a new `realtime_effect_visibility` checkpoint drift case.

**2. Opaque-driver follow-up stall not sealed (`parallel_executor.py:10343`)** — the follow-up stall branch now applies the same rule as the primary: seal + fail closed when a tool effect was observed OR the driver does not attest real-time effect visibility. An AFTER_TURN follow-up on an opaque driver that stalls with no observed effect is sealed and fails closed (no bounce/retry). Regression is a full in-process run whose opaque-driver follow-up acks then stalls silently.

**3. Verify-outcome replay coerced types (`execution_event_emitter.py:82`)** — `bound_verify_gate_outcome` now requires exactly the allowed keys with native types (real bool `passed`, string/null `reason`, string `output_tail`, list-of-string `missing_artifacts`) and rejects unknown fields; `"passed":"false"`/`1` no longer coerce to a trusted PASS. Regression exercises each coercion/shape rejection.

**4. Relative executable paths resolved against wrong cwd (`parallel_executor.py:2626`)** — relative paths with a separator now resolve against the effective child cwd (the workspace), so `./tool` binds the binary actually launched and follows a symlink-target swap under that cwd. Regression covers both.

**Validation:** `tests/unit/orchestrator` — 3649 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `960976518`.

Q00:
Addressed the three blockers from review `4731490478` on new HEAD `5ebdda3be`:

**1. Alternate-harness swallowed ambiguity (`parallel_executor.py:8683`)** — `_maybe_redispatch_alt_harness` now re-raises `AmbiguousACExecutionError` from the alternate run (it ran in the shared workspace and may have applied non-idempotent effects), instead of falling back to `None` and finalizing/escalating from stale pre-redispatch state. The `None` fallback is retained only for other exceptions (pre-provider-entry spawn failures). Regression: an alternate that raises ambiguity now propagates it.

**2. output_assertion-only contract passed unevaluated (`parallel_executor.py:11287`)** — `AcceptanceCriterionSpec` now rejects an `output_assertion` with no `verify_command` (model validator), with a defense-in-depth fail-closed in `_run_ac_verify_gate` for specs built via `model_construct`. An output-only contract can no longer report a success contract while passing without evaluating the assertion. Regression added.

**3. Watchdog policy omitted from authority (`parallel_executor.py:2821`)** — capsule runtime authority now binds the effective watchdog/termination policy (startup-output and stdout-idle timeouts, process-shutdown grace) via a provider-neutral `watchdog_identity_contract()` the runtime exposes; two `CodexCliRuntime` instances with idle timeouts 1 vs 999 now produce different capsule authority. Regressions cover the authority difference and the real Codex contract.

**Validation:** `tests/unit/orchestrator` — 3715 passed, 1 skipped (+ `tests/unit/core`/`evaluation` 1038 passed, confirming the schema-validator change has no fallout); the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `5ebdda3be`.

Q00:
Addressed the three blockers + the follow-up warning from review `4731513458` on new HEAD `21dfed012`:

**1. Post-verify failure reran the command under a separate key (`parallel_executor.py:10857`)** — the atomic gate, `_apply_verify_gate`, and the sibling-flip gate now share one per-attempt oracle key `acgate:<execution>:<ac_index>:<retry>`. If the atomic outcome was persisted but its result-carried copy was lost (e.g. a failed `session.completed`), the recovery gate consumes that durable outcome instead of re-running the command. Regression seeds the atomic intent+outcome and asserts `_apply_verify_gate` consumes it (command runs 0 times).

**2. Completed recovery trusted a cached PASS without the verify chain (`parallel_executor.py:8449`)** — command-bearing completed-attempt recovery now loads the single-shot oracle for the attempt and fails closed unless exactly one matching intent+outcome backs the completion. The prior recovery test (which supplied no intent record) now provides the full chain; a new regression asserts an orphan completion fails closed.

**3. Watchdog authority was not provider-neutral (`parallel_executor.py:2685`)** — `_runtime_watchdog_identity` now probes the well-known startup/stdout-idle/shutdown timeout attributes (and a wrapped `_transport`) in addition to an explicit `watchdog_identity_contract()`, so OpenCode/Hermes/Pi/GJC and wrapped runtimes contribute their policy; idle timeouts 1 vs 999 no longer share a capsule fingerprint. Regression covers the attribute-only path.

**4. (warning) Completed verify outcome bounded by chars not bytes (`ac_runtime_handle_manager.py:259`)** — the completed-lifecycle validator now bounds UTF-8 bytes, and the source caps each missing-artifact entry length, so a multi-byte artifact list cannot exceed the 64 KiB bound. Regression rejects an over-64-KiB-byte / under-char-limit outcome.

**Validation:** `tests/unit/orchestrator` — 3656 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `21dfed012`.

Q00:
Addressed the three blockers + the non-blocking note from review `4731540095` on new HEAD `cf2b1640d`:

**1. Child-PATH resolution ignored the child cwd (`parallel_executor.py:2625`)** — bare-name resolution now anchors RELATIVE PATH entries to the effective child cwd (not the orchestrator's), so `PATH=bin` with a workspace `bin/probe` binds the actual launched binary instead of collapsing to `unresolved` and colliding across workspaces. Regression: two workspaces with different `bin/probe` targets under a relative PATH produce different authority.

**2. Watchdog fingerprint omitted the total-turn timeout (`parallel_executor.py:2699`)** — the watchdog probe now includes a wrapped transport's `_timeout` (ClaudeWorkerTransport's total-turn kill via `asyncio.wait_for`), so runtimes with timeouts 1 vs 999 no longer share capsule authority. Regression covers the wrapped-transport `_timeout`.

**3. Execution policy omitted `max_concurrent` (`parallel_executor.py:2969`)** — capsule execution policy now records effective `max_concurrent`; concurrency 1 vs 8 no longer produce identical authority scopes. Regression asserts the difference independently of checkpoint restoration.

**Non-blocking (`ac_execution_capsule.py:210`)** — `ACSuccessContract` now mirrors `AcceptanceCriterionSpec` and rejects an `output_assertion` without a `verify_command`. Regression added.

**Validation:** `tests/unit/orchestrator` — 3660 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `cf2b1640d`.

Q00:
Addressed the two blockers + the follow-up warning from review `4731570352` on new HEAD `f57a8309c`:

**1. Worker double-executed the verify command (`atomic_prompt_builder.py:54`)** — the success-contract prompt block no longer tells the worker to run `verify_command`; it explicitly says the harness gate runs it and the worker must NOT. The orchestrator's single-shot gate is now the only authoritative execution, so a compliant worker no longer runs a possibly non-idempotent command outside the oracle. Regression asserts the "Do NOT run" wording and the absence of the old instruction.

**2. Empty PATH components unanchored (`parallel_executor.py:2635`)** — empty PATH components are now anchored to the effective child cwd (POSIX: empty = current directory), so `PATH=""` no longer collapses distinct workspace-local binaries to the same `unresolved` identity. Regression: two workspaces with local `probe` binaries under `PATH=""` produce different authority.

**3. (warning) Verify outcome truncated by code points then raised (`execution_event_emitter.py:96`)** — `bound_verify_gate_outcome` now bounds by ENCODED byte size, dropping trailing artifact entries (recording the omission) to fit the 64 KiB cap instead of raising; the completed-lifecycle outcome is routed through the same byte-fitting bound, and the completed parser accepts the optional omission counter. A valid capsule with many multi-byte artifact paths can always persist its outcome, so a durable verify intent is never left with a permanently unrecoverable outcome. Regressions assert the fitted payload stays within the byte cap and re-parses.

**Validation:** `tests/unit/orchestrator` — 3661 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New/updated regressions all pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `f57a8309c`.

Q00:
Addressed the blocker + follow-up warning from review `4731593102` on new HEAD `022474108`:

**1. Effectful active dispatch was replayable (`ac_runtime_handle_manager.py:1296`)** — the recovery loader now fails closed when the active dispatch head recorded a durable tool effect (`execution.tool.started`/`.completed`) without terminating, instead of returning its resumable handle. `_execute_atomic_ac` resends the original AC prompt on resume, so returning the handle would start another provider turn and could duplicate a non-idempotent effect; with no mechanism to continue the exact in-flight turn without resending, this is refused. Regression: `compiled → dispatched → started → tool.started` now fails closed.

**2. (warning) Execution policy omitted execution-semantic fields (`parallel_executor.py:2996`)** — capsule execution-policy authority now binds `shadow_replay_enabled`, `cross_harness_redispatch_enabled`, `ac_retry_attempts`, `lateral_escalation_enabled`, and `parked_retry_backoff_seconds`, so a restart cannot resume the same capsule under a materially different policy (e.g. shadow replay, which performs an extra provider execution). Regression asserts each field independently changes authority.

**Validation:** `tests/unit/orchestrator` — 3663 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions pass. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `022474108`.

Q00:
Addressed the three blockers + the low follow-up from review `4731613078` on new HEAD `45ec4b7c8`:

**1. Opaque active dispatch replayed the prompt (`ac_runtime_handle_manager.py:1258`)** — recovery now fails closed when an active dispatch head is resumable on a runtime that does not attest real-time tool-effect visibility (all production runtimes default opaque). An opaque effect may precede any event, and `_execute_atomic_ac` resends the AC prompt on resume, so this is refused; a runtime that attests real-time visibility (making the recorded-effect check authoritative) still resumes. Regression covers both.

**2. Authority omitted OpenCode dispatch policy (`parallel_executor.py:2893`)** — `OpenCodeRuntime` now exposes `execution_identity_contract()` binding `opencode_mode`, `llm_backend`, `skills_dir`, constructor model, and a bounded skill-dispatcher identity, so different skill-resolution/plugin routes no longer share a capsule fingerprint. Regression covers the fields and dispatcher identity.

**3. Signal-hub authority was only a boolean (`parallel_executor.py:3011`)** — capsule authority now binds the hub's durable-relay mode: a `SessionSignalHub` with a durable event store (which replays cross-process queued signals and may run follow-up turns) no longer shares authority with a non-durable one. Regression added.

**Follow-up (`execution_event_emitter.py:125`)** — `bound_verify_gate_outcome` now measures the ACTUAL serialized payload when admitting each artifact entry, so the outcome can never exceed the 64 KiB cap by a byte.

**Validation:** `tests/unit/orchestrator` — 3666 passed, 1 skipped; the only failures are the same 7 pre-existing environmental cases that shell out to a `python` binary absent on this host (only `python3`; verified identical on untouched HEAD). New regressions pass; resume-machinery tests attest real-time visibility to exercise the resume path. Ruff + format clean, MyPy clean on changed modules, `git diff --check` clean.

Please re-review the new HEAD `45ec4b7c8`.

Q00:
## Round 13 — blocker + warning + low addressed (HEAD `17fc01565`)

Thanks for the deep pass on review `4731663813`. All three findings on `45ec4b7c8` are resolved.

### R13-1 (blocker) — verifier PASS trusted after later workspace mutation
`verify_command` is now **deferred out of the concurrent atomic leaf** and applied at batch completion in `_apply_verify_gate`, over the final merged workspace — not a mid-flight leaf snapshot that a sibling could still mutate.
- Leaf: `defer_command_to_finalization = isinstance(ac_spec, AcceptanceCriterionSpec) and bool(ac_spec.verify_command)`; the leaf success-contract gate skips the command when deferral is active.
- Finalization runs the single-shot oracle over the settled workspace. The verify key (`acgate:{execution_scope}:{ac_index}:{retry_attempt}`) is unchanged, so the gate stays single-shot across atomic + final (regression: `test_contract_verify_gate_is_single_shot_across_atomic_and_final_gate`).

### R13-2 (blocker) — alt-harness generic exception assumed pre-entry
`_run_single_ac_on_backend` now wraps **only pre-entry construction** and raises `_AltHarnessPreEntryError`; `_maybe_redispatch_alt_harness` redispatches solely on that typed pre-entry signal. Any post-entry generic exception is ambiguous → **fail closed** (no silent re-run). Coverage in `test_alt_harness_hook.py` (11 cases).

### R13-3 (blocker) — executable authority sampled, not enforced at launch
Executable identity is captured at compile (`capsule_executable_identity`) and **revalidated immediately before `emit_ac_attempt_dispatched`**. Any drift between compile and launch raises `AmbiguousACExecutionError` (fail closed). Regression: `test_executable_drift_before_dispatch_fails_closed`.

### R13-warn — OpenCode dispatcher fallback recorded only class
Unidentifiable dispatchers now record `{"type": ..., "process_local_nonce": id(dispatcher), "durable_identity": False}` so two distinct dispatchers of the same class no longer collapse to one fingerprint. Regression: `test_unidentifiable_dispatcher_records_nonce`.

### R13-low — omitted-artifact accounting
`_parse_verify_gate_outcome` accepts an optional `missing_artifacts_omitted` (validated `int >= 0`).

### Validation
- `pytest tests/unit/orchestrator tests/unit/core/test_seed.py` → **3734 passed, 1 skipped**, 6 failed.
- The 6 failures are all `test_runtime_evidence.py` headless-run cases that shell out to a `python` binary absent from this sandbox (`FileNotFoundError`) — pre-existing and unrelated (confirmed against untouched HEAD).
- `ruff check` / `ruff format` / `mypy` clean on all touched sources; `git diff --check` clean.

Q00:
## Round 14 — both blockers addressed (HEAD `26d560d56`)

Thanks for the precise probes on review `4731706952`. Both follow-on defects in the deferred-finalization design are fixed.

### Blocker 1 — deferred `verify_command` completions were unrecoverable
`parallel_executor.py` recovery (`CompletedACExecutionError` handler). A command-bearing AC defers its verify_command out of the atomic leaf (R13-1), so the leaf persists `session.completed` with `verify_gate_outcome=None`. Recovery previously rejected that head **before** consulting `_load_persisted_verify_decision`, so a legitimately-deferred completion could never cross replay.

Recovery now derives the verdict **purely from the durable single-shot oracle** under the shared per-attempt key `acgate:{scope}:{ac_index}:{retry}`:
- **recorded outcome present** → consume it (never replay the command);
- **intent without outcome** → the non-idempotent command may have partially run → fail closed (`AmbiguousACExecutionError`);
- **no durable record at all** → the deferred finalization gate never started, so the command has provably not run → leave the outcome unresolved and let the batch gate run the oracle **once** over the settled workspace.

An uncorroborated cached PASS carried on the completion is **discarded, not trusted** — a forged completion still has to pass the real command at finalization.

Regressions: `test_completed_recovery_consumes_deferred_finalization_outcome` (your exact probe: deferred completion + durable intent+outcome → recovered, not rejected), `test_completed_recovery_without_verify_record_defers_to_finalization`, `test_completed_recovery_intent_without_outcome_fails_closed`, `test_completed_recovery_ignores_uncorroborated_cached_pass`.

### Blocker 2 — order-dependent gates in the live workspace
`_apply_verify_gate_isolated` now wraps every command-bearing batch/escalation/alt-harness gate (all 7 finalization sites) and the decomposition-parent gate. Each **fresh command run** is graded against a **private frozen snapshot** of the settled workspace (`isolated_workspace`), so a verifier can neither observe nor cause a cross-gate mutation: every gate sees the identical settled state regardless of order, and the delivered tree is never touched by a verifier. Cached-outcome consumption and pure `expected_artifacts` legs stay on the live delivered tree (read-only, idempotent).

Your probe now holds: AC1 verifies `marker == good`, AC2's verifier mutates `marker`; both grade their own snapshot, the live `marker` stays `good`, and AC1's acceptance reflects the immutable final workspace — regardless of gate order. When the tree cannot be frozen, the batch gate fails a would-be PASS closed; the (unflagged, proactively-hardened) parent gate degrades to in-place, consistent with `isolated_workspace`'s best-effort contract used by shadow-replay.

Regressions: `test_isolated_verify_gate_grades_immutable_snapshot_not_shared_mutation` (runs the mutating gate first, then the assertion gate — order-independent), `test_isolated_verify_gate_fails_closed_when_snapshot_unavailable`.

### Validation
- `pytest tests/unit/orchestrator tests/unit/core/test_seed.py` → **3738 passed, 1 skipped**, 6 failed.
- The 6 failures are all `test_runtime_evidence.py` headless-run cases that shell out to an absent `python` binary (`FileNotFoundError`) — pre-existing and unrelated (confirmed against untouched HEAD in earlier rounds).
- `ruff check` / `ruff format` / `mypy` clean on touched sources; `git diff --check` clean.

Q00:
Re-requesting review on HEAD `26d560d5683318dae249302b90413e19258580a7` — Round 14 addresses both blockers from review `4731706952` (deferred-verify restart recovery via the durable single-shot oracle; every command-bearing gate now grades a private frozen snapshot of the settled workspace). Details in the comment above.

Q00:
Re-requesting review on HEAD `1da3b4f2473fbfada22a420ee85a370380a8fb0f`. The R14 blocker fixes (deferred-verify restart recovery + immutable-snapshot verify gates) are in `26d560d5`; this HEAD adds only an empty re-trigger commit. Full details in the Round 14 comment above.

Q00:
Progress on review `4732205341` (HEAD `ac31c309961ebb1615d8dde890d978f744d48282`):

**Blockers 2 & 3 addressed.** Replaced the shadow-replay verification snapshot with a **no-mutation contract**:
- **B2:** `verify_command` now grades the **complete live tree** (no `isolated_workspace` exclusion), so contracts depending on `.git`/`.venv`/`node_modules`/`dist`/`build`/`target` see them. Your probe `test -f node_modules/probe` now passes at finalization. (Snapshots also can't faithfully include dep dirs — escaping symlinks make `isolated_workspace` reject them.)
- **B3:** removing the snapshot eliminates the isolation-failure path entirely. The no-mutation guard lives at the single command-run choke point (`_run_verify_gate_single_shot`), so the sibling-flip gate is covered uniformly. A verifier that mutates the acceptance-relevant tree is failed closed (verdict folded into the durable outcome before persistence, so recovery consumes the FAIL).

**Blockers 1 & 4 (final-workspace acceptance timing + premature success events) are in progress** — implementing a global final-acceptance pass after all levels + coordinator reconciliation, with terminal acceptance emitted only after that gate. Will push shortly.

Q00:
## Round 16 — all three blockers addressed (HEAD `99fd67b3a942f05e058961923013f88e8891ff51`)

Re: review `4732313961`. All three are resolved.

### B1 (final-workspace timing) + B3 (terminal acceptance published early)
Added a **global final-workspace acceptance pass** (`_apply_final_workspace_acceptance`) invoked once after **all** stages and coordinator reconciliation settle (`_execute_parallel_impl`, right after the level loop, before aggregation). It re-runs each currently-accepted command-bearing AC's `verify_command` over the **true final tree** as the authoritative acceptance:
- A per-attempt pass that a later stage/reconciliation invalidated is **flipped to FAILED**. Your two-stage probe (stage 2 invalidates stage 1's verified file) now fails stage 1.
- The terminal `execution.ac.outcome_finalized` is **(re)emitted only after this gate**, so no accepted verdict predates the final workspace.
- It also re-checks results **restored by crash recovery** (the pass runs on every execution, including cold restarts), so a per-attempt `outcome_finalized` can no longer preserve stale success permanently — the final gate has the last word.
- Single-shot on a dedicated `acgate_final:{scope}:{ac}` key (a cold restart consumes the one final outcome, no re-run); an ambiguous final chain fails that AC closed without aborting the run. Re-evaluation is side-effect-free because the no-mutation contract enforces verifier purity.

Regressions: `test_final_workspace_acceptance_flips_stale_pass_to_failed`, `test_final_workspace_acceptance_confirms_valid_pass`.

### B2 (fingerprint excluded contract paths + metadata-only)
The no-mutation fingerprint is now a **comprehensive content hash**: it SHA-256s every acceptance-relevant file — **including** `node_modules`/`.git`/`.venv`/`dist`/`build` — so any create/delete/modify/chmod is detected, including content changes that preserve size and mtime. Your `rm node_modules/probe` probe is now failed closed. Only the verify tooling's own regenerable caches (`__pycache__`/`.pytest_cache`/`.mypy_cache`/`.ruff_cache`/`.tox`) and transient git bookkeeping (index/reflogs/`*.lock`, which read-only git commands rewrite) are excluded to avoid false-positive flags on a legitimate verifier; the command still runs over the complete live tree.

Regression: `test_verify_command_deleting_dependency_is_failed_closed`.

### Validation
- `pytest tests/unit/orchestrator tests/unit/core/test_seed.py` → **3741 passed, 1 skipped**, 6 failed (all pre-existing `test_runtime_evidence.py` `python`-binary env failures, unrelated).
- ruff / format / mypy clean; `git diff --check` clean.

Q00:
Superseded by #1681.

The capsule, fresh-session dispatch, runtime-authority, verifier-authority, and recovery work in this PR remains valuable reference material, but this layer was built on terminal-event and execution-authority semantics that need to land first. We are replacing this PR with an approval-gated stack: only the current layer will be open for review, and the next layer will be created after approval.

The branch is intentionally preserved for implementation and test reuse.

PR #1674, opened 2026-07-19
docs: add hero demo video to README top (EN/KO/ZH)

Q00:
## What

Adds a **16.5s hero demo clip** to the top of all three READMEs (EN / KO / ZH), right below the nav and above the tagline.

The clip splices two impact segments from the raw 39s recording:
1. Typing `ooo interview "make me an agentic evolving video harness…"`
2. The **specialist-agent fan-out** (Advisory simplifier / contrarian / code-context / web-context → Done · 5)

## Why this way

- **48MB → 641KB** (h264, 1280w, 30fps, muted) — heavy compression as requested.
- Hosted as a **GitHub release asset** (`media-assets` tag), *not* committed to the repo → zero impact on git size and the **PyPI sdist** (this repo has hit storage bloat from in-repo binaries before).
- `<video>` tag with `controls` so a play button always shows even if GitHub strips `autoplay`.

## Please verify in preview

The one thing that only renders on GitHub: confirm the inline `<video>` player shows from the **release-download URL**. If it doesn't, the fallback is a `github.com/user-attachments/...` URL (drag-drop upload).

> Note: MP4 `<video>` does **not** render on the PyPI project page (README.md is the long_description) — a known trade-off vs. using a GIF.


sergiobuilds:
Closing without merge. All three README video tags point to the same GitHub user-attachment URL, and an independent request to that exact asset currently returns HTTP 404. The PR description also says 16.5 seconds while the final commit records 31.9 seconds. Reopen with a reachable immutable asset and aligned metadata, then verify the player in GitHub preview.

PR #1708, opened 2026-07-22
fix(hooks): honor UserPromptSubmit JSON contract

### LINKED ISSUE #1694: Codex plugin: keyword-detector.py echoes the hook JSON payload, failing UserPromptSubmit schema validation on every prompt
## Summary

When ouroboros is installed as a **Codex plugin** (via the plugin marketplace), the `UserPromptSubmit` hook (`scripts/keyword-detector.py`) fails on **every prompt** with:

```
UserPromptSubmit hook (failed)
error: hook returned invalid user prompt submit JSON output
```

## Environment

- ouroboros 0.50.5 (installed from the `https://github.com/Q00/ouroboros.git` marketplace, `hooks/hooks.json`)
- Codex CLI 0.144.6 and 0.145.0-alpha.18 (macOS desktop app / `codex exec`)
- macOS (Darwin 25.5.0)

## Root cause

`keyword-detector.py` reads stdin as raw text and **echoes it back to stdout** (the "pass through unchanged" behavior, plus the `<skill-suggestion>` block when a keyword matches):

```python
user_input = sys.stdin.read().strip()
...
# Pass through unchanged when no keyword detected
print(user_input)
```

Under both Claude Code and Codex, stdin for a `UserPromptSubmit` hook is **the JSON hook payload** (`{"session_id": ..., "prompt": ..., "hook_event_name": ...}`), not the bare prompt text. This causes two problems on Codex:

1. Keyword detection runs against the serialized JSON payload instead of the user's prompt.
2. The echoed stdout **is itself a JSON object**, so Codex parses it as the hook's structured output and rejects it against its `UserPromptSubmit` output schema (`hookSpecificOutput.hookEventName` / `additionalContext` — the same wire format Claude Code documents). Result: the hook fails on every single prompt.

Minimal repro (no Codex needed):

```console
$ echo '{"session_id":"t","prompt":"hello","hook_event_name":"UserPromptSubmit"}' \
    | python3 scripts/keyword-detector.py
{"session_id":"t","prompt":"hello","hook_event_name":"UserPromptSubmit"}
```

The output is the input payload echoed verbatim — valid JSON, wrong schema. End-to-end repro: run `codex exec --skip-git-repo-check "hi"` with the plugin enabled and watch `hook: UserPromptSubmit Failed`.

## Suggested fix

Parse the payload, run detection on `payload["prompt"]`, and emit either **nothing** (no keyword matched) or the **official structured output**, which is valid for both Claude Code and Codex:

```python
try:
    payload = json.loads(user_input)
    if isinstance(payload, dict):
        user_input = str(payload.get("prompt") or "")
except ValueError:
    pass  # raw text fallback

result = detect_keywords(user_input)

def emit(context: str) -> None:
    print(json.dumps({
        "hookSpecificOutput": {
            "hookEventName": "UserPromptSubmit",
            "additionalContext": context,
        }
    }, ensure_ascii=False))
```

…and replace the three `print(f"""{user_input}\n\n<skill-suggestion>...""")` call sites with `emit("<skill-suggestion>...")`, dropping the prompt echo entirely (both runtimes keep the original prompt; the suggestion arrives as `additionalContext`).

I patched my local Codex plugin cache exactly this way and verified with `codex exec`: all hooks now report `Completed`, and keyword matching (e.g. `ooo status`) still produces the skill suggestion. Happy to turn this into a PR if useful.

## Related note

`scripts/drift-monitor.py` (PostToolUse) prints plain-text `"Success"` to stdout. It didn't visibly error in my testing, but the same structured-output treatment (or printing to stderr like `session-start.py` does) would make it robust against the same strict validation.

sergiobuilds:
## Summary
- parse Codex and Claude `UserPromptSubmit` envelopes from stdin
- emit schema-valid `hookSpecificOutput.additionalContext` instead of echoing the hook payload
- stay silent when no suggestion applies
- preserve legacy raw-text prompts, including JSON-object-shaped prompt text

Closes #1694.

## Independent verification
1. Focused hook contract suite: `44 passed`
2. Full script unit suite plus Ruff: `220 passed`; lint and format checks passed
3. Fresh isolated Codex 0.145.0 marketplace install: `ooo status` loaded the bundled status skill and completed with the fixed hook installed

Additional direct probes confirmed zero stdout for an ordinary no-match prompt and schema-valid JSON for matched prompts.

## Credit
Retains the original bot-authored implementation commit from #1695 and adds the independent compatibility regression requested in that review.

### DIFF
--- a/scripts/keyword-detector.py
+++ b/scripts/keyword-detector.py
@@ -285,9 +285,10 @@ def _extract_prompt(hook_input: str) -> str:
     if (
         isinstance(payload, dict)
         and payload.get("hook_event_name") == "UserPromptSubmit"
-        and "prompt" in payload
+        and isinstance(payload.get("session_id"), str)
+        and isinstance(payload.get("prompt"), str)
     ):
-        return str(payload.get("prompt") or "")
+        return payload["prompt"]
     return hook_input
 
 
--- a/scripts/keyword-detector.py
+++ b/scripts/keyword-detector.py
@@ -8,8 +8,8 @@ IMPORTANT: If MCP is not configured (ooo setup not run),
 ALL ooo commands (except setup/help) redirect to setup first.
 
 Hook: UserPromptSubmit
-Input: User prompt text via stdin (piped by Claude Code)
-Output: Modified prompt with skill suggestion appended
+Input: JSON hook payload via stdin, with raw prompt text supported as a fallback
+Output: Structured hook context when a skill is suggested; otherwise no output
 """
 
 import json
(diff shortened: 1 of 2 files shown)


Q00:
## Review-gate status

Closing this follow-up without merge.

The only `ouroboros-agent` review is anchored to `12fb532a56cb18fb5f098d41d2b6e072d97123ef`, while the current head is `37c0b52e218935e2d45d41db31c81f13cd4fff57`. The current commit changes the envelope discriminator by requiring `session_id`, which is exactly the raw-JSON collision identified by that review. The recorded `APPROVE` therefore is not an assessment of the current tree and cannot serve as an independent merge gate.

The review is also internally inconsistent: it records the legacy JSON-object compatibility requirement as only partially met and requests a marker-collision regression, yet ends in `APPROVE`. The original bot-authored PR #1695 was already closed because self-review is not independent. Keep #1694 open; a replacement should receive an independent exact-head review after the full check set has completed.

sergiobuilds:
Reopened after the requested exact-head independent verification completed. Current head `37c0b52e218935e2d45d41db31c81f13cd4fff57` now requires a string `session_id`, `hook_event_name=UserPromptSubmit`, and a string `prompt`, and includes the marker-collision regression. Independent review of that exact head returned APPROVE; focused hook tests passed, the full script suite passed, and the complete GitHub check set is green after rerunning one unrelated generation-watchdog timing flake. Please rerun the formal reviewer against the current head.

sergiobuilds:
Replaced by https://github.com/Q00/ouroboros/pull/1712 so the formal reviewer can assess the already independently approved exact head on a fresh PR.

PR #1712, opened 2026-07-22
fix(hooks): honor UserPromptSubmit JSON contract

### LINKED ISSUE #1694: Codex plugin: keyword-detector.py echoes the hook JSON payload, failing UserPromptSubmit schema validation on every prompt

sergiobuilds:
## Summary
- parse Codex and Claude UserPromptSubmit JSON envelopes before keyword matching
- require the genuine envelope shape (`hook_event_name`, string `session_id`, and string `prompt`) so hook-shaped raw JSON remains compatible
- emit schema-valid hookSpecificOutput only for matching prompts and preserve no-match silence

Replaces #1708 after exact-head independent review was unavailable on the reopened PR.

Closes #1694.

## Verification
- exact-head independent review of `37c0b52e218935e2d45d41db31c81f13cd4fff57`: APPROVE
- focused hook tests: 47 passed
- full script suite passed
- Ruff and format checks passed
- full GitHub CI passed on the identical commit tree
- actual Codex plugin paths completed successfully

### DIFF
--- a/scripts/keyword-detector.py
+++ b/scripts/keyword-detector.py
@@ -16,6 +16,7 @@ import json
 from pathlib import Path
 import re
 import sys
+import tomllib
 
 
 def _configure_utf8_stdio() -> None:
@@ -156,12 +157,18 @@ KEYWORD_MAP = [
 
 
 def is_mcp_configured() -> bool:
-    """Check if MCP server is registered in ~/.claude/mcp.json."""
+    """Check if the Ouroboros MCP server is registered for Claude or Codex."""
     try:
-        mcp_path = Path.home() / ".claude" / "mcp.json"
-        if not mcp_path.exists():
+        claude_path = Path.home() / ".claude" / "mcp.json"
+        if claude_path.exists() and "ouroboros" in claude_path.read_text():
+            return True
+
+        codex_path = Path.home() / ".codex" / "config.toml"
+        if not codex_path.exists():
             return False
-        return "ouroboros" in mcp_path.read_text()
+        codex_config = tomllib.loads(codex_path.read_text())
+        mcp_servers = codex_config.get("mcp_servers")
+        return isinstance(mcp_servers, dict) and isinstance(mcp_servers.get("ouroboros"), dict)
     except Exception:
         return False
 
--- a/scripts/keyword-detector.py
+++ b/scripts/keyword-detector.py
@@ -285,9 +285,10 @@ def _extract_prompt(hook_input: str) -> str:
     if (
         isinstance(payload, dict)
         and payload.get("hook_event_name") == "UserPromptSubmit"
-        and "prompt" in payload
+        and isinstance(payload.get("session_id"), str)
+        and isinstance(payload.get("prompt"), str)
     ):
-        return str(payload.get("prompt") or "")
+        return payload["prompt"]
     return hook_input
 
 
--- a/scripts/keyword-detector.py
+++ b/scripts/keyword-detector.py
@@ -13,6 +13,7 @@ Output: Structured hook context when a skill is suggested; otherwise no output
 """
 
 import json
+import os
 from pathlib import Path
 import re
 import sys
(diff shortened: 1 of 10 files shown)


sergiobuilds:
Addressed the formal review warning before merge. `_extract_prompt()` now treats `RecursionError` from deeply nested caller-controlled JSON like other parse failures and falls back to raw input. Verification: hook tests 46 passed; script suite 222 passed; Ruff, format, and diff checks passed; a real Python 3.12 10,000-level nested JSON probe returned the raw prompt without crashing.

sergiobuilds:
Addressed the exact-head Codex MCP blocker on head `9f35682e`. The hook now parses `~/.codex/config.toml` and recognizes `[mcp_servers.ouroboros]` in addition to Claude MCP registration. Added a Codex-only routing regression that sends a real `UserPromptSubmit` envelope for `ooo status` and verifies it routes to status rather than setup. Verification: focused hook suite 48 passed; full script suite 224 passed; Ruff and format checks passed.

sergiobuilds:
Addressed the Python 3.10 hook-host blocker on exact head `2b013f5a`. `tomllib` is now optional; Python 3.10 uses a conservative fallback that requires the exact `[mcp_servers.ouroboros]` table plus a `command` or `url` transport. Added fallback acceptance and malformed-registration rejection tests. Verification: 50 focused hook tests, 226 script tests, Ruff/format, and an actual isolated Python 3.10 Codex-only `UserPromptSubmit` probe all passed.

sergiobuilds:
Addressed all three exact-head MCP-routing blockers on head `a98d6ddd`. Genuine Codex envelopes are now distinguished by the required Codex `turn_id`, so Claude-only and Codex-only registrations cannot leak across hosts. Codex lookup honors `CODEX_HOME`. Both TOML paths require an enabled non-empty `command` or `url`; the Python 3.10 fallback rejects disabled, transportless, duplicate, and multiline-string decoy registrations. Verification: 55 focused hook tests, 231 script tests, Ruff/format, plus isolated Python 3.10 probes for valid custom CODEX_HOME, disabled registration, and multiline decoy all passed.

sergiobuilds:
Fixed the Python 3.10 fallback comment-parsing blocker on exact head `a447667b`. TOML comments are now stripped only outside quoted strings and before multiline-delimiter detection, so inline comments are accepted, `#` inside transport strings is preserved, and triple quotes inside comments cannot hide a real registration. Verification: 57 focused tests, 233 script tests, Ruff/format/diff check, and isolated Python 3.10 probes for inline comments and quoted `#` transports passed.

sergiobuilds:
Replaced partial Codex TOML parsing with Codex’s authoritative effective-config query on exact head `efb9413a`. The hook now runs `codex mcp get ouroboros --json`, so CLI overrides, trusted project config, profiles, user config, system config, disabled state, and CODEX_HOME are resolved by Codex itself. Setup now writes MCP registration through the shared active-Codex-home resolver, closing the custom-home setup loop. Claude JSON is read explicitly as UTF-8. Verification: 228 script tests; 166 setup tests combined with scripts, 394 total selected tests; 13 MCP registration tests; Ruff/format/diff checks; MyPy across 467 source files; and a real Python 3.10 hook invocation against effective Codex MCP state all passed.

sergiobuilds:
Closed the literal-tilde CODEX_HOME loop on exact head `26d1c752`. The hook now normalizes a nonblank CODEX_HOME with the same strip/expanduser semantics as setup and removes blank overrides before invoking Codex. An end-to-end probe with literal `CODEX_HOME=~/custom-codex` verified setup wrote the expanded path and a real Python 3.10 hook resolved effective MCP state to `/ouroboros:status`. Verification: 394 selected tests, Ruff/format/diff checks, and MyPy across 467 source files passed.

sergiobuilds:
Addressed the current-head blockers on `b7681354`. Codex `UserPromptSubmit` payloads do not carry the parent invocation’s `--profile` or `-c` layers, and Codex executes hooks with inherited environment plus static handler env only. A child `codex mcp get` therefore cannot truthfully reconstruct parent-effective state. The Codex envelope path now avoids that false setup gate and preserves the requested skill; the legacy raw path remains Claude-scoped and still applies Claude MCP gating. Setup now resolves one active Codex home for MCP config, default profiles, worker profile, rules, skills, refresh, and guidance. Verification: 392 affected tests, Ruff/format/diff checks, MyPy across 467 source files, plus a real Python 3.10 end-to-end probe confirmed literal-tilde CODEX_HOME installs config/profiles/rules/skills only under the custom home, Codex envelopes route status, and raw legacy input routes setup when Claude is unconfigured.

sergiobuilds:
The timed-out independent review transcript exposed one final exact-semantics mismatch: setup expanded a literal `CODEX_HOME=~/custom-codex`, while Codex CLI treats that value as a literal relative path. Exact head `c1e27337` now preserves Codex path semantics in the shared home resolver and makes setup use the absolute location corresponding to that same current-working-directory interpretation for all artifacts. A real Codex CLI 0.145.0 probe now passes end to end: setup and `codex mcp get ouroboros --json` agree, config/profiles/rules/skills all land under the literal path, and no expanded `$HOME/custom-codex` tree is created. Verification: 445 affected tests, Ruff/format/diff checks, and MyPy across 467 source files.

sergiobuilds:
Closed the custom-home lifecycle finding on exact head `fd978604`. Uninstall preview/removal and configured-model discovery now resolve the same literal CODEX_HOME semantics as setup/doctor/refresh. Uninstall removes the managed MCP tables, legacy profiles, profile-v2 files, rules, and skills from the active Codex home rather than `~/.codex`. Real end-to-end setup → `ouroboros uninstall --yes --keep-data` with literal `CODEX_HOME=~/custom-codex` passed: MCP, profile, rule, and skill existed before removal and all were absent afterward. Verification: 534 affected tests, Ruff/format/diff checks, and MyPy across 467 source files.

sergiobuilds:
Closed both exact-head uninstall blockers on `e52b2986`. Managed legacy profile matching now removes nested tables such as `[profiles.ouroboros-fast.model_provider]` while preserving unrelated tables. Final-status accounting now keys off the actual custom-home preview categories, so MCP/config rewrite and profile/rule/skill deletion failures produce `Ouroboros partially removed` instead of false success. Regressions cover both failure paths and retained artifacts. A real absolute-home nested-profile probe removed the managed parent/subtable and preserved `[other]`. Verification: 536 affected tests, Ruff/format/diff checks, MyPy across 467 source files.

sergiobuilds:
Closed the nested-only CLI discovery blocker on `65cb8e52`. Preview discovery and removal now share `_is_managed_codex_header()`, covering MCP tables, exact managed profiles, and nested managed profile tables. Added an end-to-end CLI regression where the config contains only `[profiles.ouroboros-fast.model_provider]`; uninstall now discovers and removes it while preserving `[other]`. Verification: 537 affected tests, Ruff/format/diff checks, MyPy across 467 source files.

shaun0927:
Closing in favor of #1753.

This PR correctly identified the UserPromptSubmit JSON failure, but expanded from the two-file hook contract into CODEX_HOME, setup, model catalog, doctor, and uninstall lifecycle changes. That scope is not required by #1694 and introduces separate configuration/deletion risks, including managed-profile ownership ambiguity.

#1753 reapplies the verified hook fix on current `main` with only `scripts/keyword-detector.py` and its protocol tests. It preserves raw fallback behavior, emits schema-valid structured context, and uses the documented Codex `turn_id` field without reconstructing parent session configuration.

PR #1713, opened 2026-07-23
fix(seed): preserve literal pipe characters in constraints

### LINKED ISSUE #1696: SeedGenerator corrupts literal pipe characters in constraint values
## Summary

The Seed requirement extractor uses a bare pipe character as its list delimiter and then splits constraint text with split("|"). A legitimate literal pipe inside one constraint is therefore converted into multiple constraints. The extraction format does not define an escaping rule.

## Environment

- Ouroboros 0.50.5
- Reproduced with the Python SeedGenerator path; this is OS-independent

## Minimal reproduction

Pass this extracted constraint field to SeedGenerator._build_seed:

    --lang ko|en | keep exact flag

Actual Seed constraints:

    ("--lang ko", "en", "keep exact flag")

Expected Seed constraints:

    ("--lang ko|en", "keep exact flag")

The same collision can affect shell syntax, union notation, Obsidian alias syntax, or any other legitimate value containing a pipe.

## Root cause

src/ouroboros/bigbang/seed_generator.py parses constraints using requirements["constraints"].split("|"). The prompt reserves pipe as the separator but provides no escape or structured representation for literal pipe data.

## Impact

The generated Seed can silently change a user requirement while still remaining schema-valid. Downstream QA may or may not notice the semantic corruption.

## Suggested direction

Use a structured response representation or an escape-aware parser, and add a regression test proving literal pipe preservation.

berkshirehathaways:
## Summary
- preserve escaped literal pipes in Seed constraint values
- keep legacy `|`-delimited constraints unchanged for ordinary values
- document `\\|` in both extraction prompt paths
- add a regression test for `--lang ko\\|en`

## Verification
- `uv run pytest tests/unit/bigbang/test_seed_generator.py -q` (44 passed)
- `uv run ruff check src/ouroboros/bigbang/seed_generator.py tests/unit/bigbang/test_seed_generator.py`

Closes #1696

### DIFF
--- a/src/ouroboros/bigbang/requirement_distillation.py
+++ b/src/ouroboros/bigbang/requirement_distillation.py
@@ -50,6 +50,16 @@ _CONSTRAINT_RE = re.compile(
 )
 
 
+def split_escaped_pipe_values(raw_value: object) -> tuple[str, ...]:
+    if isinstance(raw_value, list | tuple):
+        values = (str(item) for item in raw_value)
+    elif isinstance(raw_value, str):
+        values = re.split(r"(?<!\\)\|", raw_value)
+    else:
+        return ()
+    return tuple(value.replace(r"\|", "|").strip() for value in values if value.strip())
+
+
 @dataclass(frozen=True, slots=True)
 class AppliedRequirementDistillation:
     """Requirements after the deterministic reference-aware promotion gate."""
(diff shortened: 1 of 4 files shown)


berkshirehathaways:
Addressed the reference-aware Seed lossiness finding in commit `26b7dc3d`.

- Added a shared escaped-pipe splitter and reused it in both LLM extraction and deterministic `build_promoted_reference_seed` paths.
- Promoted constraints and acceptance criteria now preserve literal `\|` values instead of splitting them as delimiters.
- Added a reference-aware regression test for `Only --lang ko\|en is required.`.

Validation: Ruff check, Ruff format check, and Python compileall passed. Local pytest dependencies are unavailable; please re-run the supported CI matrix for the pushed commit.

berkshirehathaways:
This PR is superseded by upstream [#1714](https://github.com/Q00/ouroboros/pull/1714), which was merged on 2026-07-23 and fixes the same #1696 issue across both paths using structured constraint values. That upstream change also resolves the reference-aware lossiness called out in the review here, so I am closing this duplicate rather than rebasing overlapping code.

berkshirehathaways:
Superseded by the merged upstream fix in #1714.

PR #1714, opened 2026-07-23
fix(seed): preserve literal pipes in constraints across both seed paths

### LINKED ISSUE #1696: SeedGenerator corrupts literal pipe characters in constraint values

sumin220:
## Summary

Literal `|` characters inside a single constraint value are corrupted on both seed-generation paths (#1696):

- `_build_seed` splits extracted CONSTRAINTS on bare `|` (`src/ouroboros/bigbang/seed_generator.py`), so `--lang ko|en` becomes two constraints.
- The promoted-reference path joins confirmed constraints/acceptance criteria with `" | "` and re-splits them in `build_promoted_reference_seed` (`src/ouroboros/bigbang/requirement_distillation.py`) — the blocker called out when #1698 was closed.

Following that close comment's guidance, each path now uses a structured representation with no delimiter round trip:

- **Extraction**: CONSTRAINTS is requested as a single-line JSON array in both prompt templates. `_parse_constraint_values` prefers JSON and falls back to legacy pipe splitting on `JSONDecodeError`, so bracket-prefixed prose like `[P0] Must work offline` keeps its legacy behavior.
- **Promotion**: `apply_requirement_distillation` emits constraints/acceptance criteria as tuples, and `build_promoted_reference_seed` consumes them directly via `_normalized_requirement_values`.

One scope note: this is structured-per-path (JSON array at the LLM text boundary, tuples for in-process promotion) rather than one shared form across both — each boundary uses its natural structure. Happy to adjust if you'd prefer a single shared representation.

Four existing assertions in `tests/unit/bigbang/test_requirement_distillation.py` pinned the pipe-joined string form of applied requirements and were updated to the structured tuples.

Relationship to #1713: that PR takes an escaping approach for the extraction path; this one follows the "one structured representation" direction from the #1698 close and also fixes the reference-aware path. Happy to coordinate whichever direction you prefer.

Closes #1696

## Test plan

- [x] 4 new regression tests: JSON-array pipe preservation, legacy bracket-prefixed fallback, promoted-reference pipe preservation for constraints and for acceptance criteria (3 fail before the fix, all pass after)
- [x] `uv run pytest tests/ -q` passes — 12888 passed, 12 skipped
- [x] `uv run mypy src/` clean — no issues in 467 source files

### DIFF
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -21,8 +21,9 @@ Example: "Build a CLI task management tool in Python"
 
 ### 2. CONSTRAINTS
 Hard limitations or requirements that must be satisfied.
-Format: pipe-separated list
-Example: "Python >= 3.12 | No external database | Must work offline"
+Format: single-line JSON array of strings. Values may contain any characters,
+including literal `|` pipes; never use a bare pipe as the list separator.
+Example: ["Python >= 3.12", "No external database", "Must work offline"]
 
 ### 3. ACCEPTANCE_CRITERIA
 Specific, measurable criteria for success.
@@ -74,7 +75,7 @@ Provide your analysis in this exact structure:
 
 ```
 GOAL: <clear goal statement>
-CONSTRAINTS: <constraint 1> | <constraint 2> | ...
+CONSTRAINTS: ["<constraint 1>", "<constraint 2>", ...]
 ACCEPTANCE_CRITERIA:
 AC: <description> | verify: <command or NONE> | artifacts: <comma-list or NONE> | expect: <output assertion or NONE>
 AC: <description> | verify: <command or NONE> | artifacts: <comma-list or NONE> | expect: <output assertion or NONE>
--- a/src/ouroboros/bigbang/requirement_distillation.py
+++ b/src/ouroboros/bigbang/requirement_distillation.py
@@ -266,8 +266,8 @@ def apply_requirement_distillation(
     ]
     filtered = {
         "goal": promoted_goals[0] if promoted_goals else "Confirmed interview requirements",
-        "constraints": " | ".join(dict.fromkeys(promoted_constraints)),
-        "acceptance_criteria": " | ".join(promoted_criteria),
+        "constraints": tuple(dict.fromkeys(promoted_constraints)),
+        "acceptance_criteria": tuple(promoted_criteria),
         "ontology_name": "ConfirmedRequirementContract",
         "ontology_description": "Only user-authorized interview requirements.",
         "ontology_fields": "",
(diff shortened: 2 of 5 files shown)


sergiobuilds:
Maintainer follow-up pushed on head `5a350593`. JSON intent now applies to every value beginning with `[`, except the existing closed bracket-prefix prose form such as `[P0] Must work offline`. Unquoted/truncated array-shaped output now raises before legacy pipe splitting and reaches the extraction retry path. Added both direct parser and end-to-end retry regressions for `[--lang ko|en | keep exact flag`. Verification: focused adversarial cases 7 passed; Big Bang plus Seed contract suites 709 passed; Ruff, format, and diff checks passed.

sumin220:
After six review rounds the remaining blocker class has converged to inputs like `[ko] preserve ko|en flag` — syntactically identical to the legacy bracket prose the #1698 close comment asked to keep (`[P0] Must work offline | [P1] Fast startup`). No syntactic rule can route one to retry while keeping the other on the legacy split.

Proposal: make the extraction boundary strictly structured — any `[`-led CONSTRAINTS value that isn't a valid JSON string array goes through the existing retry (the extraction prompt already demands a JSON array) — and keep the lenient legacy split (pipe lists and bracket prose, never raising) only where legacy data actually lives: stored requirements consumed by `_build_seed`. Legacy seeds keep working; the LLM-output boundary becomes fully structured.

Happy to push that if it matches your intent — or if prose should stay accepted at extraction time, current HEAD already does that and the remaining finding is inherent to that choice.

sumin220:
Addressed the sixth review in 255d3a90 by implementing the split proposed above: extraction-time CONSTRAINTS parsing is now strictly structured — any `[`-led value that is not a valid JSON array of strings routes through the existing retry — and the never-raising legacy pipe/prose split applies only to stored requirements consumed at seed-build time, which have no retry path.

This removes the prose-vs-array heuristic entirely (no syntactic rule can separate `[ko] preserve ko|en flag` from `[P0] Must work offline | [P1] Fast startup`); the boundary is drawn by data source instead of shape. Legacy `[P0]`-style coverage is pinned by lenient-mode regression tests, and bracket prose at extraction time now gets one retry to reformat as the JSON array the prompt already demands.

Verification: tests/unit/bigbang 653 passed, ruff and mypy clean. Happy to revert to extraction-time prose acceptance instead if that's the preferred trade-off.

sumin220:
Addressed the seventh review in 4b896ede: strict extraction now accepts exactly one shape — a valid JSON array of strings — and routes every other value (plain pipe lists included) through the extraction retry path. The structured boundary has no fallback surface left.

Lenient parsing for stored legacy requirements (plain pipe lists and `[P0]`-style bracket prose) is unchanged and pinned by lenient-mode tests; extraction fixtures representing compliant LLM output now use JSON arrays.

Verification: tests/unit/bigbang 654 passed, tests/unit/mcp + tests/unit/auto 2668 passed, ruff and mypy clean.

Q00:
@sumin220 @kigam-research 
thank you for your first contribution and report this issue. will release today!

Q00 (approved):

PR #1716, opened 2026-07-23
test: hermetic isolation + eliminate real agent-CLI spawns + faster CI

Q00:
## Summary

- **Fix flaky test isolation**: 17 runtime-inference tests failed locally with `assert 'codex' == 'opencode'` while passing on CI — they read the developer's real `~/.ouroboros/config.yaml`. Root cause was non-hermeticity (everything derives from `Path.home()`), not cross-worker pollution.
- **Eliminate real agent-CLI subprocess spawns**: a full-suite spawn audit found tests spawning real `claude`/`codex`/`opencode` binaries, making behavior depend on what's installed on the machine (and adding ~13s to one test).
- **Cut CI wall-clock**: switch xdist to `worksteal` distribution.

## Changes

**Hermetic test isolation** (`tests/conftest.py`, `config/models.py`, `persistence/event_store.py`)
- Autouse fixture isolates `Path.home()` per test (chosen over `$HOME`/`OUROBOROS_HOME` so `uv build`'s cache and the ~dozens of `patch("pathlib.Path.home")` tests keep working).
- `get_config_dir()` gains an `OUROBOROS_HOME` override seam; the no-arg `EventStore()` default DB now resolves through it (stops writing the shared real `~/.ouroboros/ouroboros.db` under `pytest -n`).
- Test fixups for import-time constants / config-dir path assertions.

**No real agent-CLI spawns** (`tests/conftest.py`, `test_setup.py`, `test_opencode_config.py`)
- Autouse guard pins the answer-refiner factory (`build_answer_refiner`) and the `ooo setup` capability probes (`_codex_uses_profile_v2`, `_debug_paths_config_dir`) to their documented CLI-unavailable fallbacks — the effective clean-CI behavior.
- The dedicated probe unit tests opt back into the real functions (import-time alias / class-scoped restore fixture).

**CI speed** (`.github/workflows/test.yml`)
- `pytest ... -n 4 --dist worksteal`.

## Notes

- Verified: 12949 tests green across multiple full runs with different scheduling (`load -n4`, `worksteal -n4/-n8`).
- A re-audit drops real agent-CLI spawns from 32 to 8; the remaining 8 are `tmp_path` stub binaries the codex/zcode CLI runtimes exec on purpose (executing the stub *is* the behavior under test) — not real installed CLIs.
- Timing (full suite): `-n4 load` 105s → `-n4 worksteal` 93s; the worst single test drops 13.77s → 0.06s once its real `claude` spawns are removed.

### DIFF
--- a/.github/workflows/dev-publish.yml
+++ b/.github/workflows/dev-publish.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Install uv
         if: steps.check.outputs.skip == 'false'
-        uses: astral-sh/setup-uv@v8.3.0
+        uses: astral-sh/setup-uv@v8.3.2
         with:
           version: "latest"
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v7.0.0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8.3.0
+        uses: astral-sh/setup-uv@v8.3.2
         with:
           version: "latest"
 
@@ -46,7 +46,7 @@ jobs:
         uses: actions/checkout@v7.0.0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8.3.0
+        uses: astral-sh/setup-uv@v8.3.2
         with:
           version: "latest"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8.3.0
+        uses: astral-sh/setup-uv@v8.3.2
         with:
           version: "latest"
 
@@ -145,7 +145,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8.3.0
+        uses: astral-sh/setup-uv@v8.3.2
         with:
           version: "latest"
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: Run pytest with coverage
         run: |
-          uv run --python "${{ matrix.python-version }}" --no-sync pytest tests/ -n 4 --cov=src/ouroboros --cov-report=xml --cov-report=term-missing -v
+          uv run --python "${{ matrix.python-version }}" --no-sync pytest tests/ -n 4 --dist worksteal --cov=src/ouroboros --cov-report=xml --cov-report=term-missing -v
 
       - name: Upload coverage artifact
         uses: actions/upload-artifact@v7
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: Run pytest with coverage
         run: |
-          uv run --python "${{ matrix.python-version }}" --no-sync pytest tests/ -n 4 --dist worksteal --cov=src/ouroboros --cov-report=xml --cov-report=term-missing -v
+          uv run --python "${{ matrix.python-version }}" --no-sync pytest tests/ -n 4 --dist worksteal --durations=25 --cov=src/ouroboros --cov-report=xml --cov-report=term-missing -v
 
       - name: Upload coverage artifact
         uses: actions/upload-artifact@v7
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8.3.0
+        uses: astral-sh/setup-uv@v8.3.2
         with:
           version: "latest"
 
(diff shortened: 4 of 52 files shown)


Q00:
Thanks for the deep review — all three blockers were valid. Addressed in `d793c50f2`:

**Blocker 1 & 2 — `OUROBOROS_HOME` untrusted-env surface + split persistence** → **removed entirely.**
The override was a production state-root contract introduced only to serve a test need — and you're right that it (a) let the untrusted project `.env` redirect config/credentials carrying executable CLI paths, and (b) split persistence (`EventStore` honored it; `BrownfieldStore`/dashboard reader/MCP doctor did not). `get_config_dir()` and `EventStore` are reverted to their `Path.home()`-based originals. Test isolation is now achieved **purely** by patching `Path.home` — no production surface change (`src/` diff vs `main` is empty).

**Blocker 3 — collection-time isolation / overclaim** → **fixed.**
`Path.home` is now patched at conftest import (i.e. *before* collection) to a session-scoped temp dir, with the per-test fixture layering a fresh dir per test on top. This covers import-time home-derived constants and collection-time writes (`~/.ouroboros/logs/ouroboros.log`, `config.loader` reading `~/.ouroboros/.env` at import). Verified with your repro:

```
$ CLEANHOME=$(mktemp -d); HOME=$CLEANHOME pytest --collect-only tests/unit/mcp/tools/test_pm_handler.py
$ find $CLEANHOME -path '*/.ouroboros/*'    # → nothing; ~/.ouroboros untouched
```
The overclaiming docstring is corrected to describe the two-layer behavior.

**Follow-up — refiner-wiring coverage** → **added.**
`test_auto_handler_wires_answer_refiner_factory_result_into_driver` re-patches `build_answer_refiner` to a sentinel and asserts it reaches `AutoInterviewDriver.answer_refiner`, so the suite-wide `None` guard can't mask a broken hand-off (also proves the guard is overridable). The CLI consumer uses the identical `build_answer_refiner()` → `AutoInterviewDriver(answer_refiner=...)` line; happy to add a symmetric CLI test if you'd like it explicit.

**Suggestion — reproducible CI timing** → CI now runs `--durations=25`.

Full suite: **12949 passed** under `worksteal -n8` (65s locally). I did not reproduce the PM-snapshot git-fixture failures you saw — they look environment-specific (git identity/network for the snapshot fixtures); happy to dig in if CI shows them.

Q00:
Round 2 addressed in `c45a9d78b` — thanks again, the dual-resolver blocker was correct.

**Blocker — `$HOME`/`expanduser` consumers** → **fixed at the single chokepoint.**
You're right that `cli/commands/{cancel,run,job,resume,tui}.py` build the default EventStore DB via `os.path.expanduser("~")`, which my `Path.home()`-only patch didn't cover. Since `Path.home()` calls `expanduser("~")` internally, **`$HOME` is the one source both read** — conftest now redirects `$HOME` (before collection) to a throwaway session dir, with the per-test fixture layering a fresh dir per test. This also covers subprocess inheritance. Verified against your exact probe:

```
$ HOME=$(mktemp -d) python -c "from ouroboros.cli.commands import cancel; \
    print(cancel.os.path.expanduser('~/.ouroboros/ouroboros.db'))"   # → under the isolated home
```
External tools keep working under the isolated home via explicit locations: `uv build` (wheel test) uses `UV_CACHE_DIR`; git fixtures set per-repo identity (no global config needed). The wheel-packaging + PM-snapshot git tests pass locally.

**Coverage — both mechanisms + subprocess inheritance** → **added** `tests/unit/test_home_isolation.py`: asserts `Path.home()` and `os.path.expanduser` agree on the isolated home, that the flagged `expanduser`-based CLI DB paths resolve under it, and that a spawned subprocess inherits it.

**Suggestion 1** → `test_pm_handler` now asserts `_DATA_DIR` was captured beneath the (conftest-exposed) session home — proves pre-collection isolation held, stronger than the suffix-only check.

**Suggestion 2** → added the symmetric CLI answer-refiner wiring test.

Full suite: **12952 passed** under `worksteal -n8`. One unrelated timing-sensitive test (`test_generation_watchdog::test_no_material_progress_timeout_emits_cancellation_mode`, a 0.07s no-progress limit) flaked under heavy local `-n8` load; it passes 3/3 in isolation, is untouched by this change, and CI runs `-n4`.

Q00:
Follow-up warning addressed in `2c9696acf`.

**`test_opencode_config.py` — class-wide probe restore could spawn a real CLI.**
You're right: the autouse `_restore_real_debug_paths_probe` fixture restored the real `_debug_paths_config_dir` for *every* `TestOpencodeConfigDir` test, but the darwin/XDG-fallback tests neither mock `subprocess.run` nor clear `OUROBOROS_OPENCODE_CLI_PATH`/`OPENCODE_CLI_PATH` — so an externally configured executable ran during the probe (your `OUROBOROS_OPENCODE_CLI_PATH=/usr/bin/touch` repro created `debug`/`paths`).

Fix: made the restore **opt-in** (`real_debug_paths_probe` fixture) — only the three tests that also mock `subprocess.run` request it, so no real CLI is ever executed. The fallback tests keep the conftest guard (`probe → None`) and still exercise darwin/XDG resolution (the probe is skipped when it returns `None`).

Verified with your repro:
```
$ OUROBOROS_OPENCODE_CLI_PATH=/usr/bin/touch pytest tests/unit/cli/test_opencode_config.py::TestOpencodeConfigDir
6 passed  —  opencode/touch spawns: NONE   (previously `touch` ran and created files)
```

PR #1679, opened 2026-07-19
fix(auto): preserve explicit Seed acceptance contracts

AlphaComposite:
## Summary

Prevents `ooo auto` from silently broadening a non-empty Seed acceptance contract with task-class defaults. Task-class inference still records the active class and still fills a genuinely empty contract, while deterministic Seed repair now preserves structured AC identity and verification evidence.

This also documents that mentioning a Seed ID/path in auto goal prose does not establish typed derivative lineage; callers must use evolution or execute the immutable Seed directly.

## Test plan

- [x] `uv run pytest -q tests/unit/auto/test_pipeline_task_class_envelope.py` — 4 passed
- [x] `uv run pytest -q tests/unit/auto/test_interview_pipeline.py -k seed_repairer` — 7 passed
- [x] Broader `tests/unit/auto tests/unit/bigbang` — 1,919 passed; 6 runtime-routing tests failed identically on clean `v0.50.5` because the user-level stage profile routes authoring/execution to different runtimes than their legacy expectations
- [x] `ruff check src tests` and `ruff format --check src tests` clean
- [x] `mypy src/ouroboros` clean
- [x] GitNexus impact: LOW for both changed methods; change detection MEDIUM, limited to expected auto/repair flows
- [x] Independent read-only review: PASS

## R-run comparison (required for `src/ouroboros/auto/` changes)

This is a deterministic contract-preservation guard and structure-preserving model copy; it does not alter per-round model calls, orchestration count, or execution scheduling.

| Metric                         | Baseline (sha) | This PR (sha) | Ratio |
|--------------------------------|----------------|---------------|-------|
| Rounds completed in 600 s      | N/A            | N/A           | n/a   |
| Per-round wall-clock (s/round) | N/A            | N/A           | n/a   |
| Terminal reason                | N/A            | N/A           | n/a   |
| EventStore event count         | N/A            | N/A           | n/a   |

Budget compliance: [ ] within 1.5× / [ ] regression flagged with mitigation / [x] N/A (substrate-only contract guard; no per-round execution change)

## Related issues

Fixes #1678

PR #1717, opened 2026-07-23
fix(auto): preserve explicit Seed acceptance contracts (rebased #1679)

### LINKED ISSUE #1678: bug(auto): task-class defaults silently broaden non-empty Seed contracts
## Summary

`ooo auto` can silently broaden a non-empty Seed acceptance contract after generation by prepending task-class defaults inferred from broad ledger keywords. In a real Gefa run on v0.50.5, a Seed with six explicit local-STT criteria was classified as `GAME_2D` and received three unrelated game criteria before QA.

## Impact

High. The deterministic post-generation step mutates the execution contract without user confirmation. QA caught this instance, but the polluted criteria were persisted and repeatedly carried through repair generations. If similarly injected criteria happen to pass review, execution could follow unauthorized scope.

A second issue compounds this: indexed `SeedRepairer` replacements convert `AcceptanceCriterionSpec` into plain strings, discarding semantic keys, verify commands, expected artifacts, output assertions, and investment metadata.

## Reproduction

Environment:

- Ouroboros `0.50.5`
- macOS / Python CLI runtime
- `ooo auto` with a non-empty six-criterion Rust Seed contract

Observed persisted sequence:

1. Generated Seed had six Gefa Phase 0 criteria.
2. `_record_generated_seed()` inferred `TaskClass.GAME_2D` from ledger text.
3. `apply_default_ac_template()` prepended the catalog's exact three game defaults:
   - headless N-tick state-change trace;
   - clean quit from the game main loop;
   - runnable build launches without missing assets.
4. The Seed now had nine criteria while its Success contract still referred to AC1–AC6.
5. Repair generations preserved the pollution and chained immediate parent IDs.

Relevant session: `auto_c92617d366e0`; final QA failed closed at score `0.32`.

## Expected behavior

- A non-empty user/generated Seed acceptance contract is not broadened by task-class inference.
- Task-class defaults remain available only as fallback when generation produced no acceptance criteria.
- Structured AC repair changes only the targeted description while retaining semantic identity and verification evidence.
- `ooo auto` documentation clearly states that mentioning a parent Seed path/ID in goal prose does not establish derivative lineage; typed evolution must use the evolution surface.

## Proposed minimal fix

1. In `AutoPipeline._record_generated_seed()`, apply task-class defaults only when `seed.acceptance_criteria` is empty; still record `active_task_class` for non-empty contracts.
2. In `SeedRepairer.repair_once()`, preserve `AcceptanceCriterionSpec` via `model_copy(update={"description": ...})` and preserve AC order/multiplicity rather than converting structured entries to strings/deduplicating them.
3. Add regressions for non-empty preservation, empty fallback, and structured repair evidence.
4. Document the unsupported derivative-by-path behavior in `skills/auto/SKILL.md`.

## Acceptance criteria

- A single task-class inference with a non-empty Seed leaves AC count, order, descriptions, and semantic identities unchanged.
- An empty Seed still receives the task-class fallback template.
- Structured repair retains semantic key, verify command, artifacts, assertion, investment metadata, order, and untouched sibling criteria.
- Existing immediate-parent repair lineage tests remain green.
- Auto/bigbang focused suites, Ruff, format, and mypy pass; any unrelated baseline failures are documented.

Q00:
## Summary

Rebases and completes @AlphaComposite's #1679 (closed because the fork branch could not carry new commits). All three original commits are preserved with their authorship; a final commit resolves the blocking normalization-boundary finding that repeated bot review flagged.

Fixes #1678. Supersedes #1679. Thanks @AlphaComposite for the original contribution — carried forward here.

## What #1679 already did (AlphaComposite)

- `AutoPipeline._record_generated_seed()` applies task-class defaults **only** when the Seed contract is empty; a non-empty contract still records `active_task_class` without broadening.
- `SeedRepairer.repair_once()` preserves `AcceptanceCriterionSpec` identity/evidence via `model_copy(update={"description": ...})` instead of flattening to strings.
- Documents that mentioning a Seed ID/path in auto goal prose does not establish typed derivative lineage.

## What this rebase adds (the remaining blocker)

The repeated blocking review was the **many-to-one canonicalization boundary** in `normalize_execution_acceptance`:

1. `Seed` materializes every legacy string as an `AcceptanceCriterionSpec`, so restoring **one structured entry per source** treated equivalent legacy criteria as independent structured inputs — one canonical AC expanded into duplicate execution work (7 execution-acceptance tests failed).
2. Distinct structured criteria collapsing to one canonical description remained several ACs sharing that description. The MCP evaluation map keys contracts by description via `setdefault`, so every verification contract but the first became unreachable.

**Fix:** collapse the sources for each canonical text into **one** contract — first non-empty scalar wins in source order, `expected_artifacts` are an order-preserving union, and the first source's semantic identity is kept so the criterion has a **stable, unique key**. Bare legacy collapses degrade to a plain string, keeping persistence byte-for-byte legacy-compatible. This gives the explicit, unique criterion-to-contract association requested at close through canonicalization, persistence, and evaluation.

## Test plan

- [x] `uv run pytest -q tests/unit/auto/test_execution_acceptance.py` — 26 passed (added legacy-collapse + multi-source regressions)
- [x] `uv run pytest -q tests/unit/auto/test_interview_pipeline.py` — pipeline-level regression asserts the merged contract stays reachable through `_seed_ac_spec_map`; rewrote the prior test that asserted the lossy duplicate behavior
- [x] `uv run pytest -q tests/unit/auto` — 1287 passed; 4 `test_surface` runtime-routing failures are the documented pre-existing codex/opencode stage-profile leak, unrelated to this change
- [x] `ruff check` / `ruff format --check` / `mypy src/ouroboros/auto/execution_acceptance.py` clean

This is a deterministic, contract-preservation change to Seed acceptance normalization and repair. It does not alter per-round model calls, orchestration count, execution scheduling, or the auto convergence loop; it only changes how already-collected acceptance criteria are canonicalized and deduplicated before hand-off. No per-round execution characteristic changes.

| Metric                         | Baseline (sha) | This PR (sha) | Ratio |
|--------------------------------|----------------|---------------|-------|
| Rounds completed in 600 s      | N/A            | N/A           | n/a   |
| Per-round wall-clock (s/round) | N/A            | N/A           | n/a   |
| Terminal reason                | N/A            | N/A           | n/a   |
| EventStore event count         | N/A            | N/A           | n/a   |

Budget compliance: [ ] within 1.5× / [ ] regression flagged with mitigation / [x] N/A (substrate-only contract guard; no per-round execution change)

Note on applicability: this change is deterministic normalization of the acceptance-criteria *set* before hand-off. It does not alter the `ooo auto` convergence loop — no change to per-round LLM calls, round cadence, orchestration count, or terminal conditions — which is exactly what the four R-run metrics above measure, hence N/A. It does change the normalized AC cardinality/order for observation/autoresearch seeds (that is the fix), but the executor's per-AC dispatch is deterministic and bounded by the same seed; no per-round wall-clock regression is possible from a pure normalization pass.

### DIFF
--- a/skills/auto/SKILL.md
+++ b/skills/auto/SKILL.md
@@ -58,6 +58,12 @@ ooo auto "Build a local-first habit tracker CLI" --complete-product
 /ouroboros:auto "Build a local-first habit tracker CLI"
 ```
 
+`ooo auto` does not accept a parent Seed ID/path or infer lineage from goal
+prose. Mentioning an existing Seed in the goal is ordinary context, not an
+authorized derivative operation. Use `ooo evolve` for typed parent evolution,
+or `ooo run` to execute an existing immutable Seed. Do not claim preserved
+lineage or AC edit scope from an `ooo auto` goal alone.
+
 ## CLI flag → MCP arg translation
 
 When the user types `ooo auto` with CLI-style flags inside chat, translate to MCP arguments before invoking `ouroboros_start_auto`:
--- a/src/ouroboros/auto/execution_acceptance.py
+++ b/src/ouroboros/auto/execution_acceptance.py
@@ -4,7 +4,12 @@ from __future__ import annotations
 
 import re
 
-from ouroboros.core.seed import AcceptanceCriterionInput, Seed, ac_text
+from ouroboros.core.seed import (
+    AcceptanceCriterionInput,
+    AcceptanceCriterionSpec,
+    Seed,
+    ac_text,
+)
 
 _AUTO_WRAPPER_CRITERIA = frozenset(
     {
@@ -213,19 +218,52 @@ def _restore_surviving_acceptance_specs(
     filtered: tuple[str, ...],
     original: tuple[tuple[AcceptanceCriterionInput, str], ...],
 ) -> tuple[AcceptanceCriterionInput, ...]:
-    """Keep structured contracts for criteria that survive normalization unchanged."""
-    original_by_text: dict[str, list[AcceptanceCriterionInput]] = {}
-    for criterion, text in original:
-        original_by_text.setdefault(text, []).append(criterion)
+    """Keep structured contracts when surviving criteria are canonicalized.
 
+    Normalization may rewrite a known-equivalent hello_auto criterion rather
+    than preserving its text verbatim.  That description-only rewrite must not
+    discard the criterion's semantic identity or verification evidence.
+    """
+    remaining = list(original)
     restored: list[AcceptanceCriterionInput] = []
     for text in filtered:
-        matches = original_by_text.get(text)
-        if matches:
-            restored.append(matches.pop(0))
-        else:
-            restored.append(text)
+        exact_index = next(
+            (
+                index
+                for index, (_criterion, original_text) in enumerate(remaining)
+                if original_text == text
+            ),
+            None,
+        )
+        if exact_index is not None:
+            criterion, _original_text = remaining.pop(exact_index)
+            restored.append(criterion)
+            continue
+
+        canonical_sources = [
+            (index, criterion)
+            for index, (criterion, original_text) in enumerate(remaining)
+            if isinstance(criterion, AcceptanceCriterionSpec)
+            and _structured_criterion_normalizes_to(original_text, text)
+        ]
+        if len(canonical_sources) == 1:
+            index, criterion = canonical_sources[0]
+            remaining.pop(index)
+            restored.append(criterion.model_copy(update={"description": text}))
+            continue
+
+        restored.append(text)
     return tuple(restored)
+
+
+def _structured_criterion_normalizes_to(original_text: str, normalized_text: str) -> bool:
+    """Return whether one structured hello_auto AC became ``normalized_text``."""
+    subject = _unwrap_seed_repairer_original_requirement(original_text).strip()
+    if _normalize_known_observation_execution_line(subject) == normalized_text:
+        return True
+    return normalized_text.startswith(
+        "Create `hello_auto.py` and `tests/test_hello_auto.py` so "
+    ) and _is_hello_auto_observation_unit_line(original_text)
 
 
 def normalize_observation_execution_criteria(
--- a/src/ouroboros/auto/execution_acceptance.py
+++ b/src/ouroboros/auto/execution_acceptance.py
@@ -163,6 +163,16 @@ _AUTORESEARCH_CANONICAL_AC = (
     "The final report artifact includes baseline val_bpb, each attempted experiment result, final best val_bpb, and the keep/discard reason for every candidate.",
 )
 
+
+def _autoresearch_canonical_position(text: str) -> int | None:
+    """Return the fixed canonical index of ``text``, or None if not canonical."""
+    stripped = text.strip()
+    for index, canonical in enumerate(_AUTORESEARCH_CANONICAL_AC):
+        if canonical.strip() == stripped:
+            return index
+    return None
+
+
 _AUTORESEARCH_NON_GOALS = (
     "Do not edit prepare.py.",
     "Do not edit files outside train.py unless scope_widening_ledger explicitly widens scope.",
@@ -357,7 +367,15 @@ def _restore_surviving_acceptance_specs(
         )
         if exact_index is not None:
             criterion, _text = _pop(exact_index)
-            _emit((1, float(exact_index)), criterion)
+            # A source whose text IS an autoresearch canonical AC belongs in the
+            # fixed canonical sequence (tier 0 at its canonical index), not the
+            # source coordinate space — otherwise a lone exact baseline AC would
+            # sort after injected experiments/reporting and invert stage order.
+            canonical_position = _autoresearch_canonical_position(text)
+            if canonical_position is not None:
+                _emit((0, float(canonical_position)), criterion)
+            else:
+                _emit((1, float(exact_index)), criterion)
             continue
 
         # A filtered text with no remaining source.  Either the normalizer
@@ -526,7 +544,14 @@ def _collapse_canonical_acceptance_specs(
 def _structured_criterion_normalizes_to(original_text: str, normalized_text: str) -> bool:
     """Return whether one structured hello_auto AC became ``normalized_text``."""
     subject = _unwrap_seed_repairer_original_requirement(original_text).strip()
-    if _normalize_known_observation_execution_line(subject) == normalized_text:
+    normalized_subject = _normalize_known_observation_execution_line(subject)
+    # ``_normalize_known_observation_execution_line`` returns its input unchanged
+    # for anything it does not recognize as a hello_auto line.  Requiring an
+    # actual transformation avoids a false identity match — e.g. an autoresearch
+    # canonical AC equal to its own ``normalized_text`` must NOT be treated as a
+    # hello collapse (which would route it into the source coordinate space and
+    # invert the canonical sequence).
+    if normalized_subject != subject and normalized_subject == normalized_text:
         return True
     return normalized_text.startswith(
         "Create `hello_auto.py` and `tests/test_hello_auto.py` so "
@@ -853,8 +878,31 @@ def _is_observation_report_wrapper(criterion: str) -> bool:
     return "observation report" in key or "plain chat summary" in key
 
 
+# Case-insensitive, whitespace-tolerant matcher for the seed-repairer wrapper
+# prefix.  Stripping via this regex (rather than the case-folded criterion key)
+# preserves the original casing of the wrapped requirement — a repaired
+# requirement referencing case-sensitive identifiers like ``RawStderr.LOG`` or
+# ``VAL_BPB`` must round-trip verbatim, not lowercased.
+_SEED_REPAIRER_WRAPPER_RE = re.compile(
+    r"^\s*"
+    + re.escape(_SEED_REPAIRER_ORIGINAL_REQUIREMENT_PREFIX.strip()).replace(r"\ ", r"\s+")
+    + r"\s+",
+    re.IGNORECASE,
+)
+
+
 def _unwrap_seed_repairer_original_requirement(criterion: str) -> str:
-    key = _criterion_key(criterion)
-    if not key.startswith(_SEED_REPAIRER_ORIGINAL_REQUIREMENT_PREFIX):
-        return criterion
-    return key.removeprefix(_SEED_REPAIRER_ORIGINAL_REQUIREMENT_PREFIX)
+    """Strip every nested seed-repairer wrapper, preserving the inner casing.
+
+    The repairer can wrap an already-wrapped criterion, so unwrapping must
+    recurse until the innermost requirement is exposed; otherwise a nested
+    wrapper is later matched as a generic placeholder and the real requirement
+    it carries is deleted.  Casing is preserved so case-sensitive identifiers
+    survive round-trip.
+    """
+    text = criterion.strip()
+    while True:
+        stripped = _SEED_REPAIRER_WRAPPER_RE.sub("", text, count=1).strip()
+        if stripped == text:
+            return text
+        text = stripped
--- a/src/ouroboros/auto/execution_acceptance.py
+++ b/src/ouroboros/auto/execution_acceptance.py
@@ -404,30 +404,56 @@ def _transfer_contract_to_emitted_canonical(
 ) -> bool:
     """Attach a covered source's contract to its canonical AC, in place.
 
-    Returns whether the canonical AC was found and updated.  The canonical
-    criterion keeps its description and gains the source's verification
-    evidence, so the executor/evaluator sees exactly one contracted AC rather
-    than a contractless canonical plus a duplicate source.  An existing contract
-    on the canonical AC is never clobbered.
+    Returns whether the source is fully represented by the canonical AC (so the
+    caller may drop it).  The canonical criterion keeps its description and gains
+    the source's verification evidence — including an explicitly-supplied
+    ``semantic_ac_key`` that runtime routing/recovery correlate on — so the
+    executor/evaluator sees exactly one contracted AC rather than a contractless
+    canonical plus a duplicate source.
+
+    A canonical AC already carrying a contract is only treated as representing
+    this source when the two contracts are identical; a *different* contract
+    returns False so the caller preserves the source as its own criterion rather
+    than silently dropping a distinct command.
     """
     target = canonical_text.strip()
+    explicit_key = (
+        source.semantic_ac_key
+        if source.semantic_ac_key is not None
+        and source.semantic_ac_key != derive_semantic_ac_key(source)
+        else None
+    )
+    transferred = AcceptanceCriterionSpec(
+        description=canonical_text,
+        semantic_ac_key=explicit_key,
+        verify_command=source.verify_command,
+        expected_artifacts=source.expected_artifacts,
+        output_assertion=source.output_assertion,
+        investment=source.investment,
+    )
     for position, (anchor, item) in enumerate(emissions):
         if ac_text(item).strip() != target:
             continue
         if isinstance(item, AcceptanceCriterionSpec) and item.has_success_contract:
-            return True
-        emissions[position] = (
-            anchor,
-            AcceptanceCriterionSpec(
-                description=canonical_text,
-                verify_command=source.verify_command,
-                expected_artifacts=source.expected_artifacts,
-                output_assertion=source.output_assertion,
-                investment=source.investment,
-            ),
-        )
+            # Only a byte-identical contract is safe to collapse onto; a distinct
+            # command must survive as its own criterion.
+            return _autoresearch_contract_matches(item, transferred)
+        emissions[position] = (anchor, transferred)
         return True
     return False
+
+
+def _autoresearch_contract_matches(
+    existing: AcceptanceCriterionSpec,
+    candidate: AcceptanceCriterionSpec,
+) -> bool:
+    """Return whether two canonical-AC contracts are identical evidence."""
+    return (
+        existing.verify_command == candidate.verify_command
+        and existing.expected_artifacts == candidate.expected_artifacts
+        and existing.output_assertion == candidate.output_assertion
+        and existing.investment == candidate.investment
+    )
 
 
 def _collapse_canonical_acceptance_specs(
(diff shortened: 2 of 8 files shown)


Q00:
Addressed the second-round blocking findings in 95a1b233b.

**Finding #1 — lossy scalar merge.** Removed the "first non-empty scalar wins" merge entirely. A scalar `AcceptanceCriterionSpec` cannot represent several independent contracts, so distinct sources are no longer merged. When a canonical text would conflate *distinct* contracts, the collapse is refused and each contract-bearing source survives **verbatim**, uniquely reachable through the description-keyed persistence/evaluation boundary. Only bare/legacy equivalents (or byte-identical contracts) still collapse to one.

**Finding #2 — autoresearch canonicalization.** Added a loss-free backstop: any contract-bearing source a normalizer replaced or dropped (the six-AC autoresearch rewrite this matcher does not model) is re-appended verbatim, so its `verify_command`/artifacts/assertion/investment/key are never discarded.

**Follow-up — repair dedup.** `SeedRepairer` now drops only byte-for-byte identical criteria; any difference in wording, contract, or identity keeps them distinct, so multiplicity is preserved without emitting duplicate execution ACs.

**Regressions added:** distinct-contract preservation + reachability, equivalent-contract collapse, structured-contract survival across autoresearch canonicalization, and identical-only repair dedup.

**Validation:** `tests/unit/auto` — 1290 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract/investment subset — 24 passed; ruff/format/mypy clean.

Q00:
Addressed the third-round blocking findings in 2766036a9. The restoration is no longer purely description-based — it now preserves semantic identity and source position, which the review correctly identified as authoritative.

**Blocker #1 — explicit identity.** `_carries_explicit_contract` now recognizes an explicitly-supplied `semantic_ac_key` (one that differs from `derive_semantic_ac_key`) as identity, and the collapse-distinctness test compares full identity (contract **+** explicit key), not the contract alone. Three criteria carrying only distinct explicit keys no longer fold into one re-derived key; each identity survives. Auto-derived keys still carry no intent, so legacy strings still collapse.

**Blocker #2 — unique reachability.** A final dedupe-by-description keeps the first criterion per description, so no two emitted criteria collide on the key `_seed_ac_spec_map` uses. Every surviving contract is uniquely reachable.

**Follow-up — source order.** Every emission is anchored to its originating source index and re-sorted, so a refused collapse or autoresearch replacement can no longer move a contract ahead of/behind its siblings. `[contract A, report, contract B]` stays A-before-B; injected canonical ACs anchor to their filtered position.

**Regressions added:** explicit-identity preservation, refused-collapse source order, unique-description reachability.

**Validation:** `tests/unit/auto` — 1293 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; semantic-identity/routing/recovery subset — 72 passed; ruff/format/mypy clean.

Q00:
Addressed the fourth-round blocking findings in 1c611e428. Normalization is now loss-free — it never deletes a caller-authorized requirement or contract to satisfy any downstream constraint.

**Blocker #1 — refused collapse dropped bare siblings.** The refusal branch now keeps **every** source verbatim at its own position, not just the contract-bearing ones. The bare requirement the canonical text would have represented (e.g. "the exact pytest command passes") survives instead of being silently consumed.

**Blocker #2 — dedupe-by-description deleted distinct contracts.** Removed entirely. Two criteria that arrive sharing a description now keep **both** contracts through to durable handoff. Per the review's own guidance ("or otherwise preserve both distinct contracts instead of deleting one"), I chose preservation. The description-keyed `_seed_ac_spec_map` is a pre-existing boundary limitation for such inputs; normalization no longer manufactures new collisions and never resolves one by discarding a valid contract.

**Non-blocking note — stale key on bare collapse.** The collapse degrade decision now evaluates the *original* identity, so a bare legacy collapse degrades to a string and Seed re-derives a fresh key from the canonical description. Added a serialized-representation assertion proving the persisted key matches the canonical text, not a collapsed source.

**Regressions strengthened** to assert source contract *cardinality and semantics*, not merely reachability of survivors: every source survives a refused collapse (3/3), both same-description contracts persist, and a bare collapse serializes with a fresh derived key.

**Validation:** `tests/unit/auto` — 1293 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Fixed the fifth-round follow-up (phantom contractless duplicate on a repeated canonical description) and the docstring nit in 80004dada. On the remaining blocker I want to flag a genuine design tradeoff for @Q00 rather than silently reverse purpose-built behavior:

**What the blocker asks:** stop `normalize_file_artifact_execution_criteria` and `normalize_autoresearch_execution_criteria` from dropping library-default / "covered" criteria, because a non-empty Seed's criteria are now caller-authored (task-class defaults only fill *empty* contracts after this PR).

**Why it isn't a clean code fix:**

1. *File-artifact library defaults* are dropped only when the goal is an exact file-state task (`_has_file_artifact_context`) **and** the criterion is exactly a catalog library default (`all public API symbols importable`, `ruff check … exit 0`, …). The drop exists so a "write one file with content X" worker isn't handed unrelated import/lint ACs that then fail evaluation. Preserving them has a real downside (irrelevant ACs reaching the worker); the caller-authored scenario the probe constructs (a file-write Seed that *also* demands full public-API importability) is self-contradictory. This is a product call, and it's covered by `drops_library_defaults_for_file_artifact` / `drops_wrapped_library_defaults_for_file_artifact`.

2. *Autoresearch "covered" clauses* use `startswith` matching. The existing green test canonicalizes criteria like "…record a baseline … **before any experiment changes evaluated**" (elaboration of the same requirement), while the probe's "…record a baseline … **signed by reviewer Alice**" adds a *distinct* requirement. Both are trailing clauses on the same prefix; no syntactic heuristic separates "elaboration (safely covered)" from "new requirement (must survive)" without either breaking the RFC-designed clean 6-AC contract (`canonicalizes_autoresearch_contract`) or under-dropping.

Everything in the restoration boundary I own is now loss-free: distinct contracts and explicit identities are never conflated, source order is preserved, refused collapses keep every sibling, and no phantom duplicates are produced. The two remaining drops live in the context normalizers and are intentional, tested behavior with real tradeoffs.

@Q00 — this needs your call: (a) make those two normalizers fully loss-free (I'll update the drop tests + add the probes as regressions), (b) keep them as-is (intentional boilerplate/irrelevant-AC cleanup), or (c) a middle ground (keep the file-artifact drop, tighten only the autoresearch `startswith` coverage). I've held off on reversing tested behavior unilaterally.

Q00:
Addressed both sixth-round blockers in 751f8fe53 and 06e2e757e. The two remaining losses were both in the autoresearch path:

**Blocker #2 — structured contract duplicated + contractless canonical.** A source truly subsumed by a canonical autoresearch AC is now mapped to that canonical criterion (`_AUTORESEARCH_COVERED_PHRASES` carries the canonical index) and its verification contract is **transferred onto the canonical AC in place**. Execution/evaluation now see exactly one contracted baseline AC — no contractless canonical + duplicate source. An existing contract on the canonical AC is never clobbered.

**Blocker #1 — plain source with an extra requirement deleted.** Autoresearch coverage is now vocabulary-aware. A covered-prefix source is treated as subsumed only when its remaining words stay inside the canonical requirement vocabulary (baseline/experiment/ledger/discard/diff/report domain). A foreign token — "stderr", "raw", "log", "signed", … — marks a distinct caller requirement the six canonical ACs don't carry, so the source is preserved rather than dropped. Standard restatements (all-domain vocabulary) still collapse into the clean canonical contract.

**Regressions added:** contract transfer onto the canonical baseline AC (six ACs, no duplicate, contract reachable via `_seed_ac_spec_map`) and preservation of a baseline restatement carrying an extra "raw stderr log" clause. The existing clean-contract autoresearch test stays green.

**Validation:** `tests/unit/auto` — 1294 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Addressed all four seventh-round blockers in 6ba6b2e04 (and confirmed the file-artifact / same-description tradeoffs are withdrawn — thanks).

1. **Multi-contract collapse.** Transferring onto an already-contracted canonical AC now returns success only when the two contracts are byte-identical; a second covered baseline with a *different* verify command is preserved as its own criterion rather than dropped. Both commands survive.

2. **Explicit key on transfer.** The transferred canonical AC now inherits the source's explicitly-supplied `semantic_ac_key` (a key differing from the derived value), so durable routing/recovery correlation is preserved instead of getting a freshly derived key.

3. **Equivalence, not vocabulary.** You're right that token membership can't establish equivalence — "record a baseline after every experiment" is all in-domain yet contradicts the canonical before-edit baseline. Coverage now requires an **exact** match against the known standard requirement restatements (demonstrable equivalence); any reworded, contradicting, or extended criterion is preserved verbatim. The vocabulary heuristic is gone.

4. **Stale repaired key.** `SeedRepairer` now re-derives a non-explicit key after a description rewrite (an auto-derived key no longer masquerades as explicit identity), while an explicitly-supplied key is still preserved verbatim.

**Regressions added:** distinct covered contracts both survive, explicit key transferred onto the canonical AC, contradicting in-vocabulary criterion preserved, and repaired auto-derived key refreshed to match the rewrite.

**Validation:** `tests/unit/auto` — 1298 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Addressed the eighth-round blockers and the ordering follow-up in bdabf8e0d.

1. **Full-identity collapse.** The transfer now refuses to collapse onto a canonical AC carrying *any* explicit authority unless the **complete** identity matches — verify_command, expected_artifacts, output_assertion, investment, **and** semantic_ac_key. Authority-only specs (investment or explicit key with no success contract) are no longer bypassed by the `has_success_contract` gate, so identical commands with different keys keep both keys, key-only criteria don't overwrite each other, and differing investments both survive.

2. **Exact generic removal.** The seed-repairer placeholder is now matched exactly, not by substring, so "A command/API check returns stable observable output **while preserving the raw stderr log** …" is a distinct requirement that survives instead of being deleted as generic.

3. **Source-ordered transfer.** A transferred contract is re-anchored to its **source index**, so sequential-execution tuple order follows the source. `[distinct source, covered baseline]` keeps the distinct source first.

**Regressions added:** distinct covered identities (same command/different keys; differing investments) both survive; generic removal is exact-only; a transferred baseline keeps its source position behind an earlier distinct source.

**Validation:** `tests/unit/auto` — 1301 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Addressed all three ninth-round blockers in ee15bbb52.

1. **Collapse key refresh.** A contract kept under a canonical hello description now re-derives its non-explicit `semantic_ac_key` from that description (an explicitly-supplied key is preserved). Canonical identity no longer depends on the collapsed source's wording, so nothing masquerades as explicit identity downstream.

2. **One ordering coordinate system.** You correctly flagged the mixed anchors. Emissions now sort on a single `(tier, position)` key — tier 0 = normalizer-injected criteria in filtered order, tier 1 = source-derived criteria in source order. A contract transferred onto an injected canonical AC keeps that AC's **tier-0 position**, so the fixed baseline→experiments→… canonical sequence is never re-anchored into the source coordinate space.

3. **Equivalent-contract collapse.** Transfer identity comparison is now on **explicit** identity (`_explicit_semantic_key`) rather than raw key. A materialized auto-derived key vs `None` is not a real identity difference, so a canonical-exact source and its known restatement carrying the same contract collapse to one AC — the command is dispatched once, not twice.

**Regressions added:** hello collapse re-derives its key; the canonical baseline/experiment sequence survives a transfer (baseline before experiments); equivalent canonical/restatement contracts collapse to a single dispatched command.

**Validation:** `tests/unit/auto` — 1303 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Re-requesting review for HEAD `ee15bbb52` — all three ninth-round blockers (collapse key refresh, single (tier, position) ordering coordinate, explicit-identity equivalent collapse) are addressed with regressions. `tests/unit/auto` 1303 passed, ruff/format/mypy clean.

Q00:
Addressed all three tenth-round blockers in 284b8dcf1, and clarified the R-run applicability in the PR body.

1. **Stable canonical ordering.** Root cause: `_structured_criterion_normalizes_to` treated a non-hello *identity* match as a hello collapse, so a source equal to a canonical autoresearch AC was routed into the source coordinate space and sorted after the injected experiments/reporting ACs. It now requires an actual hello transformation, and an exact-canonical source anchors to tier 0 at its canonical index — a lone exact baseline AC stays first, full canonical sequence intact.

2. **Case-preserving unwrap.** The seed-repairer wrapper is now stripped by a case-insensitive regex over the original text, not the case-folded key, so `RawStderr.LOG` / `VAL_BPB` round-trip verbatim instead of lowercased.

3. **Recursive unwrap.** Unwrapping recurses through every nested wrapper, so a doubly-wrapped criterion carrying a distinct requirement is exposed and preserved rather than deleted as a generic placeholder.

**Regressions added:** lone exact baseline AC keeps the full canonical order; unwrap preserves case and strips nested wrappers; a nested-wrapped raw-stderr requirement survives.

**On the R-run follow-up:** the four R-run metrics measure the `ooo auto` convergence loop (rounds in 600 s, per-round wall-clock, terminal reason, event count). This change is a deterministic normalization of the AC *set* before hand-off — it does not touch per-round LLM calls, round cadence, orchestration count, or terminal conditions, so those metrics are genuinely N/A. It does change normalized AC cardinality/order (that's the fix), but a pure normalization pass cannot cause a per-round wall-clock regression. I've spelled this out in the R-run section.

**Validation:** `tests/unit/auto` — 1306 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Addressed the eleventh-round blocker in 93b8c3622.

**Exact-match grouping.** Restoration no longer consumes only the first exact-text source. It now groups **every** source sharing an exact text and collapses them by full identity: byte-identical contracts (and a bare source subsumed by a contracted twin) collapse to one execution AC — one command, one key — while genuinely distinct identities each survive. Canonical exact matches keep their tier-0 canonical position, so two identical already-canonical baselines yield six ACs (not seven) and dispatch the command once, independent of source order.

**Regressions added:** byte-identical collapse (one command, six ACs), order-independent bare/contract subsumption, and distinct-contract preservation on the same canonical text.

**Validation:** `tests/unit/auto` — 1307 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Addressed both twelfth-round blockers in 0475fcea1.

1. **Wrapped structured passthrough no longer duplicated.** A SeedRepairer-wrapped structured criterion had its unwrapped requirement emitted by the string normalizer *and* its original contracted spec appended by the backstop (8 ACs, requirement dispatched twice). The backstop now transfers the contract **onto** the unwrapped passthrough emission — case preserved, fresh derived key — instead of appending a second copy, and identical wrapped sources collapse by full identity. One wrapped `RawStderr.LOG` criterion now yields exactly one contracted AC.

2. **No mis-transfer to a narrower canonical AC.** "up to two post-baseline experiments … sequentially … kept … reverted" spans experiment count (canonical AC 1) *and* sequential keep/revert semantics (canonical AC 2), so no single AC is one-to-one equivalent. Its transfer target is now `None`: a contracted source is preserved verbatim rather than attaching its contract to the narrower count-only AC; a bare restatement stays covered by the canonical set.

**Regressions added:** a wrapped structured requirement yields one contracted AC (case preserved, one command); a multi-stage restatement's contract is not mis-transferred to a narrower canonical AC.

**Validation:** `tests/unit/auto` — 1309 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Addressed all three thirteenth-round blockers in ac93811c0. They shared one root cause — equivalence was resolved across separate restoration passes — so the fix reconciles atomically.

1. **Distinct restatement reachability.** When a transfer conflicts with an occupied canonical AC, the source is now emitted under its **own** unwrapped description, not the canonical text. A canonical baseline (cmd_a) and a distinct restatement (cmd_b) keep separate descriptions, so `_seed_ac_spec_map` reaches both.

2. **Scan every same-text emission.** `_place_transferred_onto_emission` now scans **all** emissions sharing the target text: it collapses onto any equivalent identity (enriching a bare one if present) and reports "distinct" only when every match carries a different contract. An equivalent restatement collapses onto an existing later AC instead of duplicating it.

3. **Atomic reconciliation + idempotence.** A final `_reconcile_emissions` pass groups emissions by description and keeps one criterion per distinct identity at its earliest position. Exact-canonical and canonicalizing sources can no longer emit two identical ACs, and re-normalizing a normalized Seed is a fixed point.

**Regressions added:** mixed canonical/restatement reachability via `_seed_ac_spec_map`, equivalent collapse across multiple matches, and normalization idempotence.

**Validation:** `tests/unit/auto` — 1312 passed (the 4 `test_surface` failures are the documented pre-existing codex/opencode stage-profile leak, unrelated); downstream evaluation/contract subset — 22 passed; ruff/format/mypy clean.

Q00:
Re-requesting review for HEAD `ac93811c0` — the three thirteenth-round blockers (distinct-restatement reachability, scan-all-matches equivalent collapse, atomic reconciliation + idempotence) are addressed with regressions covering `_seed_ac_spec_map` reachability and re-normalization idempotence. `tests/unit/auto` 1312 passed, ruff/format/mypy clean.

Q00:
@ouroboros-agent please re-review HEAD `ac93811c0`. The thirteenth-round blockers (distinct-restatement reachability, scan-all-matches collapse, atomic reconciliation + idempotence) are resolved with regressions; `tests/unit/auto` 1312 passed, ruff/format/mypy clean.

PR #1720, opened 2026-07-23
fix(interview): atomically persist resumable artifacts

berkshirehathaways:
## Summary

Interview state and PM seed JSON are resume-critical, but direct overwrites can leave a truncated file after an interrupted write.

This change uses a shared private same-directory atomic writer for both paths: preserve existing permission bits, fsync content, replace atomically, fsync the POSIX parent directory where supported, and clean up descriptors and temporary files on failure. Existing interview locking and public APIs remain unchanged.

## Test plan

- [x] 650 tests in tests/unit/bigbang
- [x] 31 focused atomic-save, failure-injection, mode, fsync, and round-trip tests
- [x] Partial-write, fdopen, fsync, replace, and directory-fsync failure paths preserve prior bytes and clean up
- [x] Existing destination modes are preserved
- [x] Ruff check and format clean
- [x] MyPy clean on changed source files
- [x] Python compileall clean on changed source and tests

## Related issues

No duplicate issue or pull request matched these two direct resume-artifact save paths.

### DIFF
--- a/src/ouroboros/bigbang/interview.py
+++ b/src/ouroboros/bigbang/interview.py
@@ -115,9 +115,7 @@ def _fsync_parent_directory(file_path: Path) -> bool:
     try:
         directory_fd = os.open(file_path.parent, flags)
     except OSError as error:
-        if error.errno in (errno.EINVAL, errno.ENOTSUP):
-            return True
-        return False
+        return error.errno in (errno.EINVAL, errno.ENOTSUP)
     durability_confirmed = True
     try:
         try:
--- a/src/ouroboros/bigbang/interview.py
+++ b/src/ouroboros/bigbang/interview.py
@@ -105,9 +105,9 @@ _TOOLLESS_INTERVIEW_BASE_PROMPT = """## Role Boundaries
 """
 
 
-def _fsync_parent_directory(file_path: Path) -> None:
+def _fsync_parent_directory(file_path: Path) -> bool:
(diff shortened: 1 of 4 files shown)


berkshirehathaways:
Follow-up pushed in f1a19e33.

- Keeps pre-publication write/replace failures fatal.
- Treats post-replace directory fsync/close errors as durability-uncertain, so save_state() and save_pm_seed() report the already-published artifact consistently with reload/visible JSON.
- Adds regression coverage for both consumers under EIO.

Local verification: git diff --check and Python 3.14 compileall pass. Targeted pytest could not collect locally because the available pytest runs on Python 3.11 while this branch uses Python 3.12+ syntax, and pytest is not installed for Python 3.14. Please re-review the new head.

berkshirehathaways:
CI follow-up: corrected the Ruff SIM103 finding in 606f66d7 and pushed the new head. The atomic-save behavior and regression coverage are unchanged.

berkshirehathaways:
CI follow-up: formatted the new regression test with Ruff and pushed 59329735. The previous Ruff check was logic-clean; only ruff format --check needed this adjustment.

berkshirehathaways:
All checks are now green on 59329735 (Python 3.12/3.13/3.14, Ruff, MyPy, TypeScript, and policy checks). The PR is mergeable/clean; review status remains CHANGES_REQUESTED pending reviewer re-review.

PR #1726, opened 2026-07-24
fix(mcp): omit unset optional FastMCP arguments

### LINKED ISSUE #1538: MCP tool_wrapper forwards unset optional params as explicit None — normalize at the chokepoint
Follow-up to #1502 (merged as a correct hotfix for #1501). Two hardening items remain so the bug *class* — not just the two crash sites — is closed:

**1. Normalize at the chokepoint (`src/ouroboros/mcp/server/adapter.py` `tool_wrapper`, ~:925-943).**
The wrapper builds optional params with `annotation = list | None, default = None`, and FastMCP fills unset optionals with that explicit `None`; `normalized_kwargs.setdefault(...)` then forwards it to handlers. In-process calls skip the wrapper, so handlers see *missing key* (`.get(k, [])` → safe) while plugin-MCP calls see *key present, value None* (→ `TypeError: 'NoneType' object is not iterable`). Any future optional array/object param whose handler iterates its default will silently reintroduce the #1501 crash. Either strip/convert `None`-valued unset optionals at the wrapper so both paths look identical, or document the asymmetry as a contract.

**2. Add one regression test through the real wrapper path.**
The #1502 tests inject `None` manually into handler args — they pin "handlers tolerate explicit None" but never reproduce that the None *originates* in the wrapper, so a wrapper regression wouldn't be caught. One test driving `tool_wrapper`/`call_tool` with the param omitted (serialized envelope) closes that gap.

Context: #1502 review confirmed the fix itself addresses the root cause at the two live crash sites (`evolution_handlers.py:559`, `ralph_handlers.py:403`) and is consistent across both; this issue is about preventing recurrence for future params.

berkshirehathaways:
## Summary
- normalize FastMCP signature aliases to original MCP parameter names
- omit None placeholders only for declared optional parameters before handler dispatch
- retain non-None and required arguments
- prove the omitted-parameter behavior through the real FastMCP call_tool boundary

Closes #1538

## Verification
- pytest -q tests/unit/mcp/server/test_adapter.py: 67 passed
- ruff check and ruff format --check changed files
- mypy src/ouroboros/mcp/server/adapter.py
- manual FastMCP scenarios: omitted optional, explicit alias value, optional None, required value, and required None/missing validation

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -1019,7 +1019,17 @@ class MCPServerAdapter:
                         if alias_key in kwargs:
                             normalized_kwargs[original_key] = kwargs[alias_key]
                     for key, value in kwargs.items():
-                        normalized_kwargs.setdefault(key, value)
+                        normalized_kwargs.setdefault(alias_to_original.get(key, key), value)
+                    optional_parameter_names = {
+                        parameter.name
+                        for parameter in h.definition.parameters
+                        if not parameter.required
+                    }
+                    normalized_kwargs = {
+                        key: value
+                        for key, value in normalized_kwargs.items()
+                        if value is not None or key not in optional_parameter_names
+                    }
 
                     # Route through call_tool() to enforce security checks.
                     # FastMCP does not provide credentials, so:
--- a/tests/unit/mcp/server/test_adapter.py
+++ b/tests/unit/mcp/server/test_adapter.py
@@ -1430,6 +1430,65 @@ class TestServeTransport:
         assert isinstance(returned["duration_ms"], int)
 
     @pytest.mark.asyncio
+    async def test_fastmcp_wrapper_omits_unset_optional_arguments(self) -> None:
+        from unittest.mock import MagicMock, patch
+
+        class OptionalToolHandler(MockToolHandler):
+            @property
+            def definition(self) -> MCPToolDefinition:
+                return MCPToolDefinition(
+                    name=self._name,
+                    description="A test tool",
+                    parameters=(
+                        MCPToolParameter(name="input", type=ToolInputType.STRING),
+                        MCPToolParameter(
+                            name="optional-input",
+                            type=ToolInputType.STRING,
+                            required=False,
+                        ),
+                    ),
+                )
+
+        adapter = MCPServerAdapter()
+        handler = OptionalToolHandler(name="optional_tool")
+        adapter.register_tool(handler)
+
+        mock_fastmcp_cls = MagicMock()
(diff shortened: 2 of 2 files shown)


berkshirehathaways:
리뷰의 테스트 보강 후속사항을 반영했습니다.\n\n- optional alias에 대해 명시적 non-None 값이 원래 MCP key로 전달되는지 검증했습니다.\n- 인자를 생략했을 때 optional alias가 호출 payload에서 빠지는지도 검증했습니다.\n\nCommit: \n검증: Ruff check/format, Python 3.14 compileall 통과. GitHub CI 재실행을 확인 중입니다.

berkshirehathaways:
리뷰의 테스트 보강 후속사항을 반영했습니다.

- optional alias에 대해 명시적 non-None 값이 원래 MCP key로 전달되는지 검증했습니다.
- 인자를 생략했을 때 optional alias가 호출 payload에서 빠지는지도 검증했습니다.

Commit: 0029ed93
검증: Ruff check/format, Python 3.14 compileall 통과. GitHub CI 재실행을 확인 중입니다.

berkshirehathaways:
Added a7f388a5 to close the remaining #1538 test gap before merge. The regression now drives real FastMCP.call_tool with the optional parameter omitted, rather than only invoking the captured wrapper.\n\nObserved behavior: omitted optional arguments reach the handler as absent; explicit alias values retain the original MCP name; optional None is omitted; required inputs remain subject to FastMCP validation.\n\nVerification: adapter suite 67 passed; Ruff, formatting, and adapter mypy passed. Independent goal, QA, code, security, and context reviews found no blockers.

Q00:
Rebased onto current `main` and re-pushed as `ae8cb4c6d`.

**The CI failure was not yours.** `Test Python 3.12` failed on `tests/unit/mcp/test_job_manager_drain.py::TestJobManagerDrain::test_drain_does_not_recount_terminal_job_while_task_unwinds` — a file this PR does not touch (`git diff origin/main` shows only `mcp/server/adapter.py` and `mcp/server/test_adapter.py`). It passed on 3.13 and 3.14 in the same run.

That test uses hard wall-clock bounds (`asyncio.wait_for(..., timeout=2.0)` against `grace_seconds=0.05`) around a deliberately held terminal append, which is the classic shape for a flake under xdist load. Ran it 25× in isolation locally: 0 failures. Not hardened here, because it is unrelated to this PR and bundling it would muddy the diff — filed as a follow-up instead if it recurs.

This PR's own suite: `tests/unit/mcp/server/test_adapter.py` 67 passed; ruff and mypy clean.

Issue #1681, opened 2026-07-20
RFC: Re-sequence the Ouroboros AC runtime into approval-gated stacked PRs


Q00:
> **Status:** Stack 1 and Routing B/C are merged. Routing A remains explicitly deferred by the owner decision recorded below.
> **Current review slot:** Stack 3 — Learning A is the only unlocked implementation slot; no Learning A PR is open.
> **Locked:** Learning B remains locked until Learning A has an approved current-head review. Page / Probe / continuation remains locked until Learning A is approved and follow-up sequencing is reconfirmed here.
> **Last reconciled:** 2026-08-03 against `main@cb658aa`. This Q00-authored issue remains the sequencing source of truth.

---

## Goal

Rebuild the Ouroboros AC runtime as a sequence of small, independently reviewable changes that converge on automatic, frugal `model × harness` routing.

The desired runtime should:

- try the cheapest **eligible** route first;
- escalate model, harness, effort, or lateral strategy only when evidence justifies it;
- keep model names and provider-specific assignments out of orchestration policy;
- use advisory reasoning to rank candidates, while a deterministic Admission Kernel owns dispatch authorization;
- let the Final Gate remain the only authority that can declare an AC accepted;
- learn from raw route observations within a run first, then across runs;
- treat the 100–140k smart zone as a physical attention/session boundary, not as AC atomicity;
- use Page/Probe/continuation for large atomic ACs rather than decomposing them solely because of token pressure.

## Why #1648 and #1667 are being replaced

The feature direction is sound, but both PRs accumulated correctness responsibilities inside `parallel_executor` before the shared authority and event semantics were stable.

In particular:

- `execution.ac.outcome_finalized` currently acts as acceptance evidence, frugality-proof admission, MCP recovery evidence, attention progress, and lateral-escalation replay state;
- provisional attempt results and final-workspace acceptance are not represented as separate authorities;
- runtime, verifier, workspace, and execution-policy identity were completed only after reusable trust and checkpoint recovery already depended on them;
- trust/escalation projections are independently folded by the board, AC-tree HUD, and conductor;
- model routing, effort routing, alternate-harness redispatch, and lateral-persona escalation are separate decision systems rather than one route-selection contract;
- parking currently risks becoming monitoring plus indefinite redispatch instead of a bounded transition to an explicit blocked state.

This RFC preserves the reusable modules and tests from the experiments while changing the merge order.

## Approval gate

Only **one unapproved implementation PR may be open at a time**.

[...]

PR #1731, opened 2026-07-24
Fix Codex completion evidence projection

### LINKED ISSUE #1724: Codex tool completions are not journaled as completed evidence, causing false fat-harness FABRICATION_SUSPECTED failures
## Summary

A Codex execution can successfully edit files and run tests, emit a valid typed evidence record, and still be rejected by the fat-harness because the execution journal does not provide matchable completed-tool evidence.

In one execution aggregate:

```text
execution.tool.started:   158
execution.tool.completed: 0
typed_evidence_present:   true
typed_evidence_valid:     true
verifier_evidence_used:   []
verifier_failure_class:   FABRICATION_SUSPECTED
```

The affected ACs were rejected with `unsupported evidence claims` for exact `tests_passed` commands and files generated by those commands.

## Environment

- ouroboros-ai: 0.50.5
- codex-cli: 0.144.0
- macOS 26.5.1 arm64
- runtime backend: codex_cli
- permission mode: bypassPermissions

## Representative failure

Codex ran a command such as:

```text
node scripts/audit/run-command.mjs --finding-id F-006 --receipt-id focused-test -- node --test tests/unit/audit-run-command.test.mjs
```

and reported the exact same command in both `commands_run` and `tests_passed`. The final JSON evidence record was structurally valid, but the verifier rejected it:

```text
Fat-harness verifier failed
unsupported evidence claims:
  tests_passed: node scripts/audit/run-command.mjs ... -- node --test ...
```

Files generated by the wrapper, such as receipt JSON/stdout/stderr artifacts, were also rejected as unsupported `files_touched` claims.

## Observed evidence gap

Codex JSONL emits `item.completed` events for command executions and file changes. The Codex runtime converter has an `item.completed` branch, but the persisted execution projection records tool-bearing messages as `execution.tool.started` without a corresponding `execution.tool.completed` event.

The verifier requires transcript-grounded success evidence and exact journal matching. A tool start is insufficient to prove:

- exit code 0,
- test success,
- stdout/stderr provenance,
- or files generated by a Bash command.

Wrapper commands amplify the problem when the inner test output is written to receipt artifacts rather than exposed in the runtime message.

## Expected behavior

For a Codex `item.completed(command_execution)` event, Ouroboros should preserve and project:

- stable tool-call ID,
- exact command,
- completion state,
- exit code,
- stdout and stderr,
- success/failure status,
- and a matching `execution.tool.completed` journal event.

For `item.completed(file_change)`, the journal should record workspace-relative mutated paths.

Files generated by a successful Bash command should be attributable through an explicit artifact manifest or a command-scoped workspace diff, rather than being indistinguishable from stale files.

## Suggested fix

1. Project Codex `item.completed(command_execution)` as a completed tool event rather than another tool-start event.
2. Correlate start/completion with the Codex item ID or another stable call ID.
3. Preserve `exit_code`, `output`, `stdout`, `stderr`, and status in the evidence manifest.
4. Project completed `file_change` items as mutation evidence with workspace-relative paths.
5. For Bash-generated artifacts, add explicit artifact provenance (or pre/post workspace diff attribution).
6. Ensure wrapper commands can prove test success through their structured receipt, without trusting the agent's final prose/JSON alone.

## Regression tests

Please add recorded Codex JSONL fixture tests covering:

- `item.completed(command_execution)` with exit code 0 produces `execution.tool.completed`;
- the exact successful test command is accepted as `tests_passed`;
- a command not present in the journal is still rejected;
- a completed Codex file change supports the matching workspace-relative `files_touched` claim;
- a Bash-generated receipt can be verified through explicit artifact provenance;
- non-zero exit codes and malformed completion metadata remain fail-closed.

berkshirehathaways:
Fixes #1724\n\n## Summary\n- project native Codex completion events into correlated start/result journal receipts\n- preserve command output, exit status, and completion provenance\n- fail closed for missing, malformed, unknown, nonzero, and conflicting terminal metadata\n\n## Verification\n- pytest: parallel executor (276 passed)\n- pytest: Codex runtime + projection (100 passed)\n- ruff check/format and mypy on changed runtime/tests

### DIFF
--- a/src/ouroboros/orchestrator/codex_cli_runtime.py
+++ b/src/ouroboros/orchestrator/codex_cli_runtime.py
@@ -45,6 +45,7 @@ from ouroboros.orchestrator.adapter import (
     SubagentOrchestration,
     TaskResult,
 )
+from ouroboros.orchestrator.mcp_tools import normalize_runtime_tool_result, serialize_tool_result
 from ouroboros.providers.base import CompletionConfig
 from ouroboros.providers.codex_cli_stream import (
     iter_runtime_stream_lines,
@@ -1735,6 +1736,76 @@ class CodexCliRuntime:
                 self._merge_command_metadata(data, nested)
         return data
 
+    @staticmethod
+    def _extract_tool_call_id(item: dict[str, Any]) -> str | None:
+        for key in ("id", "item_id", "call_id"):
+            value = item.get(key)
+            if isinstance(value, str) and value.strip():
+                return value.strip()
+        return None
+
+    @staticmethod
+    def _command_result_text(metadata: dict[str, Any]) -> str:
+        parts: list[str] = []
+        for key in ("stdout", "output", "result_preview", "stderr"):
+            value = metadata.get(key)
+            if isinstance(value, str) and value.strip() and value.strip() not in parts:
+                parts.append(value.strip())
+        return "\n".join(parts)
+
+    @staticmethod
+    def _completion_failed(
+        metadata: dict[str, Any],
+        *,
+        require_terminal_evidence: bool = False,
+    ) -> bool:
+        if metadata.get("is_error_invalid") is True or metadata.get("is_error") is True:
+            return True
+        status = metadata.get("status")
+        if isinstance(status, str) and status.strip():
+            normalized_status = status.strip().lower()
+            if normalized_status in {"failed", "error"}:
+                return True
+            if normalized_status not in {"completed", "success", "succeeded"}:
+                return True
+        exit_code = metadata.get("exit_code")
+        if exit_code is not None:
+            return not isinstance(exit_code, int) or isinstance(exit_code, bool) or exit_code != 0
+        if isinstance(status, str) and status.strip():
+            return False
+        return require_terminal_evidence
+
+    def _build_tool_completion_message(
+        self,
+        *,
+        tool_name: str,
+        tool_call_id: str,
+        content: str,
+        is_error: bool,
+        handle: RuntimeHandle | None,
+        metadata: dict[str, Any] | None = None,
+    ) -> AgentMessage:
+        result_metadata = dict(metadata or {})
+        return AgentMessage(
+            type="tool_result",
+            content=content,
+            tool_name=tool_name,
+            data={
+                "subtype": "tool_result",
+                "tool_call_id": tool_call_id,
+                "is_error": is_error,
+                "runtime_event_type": "tool.completed",
+                "tool_result": serialize_tool_result(
+                    normalize_runtime_tool_result(
+                        content,
+                        is_error=is_error,
+                        meta=result_metadata,
+                    )
+                ),
+            },
+            resume_handle=handle,
+        )
+
     def _merge_command_metadata(self, data: dict[str, Any], source: dict[str, Any]) -> None:
         """Merge known command-result fields from one Codex event object."""
         text_key_map = {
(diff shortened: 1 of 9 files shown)


berkshirehathaways:
Addressed the latest review on 473cc9f6.\n\n- ID-bearing Codex command completions with no text now expose their normalized exit status to the test-evidence detector; the new regression proves exit_code=0 supports the focused test claim, while the manual negative check confirms exit_code=1 is rejected.\n- Runner tool-called events now emit only for projected invocations, so named tool results are not counted twice.\n- Progress persistence keeps the message runtime_event_type when a resumed handle carries stale metadata.\n\nVerification: 277 parallel-executor tests passed; 218 runner/session tests passed (1 skipped); Ruff and mypy passed for changed files. Independent goal, QA, code, security, and context reviews found no remaining blockers. Requesting re-review.

Q00:
@berkshirehathaways — thank you for this. Your work is not being discarded; it is being split, and the consumer half is now open as #1740 with your commits' authorship preserved.

Here is the full reasoning, including a finding that credits your consumer patch specifically.

## What happened

You and @sumin220 (#1732) independently fixed #1724/#1690. The two PRs turned out not to be competitors — they are the two halves of one pipeline:

```
Codex emits events → [adapter translates] → [consumers use] → verifier / TUI
                          #1732                  #1731
```

Both of you also wrote an adapter half. Yours and theirs overlapped; the consumer half was yours alone.

## Why the adapter half went to #1732

Not quality — sequencing and blocker status.

When @Q00 closed the earlier #1692, two specific blockers were named: terminal success must stay tri-state/fail-closed, and a `file_change` completion must not synthesize a duplicate start when a real `item.started` was observed. #1732 resolved both explicitly and reached APPROVED at head.

The three blocking findings against this PR were all located in **your adapter half**, not your consumer half:

- `codex_cli_runtime.py:1954` / `:2022` — synthetic starts carry only `tool_call_id`, so on a resumed handle whose persisted event type is `run.completed`, projection classifies the start as a session result and both journal emitters skip it, leaving an orphan completion receipt.
- eight mypy annotation errors, in the adapter-side test file.

Drop the adapter half and those findings have nothing left to attach to. That is what #1740 does.

## Your consumer half found something the review did not

This is the part worth your time.

Your patch makes message-level `runtime_event_type` take priority over stale handle metadata. That is correct, and it is also what made a naming difference load-bearing.

Your adapter stamped completions `runtime_event_type="tool.completed"`. The merged adapter uses `"tool.result"`. Because your consumer patch routes the message's own value straight into lifecycle classification, `_runtime_handle_lifecycle_state` (`adapter.py:416`) now sees it — and it maps any value containing `"completed"` outside the `message.`/`result.`/`turn.` prefixes onto the terminal `completed` state, which is in `_RUNTIME_TERMINAL_STATES`. Verified by running it:

```
tool.completed -> lifecycle=completed  terminal=True
tool.result    -> lifecycle=running    terminal=False
```

**Merged whole, your PR would have marked the entire runtime handle terminal on the first finished tool.** Neither half was dangerous alone — your adapter naming was inert until your own consumer patch gave it authority. The review bot flagged synthetic starts, tool counters, and the TUI, but did not catch this one.

It surfaced only because the halves were separated and the consumer patch was run against a different adapter. So the split was not just bookkeeping — it was the thing that exposed it.

#1740 does not merely fix the assertion; it pins the invariant (`test_message_level_tool_completion_event_type_is_not_terminal`), driving the real `CodexCliRuntime` conversion and asserting no per-tool event type classifies the runtime as terminal. That guard exists because of your patch.

## Two things you flagged that neither PR fixes

You were right that these are broken; both are pre-existing and cross-runtime, so they are being tracked separately instead of bundled into a Codex-scoped change:

1. **`tool_calls_count` inflation.** Worth being precise: your emitter/runner guards remove the duplicate *journal event*, but the counter runs through `workflow_state.py:704` (`process_runtime_message` passes `projected.tool_name` for tool *and* tool_result) into `process_message:682` (`if tool_name: tool_calls_count += 1`). Neither PR touches that path, and every runtime — Claude included — already double-counts today.
2. **Failed tools render green.** `emit_atomic_tool_completed` never sets `success`, and `tui/events.py:1000` does `success=data.get("success", True)`. Separate PR, going up shortly.

## Governance

RFC #1681 now carries a written gate boundary, added because this PR and #1732 were both held against a rule that had never been written down: an adapter's own event-conversion contract is outside the gate; shared projection/emitter/consumer semantics are sequenced. Your consumer half is the sequenced part, which is why it lands second rather than not at all.

Closing this in favour of #1740. Please review it — and if you disagree with the split or think something of yours was dropped that should not have been, say so on #1740 and it goes back in.

PR #1732, opened 2026-07-24
fix(codex): journal correlated tool completions from item lifecycle events

### LINKED ISSUE #1690: Codex CLI item completion events do not reach execution evidence
## Summary

On Ouroboros 0.50.5, Codex CLI JSONL `item.started` and `item.completed` events are not projected as a correlated tool start and tool result pair. Successful command output can therefore disappear from the authoritative execution evidence.

## Impact

The worker can create the requested files and pass its tests, while the verifier receives no correlated completion evidence. This can produce `FABRICATION_SUSPECTED` or a failed acceptance criterion even though the command really completed.

The problem is not specific to one project. It affects executions using the Codex CLI runtime event stream.

## Steps to reproduce

1. Use the Codex CLI runtime with Ouroboros 0.50.5.
2. Feed an `item.started` event for a `command_execution` item with an item `id`.
3. Feed the matching `item.completed` event with the same `id`, output, exit code 0, and completed status.
4. Project and persist the resulting runtime messages.
5. Observe that no correctly correlated `execution.tool.started` plus successful completion evidence pair is available to the deliver gate.

A failed `file_change` or `mcp_tool_call` also needs special care: failure status and result text must not be dropped and then defaulted to success. A multi-file `file_change` needs one unambiguous correlation identity per path.

## Expected behavior

- `item.started` produces a tool-start message.
- `item.completed` produces a tool-result message.
- Both retain a stable correlation identity derived from the Codex item ID.
- Output, exit code, status, `success`, and `ok` survive projection.
- Failed operations remain failed and cannot be admitted as successful mutation evidence.
- Multi-file changes have matching unique identities per path.

## Actual behavior

The stock 0.50.5 adapter handles only the completion branch for these items and does not construct the complete correlated result contract. In a real execution this produced many start records but zero usable completion records for verifier evidence.

- [ ] Add regression coverage for command start/completion pairing and result text.
- [ ] Add regression coverage for failed command, file change, and MCP completion.
- [ ] Add regression coverage for multi-file change correlation.
- [ ] Preserve the normalized runtime event type and tool call ID in persisted metadata.
- [ ] Verify a real or representative execution can provide successful test evidence to the deliver gate.

## Environment

- Ouroboros: 0.50.5
- Codex CLI: 0.144.6
- Python: 3.12.12
- OS: Ubuntu 24.04.3 LTS under WSL2
- Kernel: 6.6.87.2-microsoft-standard-WSL2

## Additional validation

A local patch with focused regression tests and a disposable end-to-end canary has confirmed the proposed event contract. A small focused PR can follow after maintainer feedback.

sumin220:
## Summary

Codex CLI item lifecycle events now convert into a correlated start/result contract, so command completions become `execution.tool.completed` journal evidence.

- `item.started` for command_execution / mcp_tool_call / file_change / web_search now emits a tool-start message carrying `tool_call_id` — previously the event was dropped entirely.
- `item.completed` for those types now projects as a `tool_result` message with the item id preserved, exit/status metadata, and a tri-state `is_error`, so `project_runtime_message` classifies it as a result and the dispatcher emits `execution.tool.completed`. Previously completions were shaped as a second start, producing the starts-only journals reported in #1724 (158 starts / 0 completions) and FABRICATION_SUSPECTED rejections of truthful claims downstream.
- Success is fail-closed per the #1692 review: only explicit signals (exit code 0, completed/success status) claim success; failed status or non-zero exit set `is_error`; malformed metadata claims nothing. For `command_execution`, lifecycle status alone never claims success: only a validated zero exit or an explicit success/ok flag counts, and a missing or malformed exit code fails closed with the raw status preserved as `reported_status`.
- A completed-only legacy stream synthesizes a start+result pair, but a real `item.started` is never duplicated (the other #1692 review blocker).
- `file_change` completions drop the hardcoded `subtype="success"` stamp and correlate per path (`{item_id}:{path}`), so failed changes are no longer laundered to success.

Closes #1690. Fixes the completion-evidence gap reported in #1724; the artifact-manifest and wrapper-receipt asks there are left as separate verifier-side follow-up.

## Sequencing vs RFC #1681

The #1692 close asked for attempt-evidence projection changes to wait for the authority/event contract RFC. This PR is deliberately adapter-local: only the conversion contract inside `codex_cli_runtime.py` changes — `runtime_message_projection.py`, `leaf_dispatcher.py`, `execution_event_emitter.py`, `runner.py`, and `deliver_gate.py` are untouched consumers (the journaling test drives the real executor path to prove the emitted shapes are consumable end to end). If you consider the adapter conversion contract within the #1681 gate, happy to hold this until the RFC lands.

## Test plan

- [x] 72 regression tests in `tests/unit/orchestrator/test_codex_item_completion_evidence.py`, each verified failing before its fix (by stashing the source change) and passing after — covering lifecycle correlation and per-invocation identity, completed-only synthesis without duplicate starts, per-item-type fail-closed verdict resolution, MCP payload preservation and credential redaction, and the review-round regressions accumulated across the audit
- [x] 5 existing conversion tests updated from the old start-shaped contract to the correlated tool_result contract
- [x] `uv run pytest tests/ -q` — 13228 passed, 12 skipped
- [x] `tests/unit/orchestrator` 3244 passed; `tests/unit/harness` 262 passed (includes the deliver-gate suites); `tests/unit/mcp` 1386 passed; `tests/unit/providers` 565 passed
- [x] `uv run ruff check` / `ruff format --check` / `mypy src/ouroboros` clean

Note: the regression tests import `_event_has_explicit_tool_success` from `ouroboros.harness.deliver_gate` (read-only) to make the fail-closed contract executable against the real gate predicate; happy to drop those assertions if importing a private harness helper into orchestrator tests is unwanted.

### DIFF
--- a/src/ouroboros/orchestrator/codex_cli_runtime.py
+++ b/src/ouroboros/orchestrator/codex_cli_runtime.py
@@ -1926,7 +1926,9 @@ class CodexCliRuntime:
         if item_type == "command_execution":
             return self._extract_command(item)
         if item_type == "mcp_tool_call":
-            return self._mcp_tool_name(item)
+            tool_input = self._extract_tool_input(item)
+            arguments = json.dumps(tool_input, ensure_ascii=False, sort_keys=True, default=str)
+            return f"{self._mcp_tool_name(item)}\x00{arguments}"
         if item_type == "file_change":
             return "\n".join(self._extract_paths(item))
         if item_type == "web_search":
@@ -2072,11 +2074,18 @@ class CodexCliRuntime:
                     if item_type != "command_execution":
                         has_success = True
 
-            error_envelope = source.get("error")
-            if (isinstance(error_envelope, dict) and error_envelope) or (
-                isinstance(error_envelope, str) and error_envelope.strip()
-            ):
-                has_failure = True
+            if "error" in source:
+                error_envelope = source.get("error")
+                # Any present, meaningfully non-empty error envelope is a
+                # failure — including malformed non-null shapes (e.g. an int
+                # or list). Only an explicitly empty/false envelope
+                # (None/{}/[]/""/0/False) is treated as "no error" (round
+                # eight, blocker 3).
+                if isinstance(error_envelope, str):
+                    if error_envelope.strip():
+                        has_failure = True
+                elif error_envelope:
+                    has_failure = True
 
             for key in ("success", "ok"):
                 flag = source.get(key)
@@ -2225,12 +2234,26 @@ class CodexCliRuntime:
         is_error = self._resolve_item_completion_is_error(item_type, item)
         item_id = self._item_lifecycle_id(item)
         if item_id is not None:
-            # Dedup on the id AND the terminal verdict: an exact replay is
-            # suppressed, but a conflicting same-id completion (e.g. a later
-            # failure after a success) has a different key and stays visible
-            # so verification fails closed (review round seven, blocker 1).
+            # Dedup on the id, input signature, verdict, AND every
+            # evidence-bearing completion field: an exact replay is
+            # suppressed, but any non-identical completion (a conflicting
+            # verdict, or the same verdict with different output) has a
+            # different fingerprint and stays visible so verification fails
+            # closed (review rounds seven & eight).
+            evidence_fields: dict[str, Any] = {
+                key: metadata.get(key)
+                for key in ("output", "stdout", "stderr", "result_preview", "exit_code", "status")
+            }
+            # The raw error envelope is evidence too: two malformed errors that
+            # produce no result text must not collapse to one fingerprint.
+            if "error" in item:
+                evidence_fields["error"] = item.get("error")
+            evidence_fingerprint = json.dumps(
+                evidence_fields, ensure_ascii=False, sort_keys=True, default=str
+            )
             dedup_key = (
-                f"{item_id}\x00{self._item_lifecycle_signature(item_type, item)}\x00{is_error}"
+                f"{item_id}\x00{self._item_lifecycle_signature(item_type, item)}"
+                f"\x00{is_error}\x00{evidence_fingerprint}"
             )
             if dedup_key in scope.completed_item_keys:
                 return []
--- a/src/ouroboros/orchestrator/codex_cli_runtime.py
+++ b/src/ouroboros/orchestrator/codex_cli_runtime.py
@@ -1860,6 +1860,12 @@ class CodexCliRuntime:
             return item_id.strip()
         return None
 
+    @staticmethod
+    def _extract_web_search_query(item: dict[str, Any]) -> str:
+        """Extract exactly the search query from a web_search thread item."""
+        query = item.get("query")
+        return query.strip() if isinstance(query, str) else ""
+
     def _item_lifecycle_signature(self, item_type: str, item: dict[str, Any]) -> str:
         """Build a best-effort identity for id-less legacy thread items."""
         if item_type == "command_execution":
@@ -1869,6 +1875,11 @@ class CodexCliRuntime:
             return name.strip() if isinstance(name, str) else ""
         if item_type == "file_change":
             return "\n".join(self._extract_paths(item))
+        if item_type == "web_search":
+            # The query is the only stable identity: volatile fields such as
+            # status must not change the signature between started/completed,
+            # or an id-less completion would synthesize a duplicate start.
+            return self._extract_web_search_query(item)
         return self._extract_text(item)
 
     def _item_tool_calls(self, item_type: str, item: dict[str, Any]) -> list[_CodexToolCall]:
@@ -1915,7 +1926,7 @@ class CodexCliRuntime:
             ]
 
         if item_type == "web_search":
-            query = self._extract_text(item)
+            query = self._extract_web_search_query(item)
             return [
                 _CodexToolCall(
                     tool_name="WebSearch",
@@ -1981,7 +1992,7 @@ class CodexCliRuntime:
         # never let an outer success status shadow a nested failure (review
         # blocker: failure signals take precedence wherever they appear).
         for source in self._iter_item_metadata_sources(item):
-            for exit_key in ("exit_code", "exitCode"):
+            for exit_key in ("exit_code", "exitCode", "returncode", "return_code"):
                 exit_code = source.get(exit_key)
                 if isinstance(exit_code, int) and not isinstance(exit_code, bool):
                     if exit_code == 0:
--- a/src/ouroboros/orchestrator/codex_cli_runtime.py
+++ b/src/ouroboros/orchestrator/codex_cli_runtime.py
@@ -192,14 +192,15 @@ class _CodexItemCorrelationScope:
     so a new thread begins with a clean correlation space.
     """
 
-    started_item_ids: set[str] = field(default_factory=set)
+    started_item_signatures: dict[str, str] = field(default_factory=dict)
     started_unkeyed_item_counts: dict[tuple[str, str], int] = field(default_factory=dict)
-    completed_item_ids: set[str] = field(default_factory=set)
+    completed_item_keys: set[str] = field(default_factory=set)
+    current_thread_id: str | None = None
 
     def clear(self) -> None:
-        self.started_item_ids.clear()
+        self.started_item_signatures.clear()
         self.started_unkeyed_item_counts.clear()
-        self.completed_item_ids.clear()
+        self.completed_item_keys.clear()
 
 
 class CodexCliRuntime:
@@ -1870,6 +1871,20 @@ class CodexCliRuntime:
         return None
 
     @staticmethod
+    def _mcp_tool_name(item: dict[str, Any]) -> str:
+        """Resolve an MCP tool identity from native tool+server or legacy name."""
+        tool = item.get("tool")
+        if isinstance(tool, str) and tool.strip():
+            server = item.get("server")
+            if isinstance(server, str) and server.strip():
+                return f"{server.strip()}.{tool.strip()}"
+            return tool.strip()
+        name = item.get("name")
+        if isinstance(name, str) and name.strip():
+            return name.strip()
+        return "mcp_tool"
+
+    @staticmethod
     def _extract_mcp_result_text(item: dict[str, Any]) -> str:
         """Normalize MCP result/error envelopes into result text."""
         error_envelope = item.get("error")
@@ -1894,6 +1909,8 @@ class CodexCliRuntime:
             structured = result.get("structured_content")
             if isinstance(structured, str) and structured.strip():
                 return structured.strip()
+            if isinstance(structured, (dict, list)) and structured:
+                return json.dumps(structured, ensure_ascii=False, sort_keys=True)
         if isinstance(result, str) and result.strip():
             return result.strip()
         return ""
@@ -1909,8 +1926,7 @@ class CodexCliRuntime:
         if item_type == "command_execution":
             return self._extract_command(item)
         if item_type == "mcp_tool_call":
-            name = item.get("name")
-            return name.strip() if isinstance(name, str) else ""
+            return self._mcp_tool_name(item)
         if item_type == "file_change":
             return "\n".join(self._extract_paths(item))
         if item_type == "web_search":
@@ -1938,8 +1954,7 @@ class CodexCliRuntime:
             ]
 
         if item_type == "mcp_tool_call":
-            raw_name = item.get("name")
-            tool_name = raw_name if isinstance(raw_name, str) else "mcp_tool"
+            tool_name = self._mcp_tool_name(item)
             return [
                 _CodexToolCall(
                     tool_name=tool_name,
@@ -1986,10 +2001,11 @@ class CodexCliRuntime:
     ) -> None:
         """Record that a thread item's tool start was already projected."""
         item_id = self._item_lifecycle_id(item)
+        signature = self._item_lifecycle_signature(item_type, item)
         if item_id is not None:
-            scope.started_item_ids.add(item_id)
+            scope.started_item_signatures[item_id] = signature
             return
-        key = (item_type, self._item_lifecycle_signature(item_type, item))
+        key = (item_type, signature)
         scope.started_unkeyed_item_counts[key] = scope.started_unkeyed_item_counts.get(key, 0) + 1
 
     def _consume_item_started(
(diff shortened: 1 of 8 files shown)


sumin220:
Addressed the first review in a0cff548. Blocker 1: `_resolve_item_completion_is_error` now scans every metadata source directly with failure precedence, so an outer `status=completed` can no longer shadow a nested `result.status=failed` — reproduced by a new regression test before the fix. Blocker 2: correlation state moved off the shared runtime into a per-stream `_CodexItemCorrelationScope` (fresh scope per streamed process, cleared on every `thread.started`), so a second thread's completed-only item synthesizes its own start again — also reproduced by a new regression test, and the inheriting runtimes (gemini/goose/grok/zcode/copilot) share the updated signature. Full suite 13160 passed, 12 skipped; ruff and mypy clean.

sumin220:
Addressed the second review in 8be3855c: cancelled/canceled/aborted/interrupted/killed/timeout/timed_out now sit in the failure-status set, so failure precedence covers every non-success terminal state — an outer cancelled item with a stale nested completed status resolves to is_error=true and fails `_event_has_explicit_tool_success`. Reproduced by a regression test before the fix (verified by stashing the source change). Orchestrator suite 3247 passed; ruff and mypy clean.

sumin220:
Addressed the remaining second-review findings in 9b6797b3 (the cancellation blocker was already fixed in 8be3855c). The failure-verdict resolver now scans every exit-code alias the metadata merge accepts (exit_code/exitCode/returncode/return_code), so returncode=1 with status=completed resolves to is_error=true. web_search items extract item['query'] explicitly and the id-less lifecycle signature is the query alone, so status changes no longer synthesize a duplicate start — keyed and id-less pairs are both covered by regression tests, each verified failing against the unfixed source. Also formatted the test file that failed the Ruff format CI gate. Orchestrator suite 3250 passed; ruff check/format and mypy clean.

sumin220:
Completed the round-two coverage asks in b48507fd (tests only). The keyed web_search lifecycle pair is now exercised alongside the id-less pair — one exact query, one correlated start/result, no duplicate — and parametrized tests pin the remaining non-success terminal statuses (including casing/whitespace variants), the nested exit-code alias placement, and the conflicting-alias case, asserting the verdict stays consistent with the journaled exit_status. Orchestrator suite 3260 passed.

sumin220:
Addressed the third review in 7b99b244. Blocker 1: keyed completions are now deduplicated per correlation scope and the synthesized start is recorded in lifecycle state, so replaying one keyed item.completed emits nothing instead of a second start/result pair. Blocker 2: declined joins the failure statuses — alone or with a stale nested completed it resolves to is_error=true and fails the success gate. Warning: conflicting exit aliases persist one verdict-consistent value (the failing code overrides a stale zero), so journaled exit_status can no longer contradict is_error. Four regression tests reproduce the probes, each verified failing against the previous commit. Full suite 13177 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the fourth review in 9eaf6c94. Blocker 1: command_execution no longer claims success from lifecycle status alone — only a validated zero exit or an explicit success/ok flag counts, so completed with a missing or malformed exit code fails closed (other item types keep completed as a success signal). Blocker 2: an id-less synthesized start is no longer retained as pending, so consecutive identical completed-only items each emit a full start/result pair with no orphan results; keyed replay suppression is unchanged. Also removed the unreachable declined entry from the terminated-handle guard (a scripted-edit artifact) — declined remains a failure status for item verdicts. Both probes reproduced by regression tests verified failing against the previous commit. Full suite 13179 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the fifth review in 04cb29f1. Blocker 1: when the command verdict is unresolved, the emitted message no longer forwards a bare success-implying status (it moves to reported_status), so the in-memory evidence verifier and the journal gate now reach the same fail-closed conclusion — the exit_code="1" probe is pinned by a regression test asserting _runtime_message_has_success_signal is False for that message. Blocker 2: MCP error envelopes are failure signals with precedence (completed + error.message resolves to is_error=true with the reason as result text), and successful MCP results normalize result.content blocks into the journaled result text. Three regression tests reproduce the probes, each verified failing against the previous commit. Full suite 13182 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the sixth review in 3ab65d84: replayed keyed item.started events are now suppressed, mirroring the completion-side replay suppression, so exact correlation always sees one matching start per id — the start → replayed-start → completion sequence is pinned by a regression test verified failing against the previous commit. Id-less starts keep per-count pairing for legitimate repeated invocations. Also updated the PR body per the documentation suggestion: the summary now describes the current fail-closed behavior for malformed command exit codes. Full suite 13183 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the seventh review in 3d6ab8e1. Blocker 1: completion dedup keys on the id plus the terminal verdict, so an exact replay is suppressed but a conflicting same-id completion (a later failure after a success) stays visible and verification fails closed. Blocker 2: keyed pairing validates the stable tool-input signature, not just the id — a same-id completion for a different command synthesizes its own pair instead of inheriting an unrelated start. Warnings: thread.started clears correlation only when the thread identity actually changes (same-thread header replay no longer orphans an in-flight start), and native MCP items resolve tool identity from tool+server with name as a legacy fallback. Follow-ups also done: reported_status is persisted into the journaled result meta, and mapping/list structured_content is serialized into the result text. Six regression tests reproduce the probes, each verified failing against the previous commit. Full suite 13189 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the eighth review in 07a38723. Blocker 1: MCP correlation signatures now include normalized arguments, so a same-id completion whose arguments differ from the start no longer inherits it as evidence. Blocker 2: completion dedup fingerprints every evidence-bearing field (output/stdout/stderr/result_preview/exit_code/status plus the raw error envelope), so two same-verdict completions with different output both stay visible while a genuinely identical replay is still suppressed. Blocker 3: any present, meaningfully non-empty error envelope fails closed — including malformed non-null shapes like an int or list — while an explicitly empty/false envelope stays no-error. Three regression tests reproduce the probes (each verified failing against the previous commit), and adversarial variant probes (matching vs mismatched MCP args, empty vs malformed error envelopes, id-less signatures) all behave. Full suite 13192 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the ninth review in 27eabf2d. Blocker 1: file-change correlation signatures now pair each changed path with its mutation kind, so an add start no longer correlates with a delete completion for the same path/id. Blocker 2: completion dedup now fingerprints the entire completion item envelope rather than a cherry-picked field subset — any distinct same-id completion (a changed web_search action, non-text MCP result content, nested metadata, or a secondary exit alias) stays visible and only a genuinely identical replay is dropped. Also aligned the correlation-scope docstring with the thread-identity-change clearing behavior. Three regression tests reproduce the probes (each verified failing against the previous commit), and adversarial variant probes (matching kinds, nested metadata, structured MCP content, item-level kind, identical replay) all behave. Full suite 13195 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the tenth review in a5693035. Blocker 1: command correlation now includes the normalized working directory and persists it in the tool input, so a same-id completion in a different cwd no longer inherits an unrelated start. Blocker 2: file-change signatures serialize the complete normalized change operation (mutation kind of any shape, patch content, move destination), so structured update/delete and differing diffs no longer collapse to one identity. Blocker 3: MCP result normalization preserves supported content blocks (text/image/resource) and structured data through the tool_result content contract instead of hardcoding an empty list, with the internal carrier key stripped from message data. Four regression tests reproduce the probes (each verified failing against the previous commit), plus adversarial variant probes. Full suite 13199 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the eleventh review in abd54cbd. Blocker: MCP result content blocks are normalized to the shared projection contract (flat text/data/mime_type/uri) — camelCase mimeType flattened, nested resource uri/text lifted, structured content carried in the projection-preserved data field — so image mime/data, resource uri, and structured payload survive project_runtime_message with their values intact (asserted after projection, not just by type name). Follow-up: each path in an id-less multi-file file_change now gets a stable 'filechange:{path}' synthetic correlation id instead of None, so the durable gate correlates each result with its own start; and nested result.error.message is extracted as the result text so journaled failures carry their reason. Also aligned the remaining default-scope docstring. Three regression tests plus adversarial probes; full suite 13202 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the twelfth review in a42376b5. Blocker 1: id-less item correlation now allocates a monotonic per-scope nonce queued FIFO per signature, so a start and its completion share one invocation id while two separate edits of the same path get distinct ids — exact correlation no longer sees multiple starts under one deterministic id (_item_tool_calls takes an explicit correlation key so keyed and id-less flow uniformly). Blocker 2: _file_change_signature now fingerprints every field _extract_paths turns into a call (top-level path/file_path/target_file) plus the full normalized changes, so a differing top-level path is not masked by identical changes. Blocker 3: MCP normalization copies resource.blob into the flat content data field and transfers result _meta into tool_result.meta, so binary payload and audit metadata survive projection. Four regression tests plus adversarial probes (keyed lifecycle/replay, id-less consecutive/paired/multifile uniqueness). Full suite 13206 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the thirteenth review in f0af7f26. Blocker 1: the verdict resolver now honors MCP's authoritative result.isError (and is_error) with failure precedence — {status: completed, result: {isError: true}} resolves to is_error=true and never becomes accepted success, while isError=false is an explicit success signal. Blocker 2: structured payload extraction accepts the canonical camelCase structuredContent wire field alongside structured_content, so standard MCP results survive projection (value asserted after projection). Also removed the always-true conditional assertion from the round-twelve test. Two regression tests plus adversarial probes (result/top-level isError, isError=false, camelCase structuredContent, dual-spelling precedence, isError-vs-success conflict). Full suite 13208 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the fourteenth review in a463eef9, and preemptively hardened the same forged-success contract family after an adversarial audit.

Round-fourteen blockers:
- Verdict resolution now follows a per-item-type authority contract: failure takes precedence anywhere, a present-but-malformed verdict field (non-bool isError/success/ok, non-int exit) poisons the success claim (fail closed), and success is claimed only from a signal authoritative for that item type. A command with only isError=false, or an MCP result.isError='false', no longer forges success.
- Opaque MCP result _meta is namespaced under tool_result.meta.mcp_meta so it can never populate the authoritative keys the deliver gate trusts (e.g. exit_status).
- A resumed handle's stale terminal runtime_event_type is neutralized on the result (and start) message so projection cannot inherit run.completed and forge success.
- Completed-only keyed items record their synthesized start so a later/replayed keyed item.started is suppressed.

Preemptive hardening (the audit found these before they shipped): exit_code/exit_status are demoted to audit-only keys whenever the resolver produced no verdict (is_error is None), so a leaked exit 0 on a non-command item or a malformed-voided command can no longer read as success in either the in-memory verifier or the deliver gate — is_error is now the single authoritative success channel end to end; failure tokens under alternate wire spellings (exit/statusCode/errorCode) are detected as failures only; and MCP identity recognizes the toolName wire spelling.

Eleven regression tests reproduce the review probes and the audited next-round edges (each verified failing against the previous commit). Full suite 13219 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the fifteenth review in 6719ee59. Blocker 1: a present-but-non-integer failure-code alias (statusCode='500', errorCode='E_FAIL', bool) now poisons the success claim instead of being ignored, so a simultaneous status=completed cannot promote it — cleanly zero/absent stays a no-op, nonzero int stays a failure. Blocker 2: command identity and persisted tool input include input.workdir alongside the other cwd aliases, so a different nested workdir no longer mispairs. Blocker 3: opaque MCP result _meta is recursively redacted before persistence — authorization/api_key/token/secret/password/credential-style keys become [REDACTED] (depth-bounded) — so credentials in untrusted audit data never reach the durable journal, while non-sensitive audit fields are preserved. Blocker 4: completion dedup stores a fixed-size SHA-256 digest of the envelope instead of the full serialized item, so multi-megabyte output cannot accumulate unbounded in scope state. Follow-up: handle neutralization uses the message-appropriate neutral event type, so a resumed tool start keeps tool.started. Five regression tests plus adversarial probes; full suite 13224 passed, 12 skipped; ruff check/format and mypy clean.

sumin220:
Addressed the sixteenth review in 235498b7. Blocker 1: the credential sanitizer now includes a bare 'token' pattern, covering token/refresh_token/idToken so OAuth tokens in MCP _meta are redacted before persistence. Blocker 2: a present non-string status is treated as malformed and poisons the success claim, so a command with exit_code=0 and status=7 fails closed. Follow-up: lifecycle signatures are scoped by tool type, so a reused id across different tool types no longer correlates (the mismatched completion synthesizes its own pair). P2: the MCP result summary now includes a resource block's nested text, so message content and tool_result_text are no longer empty for resource-only results. Refreshed the PR body test counts per the doc note. Four regression tests, each verified failing against the previous commit. Full suite 13228 passed, 12 skipped; ruff check/format and mypy clean.

Q00:
@sumin220 — thank you for this, and for the explicit sequencing question in the PR description. That question is now answered, and the answer is written into RFC #1681 rather than left as a one-off judgement call.

## Gate ruling: outside the gate

You asked whether the adapter conversion contract falls inside the #1681 gate, and offered to hold. It does not, and you should not have to hold.

RFC #1681 now carries a written boundary:

- **Outside the gate** — a single runtime adapter's own event-conversion contract, where the consumers (`runtime_message_projection.py`, `leaf_dispatcher.py`, `execution_event_emitter.py`, `runner.py`, `deliver_gate.py`) are untouched.
- **Inside the gate** — changing shared projection/emitter/dispatcher semantics for all runtimes at once, or introducing a new event type or acceptance authority.

Your PR is squarely in the first category, and you said so yourself in the "Sequencing vs RFC #1681" section — that framing was correct. The rationale now recorded in the RFC is that a runtime which mistranslates its own vendor's events is emitting false evidence *today*; holding that fix preserves a broken foundation, not a stable one.

## What was independently verified before merge

Beyond the green CI and the approval at head, one thing was checked specifically because CI could not have caught it:

Your change adds an `item_scope` keyword to `CodexCliRuntime._convert_event`, and five subclasses were updated to match (`GrokCliRuntime`, `CopilotCliRuntime`, `GooseCliRuntime`, `ZcodeCLIRuntime`, `GeminiCLIRuntime`). A sixth file, `opencode_runtime.py`, also defines `_convert_event` and is **not** in your diff. Since `test_opencode_*` are known pre-existing failures in this repo, a missed subclass there would have been an unguarded runtime `TypeError`.

It is safe: `OpenCodeRuntime` is a standalone class that does not inherit from `CodexCliRuntime`, and its `_convert_event` takes an entirely different signature (`context: OpenCodeEventContext`). Your subclass coverage is complete. Recording it here so nobody has to re-derive it.

## Two known gaps this does not close

Neither is caused by your change, and neither should have been bundled into it. Both are pre-existing and affect every runtime, so both are being tracked separately:

1. **Inflated tool counts.** `workflow_state.py:704` (`process_runtime_message`) passes `projected.tool_name` for tool *and* tool_result messages, and `process_message` at :682 does `if tool_name: tool_calls_count += 1`. Your correlated result messages carry `tool_name`, so one command counts as two. Claude and the other runtimes already have this today.
2. **Failed tools render green.** `emit_atomic_tool_completed` never sets `success` in the event payload, and `tui/events.py:1000` does `success=data.get("success", True)`. For Codex this changes "no completions shown at all" into "failures shown as successes", so it is a real follow-up — just not a Codex-scoped one.

## Sequencing with #1731

@berkshirehathaways independently fixed the same issue in #1731. Its adapter half is superseded by yours; its consumer-side half (4 files that make your new evidence actually reachable by the verifier) is being carried forward as its own slice on top of this merge. Details in #1731.

Merging now. Note that #1690 and #1724 stay open until a real Codex run shows `execution.tool.completed > 0` and the false `FABRICATION_SUSPECTED` rejections stop — unit tests cannot demonstrate that.

Issue #1741, opened 2026-07-25
tool_calls_count double-counts every tool: results carry tool_name and increment the counter


Q00:
## Summary

`tool_calls_count` counts every message that carries a `tool_name`, including tool *results*. One tool invocation therefore reports as two. This affects every runtime, not one adapter.

## Mechanism

`WorkflowStateTracker.process_runtime_message` passes the projected tool name regardless of whether the message is a call or a result:

```python
# src/ouroboros/orchestrator/workflow_state.py:704
def process_runtime_message(self, message: AgentMessage) -> None:
    projected = project_runtime_message(message)
    message_data = {**message.data, **projected.runtime_metadata}
    self.process_message(
        projected.content,
        message_type=projected.message_type,
        tool_name=projected.tool_name,      # <-- set for tool AND tool_result
        is_input=message.type == "user",
        message_data=message_data,
    )
```

and the counter keys only on truthiness:

```python
# src/ouroboros/orchestrator/workflow_state.py:682
if tool_name:
    self._state.tool_calls_count += 1
    self._state.last_tool = tool_name
    self._update_activity_from_tool(tool_name, content)
```

`projected.tool_name` is populated for results by construction — `is_tool_result` is defined as `message_type == "tool_result" and self.tool_name is not None` (`runtime_message_projection.py:72`). Fed from `runner.py:6567` and `runner.py:8112` for every runtime message.

## Impact

`tool_calls_count` is roughly 2× the real invocation count wherever a runtime emits correlated start/result pairs. Non-parallel workflow progress reports inflated tool activity.

This is **pre-existing and not Codex-specific** — Claude and the other runtimes already emit tool results carrying `tool_name`. #1732 makes Codex join the same behaviour; it does not introduce it.

## Why this is filed rather than fixed inline

The obvious fix (`if tool_name and message_type == "tool"`) changes reported counts for **every** runtime, so it is a cross-cutting behaviour change that deserves its own review and its own regression coverage. It was deliberately kept out of #1732 and #1740, both of which are Codex-scoped.

Note the interaction: `tool_name` is also used for `last_tool` and `_update_activity_from_tool`. Simply passing `None` for results would lose activity tracking on completions, so the guard belongs at the counter, not at the call site.

## Suggested fix

Count only projected invocations, keep `last_tool`/activity updates as they are, and add coverage asserting one correlated start/result pair increments the counter exactly once — for at least one Codex and one Claude message stream.

## Origin

[...]

PR #1742, opened 2026-07-25
fix(tui): fail closed on reported tool failures instead of rendering them green

Q00:
## Summary

`execution.tool.completed` was converted to a TUI message with `success=data.get("success", True)` — but **no producer ever sets `success`**. `ExecutionEventEmitter.emit_atomic_tool_completed` (`execution_event_emitter.py:986`, the only producer of this event type) builds its payload from the runtime identity plus `serialize_runtime_message_metadata` output, and neither contributes that key. The default always won, so every failed tool displayed as a success.

## Reproduced against the real path

Not a hand-built payload — the actual adapter and projection:

```
Codex command_execution, exit_code=1
  -> event data carries is_error=True
  -> create_message_from_event(...) -> ToolCallCompleted(success=True)
```

## Fix

Resolution order in `_resolve_tool_completion_success`:

1. explicit boolean `success` wins, if a producer ever sets one
2. `is_error_invalid` (a malformed `is_error` claim) → **failure**, rather than silently ignored
3. `is_error` boolean → `not is_error`
4. otherwise → optimistic

`is_error` is the authoritative verdict channel, and step 2 matches the fail-closed contract the runtime adapters already use.

### Why step 4 is not fail-closed

Deliberate, and worth stating plainly. The serialized metadata forwards neither `tool_result` nor `reported_status` — verified, not assumed — so a verdict-less completion carries no signal to fail closed on. Flipping every signal-less runtime to a failed display would be a louder wrong answer than the one being fixed.

Widening that payload so the unknown case is representable is a **shared projection change**, which lands inside the RFC #1681 gate boundary. Out of scope here by that rule.

## Scope

Cross-runtime and pre-existing — **not** caused by #1732. But Codex previously emitted no completion events at all (`started: 158 / completed: 0`, #1724), so for that runtime the adapter fix turns "nothing shown" into "failures shown as successes". That is why it is worth fixing now rather than later.

Kept out of #1732 and #1740, both of which are Codex-scoped, because this changes display behaviour for every runtime.

## Verification

- 7 new regression tests; **confirmed failing before the fix** by reverting the one-line call site — 3 of the 7 fail (`reported_error_renders_as_failure`, `malformed_error_claim_fails_closed`, and the real-Codex `exit_code=1` case), the other 4 pin behaviour that must not change
- Two of the tests drive the real `CodexCliRuntime` conversion + `serialize_runtime_message_metadata` end to end, parameterized over exit 0 and exit 1
- `tests/unit/tui` — 217 passed
- `ruff check` / `ruff format --check` clean across `src/` and `tests/`; `mypy` clean

## Related

- #1741 — `tool_calls_count` double-counting, the other cross-runtime defect found in the same review, filed rather than bundled
- #1740 — Codex completion consumer slice
- #1732 — Codex adapter half (merged)

### DIFF
--- a/src/ouroboros/tui/events.py
+++ b/src/ouroboros/tui/events.py
@@ -793,6 +793,41 @@ def _subtask_root_ac_index(data: dict[str, Any]) -> int:
     return ac_index if ac_index is not None else legacy_ac_index or 0
 
 
+def _resolve_tool_completion_success(data: dict[str, Any]) -> bool:
+    """Resolve tool-completion success without laundering a reported failure.
+
+    No producer sets ``success`` on ``execution.tool.completed``:
+    ``ExecutionEventEmitter.emit_atomic_tool_completed`` builds its payload from
+    the runtime identity plus ``serialize_runtime_message_metadata`` output, and
+    neither contributes that key. A plain ``data.get("success", True)`` therefore
+    rendered every failed tool as succeeded — a Codex command exiting 1 arrives
+    carrying ``is_error=True`` and was still displayed as a success.
+
+    ``is_error`` is the authoritative verdict channel, and a malformed claim is
+    treated as failure rather than silently ignored, matching the fail-closed
+    contract the runtime adapters use.
+
+    A completion carrying no verdict at all stays optimistic: the serialized
+    metadata forwards neither ``tool_result`` nor ``reported_status``, so there
+    is no signal to fail closed on, and flipping every signal-less runtime to a
+    failed display would be a louder wrong answer than the one being fixed.
+    Widening the payload is a shared-projection change and is deliberately out
+    of scope here.
+    """
+    raw_success = data.get("success")
+    if isinstance(raw_success, bool):
+        return raw_success
+
+    if data.get("is_error_invalid") is True:
+        return False
+
+    is_error = data.get("is_error")
+    if isinstance(is_error, bool):
+        return not is_error
+
+    return True
+
+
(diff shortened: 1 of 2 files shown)


Q00:
Blocker fixed in `d05715b3c`.

You were right, and my earlier reasoning was wrong in a way worth naming: I probed `serialize_runtime_message_metadata` directly, without its `tool_result` keyword argument, saw no `tool_result` key, and concluded the nested check was dead code. That is not the production path — the live path is `ExecutionEventEmitter.runtime_event_metadata` → `project_runtime_message(...).runtime_metadata`, which does persist it. Generalizing from a single Codex probe to every runtime was the error.

Reproduced your finding through the real path before fixing:

```
OpenCode-style nested failure (no top-level is_error)
  top_is_error=<absent>  nested=True  -> TUI success=True
```

Resolution now walks both envelopes in priority order — top level first, then `tool_result` — and a malformed verdict in either is treated as failure rather than as an absent signal.

Five new tests: nested failure, nested success, nested malformed claim, top-level precedence over a conflicting nested value, and an end-to-end case that builds the payload through the production metadata path and asserts (`assert "is_error" not in metadata`) that the envelope genuinely carries no top-level verdict — so the test cannot silently start passing for the wrong reason.

Confirmed the three nested cases fail against the previous top-level-only resolution and pass after. `tests/unit/tui` 222 passed; ruff/format/mypy clean.

The optimistic fallback for a completion carrying no verdict in *either* envelope is retained deliberately, as described in the PR body.

PR #1703, opened 2026-07-21
feat(interview): data_context advisory lane — MCP-backed data evidence (#1671)

cohemm:
## Summary

Implements the `data_context` advisory lane proposed and refined in discussion #1671, plus the wiring and contract work that live verification showed was required to make it actually function end-to-end.

### 1. `data_context` lane as a read-only proposer (5e481358)

Sixth question-advisory lane (`capability: call_mcp`, `required: false`), following the MVP shape converged on in the discussion:

- **Relevance decided from the question text before any tool call** (metered-source cost protection).
- **Direct execution only for local, free, read-only lookups**; metered / external / side-effect-ambiguous sources come back as `proposed_queries` for the parent session to run after user confirmation.
- Machine-readable `data_policy` block (read_only, aggregate_only, forbidden_operation_patterns, evidence policy with aggregates-only / no-raw-rows / PII-scrub caps) shipped on the payload; the prompt is the fallback, not the mechanism.
- Error-shaped tool output (e.g. HTTP 200 with an `{error}` envelope) yields a no-evidence finding, never evidence.

**Completion-gate fix** (the discussion's MVP prerequisite): `FanoutRecord` now tracks `required_keys`; only `required: true` lanes gate `submit_fanout_results` completion. A host that cannot run an optional lane no longer pins the fanout at `status="partial"` forever — missing optional lanes are reported as `missing_optional_keys` on the complete outcome. Legacy records without the field keep the all-keys gate. Contract stays `interview_question_advisory_fanout.v1` (extended in place) with explicit `lane_compatibility_rules`: unknown `lane_id` → generic dispatch or skip; unsupported `capability` → dispatch and return the no-op finding (the no-op IS the completion signal).

### 2. Fanout re-entry wired into the stdio server (98c0872a)

Live verification surfaced a pre-existing gap: `create_ouroboros_server` never registered `SubmitFanoutResultsHandler` and never injected a `FanoutRegistry` into the interview/lateral handlers — so on the primary MCP surface no `question_advisory_fanout_id` was stamped and the re-entry contract documented in `skills/interview/SKILL.md` was uncallable. One shared registry (rooted at `state_dir/fanout`) is now injected into both `InterviewHandler` instances and `LateralThinkHandler`, and the re-entry tool is registered.

### 3. `data_evidence_answer.v1` — informed-consent answer contract (33f64377)

[...]

PR #1749, opened 2026-07-26
feat(interview): data_context advisory lane — MCP-backed data evidence

cohemm:
Replaces #1703. Same feature, rebased onto current `main`, squashed to two
commits, with three commits that landed on the old branch from a concurrent
process dropped. Content is byte-identical to that branch's tree at
`11b1c536`, verified by tree hash.

## Purpose

An interview question whose honest answer is a number currently gets a guess.
The user either answers from memory or leaves the session to look it up, and a
guessed number becomes the requirement's evidence. This lane fetches the
number through the host's own data MCP tools and puts it beside the question.

The governing constraint is that **data is material for the user's judgment
and never the answer.** The user reads the evidence and answers in their own
words, and only that answer carries requirement authority.

## What it does

`data_context` joins the existing question-advisory fan-out and follows the
same lane contract as `code_context` and `web_context`: dispatched
unconditionally, relevance decided by the child from the question text alone
before any tool call, and an irrelevant question answered with a no-op
finding. The unconditional dispatch is deliberate — the no-op finding is the
completion signal, so conditional dispatch would make the fan-out
uncompletable.

What the lane adds on top of that contract is the execute/propose split. Only
obviously local, free, read-only lookups are executed by the child. Anything
metered, external, or side-effect-ambiguous comes back as a structured
`read_request` for the parent session to run after the user confirms it.
Mutating operations are never run.

Evidence carries no free-text field. Grouping and dimension keys must end in a
declared category head, and anything that cannot be expressed as an aggregate
— a row list, a name, an identifier, an error message — is a no-evidence
finding rather than evidence. Error-shaped tool output is not evidence either:
every item must declare a succeeded execution.

Supporting work: durable fan-out registration and re-entry through the primary
MCP server, owner-only atomic persistence with a retention bound, terminal
replay, and a provenance boundary that keeps an adopted observation from being
promoted into a requirement on any generation path.

## What is new relative to #1703

`initial_context` had no single provenance authority.

`InterviewRound` decides provenance once at ingestion and carries it as a
typed field. Its own comment explains why: re-classifying at every consumer is
how the extraction entrances kept multiplying. That treatment was never
applied to `initial_context`, which enters at session creation rather than
through `record_response`.

The result was two authorities disagreeing on one string.
`prompt_safe_initial_context_with_provenance` — the function consumers ask —
returned a hardcoded `human` for any context short enough to use as-is, while
`requirement_distillation` classified the same string itself and got
`data_fact`. A short `[from-data]` context was therefore human to one caller
and an observation to the next, and no test pinned either answer.

`initial_context_provenance` is now the single classification point, derived
rather than stored: the marker is in-band, so a pure function of the text
cannot drift from it and no caller can forge it.

Rounds get the same treatment from the other direction. The MCP handlers build
`InterviewRound` directly, bypassing the door, and each site had to remember to
classify — the same "caller must remember" shape that let a decorated answer be
stamped human. The field now classifies itself when the constructor did not
decide. `model_fields_set` is what makes that safe: an explicit `human` (a user
decision that quotes an observation) stays human, an explicit `data_fact` (the
PM decoration path) stays `data_fact`, and a round persisted before the field
existed is classified on load.

## Success criteria

- The two authorities report the same class for a short `[from-data]` initial
  context. Pinned across every marker class rather than the one marker a probe
  happens to use.
- The oversized-context path still defers to the substitute round's typed
  field, so the agreement above is not won by making the short path
  authoritative everywhere.
- A round built without an explicit provenance classifies itself correctly, and
  an explicit value wins in both directions.
- `tests/unit`: 13,646 passed. Two failures are pre-existing and unrelated —
  both build a wheel via `uv build`, which cannot reach the network in this
  environment, and both fail identically with this branch's changes stashed.


shaun0927:
Closing this combined PR after SSOT and exact-head review.

The latest head fixes the credential-scrubbing and zero-round transcript findings, but the merge-blocking authority defect remains: `data_context` lets an untrusted child discover and directly invoke runtime MCP tools based on its own judgment that they are local/free/read-only. There is no trusted pre-invocation allowlist or capability restriction on the dispatched payload. Post-execution schema validation cannot enforce the stated “never mutate / do not execute ambiguous tools” invariant after a call has already happened.

This PR also has no accepted narrow issue/RFC authority, spans 46 files and roughly 20k added lines across interview provenance, MCP execution, persistence/replay, plugin generation, and `auto/`, and the required auto performance-evidence gate is failing. That scope is too broad to repair safely in-place.

Please split any still-needed work into independently reviewable slices:
1. provenance-only fixes with a concrete issue and existing behavior locked by tests;
2. persistence scrubber hardening as its own boundary fix;
3. a data-evidence lane that is proposal-only until the host enforces a trusted read-only tool roster before invocation.

A future execution-capable lane needs an accepted design issue, explicit capability ownership, pre-invocation authorization tests, and the required R-run/performance evidence before implementation.

Issue #1755, opened 2026-07-26
Observation-marked answers reach requirement extraction and become Seed requirements


cohemm:
## Problem

The interview advertises answer prefixes to the caller. `InterviewEngine`
builds this into the question-generator header on every turn:

```
Answer prefixes the caller may use:
- [from-code]: Existing codebase state (factual, read from files).
- [from-user]: Human decisions/judgments.
- [from-research]: Externally researched information (API docs, pricing, compatibility).
```

The distinction is real and the prompt states it: `[from-user]` is a human
decision, the other two are adopted external facts. Nothing enforces it.

`SeedGenerator._build_interview_context` walks `state.rounds` and appends every
answer verbatim, whatever prefix it leads with, and hands the result to the
extraction LLM. `core/requirement_candidate.py` contains no provenance
classification at all. So an answer the caller explicitly labelled as an
observation is indistinguishable, downstream, from a decision the user made.

The consequence is a durable artifact built from a point-in-time fact. An
extractor paraphrases: `[from-research] Competitor X ships SSO in their $99
tier` comes back as the Seed acceptance criterion "Support SSO in the paid
tier". A paraphrase cannot be matched against the marker afterwards, so the
observation has become a requirement and nothing downstream can tell.

This is independent of any data lane. `[from-code]` and `[from-research]` are
shipped conventions today, and both reach requirement extraction today.

## Desired behavior

An adopted observation informs the user's judgment and never becomes a
requirement on its own. Whatever requirement the user derived from it appears
in their own answer, in their own words, and that is what extraction reads as
a requirement.

An observation reaching a later *question* is not the defect — it is the point.
An observation is collected in order to sharpen the next question, and that is
where it is meant to arrive. Requirement extraction is the destination it was
never collected for, and that is the whole of this issue.

So the same content plays two roles depending on where it sits, and the rule is
per-role, not per-string: in an answer slot it carries requirement authority and
has to come out; in a question slot it is interpretive context and stays.

## Proposed solution

Decide provenance once, where the answer enters, and carry it as a typed field
rather than re-reading the marker at every consumer. The marker stays as the
caller-facing convention and the display projection.

Withhold observation content at the extraction *input* — the point where a
paraphrase is still preventable. A withheld answer contributes a fixed note
instead of its content. Question lines are left exactly as they render today.

One walk, read by every requirement-producing consumer, rather than the rule
being re-applied per surface.

## Non-goals

[...]

Issue #1754, opened 2026-07-26
RFC: data_context advisory lane — data evidence beside the question


cohemm:
## Revision — 2026-08-02: the lane executes

This RFC was written for a proposal-only lane. That decision is superseded. The
child now discovers what data stores are described to it, takes the measurement,
and returns the aggregate stamped with when it ran. The user-confirmation gate
is removed; the lane calls tools like every other advisory lane.

**Why.** The first line of this document names what it set out to stop: an
interview question whose honest answer is a number got a guess, and the guess
became the Seed's evidence. Forbidding execution was the instrument reached for.
It was the wrong one — it made the guess likelier, not rarer.

Observed on a live interview: `architecture_implications`, under no constraint,
queried a metrics store fifteen times and produced the run's most decisive
finding. `data_context` obeyed this RFC, proposed, and stopped. It cost a user
round trip and returned less. The gate was per-lane, so it was neither
consistent nor effective — the lane holding the line was the only one asked to.

**The standing consent** is the one the sibling lanes already run on: the user
registered the MCP server, and registering it is the willingness to have it
called.

**What survives, because none of it depended on the ban.** A number is material
for the user's judgment and never the interview answer — the user answers in
their own words, there is no `[from-data]` answer path, `[from-data]` stays
withheld from requirement extraction, and the record keeps no child-authored
content. That was always the guard; the missing value field was scaffolding
around it. Two closed answer states, aggregates only, categorical grouping, and
bounded values all still hold. `observed_at` is new and required: a measurement
is point-in-time and a Seed is not. *(Superseded the same day — see the second
revision below: the field is removed.)*

**Accepted knowingly.** `operation: "read"` is a label on the request, not a
constraint on the tool, and MCP carries no cost or mutation metadata. A metered
or mutating call can arrive named like a free local read, and this lane will
make it. The confirmation gate that justified could not price what it was
gating, and a disclaimer on every read is a thing users learn to click through.

**A second accepted risk, of the same shape.** A measurement can cross two
issuances of the same question text: `question_identity` is a digest of the
question, so a result taken for one issuance satisfies the other. Under the
proposal-only design this carried nothing time-sensitive; now it does.

It is accepted rather than closed. The number measured *the same question*, its
age travels with it — `observed_at` is required and the host is told to say when
— and the decision it could corrupt is already guarded: a measurement arriving
after the user has answered is dropped rather than shown. What is genuinely lost
is the record that this round's lane did not run. *(The grounds simplified with
[...]

PR #1757, opened 2026-07-26
fix(core): interview artifacts are owner-only from the moment they exist

### LINKED ISSUE #1756: Interview transcripts, Seeds, and Auto state are written world-readable
## Problem

Every artifact Ouroboros writes for an interview is created at the umask
default — typically `0644`, world-readable — and stays that way for its whole
lifetime.

Those artifacts are not incidental. The interview transcript holds what the
user typed, verbatim, and lives under `~/.ouroboros`. The Seed holds the
requirements distilled from it. Auto pipeline state holds the same answers for
as long as the session record exists. On a shared machine, a CI box, or a
container with more than one account, all of it is readable by anyone.

Eight writers persist this class of artifact and none of them narrows the mode:

| Writer | Artifact |
|---|---|
| `InterviewEngine.save_state` | interview transcript (stdio path) |
| `_plugin_save_state` | interview transcript (plugin path) |
| `SeedGenerator.save_seed` / `save_seed_sync` | Seed |
| `PMInterviewEngine.save_pm_seed` | PM Seed |
| `cli/commands/pm.py:_save_cli_pm_meta` | PM metadata (CLI path) |
| `mcp/tools/pm_handler.py:_save_pm_meta` | PM metadata (MCP path) |
| `FanoutRegistry.register` | fan-out record (`synthesizer_input` verbatim) |
| `auto/adapters.py`, `auto/state.py` | Auto-generated Seed, Auto pipeline state |

PM metadata holds `pending_reframe`, `deferred_items`, `decide_later_items` and
question classifications — interview content by another name. A fan-out record
holds the producer's request verbatim: the code-investigation question, the
lateral persona panel entries.

`interview.py` has a local `_atomic_write_text` that is careful about atomicity
and about *preserving* an existing mode, but a file it creates starts at the
umask default and preservation then keeps it there. The plugin path and the
Auto writers call `write_text` directly and get no atomicity either.

## Desired behavior

An artifact holding interview content is readable by its owner only, from the
moment it is created, whichever writer produced it.

Atomicity and durability reporting stay as they are today: a failed write must
leave the previous file intact, and an unconfirmed directory flush must be
visible rather than reported as success.

## Proposed solution

One primitive — `core/owner_only.py` — that creates at `0600`, writes through a
temp file plus `os.replace`, fsyncs the parent directory, and returns whether
durability was confirmed. Every writer above calls it.

An existing file created before this change is narrowed on the next write
rather than left at its inherited mode, since preserving `0644` would mean the
fix never reaches an installed user.

Directories are deliberately not re-permissioned. `seeds_dir` and Auto's `root`
may be caller-supplied and possibly shared, and narrowing a directory this
package does not own is not its call. `state_dir`, which Ouroboros does own, is
created secured.

## Non-goals

- The `data_context` lane (#1754) and interview provenance (#1755). This slice
  is only about file modes and is independent of both.
- `cli/commands/setup.py`, which has its own same-named helper for config files
  that are meant to be readable.

## Acceptance criteria

- Every writer in the table produces `0600`, including on the plugin transport
  and both Auto paths.
- A file that already exists at `0644` is narrowed on the next write.
- A failed `os.replace` leaves the previous file intact and its content
  unchanged.
- An unconfirmed directory flush is logged, and the caller's success return is
  unchanged.
- Caller-supplied directories keep their own permissions.

cohemm:
Closes #1756.

First of three slices carved out of the closed combined PR #1749. This one is
independent of the other two: it is about file modes only, and touches neither
the advisory lane (#1754) nor interview provenance (#1755).

## Purpose

An interview transcript holds what the user typed, verbatim. A Seed holds what
was distilled from it. Auto pipeline state holds the same answers for as long
as the session record exists. All three are written at the umask default —
typically `0644` — and stay world-readable for their whole lifetime.

## What changed

`core/owner_only.py` is one primitive: create at `0600`, temp file plus
`os.replace`, fsync the parent directory, return whether durability was
confirmed. Five writers now call it.

| Writer | Was |
|---|---|
| `InterviewEngine.save_state` | `_atomic_write_text`, mode preserved |
| `_plugin_save_state` | `write_text`, no atomicity |
| `SeedGenerator.save_seed` / `save_seed_sync` | `write_text` |
| `PMInterviewEngine.save_pm_seed` | `_atomic_write_text` |
| `auto/adapters.py`, `auto/state.py` | `write_text` |

`interview.py`'s local `_atomic_write_text` is removed. It was careful about
atomicity and about *preserving* an existing mode, but a file it creates starts
at the umask default and preservation then keeps it there.
`cli/commands/setup.py` keeps its own same-named helper — those config files are
meant to be readable.

An existing `0644` file is narrowed on the next write rather than preserved.
Preserving it would mean the fix never reaches anyone already installed.

Directories are deliberately left alone. `seeds_dir` and Auto's `root` may be
caller-supplied and shared, and narrowing a directory this package does not own
is not its call. `state_dir`, which Ouroboros does own, is created secured.

## Success criteria

- Every writer above produces `0600`, including the plugin transport and both
  Auto paths.
- A file that already exists at `0644` is narrowed on the next write.
- A failed `os.replace` leaves the previous file intact with its content
  unchanged; an unconfirmed directory flush is logged and the caller's success
  return is unchanged.
- Caller-supplied directories keep their own permissions.
- `tests/unit`: 13,281 passed. Three failures are pre-existing and unrelated —
  two build a wheel via `uv build`, which cannot reach the network in this
  environment, and one asserts on orphaned-session cleanup; all three fail
  identically with this branch's changes stashed.

Two existing tests pinned the old contract and are updated rather than deleted:
mode preservation becomes mode narrowing, while the atomicity and fd-hygiene
assertions around them are kept and now cover the new writer.

## R-run comparison

|---|---|---|---|

Budget compliance: [x] N/A (storage-only auto persistence change; no R-run behavior change)

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -12,9 +12,11 @@ from pathlib import Path
 from typing import Any
 from uuid import uuid4
 
+import structlog
 import yaml
 
 from ouroboros.auto.interview_driver import InterviewBackend, InterviewTurn
+from ouroboros.core.owner_only import write_owner_only
 from ouroboros.core.requirement_candidate import RequirementDistillation
 from ouroboros.core.seed import Seed, ac_texts
 from ouroboros.mcp.errors import MCPServerError
@@ -35,6 +37,8 @@ from ouroboros.orchestrator.session import SessionRepository, SessionStatus
 from ouroboros.persistence.event_store import EventStore
 from ouroboros.providers.base import CompletionConfig, LLMAdapter, Message, MessageRole
 from ouroboros.resilience.lateral import ThinkingPersona
+
+log = structlog.get_logger(__name__)
 
 _SAFE_SEED_ID_FILENAME_CHARS = frozenset(
     "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_-"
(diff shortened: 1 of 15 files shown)


shaun0927 (approved):
Approved at current head `6490240b8`. The shared owner-only writer now covers interview, Seed, Auto, plugin, PM Seed, and both PM metadata paths; caller-owned directory permissions remain unchanged, durability uncertainty is surfaced, the storage-only R-run declaration passes, and the full live CI matrix is green.

Issue #1729, opened 2026-07-24
[Feature] Extend the CONSTRAINTS JSON-array extraction contract (#1714) to the remaining pipe-delimited seed fields


sumin220:
### Problem

#1696/#1714 fixed CONSTRAINTS: extraction now demands a single-line JSON array of strings, validates it strictly at parse time (malformed output triggers the extraction retry), and keeps a lenient pipe-split fallback for stored legacy data. Every other list-valued field in the extraction format still uses the raw pipe/colon-packed encoding and silently corrupts when a value contains a literal `|` or, for the colon-triple fields, a literal `:` in the description.

All of the following are real outputs from `SeedGenerator._build_seed` on current main:

- `ONTOLOGY_FIELDS: retry_mode:string:Retry policy, one of fixed|exponential|none` parses to a single field whose description is truncated to `Retry policy, one of fixed`; the fragments `exponential` and `none` are silently dropped.
- `EVALUATION_PRINCIPLES: correctness:Outputs match spec: no regressions allowed:0.9` truncates the description to `Outputs match spec` and silently discards the intended weight 0.9 (the float parse of ` no regressions allowed` fails, so weight falls back to 1.0). With a pipe instead — `safety:Never run rm -rf | sudo in verify commands:1.0` — the parser fabricates a second principle named `sudo in verify commands` with description `1.0`.
- `EXIT_CONDITIONS: timeout:Stop after 30:00 minutes of wall clock:elapsed time exceeded` yields description `Stop after 30` and criteria `00 minutes of wall clock:elapsed time exceeded`.
- `CONTEXT_REFERENCES: /repo/api:primary:Handles auth | rate limiting via middleware | C:\legacy\svc:reference:Old service` truncates the first summary to `Handles auth`, silently drops the middle fragment, and shreds the drive-letter path into `path='C', role='\legacy\svc', summary='reference:Old service'`.
- `EXISTING_PATTERNS: Use Result.ok() | Result.err() unions for fallible calls | Repository pattern` splits one pattern into two nonsense entries.
- `EXISTING_DEPENDENCIES: typer (CLI | completion extras) | structlog` parses to `('typer (CLI', 'completion extras)', 'structlog')`.

Unlike CONSTRAINTS, none of this is caught at extraction time: `_parse_extraction_response` accepts these lines verbatim, so the retry path never fires and the corruption lands in the immutable Seed — the ground truth that evaluation, exit gating, and brownfield execution all read from.

### Why now

[...]

PR #1762, opened 2026-07-26
feat(bigbang): extract EVALUATION_PRINCIPLES and EXIT_CONDITIONS as JSON object arrays (#1729 slice 2)

sumin220:
## Summary

Slice 2 of #1729: EVALUATION_PRINCIPLES and EXIT_CONDITIONS move to single-line JSON object arrays at extraction time, following the same strict-extraction / lenient-legacy contract merged in #1714 and #1730.

Today a description like `Keep a 1:1 request:response mapping` breaks the colon split, and a pipe inside any field silently truncates the entry. With this change:

| Lane | Behavior |
|---|---|
| Extraction (strict) | Must be a valid JSON array of `{"name", "description", "weight"/"criteria"}` objects; anything else raises so the existing reformat-retry path asks the LLM to resend. No fallback surface. |
| Stored requirements (lenient) | Bracket-led valid JSON arrays parse to models; everything else keeps the historical `name:description:...` pipe split; pre-parsed model tuples pass through; never raises. |

Details:
- `criteria` (prompt spelling) and `evaluation_criteria` (model field name) are both accepted for exit conditions.
- Weight must be a JSON number in strict mode (bools rejected, malformed triggers retry) and is clamped to [0.0, 1.0] exactly as the legacy path always did.
- The extraction prompt, the reformat template, and the seed-architect agent template now request the object-array format, and the internal requirement-distillation producer emits model tuples instead of legacy colon strings — no in-repo producer emits the legacy format anymore.
- Also closes the low-priority follow-up from the #1730 review: strict string-array parsing now rejects pre-decoded non-string entries instead of coercing them.

## Sequencing vs #1729 needs-design

The bot labeled #1729 needs-design for the multi-PR rollout. Slice 1 (#1730) has since merged, and this slice extends that same merged parser boundary to the next two fields without touching shared schemas or consumers outside `bigbang/`. Slices 3 (ONTOLOGY_FIELDS) and 4 (CONTEXT_REFERENCES) stay separate and sequential. If you'd rather settle the rollout design on #1729 first, happy to hold this PR — nothing else depends on it.

## Test plan

- [x] 13 new regression tests covering the strict/lenient boundary for both fields: delimiters inside data, criteria key alias, weight defaulting/clamping, malformed-entry rejection (non-objects, missing/blank fields, string/bool weights), legacy colon/pipe split preservation (including the criteria segment absorbing further colons), never-raise lenient fallback, pre-decoded pass-through, and the extraction retry integration
- [x] `uv run pytest tests/unit/bigbang -q` — 685 passed
- [x] `uv run pytest tests/ -q -n 4` — 13628 passed, 12 skipped

Refs #1729 (slice 2 of 4; #1730 was slice 1)

### DIFF
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -56,11 +56,11 @@ Field types should be one of: string, number, boolean, array, object
 
 ### 5. EVALUATION_PRINCIPLES
 Principles for evaluating output quality.
-Format: name:description:weight (pipe-separated, weight 0.0-1.0)
+Format: single-line JSON array of objects with "name", "description", and "weight" (0.0-1.0) so colons and pipes inside the text survive as data
 
 ### 6. EXIT_CONDITIONS
 Conditions that indicate the workflow should terminate.
-Format: name:description:criteria (pipe-separated)
+Format: single-line JSON array of objects with "name", "description", and "criteria"
 
 ### 7. BROWNFIELD CONTEXT (if applicable)
 If the interview mentions existing codebases, extract:
@@ -82,8 +82,8 @@ AC: <description> | verify: <command or NONE> | artifacts: <comma-list or NONE> 
 ONTOLOGY_NAME: <name>
 ONTOLOGY_DESCRIPTION: <description>
 ONTOLOGY_FIELDS: <name>:<type>:<description> | ...
-EVALUATION_PRINCIPLES: <name>:<description>:<weight> | ...
-EXIT_CONDITIONS: <name>:<description>:<criteria> | ...
+EVALUATION_PRINCIPLES: [{"name": "<name>", "description": "<description>", "weight": <0.0-1.0>}, ...]
+EXIT_CONDITIONS: [{"name": "<name>", "description": "<description>", "criteria": "<criteria>"}, ...]
 PROJECT_TYPE: greenfield|brownfield
 CONTEXT_REFERENCES: <path>:<role>:<summary> | ...
 EXISTING_PATTERNS: ["<pattern 1>", "<pattern 2>", ...]
--- a/src/ouroboros/bigbang/requirement_distillation.py
+++ b/src/ouroboros/bigbang/requirement_distillation.py
@@ -272,11 +272,18 @@ def apply_requirement_distillation(
         "ontology_description": "Only user-authorized interview requirements.",
         "ontology_fields": "",
         "evaluation_principles": (
-            "confirmed_requirements:Evaluate only promoted user-authorized requirements:1.0"
+            EvaluationPrinciple(
+                name="confirmed_requirements",
+                description="Evaluate only promoted user-authorized requirements",
+                weight=1.0,
+            ),
         ),
         "exit_conditions": (
-            "confirmed_requirements_met:All promoted requirements are satisfied:"
-            "Every promoted acceptance criterion passes"
+            ExitCondition(
+                name="confirmed_requirements_met",
+                description="All promoted requirements are satisfied",
+                evaluation_criteria="Every promoted acceptance criterion passes",
+            ),
         ),
         "project_type": "greenfield",
     }
--- a/src/ouroboros/bigbang/seed_generator.py
+++ b/src/ouroboros/bigbang/seed_generator.py
@@ -18,6 +18,7 @@ from pathlib import Path
 import re
 from typing import Any
 
+from pydantic import ValidationError as PydanticValidationError
 import structlog
 import yaml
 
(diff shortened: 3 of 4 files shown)


sumin220:
Addressed the first review in 56585d41. Non-finite weights (NaN/Infinity — accepted by Python's json.loads but not JSON numbers) are now rejected in strict mode so the retry path reformats them, and keep the historical 1.0 fallback in lenient mode. Arbitrary-precision integers are clamped in integer space before float conversion, so an oversized weight can no longer raise an OverflowError past the retry path and crash seed generation — the clamp result matches what any out-of-range number already gets. The legacy colon/pipe weight segment routes through the same helper. Five regression tests including both end-to-end paths (NaN triggers extraction retry; a 400-digit weight generates a seed with weight 1.0). Full suite 13633 passed, 12 skipped; ruff check/format and mypy clean.

shaun0927 (approved):
Approved at exact head 56585d414567ef3c1da745f3edd93dae73ded72d. Strict object-array extraction now rejects non-finite JSON constants, handles oversized numeric weights without escaping retry or lenient fallback, preserves legacy stored-data compatibility and aliases, and passes the complete Python 3.12/3.13/3.14, coverage, Ruff, MyPy, TypeScript, boundary, envelope, and performance matrix. Independent exact-head review found no blocking issue.

shaun0927:
@ouroboros-agent Please rereview exact head `56585d414567ef3c1da745f3edd93dae73ded72d`. The prior request-changes review targets `4c9a4128504fd20757a93ed8ca7712914b503234`; exact-head CI is fully green and an independent current-head review found no issues.

shaun0927:
@ouroboros-agent rereview request

Repair pushed from exact requested head evidence:
- Before worktree creation: PR #1762 headRefOid was 56585d414567ef3c1da745f3edd93dae73ded72d.
- Before push: gh pr view #1762 headRefOid was still 56585d414567ef3c1da745f3edd93dae73ded72d.
- Before push: git ls-remote sumin220 refs/heads/feat/1729-slice2-object-arrays was still 56585d414567ef3c1da745f3edd93dae73ded72d.
- Pushed normally, no force/rebase/approval/dismissal/merge.
- New PR head: d35c1861314655ed87a984b5b012255fb798ae39.

Validation:
- PYTHONPATH=src python3.12 -m pytest -q tests/unit/bigbang/test_seed_generator.py -> 81 passed
- PYTHONPATH=src python3.12 -m pytest -q tests/unit/bigbang -> 697 passed
- PYTHONPATH=src python3.12 -m pytest -q tests/unit/core/test_seed.py tests/unit/bigbang/test_requirement_distillation.py tests/unit/mcp/tools/test_generate_seed_force_flag.py -> 92 passed
- ruff format --check . && ruff check . -> 1142 files already formatted; all checks passed
- PYTHONPATH=src pipx run mypy src/ouroboros/bigbang/seed_generator.py -> success, no issues
- git diff --check -> clean

sumin220:
Addressed the second review — and thanks @shaun0927 for pushing d35c1861 while I was preparing the same fixes locally; I've kept your version as is since it also broadens the lenient guard to TypeError/PydanticValidationError. On top of it, b527c056 adds the one scenario the new tests didn't cover: an explicit null weight parsed leniently defaults to 1.0 without raising (plus a blank-after-join criteria segment). So at head: blank strict fields retry with [] as the explicit empty spelling, omitted weight stays 1.0 while explicit null retries, and the legacy fallback skips blank segments instead of raising. Full suite 13639 passed, 12 skipped; ruff and mypy clean.

sumin220:
Addressed the third review in 39592fbc. Blocker: JSON integers past a fixed 400-digit bound now decode to a signed infinity via a parse_int hook instead of tripping CPython's int-conversion digit limit — whose plain ValueError escaped the lenient never-raises contract — so a 5000-digit weight parses without raising in lenient mode and clamps instead of retrying in strict mode. Follow-up: genuine numeric overflow now saturates by sign in both modes and in the legacy string path, restoring the historical 0.0 clamp for negative overflow that round one had regressed; the non-standard NaN/Infinity tokens keep their existing semantics (strict retries, lenient defaults) via a parse_constant marker, since those are not JSON numbers. Regression tests cover 5000-digit weights in both modes and negative-overflow sign preservation across all three paths. Full suite 13643 passed, 12 skipped; ruff and mypy clean.

Q00 (approved):
Approved at head 39592fb. Verified the strict/lenient boundary against the #1729 probe inputs: legacy colon/pipe lines are rejected at extraction and route through the reformat retry, colons and pipes survive inside object fields end to end, and stored legacy requirements keep the historical split unchanged. The one residual nuance from the final bot review (lenient `-Infinity` token saturating to 0.0 instead of the documented 1.0 default) is tracked as a maintainer follow-up in #1766 and doesn't block this. Thanks for grinding through the review rounds — merging.

PR #1751, opened 2026-07-26
fix(orchestrator): recognize artifact verification evidence

lee3Q:
## Summary

- Recognize artifact file writes performed through `pathlib.Path.write_text()` / `write_bytes()` as transcript-backed file evidence.
- Normalize direct `python -c` verifier payloads so multiline runtime commands and concise evidence claims can match.
- Accept explicit successful ad-hoc artifact verification output such as `equality: True` / `verified: True` as `tests_passed` evidence, while failing closed on negative output.
- Treat workspace-isolation acceptance criteria such as "only the target file changed" as validation ACs that do not require unrelated file/test mutation evidence.

## Verification

- Rebased/applied cleanly on `Q00/ouroboros` `origin/main@14af5553866af58a856e1829fa866619482cfb70` in isolated worktree `/private/tmp/ozo-ba050383-pubgate-20260727`.
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run --python 3.13 python -m pytest tests/unit/orchestrator/test_parallel_executor.py -q`
  - `349 passed in 15.33s`
  - repeat: `349 passed in 7.34s`
- `uv run --python 3.13 ruff check src/ouroboros/orchestrator/evidence/ac_classification.py src/ouroboros/orchestrator/evidence/claims.py src/ouroboros/orchestrator/evidence/shell_parsing.py src/ouroboros/orchestrator/evidence/test_detection.py tests/unit/orchestrator/test_parallel_executor.py`
- `uv run --python 3.13 ruff format --check src/ouroboros/orchestrator/evidence/ac_classification.py src/ouroboros/orchestrator/evidence/claims.py src/ouroboros/orchestrator/evidence/shell_parsing.py src/ouroboros/orchestrator/evidence/test_detection.py tests/unit/orchestrator/test_parallel_executor.py`
- `git diff --check`
- Direct production verifier probe using `_verify_atomic_evidence_against_runtime_messages`:
  - `Path('ozo_gui_smoke.txt').write_bytes(b'OZO')` plus Python `read_bytes()` equality verifier returned `pass_verdict True`.
  - The same typed evidence with `equality: False` returned `fail_verdict False FABRICATION_SUSPECTED`.
- Ouroboros QA on the publication-gate evidence: `qa-4ac023e8` PASS `0.83`.

Evidence memo: `/private/tmp/ozo-session140-ba050383-publication-gate-evidence.md`

## Scope Notes

[...]

Issue #1747, opened 2026-07-26
Verifier: attribute Bash-generated artifacts and validate structured wrapper receipts


sergiobuilds:
> [!IMPORTANT]
> This remains the canonical combined verifier tracker. The Q00-approved PR #1764 was explicitly a narrow part of this issue; it did not complete either trust boundary below.

## Goal

Complete the execution-bound evidence contract left after Codex completion journaling and the narrow pathlib receiver fix. A verifier must not accept a stale workspace file or agent-authored prose as proof that the accepted command attempt created an artifact or that a wrapper's inner test passed.

### Phase 1: command-scoped Bash artifact provenance

- [ ] Define a versioned artifact manifest or equivalent bounded pre/post attribution contract tied to the accepted command completion and attempt identity.
- [ ] Resolve paths from the authoritative workspace and reject absolute, escaping, symlink-escaping, missing, duplicate, and malformed entries.
- [ ] Reject artifacts from pre-existing state, an earlier retry, another attempt, or another session.
- [ ] Require a successful correlated command completion before artifact evidence can be authoritative.
- [ ] Exercise success, stale, retry, cross-session, malformed, and path-boundary cases through the real journal-to-verifier path.

### Phase 2: structured wrapper receipts

Phase 2 follows Phase 1 and may ship as a separate PR while remaining in this issue.

- [ ] Define a closed, versioned receipt with the inner command, inner and outer exit status, stdout/stderr references, test result, and accepted-attempt correlation.
- [ ] Attribute the receipt file itself through the Phase 1 artifact contract.
- [ ] Reject missing, malformed, stale, mismatched, duplicate, path-escaping, earlier-retry, and cross-session receipts.
- [ ] Prevent a non-zero inner or outer exit from producing accepted `tests_passed` evidence.
- [ ] Cover successful wrappers and every rejection class through the real journal-to-verifier path.

## Sequencing and non-goals

Use separate reviewable PRs if needed, both referencing this issue. Do not broaden command heuristics, trust final agent prose, or treat local/unmerged experiments as completion evidence. No required work in this tracker is being marked complete by this reconciliation.

## Origin

This is the remaining verifier work split from #1724. PRs #1732 and #1740 supplied correlated command completions, #1745 improved command/test manifest matching, and #1764 fixed one narrow pathlib receiver case. None supplied the general artifact or wrapper-receipt contracts above.

shaun0927:
## Canonical reconciliation

[...]

PR #1764, opened 2026-07-27
fix(orchestrator): bind pathlib write evidence to claimed path

lee3Q:
## Summary

Follow-up to the closed PR #1751. This PR implements only the first narrow
maintainer-requested slice:

Part of #1747.

- prevent `pathlib` command text from satisfying `files_touched` unless
  execution-bound file evidence is present first;
- keep Python/pathlib-looking command text authoritative for fail-closed
  rejection, so it cannot fall through to generic shell mutation heuristics;
- preserve structured Edit/Write/Bash output file proof paths.

## Why this is narrower than #1751

PR #1751 was closed because it broadened several high-trust evidence heuristics
at once. This PR does not revive that combined expansion. It handles only the
pathlib receiver-binding issue from the maintainer close-out.

## Verified

- Based on current `origin/main@f45670ef65973def1eae5cd9853502fa067198a9`.
- Focused verifier suite passed:
  - `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run --python 3.12 python -m pytest tests/unit/orchestrator/test_parallel_executor.py -q`: `417 passed in 19.56s`
  - focused pathlib/wrapper regressions: `57 passed, 360 deselected in 2.33s`
- Scoped lint passed:
  - `uv run --python 3.12 ruff check src/ouroboros/orchestrator/evidence/claims.py tests/unit/orchestrator/test_parallel_executor.py`
- Scoped format check passed:
  - `uv run --python 3.12 ruff format --check src/ouroboros/orchestrator/evidence/claims.py tests/unit/orchestrator/test_parallel_executor.py`
- `git diff --check` passed.
- PR CI passed on prior pushed heads; latest head is awaiting CI.
- OZO evidence QA:
  - `qa-d0331306` PASS 0.86; wrapper timed out after printing the verdict.

Evidence memo:

`/private/tmp/ozo-session155-pr1764-repair-evidence.md`

## Regression coverage

- Rejects Python/pathlib command text as standalone `files_touched` proof,
  including trusted-looking `-I -S -c` forms, because final filesystem state
  cannot prove execution-time receiver identity.
- Rejects a command that merely reads the claimed file but writes a different
  file.
- Rejects an absolute out-of-workspace `Path(...).write_text(...)` with the same
  basename as the claimed workspace file.
- Rejects relative or arbitrary absolute symlinked Python executables, symlinked
  command cwd values, and symlink-to-regular replacement cases that could spoof
  execution identity.

## Explicit non-claims

This PR does not address the other #1751 review blockers:

- direct interpreter-command equivalence for non-`python -c` shapes;
- ad-hoc verifier completion-status contracts;
- TraceGuard journal contract alignment;
- mixed implementation/isolation AC classification.


Q00:
@ouroboros-agent rereview exact HEAD 57023c35bc3ede4d48b860429cf2e10623f506f6. Maintainer repairs are pushed; please publish an explicit exact-head verdict after required checks complete.

Q00:
Addressed the current-head review blocker and compatibility warning in 6fea2ea2a. Recognized Python `-c` invocations now terminate command-text classification before generic shell mutation heuristics, including reconstructed pathlib plus inert positional `touch` argv. Literal-target shell redirection still accepts expanded payload content (`printf "$VALUE" > claimed.py`) via punctuation-aware shell tokenization. Removed the unused AST/static-target path. Verification: 423 focused/relevant tests passed; scoped Ruff and MyPy passed. Please re-review exact HEAD 6fea2ea2af27fecd85793b765b67ae725af89a97.

Q00:
Addressed review 4828423128 at exact HEAD 6304adc3fa948b1cbcb6f1c090c171bd233d3206. Supported `bash`/`zsh`/`sh -c` bodies returned by the shared shell parser now receive the same authoritative Python `-c` classification before generic mutation matching. Unwrapping is bounded, cycle-safe, and limited to already-supported wrapper syntax. Added an execution-backed `/bin/bash -lc` regression proving reconstructed pathlib changes only `other.py` while inert positional `touch claimed.py` cannot authenticate the stale claim. Verification: 424 relevant tests passed; scoped Ruff and MyPy passed. Please re-review this exact HEAD.

Q00:
Addressed review 4828512182 at exact HEAD d1c1a39a1fa30525603d4be93abb070524faa97d. Generic shell mutation targets are now tokenized as literal targets, resolved from the recorded effective command cwd, contained in the task workspace, and compared to the canonical workspace-relative claim. An execution-backed nested-cwd regression proves `sub/touch claimed.py` cannot authenticate root `claimed.py` but does authenticate `sub/claimed.py`. Python executable identity becomes authoritative only when `-c` is present, preserving execution-backed `python generator.py > claimed.py` literal-redirection proof. Supported wrapper, expansion, symlink/error containment, and direct Python contracts remain green. Verification: 426 relevant tests passed; scoped Ruff and MyPy passed. Please re-review this exact HEAD.

Q00:
Addressed review 4828586873 and merged current main at exact HEAD 37a9735d933e2ab1eadb8e7ccf7a56fde02dad33 (non-force). The authoritative Python `-c` classifier now recognizes the immutable `/usr/bin/env` wrapper and parses `-c` only in Python option position. Shell tokenization masks quoted/escaped `>` before punctuation parsing, so inert argv cannot become redirection. For generic shell compatibility without trusting mutable final state, the production LeafDispatcher records lexical targets at the Bash call and attaches immediate post-completion `lstat` identity; verifier accepts command-text mutation proof only when that internally captured regular-file device/inode/type still matches. The exact env/reconstructed-pathlib/quoted-`>` execution changes only `other.py` and is rejected. The exact external-symlink execution followed by replacement with a workspace regular file is rejected by type/inode mismatch. A production-shaped call/completion redirect regression confirms valid literal redirection remains accepted through the capture path. Verification on merged main: 428 focused tests passed; full `tests/unit/orchestrator` 4448 passed, 1 skipped; scoped Ruff and MyPy passed. Please re-review this exact HEAD.

Q00:
Superseding my prior rereview request: the post-only identity design was strengthened after an independent parent-component swap probe. Exact HEAD is now c8c3fd93ce1b210d4c6be0746c0c14f71298ec8d (current main 248cc9719 is merged, non-force). Leaf dispatch leases each receiver parent from call through completion using a workspace-rooted `openat` chain with `O_DIRECTORY|O_NOFOLLOW`, and reads pre/post receiver state through that same held dirfd. Existing receivers prove mutation only when pre is regular, post has the same device/inode/filetype, and size/mtime/ctime changed; pre-absent to post-regular creation is allowed. Symlink-to-regular and unrelated regular-inode replacement fail closed. Parent-directory swap to an external symlink, execution against the external target, and restoration before completion leaves the held receiver unchanged and emits no effect. Lease ownership uses mutable fd invalidation and a tracker context manager, ensuring exactly-once close across completion, duplicate/idless ambiguity, normal stream end, adapter exception, cancellation, and unsupported-platform fallback. Reused-fd and 64-cycle leak probes pass. Correlation tracks every tool call: duplicate IDs, mixed Bash/Edit ownership, overlapping idless calls, explicit non-Bash completions, and nameless ambiguous completions cannot consume a Bash lease. `/usr/bin/env` Python `-c` and quoted `>` provenance regressions remain rejected, while production-shaped normal literal redirect remains accepted. Verification: 445 focused tests passed; final full `tests/unit/orchestrator` 4465 passed, 1 skipped; scoped Ruff and MyPy passed. Please review exact HEAD c8c3fd93ce1b210d4c6be0746c0c14f71298ec8d.

Q00:
Exact-HEAD rereview requested for `0eadfc0cc5eb53ffed96e604b8f785f575cb8d66`.

This follow-up closes the remaining zero-trust findings:
- strips all adapter-supplied `filesystem_effects` before correlation and reattaches only locally measured execution-span lease effects;
- collects correlation aliases from top-level data/meta and nested `tool_result`/meta, rejecting conflicting IDs without id-less fallback or lease donation;
- permits a trusted Python `-c` command's independent literal shell redirection only when the leased receiver changed, while Python-internal pathlib targets and quoted/escaped `>` remain rejected.

Verification on this exact SHA:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py` — 455 passed
- `uv run pytest -q tests/unit/orchestrator` — 4,473 passed, 1 skipped
- `uv run mypy src` — passed (486 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,159 files)
- `git diff --check` — passed

Q00:
Exact-HEAD rereview requested for `b00b0f3dd2ad1605e1bc45dce43d7f8be9ccb861`.

This follow-up fixes the independently reproduced commented-redirection bypass. Shell mutation parsing now stops only at a real unquoted, unescaped word-start `#`; quoted hashes, escaped hashes, and token-internal values such as `foo#bar` remain literal argv data. The execution-backed regression proves `python -c "...Path('claimed.py').write_text(...)" # > claimed.py` cannot acquire shell receiver evidence, while literal-hash payloads followed by a real redirect still do.

Verification on this exact SHA:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py` — 459 passed
- `uv run pytest -q tests/unit/orchestrator` — 4,479 passed, 1 skipped
- `uv run mypy src` — passed (486 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,159 files)
- `git diff --check` — passed

Q00:
Exact-HEAD rereview requested for `11acba506cfce6bcf1a766c87cfd254613f1c011`.

This follow-up closes the escaped-LF continuation bypass reproduced against `b00b0f3dd`. The shell comment scanner now removes POSIX backslash-LF continuations without changing the preceding word-boundary state, so a following word-start `#` still comments out `> claimed.py`. A production-shaped regression exercises the real subprocess between Bash call/result observations and proves an internal Python write cannot gain shell-receiver evidence. A direct CRLF probe remains fail-closed.

Verification on this exact SHA:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py` — 461 passed
- `uv run pytest -q tests/unit/orchestrator` — 4,481 passed, 1 skipped
- `uv run mypy src` — passed (486 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,159 files)
- `git diff --check` — passed

Q00:
Exact-HEAD rereview requested for `dd6dbf1dc2c5fe6531ca680c81f0a130d0ea4d42`.

This follow-up closes the multi-receiver blocker and its adjacent parser boundaries:
- enumerates every stable receiver for multi-target `touch`, `truncate`, and `tee` commands;
- removes leading/interspersed/trailing input/output redirection clauses without truncating later utility operands, while excluding fd duplication from file targets;
- parses bounded utility option grammars so required option values, optional attached values, `--`, and dash-prefixed filenames cannot be confused with receivers; unknown forms fail closed;
- recognizes exact sed/Perl in-place switches and enumerates every input file, but intentionally preserves the existing inode-continuity rejection when those tools atomically replace files;
- rejects the reproduced `perl -MFile::Path -e ... claimed.py` false-`-i` case with a real subprocess/lease regression;
- treats single-quoted Perl regex `$` as literal rather than unquoted shell expansion.

Verification on this exact SHA:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py` — 494 passed
- `uv run pytest -q tests/unit/orchestrator` — 4,513 passed, 1 skipped
- `uv run mypy src` — passed (486 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,159 files)
- `git diff --check` — passed

Q00:
Exact-HEAD rereview requested for `c5ba68ad87bcb2ecf044f951df5830fd8b43cddb`.

This follow-up preserves lexical adjacency across shell tokenization:
- only an IO-number directly adjacent to a redirection operator is marked and removed;
- whitespace-separated numeric filenames remain utility receivers, proven with real `touch 123 > log.txt` and `tee 123 < source.txt` call→execution→completion regressions;
- adjacent `2>error`, `2>&1`, and `123<source` remain descriptor syntax rather than file operands;
- `>|` is recognized as clobber redirection instead of compound pipeline control, with a production-shaped positive regression;
- quote, escape, comment, continuation, fd-duplication, and multi-receiver boundaries from prior exact reviews remain covered.

Verification on this exact SHA:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py` — 501 passed
- `uv run pytest -q tests/unit/orchestrator` — 4,521 passed, 1 skipped
- `uv run mypy src` — passed (486 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,159 files)
- `git diff --check` — passed

Q00:
Exact-HEAD rereview requested for `b620b2d82b970e12f7c495953fa2976aed5885e9`.

This test-only follow-up fixes the Python 3.14 Linux CI failure without changing production trust semantics. The inode-replacement regression now executes platform-native sed syntax: BSD/Darwin uses `sed -i '' ...`, while Linux/GNU uses suffix-free `sed -i ...`. Separate parser regressions retain coverage for both spellings and prove both receivers are enumerated. Actual sed/Perl inode replacement remains fail-closed.

Verification on this exact SHA:
- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor.py` — 502 passed
- platform-native in-place focused regressions — 4 passed
- `uv run mypy src` — passed (486 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,159 files)
- `git diff --check` — passed

Q00 (approved):
Approved at exact HEAD b620b2d82b970e12f7c495953fa2976aed5885e9 after independent adversarial review. Numeric filenames are distinguished from adjacent IO-number redirects, >| is handled without admitting real pipelines, and sed/perl inode replacement remains fail-closed. Platform-native sed execution is covered on GNU/Linux and BSD/Darwin. Exact-head Python 3.12/3.13/3.14, Ruff, MyPy, TypeScript, and policy checks are green.

PR #1767, opened 2026-07-27
feat(bigbang): extract ONTOLOGY_FIELDS as JSON object arrays (#1729 slice 3)

sumin220:
## Summary

Slice 3 of #1729: ONTOLOGY_FIELDS moves to a single-line JSON object array at extraction time, extending the exact contract merged for CONSTRAINTS (#1714), the brownfield lists (#1730), and evaluation principles / exit conditions (#1762).

Today a field description like `Ratio of open:closed tasks` breaks the colon split, and a pipe anywhere silently truncates the entry. With this change:

| Lane | Behavior |
|---|---|
| Extraction (strict) | Must be a valid JSON array of `{"name", "type", "description"}` objects; anything else — legacy pipe lines, blank values, non-objects, missing or blank fields — raises so the existing reformat-retry path asks the LLM to resend. `[]` is the explicit empty spelling. |
| Stored requirements (lenient) | Bracket-led valid JSON arrays parse to models; everything else keeps the historical `name:type:description` split with the description absorbing further colons; malformed or blank-segment entries are skipped instead of raising (the legacy loop previously raised a model ValidationError on e.g. `x::d`); pre-parsed model tuples pass through; never raises. |

Details:
- The model field spelling `field_type` is accepted alongside the prompt's `type`, and an optional boolean `required` is honored (a non-boolean value poisons the strict claim rather than being ignored).
- Reuses the shared decode/validation helpers hardened in #1762's review rounds (bounded integer decoding, non-finite token markers, blank-field strict rejection), so those edge classes are covered here from the start.
- The extraction prompt, the reformat template, and the seed-architect agent template request the new format.

## Sequencing vs #1729

Same shape as the previous slices: adapter-local to `bigbang/` parsing plus its prompt templates, no shared schema or consumer changes. Slice 4 (CONTEXT_REFERENCES) stays separate and will follow once this lands. Happy to hold if you'd rather sequence differently.

## Test plan

- [x] 4 new regression tests: strict parsing with delimiters inside data plus the `field_type`/`required` handling, strict rejection of legacy/blank/malformed values, lenient legacy-split preservation and never-raise fallbacks (including blank segments the old loop raised on), and the end-to-end extraction retry
- [x] `uv run pytest tests/unit/bigbang -q` — 704 passed
- [x] `uv run pytest tests/ -q -n 4` — 14198 passed, 12 skipped

Refs #1729 (slice 3 of 4; #1730 was slice 1, #1762 slice 2)

### DIFF
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -50,7 +50,7 @@ Example: "AC: Tasks can be created | verify: python -m pytest tests/test_tasks.p
 The data structure/domain model for this work:
 - **ONTOLOGY_NAME**: A name for the domain model
 - **ONTOLOGY_DESCRIPTION**: What the ontology represents
-- **ONTOLOGY_FIELDS**: Key fields in format: name:type:description (pipe-separated)
+- **ONTOLOGY_FIELDS**: Key fields as a single-line JSON array of objects with "name", "type" (string, number, boolean, array, object), and "description"
 
 Field types should be one of: string, number, boolean, array, object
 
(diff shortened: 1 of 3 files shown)


sumin220:
The Python 3.12 failure is tests/unit/evolution/test_generation_watchdog.py::test_busy_run_without_material_progress_times_out — an unrelated timing-sensitive watchdog test; 3.13 and 3.14 pass on the same head and the full suite passes locally. Could someone with write access re-run that job? I'd rather not push an empty commit and dismiss the review.

PR #1775, opened 2026-07-27
fix(core): fail open on unreadable mechanical.toml and malformed Info.plist

### LINKED ISSUE #1772: `_load_project_overrides` documents "Never raises" but lets `OSError` escape on an unreadable `mechanical.toml`
---

`src/ouroboros/evaluation/languages.py:133-150` promises a total function, then breaks that promise on the only I/O it performs.

```python
def _load_project_overrides(working_dir: Path) -> dict[str, Any] | None:
    """Load ``.ouroboros/mechanical.toml`` if it exists.

    Returns the parsed TOML dict, or ``None`` when the file is missing or
    malformed. Never raises.
    """
    config_path = working_dir / ".ouroboros" / "mechanical.toml"
    if not config_path.exists():
        return None

    import tomllib

    try:
        with open(config_path, "rb") as f:
            return tomllib.load(f)
    except tomllib.TOMLDecodeError as e:
        log.warning("mechanical.toml_parse_error", path=str(config_path), error=str(e))
        return None
```

`exists()` returning `True` does not imply the path is openable. `open()` raises `OSError` — `PermissionError` for an unreadable mode, `IsADirectoryError` when a directory sits in that place, or `FileNotFoundError` if the file is removed between the `exists()` probe and the `open()`. None of those are `tomllib.TOMLDecodeError`, so they propagate out of a function whose docstring says they cannot.

The caller is `resolve_mechanical_config` (`languages.py:279`), so instead of Stage 1 falling back to built-in defaults, mechanical-config resolution aborts with a traceback.

### Why this is a contract violation rather than a judgment call

The docstring states the contract in three ways — "Returns ... or ``None``", "when the file is missing or malformed", "Never raises" — and the surrounding design depends on it: `build_mechanical_config` has no error path for this helper, and the existing test `test_malformed_toml_is_ignored` (`tests/unit/evaluation/test_languages.py`) pins exactly the "bad config degrades to defaults" behavior for the parse-failure half of the contract. Only the read-failure half is unguarded.

### Reproduction

```python
from pathlib import Path
from ouroboros.evaluation.languages import build_mechanical_config

project = Path("/tmp/demo")
(project / ".ouroboros").mkdir(parents=True, exist_ok=True)
(project / ".ouroboros" / "mechanical.toml").mkdir()   # directory in the file's place

build_mechanical_config(project)   # IsADirectoryError, expected: config with all-None commands
```

### Suggested fix

Catch `OSError` alongside the existing decode error and degrade to defaults, keeping the existing `mechanical.toml_parse_error` log event for genuine parse failures so current log consumers are unaffected. Roughly:

```python
    except OSError as e:
        log.warning("mechanical.toml_read_error", path=str(config_path), error=str(e))
        return None
    except tomllib.TOMLDecodeError as e:
        log.warning("mechanical.toml_parse_error", path=str(config_path), error=str(e))
        return None
```

Regression test belongs beside `test_malformed_toml_is_ignored` in `tests/unit/evaluation/test_languages.py`. Using a directory in the config's place keeps the test deterministic and independent of the privileges of the user running the suite (a chmod-based test is a no-op under root).

Happy to open a PR for this.

### LINKED ISSUE #1773: Malformed `Info.plist` XML leaks `xml.parsers.expat.ExpatError` out of `resolve_zcode_electron_node_path`
---

`src/ouroboros/zcode_cli_launcher.py:65-71` converts unreadable bundle metadata into an actionable `RuntimeError`, but its `except` tuple misses the exception `plistlib` actually raises for malformed XML.

```python
    info_plist = contents_dir / "Info.plist"
    try:
        with info_plist.open("rb") as stream:
            bundle_info = plistlib.load(stream)
    except (OSError, plistlib.InvalidFileException) as exc:
        msg = f"ZCode app bundle metadata is present but {info_plist} is unreadable: {exc}"
        raise RuntimeError(msg) from exc
```

`plistlib.InvalidFileException` is raised for a *binary* plist with a bad header. For an XML plist that is merely truncated or malformed, `plistlib.load` propagates `xml.parsers.expat.ExpatError`, whose MRO is `(ExpatError, Exception, BaseException, object)` — neither an `OSError` nor a `ValueError`, so the guard does not catch it.

### Reproduction

```
ZCode.app/Contents/Info.plist  <-  b"<?xml version='1.0'?><plist><dict>"
```

```
xml.parsers.expat.ExpatError: no element found: line 1, column 34
```

expected:

```
RuntimeError: ZCode app bundle metadata is present but .../Info.plist is unreadable: ...
```

The raw parser error reaches both callers — `providers/zcode_cli_adapter.py:52` and `orchestrator/zcode_cli_runtime.py:49` — instead of the message that tells the user which bundle file to look at.

### Why this reads as an oversight rather than intended behavior

Every other failure mode in the same function is already normalized to `RuntimeError`: `json.JSONDecodeError` for the node-bundle metadata (line 41), a non-dict plist payload (line 73), a metadata/script mismatch (line 63), and executable path traversal. Truncated XML is the one gap, and `tests/unit/test_zcode_cli_launcher.py` already pins the neighbours with `test_app_bundle_unreadable_node_metadata_fails_closed` and `test_app_bundle_non_dictionary_plist_fails_closed`.

### Suggested fix

Add `xml.parsers.expat.ExpatError` to the existing tuple:

```python
    except (OSError, plistlib.InvalidFileException, ExpatError) as exc:
```

Regression test slots in beside `test_app_bundle_non_dictionary_plist_fails_closed`.

Happy to open a PR for this.

MyoungSoo7:
## Summary

Two fail-open guards caught a narrower set of exceptions than the calls they wrap can raise, so both crashed instead of degrading. Each fix widens one `except` clause; no behavior changes on the happy path.

- **`_load_project_overrides`** (`src/ouroboros/evaluation/languages.py`) documents *"Never raises"* but only caught `tomllib.TOMLDecodeError`. `Path.exists()` returning `True` does not imply the path is openable, so a `PermissionError` / `IsADirectoryError` — or a delete racing the probe — escaped into `resolve_mechanical_config` instead of falling back to built-in defaults. Genuine parse failures keep the existing `mechanical.toml_parse_error` log event so current consumers are unaffected; read failures get their own event.

- **`resolve_zcode_electron_node_path`** (`src/ouroboros/zcode_cli_launcher.py`) normalizes every other bundle-metadata failure to an actionable `RuntimeError`, but `plistlib.InvalidFileException` only covers *binary* plists with a bad header. Malformed **XML** raises `xml.parsers.expat.ExpatError`, which is neither an `OSError` nor a `ValueError`, so a raw parser error reached both callers (`providers/zcode_cli_adapter.py:52`, `orchestrator/zcode_cli_runtime.py:49`).

Both regression tests were confirmed to **fail on the parent commit** and pass here.

## Test plan

- [x] `uv run pytest tests/ -q` passes — `13840 passed, 6 skipped` on this branch
- [x] `uv run mypy src/` clean — `Success: no issues found in 476 source files`
- [x] `python3 scripts/check-auto-boundary.py` — `OK (39 files scanned, 0 findings)`
- [x] New tests verified as real regressions: with the two source hunks reverted, `test_unreadable_toml_is_ignored` and `test_app_bundle_malformed_plist_xml_fails_closed` both fail (`ExpatError: no element found: line 1, column 34`)

The unreadable-config test places a directory at the config path rather than using `chmod`, so it stays deterministic and does not silently no-op when the suite runs as root.

## Related issues

Fixes #1772
Fixes #1773

### DIFF
--- a/src/ouroboros/evaluation/languages.py
+++ b/src/ouroboros/evaluation/languages.py
@@ -145,6 +145,12 @@ def _load_project_overrides(working_dir: Path) -> dict[str, Any] | None:
     try:
         with open(config_path, "rb") as f:
             return tomllib.load(f)
+    except OSError as e:
+        # ``exists()`` above does not guarantee the file is readable: an unreadable
+        # mode, a directory in its place, or a delete between the two calls all raise
+        # here. Fall back to built-in defaults so the docstring's contract holds.
+        log.warning("mechanical.toml_read_error", path=str(config_path), error=str(e))
+        return None
     except tomllib.TOMLDecodeError as e:
         log.warning("mechanical.toml_parse_error", path=str(config_path), error=str(e))
         return None
--- a/src/ouroboros/evaluation/languages.py
+++ b/src/ouroboros/evaluation/languages.py
@@ -137,18 +137,18 @@ def _load_project_overrides(working_dir: Path) -> dict[str, Any] | None:
     malformed. Never raises.
     """
     config_path = working_dir / ".ouroboros" / "mechanical.toml"
-    if not config_path.exists():
-        return None
 
     import tomllib
 
     try:
         with open(config_path, "rb") as f:
             return tomllib.load(f)
+    except FileNotFoundError:
+        return None
     except OSError as e:
-        # ``exists()`` above does not guarantee the file is readable: an unreadable
-        # mode, a directory in its place, or a delete between the two calls all raise
-        # here. Fall back to built-in defaults so the docstring's contract holds.
+        # Probe and read through one guarded operation. A separate ``exists()``
+        # preflight can itself raise for an inaccessible parent and introduces a
+        # race before ``open()``.
         log.warning("mechanical.toml_read_error", path=str(config_path), error=str(e))
         return None
     except Exception as e:
--- a/src/ouroboros/evaluation/languages.py
+++ b/src/ouroboros/evaluation/languages.py
@@ -151,7 +151,13 @@ def _load_project_overrides(working_dir: Path) -> dict[str, Any] | None:
         # here. Fall back to built-in defaults so the docstring's contract holds.
         log.warning("mechanical.toml_read_error", path=str(config_path), error=str(e))
         return None
-    except tomllib.TOMLDecodeError as e:
+    except Exception as e:
+        # "Never raises" is the contract, and this file is optional operator-authored
+        # input, so anything that stops it becoming a dict must degrade to defaults
+        # rather than abort mechanical-config resolution. Narrowing this to the known
+        # exception types has repeatedly missed one: tomllib raises TOMLDecodeError for
+        # bad syntax, UnicodeDecodeError for non-UTF-8 bytes, and RecursionError for
+        # pathologically nested documents — and only the first two share a base class.
         log.warning("mechanical.toml_parse_error", path=str(config_path), error=str(e))
         return None
 
--- a/src/ouroboros/zcode_cli_launcher.py
+++ b/src/ouroboros/zcode_cli_launcher.py
@@ -34,11 +34,17 @@ def resolve_zcode_electron_node_path(cli_path: str | Path | None) -> str | None:
 
     metadata_path = path.with_name(ZCODE_NODE_BUNDLE_METADATA)
     try:
-        metadata = json.loads(metadata_path.read_text(encoding="utf-8"))
+        metadata_text = metadata_path.read_text(encoding="utf-8")
     except OSError as exc:
         msg = f"ZCode app-bundle CLI metadata is missing or unreadable: {metadata_path}: {exc}"
         raise RuntimeError(msg) from exc
-    except json.JSONDecodeError as exc:
+    except Exception as exc:
+        msg = f"ZCode app-bundle CLI metadata is invalid JSON: {metadata_path}: {exc}"
+        raise RuntimeError(msg) from exc
+
+    try:
+        metadata = json.loads(metadata_text)
+    except Exception as exc:
         msg = f"ZCode app-bundle CLI metadata is invalid JSON: {metadata_path}: {exc}"
         raise RuntimeError(msg) from exc
 
--- a/src/ouroboros/zcode_cli_launcher.py
+++ b/src/ouroboros/zcode_cli_launcher.py
@@ -6,6 +6,7 @@ import json
 import os
 from pathlib import Path
 import plistlib
+from xml.parsers.expat import ExpatError
 
 ZCODE_SCRIPT_SUFFIXES = (".cjs", ".js", ".mjs")
 ZCODE_NODE_BUNDLE_METADATA = ".node-bundle-meta.json"
@@ -66,7 +67,7 @@ def resolve_zcode_electron_node_path(cli_path: str | Path | None) -> str | None:
     try:
         with info_plist.open("rb") as stream:
             bundle_info = plistlib.load(stream)
-    except (OSError, plistlib.InvalidFileException) as exc:
+    except (OSError, plistlib.InvalidFileException, ExpatError) as exc:
         msg = f"ZCode app bundle metadata is present but {info_plist} is unreadable: {exc}"
         raise RuntimeError(msg) from exc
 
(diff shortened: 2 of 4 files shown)


MyoungSoo7:
Pushed `91a6f8e` addressing the blocker.

Probing the remaining surface of `tomllib.load()` turned up a third failure mode beyond the reported `UnicodeDecodeError`, and it shares no base class with the other two:

| input | exception | base |
|---|---|---|
| invalid UTF-8 | `UnicodeDecodeError` | `ValueError` |
| bad syntax | `TOMLDecodeError` | `ValueError` |
| deeply nested document | `RecursionError` | `RuntimeError` |

So `except (OSError, UnicodeDecodeError, TOMLDecodeError)` — and even `except (OSError, ValueError)` — would still have left the documented contract broken. Rather than enumerate a third time, the parse arm now catches anything that stopped the bytes becoming a dict, with the reasoning recorded in a comment. The read arm stays `OSError` so it keeps its distinct log event. The file is optional operator-authored input whose fallback is well defined, so failing open is the intended behavior here.

Regression coverage added for both new cases; all three (`test_unreadable_toml_is_ignored`, `test_non_utf8_toml_is_ignored`, `test_pathologically_nested_toml_is_ignored`) fail on the parent commit.

---

Separately, on the failing **Test Python 3.12** check — that is a pre-existing test-isolation bug in the suite, not something this PR introduced. Filing it separately, but flagging here since it is what turned this PR's CI red.

`tests/unit/cli/test_mcp_shell_env.py::test_shell_env_whitelist_admits_proxy_and_gateway_vars` leaks `HTTPS_PROXY=http://proxy:8080`, `NO_PROXY`, `ANTHROPIC_BASE_URL` and `GH_TOKEN` into `os.environ` for the remainder of the worker. Any later test that shells out to the network then fails. Deterministic two-test repro:

```
pytest tests/unit/cli/test_mcp_shell_env.py::test_shell_env_whitelist_admits_proxy_and_gateway_vars \
       tests/unit/orchestrator/test_profile_loader.py::TestWheelPackaging::test_wheel_contains_profile_yamls
```

```
uv build failed (rc=2):
  Caused by: Failed to fetch: `https://pypi.org/simple/hatchling/`
  Caused by: tunnel error: failed to create underlying connection
  Caused by: dns error
```

Cause: `monkeypatch.delenv(key, raising=False)` records nothing when the key is absent, so the pre-emptive `delenv` at line 143 registers no undo. The production call then writes the vars directly into `os.environ`, and the cleanup `delenv` at line 168 records the **polluted** value as the original — which teardown faithfully restores. The test's own cleanup is what creates the leak. It only bites when both tests land on the same xdist worker, which is why it looks intermittent.

MyoungSoo7:
Pushed `7d5bfb6`. The finding is correct — probing the rest of the surface shows `plistlib` spreads its failures across unrelated hierarchies:

| input | exception | base |
|---|---|---|
| binary plist, bad header | `plistlib.InvalidFileException` | `ValueError` |
| malformed XML | `xml.parsers.expat.ExpatError` | `Exception` |
| well-formed XML, bad `<integer>` / `<real>` | `ValueError` | — |
| well-formed XML, bad `<date>` | `AttributeError` | — |

There is no useful common base, so each round of enumerating types missed one. This is the same conclusion already reached for `mechanical.toml` earlier in this PR, and I should have applied it here at the same time rather than leaving a second enumerated tuple in place.

`resolve_zcode_electron_node_path` normalizes every other bundle-metadata failure to `RuntimeError` — the JSON decode error, the non-dict payload, the metadata/script mismatch, the path traversal. Leaking a parser internal from this one arm was the anomaly, so the parse arm now catches broadly with the reasoning recorded in a comment. The now-unused `ExpatError` import is dropped.

Regression coverage added as a parametrized case over the three semantically-invalid payloads; all three fail on the parent commit and pass here.

Verification: `13844 passed, 6 skipped` on this branch, plus ruff check / ruff format --check / mypy / check-auto-boundary all clean. The single remaining failure in a full-suite run is `TestWheelPackaging::test_wheel_contains_profile_yamls`, which is the pre-existing test-isolation bug filed as #1793 (a proxy env var leaks out of `test_mcp_shell_env.py` and breaks any later test that shells out to the network) — unrelated to this PR, and it passes in isolation.

Q00:
@ouroboros-agent Please re-review current HEAD f1309b039. mechanical.toml now avoids the unguarded exists/stat preflight and reads through one fail-open boundary. ZCode JSON now normalizes invalid UTF-8 and parser recursion failures to actionable RuntimeError, with focused regressions. Related suites: 141 passed; Ruff and MyPy passed.

PR #1735, opened 2026-07-24
feat(orchestrator): add deterministic Routing B admission kernel

Q00:
## Summary

- add a versioned provider-neutral `RouteCandidate` registry for model × harness × effort × persona/tool policy × authority identity
- add a deterministic Admission Kernel that applies hard pins/capability constraints and selects the cheapest eligible route
- keep Advisor output advisory-only: it can break equal-cost ties but cannot authorize, create, or make an expensive route beat a cheaper one
- add fail-closed contract parsing, stable rejection codes, blocked decisions, and deterministic serialization
- document the authority split and the next Routing C escalation slice

## Scope boundaries

This PR adds the pure route contract and policy boundary only. It does not call providers, change live dispatch, implement escalation, or declare AC acceptance. The Final Gate remains the only acceptance authority.

## Validation

- `uv run ruff check src/ouroboros/orchestrator/route_policy.py tests/unit/orchestrator/test_route_policy.py`
- `uv run mypy src/ouroboros/orchestrator/route_policy.py tests/unit/orchestrator/test_route_policy.py`
- `uv run pytest -q tests/unit/orchestrator/test_route_policy.py` (16 passed)
- `uv run pytest -q tests/unit/orchestrator` (3330 passed, 1 skipped)

## Latest validation

- Exact-head source fixes are on `8204b9c6`: dot-delimited credential suffixes fail closed, Advisor normalization failures fall back safely, and `validate_admission()` revalidates against the live registry and exact requirements before effects.
- Focused route/security tests: 71 passed. Full orchestrator/core suite: 3914 passed, 1 skipped locally.

## Current exact-head validation (f5277160)

CI is green on Python 3.12/3.13/3.14, coverage, Ruff, MyPy, TypeScript, boundary, envelope, and performance. The source tree is unchanged from 094b8e7c; f5277160 is a review-queue retrigger only. Please review this exact head.

PR #1722, opened 2026-07-23
feat(orchestrator): separate AC attempts from final acceptance

Q00:
## Summary
- separate provisional `execution.ac.attempt_judged` telemetry from Final Gate `execution.ac.acceptance_finalized`
- enforce one durable acceptance per `(authority_correlation_id, root_ac_index)` with idempotent duplicate handling and fail-closed conflicts
- publish final acceptance only after terminal session CAS; keep paused episodes unfinalized
- migrate frugality proof, retrospective, attention relay, and MCP recovery consumers

## Verification
- Foundation B targeted suite: 672 passed
- `ruff check src tests`
- `mypy src/ouroboros`
- `git diff --check`

## Scope
Foundation B only. Dispatch, routing, capsules, cross-run trust, and provider policy remain deferred to the next roadmap slice.

PR #1778, opened 2026-07-27
ci: require every pull request to reference an issue

### LINKED ISSUE #1777: Merged PRs routinely leave their issues open: no gate requires a PR-issue link
## Problem

A board audit on 2026-07-28 closed 11 open issues. **6 of them were already fully implemented on `main`** — the work had landed weeks earlier, but no PR ever linked back, so the issue stayed open and inflated the apparent backlog.

The six: #1467, #1469, #1470, #1396, #1462, #1450. Every one of them had merged PRs attached only as timeline cross-references, never as closing links.

## Measurement

Of the 80 PRs merged between 2026-06-28 and 2026-07-28:

| Body content | Count | Share |
|---|---:|---:|
| `Closes` / `Fixes` / `Resolves #N` | 17 | 21% |
| Reference only (`Part of #N`, `Refs #N`) | 46 | 58% |
| No issue number at all | 17 | 21% |

Two distinct failures are visible here:

1. **21% of PRs carry no issue trail whatsoever.** Nothing connects the change to the reason it was made. Example: #1735 `feat(orchestrator): add deterministic Routing B admission kernel` is literally Stack 2 of RFC #1681 and references no issue.
2. **58% reference without closing.** This is correct behaviour for epic slices — a slice must not auto-close its parent — but it means closure depends entirely on someone remembering. Nobody did, six times.

`.github/PULL_REQUEST_TEMPLATE.md` already has a `## Related issues` section with a `Fixes #1234` example. It is a comment; nothing checks it.

## Proposal

Add a `PR Hygiene` workflow that fails a pull request carrying neither a closing link nor a plain issue reference.

Deliberately **not** requiring a closing keyword: forcing `Closes` would have produced 46 false closures against the measured data. Closing stays a human decision; the gate only guarantees the trail exists. A notice annotation nudges reference-only PRs to consider whether they complete the issue.

Detection combines GitHub's `closingIssuesReferences` (which also covers issues linked through the Development sidebar — invisible to any body regex) with a body-only `#N` scan. The title is not accepted: `fix(#1638): ...` reads like a link but creates no GitHub reference, which is exactly how the audited PRs looked linked while leaving their issues orphaned.

## Acceptance criteria

- [ ] A pull request with no closing link and no `#N` in its body fails a required-status-eligible check.
- [ ] A pull request referencing an issue without closing it passes, and receives a notice annotation.
- [ ] Dependabot, `chore(deps)` / release PRs, and PRs labelled `no-issue` pass.
- [ ] Exempt PRs report a green check, not `skipped` (a skipped required check deadlocks branch protection).
- [ ] Replaying the rule over the last 80 merged PRs yields 63 pass / 17 blocked.

## Out of scope

- Making the check a required status check in branch protection. Run it advisory for a couple of weeks first and measure the friction.
- Retroactively linking already-merged PRs.
- Automatically closing issues whose referencing PRs have all merged. Worth a separate issue — it targets the 58% band directly.

Q00:
## Summary

Adds a `PR Hygiene` workflow that fails a pull request carrying neither a closing link nor a plain issue reference, so merged work stops leaving its issue open.

Closes #1777.

## Why

A board audit on 2026-07-28 closed 11 issues; **6 were already fully implemented on `main`** (#1467 #1469 #1470 #1396 #1462 #1450). Every one had merged PRs attached only as timeline cross-references. Of the 80 PRs merged in the preceding 30 days, 17 carried no issue reference at all — including #1735, which is Stack 2 of RFC #1681 and links to nothing.

## Design decisions

**A plain reference is enough; `Closes` is not required.** 46 of those 80 PRs reference an issue without closing it because they are epic slices, and a slice must not auto-close its parent. Demanding a closing keyword would have forced 46 false closures. Closing stays a human decision — this gate only guarantees the trail exists. Reference-only PRs get a `::notice::` nudging them to use `Closes` if they do complete the issue.

**Detection combines two signals.** GitHub's `closingIssuesReferences` covers both body keywords and issues linked through the Development sidebar, which no body regex can see; a body-only `#N` scan catches `Refs`/`Part of`.

**The title is deliberately not accepted.** `fix(#1638): ...` reads like a link but creates no GitHub reference — precisely how the audited PRs looked linked while their issues sat orphaned.

**Exemptions run inside the step, not as a job-level `if`.** An exempt PR must report a green check; a `skipped` job stays pending forever under a required-status-check branch protection rule and would deadlock the merge.

## Exemptions

- `no-issue` label (created alongside this PR)
- `dependabot[bot]`
- `chore(deps)` / `chore(release)` / `release:` titles

`ouroboros-agent` is **not** exempt — bot PRs are held to the same trail requirement. Say the word if that should change.

## Test plan

- [x] YAML parses (`yaml.safe_load`)
- [x] Rule replayed against the 80 PRs merged 2026-06-28 → 2026-07-28: **63 pass, 17 blocked**. The 17 split into genuinely issue-less work (`docs: FUNDING.yml` #1633, `test(plugin)` #1641, `perf(tests)` #1598 → `no-issue` label) and PRs that should have linked (#1735, #1722, #1720)
- [x] This PR exercises its own gate via `Closes #1777`
- [ ] `uv run pytest tests/ -q` — not applicable, no Python changed
- [ ] `uv run ruff check src/ tests/` — not applicable, no Python changed

## R-run comparison

## Follow-up, not in this PR

- Promoting the check to a required status check — run it advisory for two weeks and measure friction first.
- A sweep that flags open issues whose referencing PRs have all merged. That targets the 58% reference-only band, which is where all six orphaned issues came from.

### DIFF
--- a/.github/workflows/pr-hygiene.yml
+++ b/.github/workflows/pr-hygiene.yml
@@ -186,8 +186,13 @@ jobs:
             fi
             checked=$((checked + 1))
 
-            resolve_issue "${number}"
-            case $? in
+            # `|| status=$?` is load-bearing. A bare call under `set -e`
+            # terminates the step on status 1 or 2, so one invalid candidate
+            # ahead of a valid one would reject the PR and an outage would
+            # exit red instead of following the fail-open path below.
+            status=0
+            resolve_issue "${number}" || status=$?
+            case "${status}" in
               0)
                 echo "OK: the rendered body links to issue #${number}."
                 echo "::notice::This PR references an issue without closing it. If it completes the issue, use 'Closes #N' so the board stays accurate."
--- a/.github/workflows/pr-hygiene.yml
+++ b/.github/workflows/pr-hygiene.yml
@@ -80,9 +80,13 @@ jobs:
           # contributor's problem, and a hygiene gate must not hold a merge
           # hostage to infrastructure -- but it also must not fail open
           # silently, so every fallback announces itself.
+          # Diagnostics go to stderr. `closing=$(api_retry ...)` captures
+          # stdout, so a warning printed there would be concatenated with the
+          # count and turn a successful retry into `integer expression
+          # expected` -- silently discarding a valid closing link.
           api_retry() {
             "$@" && return 0
-            echo "::warning::GitHub API call failed; retrying once."
+            echo "::warning::GitHub API call failed; retrying once." >&2
             sleep 5
             "$@" && return 0
             return 1
@@ -146,30 +150,57 @@ jobs:
           # shorthand it autolinks itself -- but an author-supplied destination
           # is rendered verbatim, so `[x](/owner/repo/issues/0)` yields an
           # anchor for an issue that does not exist. Resolve each candidate.
-          # Capped so a link-heavy body cannot fan out into API calls.
+          resolve_issue() {
+            # 0 = a real issue, 1 = definitively not, 2 = could not tell.
+            if payload=$(gh api "repos/${OWNER}/${REPO}/issues/$1" 2>"${RUNNER_TEMP}/api-err"); then
+              # The REST issues endpoint also serves pull requests; only a
+              # pull request carries a `pull_request` object.
+              if [ "$(printf '%s' "${payload}" | jq -r 'has("pull_request")')" = "true" ]; then
+                echo "  #$1: resolves to a pull request, not an issue"
+                return 1
+              fi
+              return 0
+            fi
+            if grep -qi "HTTP 404" "${RUNNER_TEMP}/api-err"; then
+              echo "  #$1: linked, but no such issue in this repository"
+              return 1
+            fi
+            # Rate limiting, 5xx, DNS. Treating these as "no such issue" would
+            # turn a GitHub outage into a red check on a correctly linked PR.
+            echo "::warning::Could not resolve #$1: $(tr -d '\n' < "${RUNNER_TEMP}/api-err")" >&2
+            return 2
+          }
+
+          # Bounded so a link-heavy body cannot fan out into API calls. Hitting
+          # the cap is *undecidable*, not a failure: the eleventh candidate may
+          # be the valid one, and candidates arrive in numeric order rather
+          # than by likelihood.
+          RESOLVE_CAP=20
           checked=0
+          undecided=0
           for number in ${candidates}; do
-            if [ "${checked}" -ge 10 ]; then
-              echo "::warning::More than 10 linked issues; stopped resolving after 10."
+            if [ "${checked}" -ge "${RESOLVE_CAP}" ]; then
+              echo "::warning::More than ${RESOLVE_CAP} linked issues; stopped resolving. Treating the result as undetermined rather than as a missing link."
+              undecided=1
               break
             fi
             checked=$((checked + 1))
 
-            if ! payload=$(gh api "repos/${OWNER}/${REPO}/issues/${number}" 2>/dev/null); then
-              echo "  #${number}: linked, but no such issue in this repository"
-              continue
-            fi
-            # The REST issues endpoint also serves pull requests; only a pull
-            # request carries a `pull_request` object.
-            if [ "$(printf '%s' "${payload}" | jq -r 'has("pull_request")')" = "true" ]; then
-              echo "  #${number}: resolves to a pull request, not an issue"
-              continue
-            fi
-
-            echo "OK: the rendered body links to issue #${number}."
-            echo "::notice::This PR references an issue without closing it. If it completes the issue, use 'Closes #N' so the board stays accurate."
-            exit 0
+            resolve_issue "${number}"
+            case $? in
+              0)
+                echo "OK: the rendered body links to issue #${number}."
+                echo "::notice::This PR references an issue without closing it. If it completes the issue, use 'Closes #N' so the board stays accurate."
+                exit 0
+                ;;
+              2) undecided=1 ;;
+            esac
           done
+
+          if [ "${undecided}" -eq 1 ]; then
+            echo "::warning::Could not determine whether a linked issue exists; skipping the issue-link check."
+            exit 0
+          fi
 
           # --- fail ------------------------------------------------------------
           echo "::error::No issue link found. Do one of the following:"
--- a/.github/workflows/pr-hygiene.yml
+++ b/.github/workflows/pr-hygiene.yml
@@ -101,12 +101,42 @@ jobs:
           # audited PRs looked linked while leaving their issues orphaned.
           printf '%s' "${PR_BODY}" > "${RUNNER_TEMP}/pr-body.md"
 
-          if python3 scripts/check-pr-issue-link.py --body-file "${RUNNER_TEMP}/pr-body.md"; then
+          candidates=$(python3 scripts/check-pr-issue-link.py \
+            --body-file "${RUNNER_TEMP}/pr-body.md" || true)
+
+          # `#N` is a shared namespace: issues and pull requests draw from it,
+          # and nothing stops an author writing `Refs #0` or `See PR #1735`.
+          # A number that is missing, or that resolves to a pull request, is
+          # not an issue trail. Resolve each candidate before accepting it.
+          # Capped so a body full of numbers cannot fan out into API calls.
+          checked=0
+          for number in ${candidates}; do
+            if [ "${checked}" -ge 10 ]; then
+              echo "::warning::More than 10 candidate references; stopped resolving after 10."
+              break
+            fi
+            checked=$((checked + 1))
+
+            if ! payload=$(gh api "repos/${OWNER}/${REPO}/issues/${number}" 2>/dev/null); then
+              echo "  #${number}: no such issue or pull request in this repository"
+              continue
+            fi
+            # The REST issues endpoint also serves pull requests; only a pull
+            # request carries a `pull_request` object.
+            if [ "$(printf '%s' "${payload}" | jq -r 'has("pull_request")')" = "true" ]; then
+              echo "  #${number}: resolves to a pull request, not an issue"
+              continue
+            fi
+
+            echo "OK: body references issue #${number}."
             echo "::notice::This PR references an issue without closing it. If it completes the issue, use 'Closes #N' so the board stays accurate."
             exit 0
-          fi
+          done
 
           # --- fail ------------------------------------------------------------
+          if [ -n "${candidates}" ]; then
+            echo "::error::The PR body mentions $(printf '%s' "${candidates}" | wc -w | tr -d ' ') number(s), but none resolve to an issue in this repository."
+          fi
           echo "::error::No issue link found. Do one of the following:"
           echo "::error::  1. Add 'Closes #123' to the PR body (this PR completes the issue)."
           echo "::error::  2. Add 'Refs #123' or 'Part of #123' (this PR is one slice of it)."
--- a/.github/workflows/pr-hygiene.yml
+++ b/.github/workflows/pr-hygiene.yml
@@ -11,9 +11,15 @@ name: PR Hygiene
 # not just a closing keyword. Most PRs here are slices of an epic and must NOT
 # auto-close their parent; forcing `Closes` on them would produce false closures.
 # Closing remains a human decision — this gate only guarantees the trail exists.
+#
+# Body scanning lives in scripts/check-pr-issue-link.py, which strips HTML
+# comments, code fences, inline code, and URLs first. Without that step the
+# unedited PULL_REQUEST_TEMPLATE.md — whose instructions mention #1234, #1256,
+# and #1258 inside comments — would satisfy the gate on its own.
 
 on:
   pull_request:
+    branches: [main]
     types: [opened, edited, synchronize, reopened, labeled, unlabeled]
 
 permissions:
@@ -26,6 +32,13 @@ jobs:
     name: Issue link present
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v7.0.0
+
+      - name: Set up Python
+        uses: actions/setup-python@v6.3.0
+        with:
+          python-version: "3.12"
+
       - name: Check for a linked or referenced issue
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -42,7 +55,7 @@ jobs:
           # --- exemptions -----------------------------------------------------
           # Written as a runtime check rather than a job-level `if` so that an
           # exempt PR still reports a green check instead of "skipped", which
-          # would block a required-status-check branch protection rule.
+          # would deadlock a required-status-check branch protection rule.
           case ",${PR_LABELS}," in
             *,no-issue,*)
               echo "Exempt: 'no-issue' label present."
@@ -64,7 +77,7 @@ jobs:
 
           # --- signal 1: GitHub's own closing link ----------------------------
           # Covers both `Closes #123` in the body and issues linked manually
-          # through the Development sidebar, which no body regex can see.
+          # through the Development sidebar, which no body scan can see.
           closing=$(gh api graphql \
             -f query='
               query($owner:String!, $repo:String!, $pr:Int!) {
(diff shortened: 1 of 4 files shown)


Q00:
Both blocking findings confirmed and fixed in 5753764.

## Blocker 1 — references hidden in non-rendered Markdown

Verified before fixing. The repository's own template passes the old rule:

```
$ grep -nE '#[0-9]+' .github/PULL_REQUEST_TEMPLATE.md
17:RFC #1256 §I5 requires every PR that touches `src/ouroboros/auto/` ...
20:slowdown observed in the PR-A/B/β/γ merge train (#1258).
32:A baseline R-run is at `~/.ooo-observability/` or in #1258 evidence;
53:<!-- Link issues this PR closes or references, e.g. "Fixes #1234", "Refs #1256". -->
```

All four are inside HTML comments. The gate was a no-op for any template-created PR.

Body scanning now lives in `scripts/check-pr-issue-link.py`, following the `scripts/check-*.py` + `tests/unit/scripts/` convention the auto-boundary and perf-budget guards already use. It strips HTML comments, fenced blocks (backtick and tilde), inline code spans, and URLs before scanning, so it sees what a reviewer sees. Cross-repo references (`astral-sh/ruff#12345`) and Markdown headings stopped counting as a side effect.

Regression coverage is the one you asked for — `test_unedited_pull_request_template_is_rejected` reads the real `.github/PULL_REQUEST_TEMPLATE.md`. There is also `test_template_actually_contains_issue_numbers`, so that deleting the `#N` examples from the template can't make the first test pass vacuously. 15 tests, all passing.

## Blocker 2 — sidebar linking leaves the check stale

Correct, and the old error message advertised exactly that dead end. Linking through the Development sidebar fires no `pull_request` activity, so the failed check would sit red while the contributor believed they had remediated it.

The sidebar is no longer offered as a remedy. The failure output now reads:

```
Editing the PR body re-runs this check automatically. Linking through
the Development sidebar does NOT — GitHub fires no pull_request event
for it, so re-run this job from the Checks tab afterwards.
```

`closingIssuesReferences` still runs first, so a sidebar link is honoured once the job re-runs — it just isn't presented as a self-triggering fix.

## On the unreproducible 80-PR replay

Fair. The original number came from an ad-hoc local query with no artifact. The replay is now reproducible from the committed script:

```
$ gh pr list --state merged --limit 80 --json number,title,body,mergedAt \
    --jq '.[] | select(.mergedAt > "2026-06-28")' > /tmp/prs.jsonl
$ python3 - <<'PY'
import json, subprocess, sys
rows = [json.loads(l) for l in open('/tmp/prs.jsonl') if l.strip()]
exempt = blocked = ok = 0
for r in rows:
    if r['title'].startswith(('chore(deps)', 'chore(release)', 'release:')):
        exempt += 1; continue
    p = subprocess.run([sys.executable, 'scripts/check-pr-issue-link.py',
                        '--body', r['body'] or ''], capture_output=True)
    blocked += bool(p.returncode); ok += not p.returncode
print(f'exempt {exempt} / pass {ok} / blocked {blocked}')
PY
exempt 5 / pass 58 / blocked 17
```

Same 17 as before the fix. Comment-stripping newly catches four dependabot bumps whose only `#N` were foreign-repo anchor tags, and those are exempt upstream regardless.

Q00:
Both blockers confirmed and fixed in 55e2ada.

## Blocker 1 — non-linking constructs satisfied the gate

Reproduced against the pre-fix script before changing anything:

```
&#1234;                    -> [1234]   numeric character entity
&#x4d2;                    -> [1234]   hex character entity
<pre>error at #4242</pre>  -> [4242]   raw-text HTML block
<code>#4242</code>         -> [4242]
"    traceback at #4242"   -> [4242]   indented code block
```

The entity case is the sharpest one you raised: `&#1234;` renders as `Ӓ`, not as a link — its digits merely happen to follow a `#`.

`visible_text()` now also strips raw-text HTML blocks *with their content* (`pre`, `code`, `script`, `style`, `kbd`, `samp`, plus an unclosed trailing one), indented code blocks (four spaces or a tab), remaining tags, and finally character entities. Order matters: entities are removed after tags, because stripping `<...>` first would leave a bare `&#1234;` looking like prose. `&` was also added to the negative lookbehind as defence in depth.

Indented-code stripping is deliberately conservative — it can also swallow a reference buried in a deeply indented list. That costs a contributor one body edit; the opposite error costs the gate. This trade-off is written into the code comment.

## Blocker 2 — matched numbers were never resolved

Correct, and your three examples all passed before: `Refs #0`, a nonexistent number, and `See PR #1735`.

The script no longer decides acceptance. It reports *candidates* — one number per line, sorted and deduplicated — and the workflow resolves each against `repos/{owner}/{repo}/issues/{n}`, rejecting anything that 404s or carries a `pull_request` object. Verified live against this repository:

```
#1777   -> issue,        accepted
#1735   -> pull request, rejected
#0      -> absent,       rejected
#999999 -> absent,       rejected
```

Resolution is capped at ten candidates so a number-heavy body cannot fan out into API calls. The cap emits a `::warning::` rather than silently narrowing coverage.

## Coverage

26 tests, up from 15. New cases: decimal and hex entities, `<pre>`/`<code>`/`<samp>`, an unclosed `<pre>`, space- and tab-indented code, plus `test_a_real_reference_survives_alongside_every_masked_form`, which puts a genuine `Closes #1777` in a body that also contains all six masked forms and asserts only `1777` is reported — so the new stripping cannot become so eager that it swallows the real link.

Replaying over the same 80 merged PRs: **5 exempt / 58 pass / 17 blocked**, unchanged by either fix.

Q00:
All three confirmed and fixed in 9cab94b. Reproduced against the current script first:

```
<!-- Refs #1777          -> [1777]   unterminated HTML comment
[hidden]: #1777          -> [1777]   link reference definition
>     trace #1777        -> [1777]   blockquoted indented code
```

You were right on each: an unterminated `<!--` comments out the rest of the document, a link reference definition renders only where its label is used, and four spaces after a `>` marker is a code block inside a quote. None produce a visible link.

Three additions to `visible_text()`:

- `_UNCLOSED_HTML_COMMENT` — removes a trailing `<!--` and everything after it, mirroring the unclosed-raw-text-HTML rule added last round.
- `_LINK_REFERENCE_DEFINITION` — removes definition lines.
- `_BLOCKQUOTE_PREFIX` — strips `>` markers **before** code detection, so blockquoted code reaches the indented-code and fence rules.

The blockquote rule needed care in both directions, and both are now pinned:

- It strips the *prefix*, not the line. Quoted **prose** is visible and must keep counting — `> Closes #1777 per review.` is a real trail. `test_blockquoted_prose_still_counts`.
- It is line-anchored, so prose arrows are untouched — `The flow is A -> B, and this Closes #1777.` still reports `1777`. `test_arrow_operators_are_not_blockquote_markers`.

Also added `test_reference_survives_an_unclosed_comment_later_in_the_body`, so the new stripping cannot swallow a genuine `Closes #1777` that precedes a malformed comment.

33 tests, up from 26. Replaying over the same 80 merged PRs: **5 exempt / 58 pass / 17 blocked**, unchanged across all three rounds of fixes.

On the general shape of these findings — this is the third round where the gap was "text that looks like a reference but renders as something else". If a fourth surfaces, I would rather switch `visible_text()` to a real CommonMark parse than keep extending the regex set; regexes are the wrong tool for Markdown context and I do not want to keep proving that one construct at a time. Say the word if you would prefer that now instead.

Q00:
Confirmed, and I took your recommendation rather than patching again — 2761fcf.

Both cases reproduced against the previous scanner: `[details](#1777)` emitted `1777` from the link destination, and the multiline reference definition emitted it from the continuation line.

## The pattern was the finding

Four rounds, four constructs, same shape:

| round | construct |
|---|---|
| 1 | HTML comments (the unedited template) |
| 2 | character entities, raw-text HTML, indented code |
| 3 | unclosed comments, reference definitions, blockquoted code |
| 4 | inline link destinations, multiline definition titles |

Markdown context is not a regular language, so each round bought exactly one construct. I said last round that a fourth would mean switching to a real parse; you independently asked for the same thing.

## What I did instead of a CommonMark dependency

Asked GitHub. The workflow renders the body through `POST /markdown` with `mode=gfm` and `context=owner/repo`, and the script now reads issue anchors out of that HTML with stdlib `html.parser`.

Verified against the live renderer across fourteen constructs — only prose autolinks:

```
Closes #1777.               -> issues/1777
See [details](#1777).       -> (none)
<!-- Refs #1777 -->         -> (none)
<!-- Refs #1777             -> (none)
[hidden]: #1777             -> (none)
[details]: /somewhere
  "Refs #1777"              -> (none)
```fenced```, indented, blockquoted code, `inline`, &#1777;, <pre> -> (none)
> Closes #1777 per review.  -> issues/1777
```

## It also removes the REST loop from round 3

Two things the scanner could never answer come free from the renderer:

- **type** — GitHub emits `/issues/N` for an issue and `/pull/N` for a pull request, so `See PR #1735` stops counting;
- **existence** — GitHub autolinks only numbers that exist. `#0` and `#999999` render as plain text.

So the per-candidate `gh api repos/.../issues/N` resolution added last round is gone. Net: ~60 lines of accumulated stripping regexes replaced by an anchor collector.

## Fixtures are GitHub's actual output

Every fixture in `test_rendered_context_decides` is the HTML GitHub returned for the Markdown noted beside it, captured via the same endpoint. They document the renderer's behaviour rather than asserting my model of it — which is the failure mode that produced the previous three rounds.

## End-to-end

- This PR's own body → 8 issue references, with `#1735` correctly excluded as a pull request.
- The unedited `PULL_REQUEST_TEMPLATE.md` → still rejected, which is the regression this gate exists for.

A rendering failure fails **open** with a `::warning::`. A GitHub API outage is not the contributor's problem, and a hygiene gate should not block a merge on infrastructure.

26 tests.

Q00:
All four addressed in eb74af5.

## Blocker 1 — author-supplied destinations were trusted

You are right about the boundary, and I had the model wrong: GitHub existence- and type-checks the shorthand *it* autolinks, but a destination the author writes is rendered verbatim. So my "the renderer vouches for it" claim only held for half the cases.

Verified end to end against a body containing all three of your cases:

```markdown
An explicit link to [nothing](https://github.com/Q00/ouroboros/issues/0) and an
empty one [](https://github.com/Q00/ouroboros/issues/1777), plus a PR link
[pr](https://github.com/Q00/ouroboros/pull/1735).
```

```
script candidates: 0
  #0 -> no such issue in this repository (rejected)
result: nothing accepted
```

The empty anchor and the `/pull/` link never became candidates; `0` did, and the repository resolution rejected it.

Two changes: anchors must carry visible content, and the per-candidate resolution loop is back. I removed it in the previous commit believing the renderer subsumed it — that was wrong for author-written hrefs, and it now has a reason to exist.

"Visible content" means text **or** an element: an `` badge or a nested `<code>` still counts, since a reader can see and click both. Pinned by `test_non_text_anchor_content_still_counts_as_visible`.

## Blocker 2 — `develop`

Added. Honest note on scope: `develop` does not currently exist on the remote and all 80 recent PRs target `main`, so this closes the stated gap rather than an observed one — but the workflow claimed to cover every PR while `test.yml`/`lint.yml` list `[main, develop]`, and that inconsistency was real.

## Warning — URL normalization

Fixed, and I agree this one matters more than its priority suggests: a false red blocks a valid PR. Owner and repository now compare case-insensitively, a trailing slash is tolerated, `www.` is accepted, and the href is stripped before matching. `test_equivalent_urls_resolve` covers all four.

## Warning — `set -e` short-circuiting the fallback

Correct, and the inconsistency was worse than the outage: the two API calls behaved differently under identical conditions. Both now go through one `api_retry` helper (retry once, then warn and fail open), and the assignment is separated from its error check so `set -e` cannot skip the fallback.

## P2 — fail-open reporting green

Kept, with the retry and a `::warning::` on every fallback path so a skip is visible rather than silent. My reasoning: a hygiene gate should not hold a merge hostage to a GitHub outage, and a contributor cannot act on one. I agree it deserves revisiting before this is ever promoted to a required check — at that point an explicit neutral outcome would preserve status semantics better than green. Flagged in the commit message rather than silently decided.

36 tests; `bash -n` on the extracted step script passes.

Q00:
All five addressed in ad82a9f.

## Blocker 1 — the retry warning corrupted the captured count

Sharp catch. `api_retry` printed to stdout while `closing=$(api_retry ...)` captured stdout, so a failed first request followed by a **successful** retry produced `::warning...\n1` and `[ "${closing}" -gt 0 ]` reported `integer expression expected`. A sidebar-only link could fail *because* recovery worked. Diagnostics moved to stderr.

## Blocker 2 — outage read as absence

Correct, and it inverted the workflow's own contract. `resolve_issue` now returns three states instead of two:

```
0  a real issue
1  definitively not (404, or a /pull/ payload)
2  undecidable (5xx, rate limit, DNS)
```

An undecidable run warns and passes rather than failing on evidence it never gathered.

## Warning 3 — the cap could hide the valid candidate

Right, and the ordering makes it worse: candidates arrive numerically, not by likelihood, so ten low invalid destinations shadow a valid eleventh. The cap is now 20, but the real fix is that hitting it is **undetermined** rather than "no issue found" — same treatment as an API failure.

## Warning 4 — invisible labels

Both probes were valid. `<a><span></span></a>` passed because I marked *any* nested element visible, and a zero-width-only label passed because `.strip()` does not remove U+200B. Visibility now means text with at least one glyph after zero-width characters (U+200B–200D, U+2060, U+FEFF, U+00AD, U+180E) and whitespace are removed, or a genuinely self-rendering element (`img`, `picture`, `svg`, `video`). `<code>#1777</code>` still counts, through its text.

## Warning 5 — unbounded `int()`

Confirmed: a 4,301-digit destination raises `ValueError` past CPython's integer-string limit, crashing the parser and reddening a valid PR. The pattern is bounded to nine digits.

## The shell helpers are now tested, not reasoned about

You asked for a first-failure/second-success shell test, and I took it further than the one case. `tests/unit/scripts/test_pr_hygiene_workflow.py` extracts `api_retry` and `resolve_issue` **from the workflow YAML** and runs them against a stub `gh` — so the tests exercise the workflow's own copy rather than a duplicate that can drift out of sync.

Covered: first-failure/second-success keeps stdout a bare integer with the warning on stderr; two failures return non-zero; and 404 / pull request / HTTP 500 / rate limit / DNS failure each map to the right one of the three states.

289 tests in `tests/unit/scripts/`.

## On the repeated P2

Agreed in principle, and the surface is now larger, not smaller — render failure, closing-reference failure, undecidable resolution, and the cap all report green with a warning. I did not fix it here because GitHub Actions exposes no *neutral* conclusion to a `run:` step; getting one requires a composite action or an explicit Checks API call. That is the right change to make before this becomes a required check, and I would rather do it as its own PR than bolt it on here. Recorded in the commit message rather than quietly dropped.

Q00:
Fixed in d9a9a07, and you also diagnosed why it reached you.

## Blocker — `set -e` killed the loop

Confirmed by driving the workflow's own loop with a stub returning 1 for the first candidate and 0 for the second:

```
bare call        exit=1, issue #1777 never reached  -> PR falsely rejected
|| status=$?     exit=0, issue #1777 recognized
```

So one invalid candidate ahead of a valid one rejected the PR, and an outage exited red instead of taking the fail-open path written three lines below it. `resolve_issue "${number}" || status=$?` captures the status without arming `set -e`, with the reason recorded next to it so the guard is not "simplified" away later.

## You were right about the test being the root cause

`TestResolveIssue` ran with `set -uo pipefail` — I omitted the `-e` the workflow actually uses, so the helper behaved differently in the test than in production. That is the specific gap that let this ship, and it is worth more than the fix.

Two changes:

- the header now matches the workflow's `set -euo pipefail` verbatim;
- `TestResolutionLoopUnderSetE` extracts **the whole loop**, not just a helper, and drives it under `set -e` across four arrangements: invalid-then-valid, all-invalid, outage-alone, and outage-then-valid.

Testing the helper in isolation was never going to catch a control-flow bug in the caller.

## Follow-up 1 — invisible characters

Accepted, and the denylist was the wrong shape for the same reason the Markdown regexes were: it enumerated instead of classifying, so U+2061 got through and the next one would have too. Visibility is now decided by Unicode general category — `Cc`, `Cf`, `Cs`, `Co`, `Cn`, `Zs`, `Zl`, `Zp` render nothing; everything else does. Verified that emoji (`So`), CJK (`Lo`), and U+3164 HANGUL FILLER (`Lo`) still count while U+2061, U+200B, U+00AD, and plain whitespace do not.

## Follow-up 2 — `jq`

Fixed; the suite skips when either `bash` or `jq` is missing, since the workflow uses both.

293 tests in `tests/unit/scripts/`.

PR #1784, opened 2026-07-27
fix(config): parse .env with python-dotenv instead of Python string syntax

### LINKED ISSUE #1783: .env values are parsed with Python string syntax, silently corrupting single-quoted paths
## Summary

`_parse_env_value` in `src/ouroboros/config/loader.py:146` parses `.env` values by handing them to `ast.literal_eval`:

```python
if candidate[0] in {'"', "'"} and candidate[-1:] == candidate[0]:
    try:
        parsed = ast.literal_eval(candidate)
    except (SyntaxError, ValueError):
        return candidate[1:-1]
    return str(parsed)
```

That applies **Python string grammar** to a POSIX-shaped `.env` file. The two grammars disagree, and the disagreement destroys data.

## Reproduction

```
$ python -c "
from ouroboros.config.loader import _parse_env_value
v = _parse_env_value(r\"'C:\new\table'\")
print(len(v), any(ord(c) < 0x20 for c in v))
"
10 True
```

The value `C:\new\table` is 12 characters. It comes back as 10, with `\n` turned into a newline and `\t` into a tab. **Any single-quoted Windows path in a `.env` is silently corrupted.**

In POSIX shells, in `python-dotenv`, and in every mainstream `.env` implementation, single quotes are literal — no escape processing. Verified against `python-dotenv` 1.2.2:

| Line | current | python-dotenv |
|---|---|---|
| `X='C:\new\table'` | 10 chars, control bytes | 12 chars, literal |
| `X='a\nb'` | 3 chars, real newline | 4 chars, literal `a\nb` |
| `X="abc" "def"` | `abcdef` (Python implicit concatenation) | line rejected |
| `X="tail" junk` | `"tail" junk` | line rejected |
| `X='it''s'` | `its` | line rejected |

The `"abc" "def"` row is the sharpest: Python's implicit string concatenation silently invents a value that appears nowhere in the file.

## Scope of the risk

`.env` is a documented trust boundary here — `_UNTRUSTED_ENV_DENYLIST` blocks a cloned repository's `.env` from setting `PATH`, `NODE_OPTIONS`, CLI path overrides, and permission-mode keys, all RCE sinks.

**That guard is key-based and is not weakened by this bug.** The damage is confined to values: API keys, tokens, and paths that contain a backslash or a quote arrive mangled, and the failure is silent — the user sees an authentication or path error far from the cause.

## Proposal

Delegate parsing to `python-dotenv`, which owns this grammar, and keep every policy decision in our code:

- key-based denylist for untrusted sources, unchanged;
- placeholder-API-key skipping, unchanged;
- "never override an already-set real environment value" precedence, unchanged;
- `interpolate=False`, so `${VAR}` stays literal exactly as today. Leaving interpolation on would be both a behaviour change and a new capability for an untrusted project `.env` to read the real process environment into a value.

`python-dotenv` currently reaches installs only as a transitive dependency of the `mcp` / `claude` / `all` extras, so it must be declared as a **core** dependency: `config/loader.py` runs on every install, including a bare `pip install ouroboros-ai`.

It should be **exact-pinned**, following the precedent set for the AI stack in #1347: this package parses a documented trust boundary, so an unreviewed upgrade is a supply-chain path.

## Acceptance criteria

- [ ] Single-quoted values are literal; `X='C:\new\table'` round-trips as 12 characters with no control bytes.
- [ ] Double-quoted escape handling is unchanged.
- [ ] Malformed lines are rejected rather than silently yielding an invented value.
- [ ] `${VAR}` and `$VAR` remain literal (no interpolation).
- [ ] Every existing test in `tests/unit/config/test_loader_env.py` passes untouched, including the full denylist suite.
- [ ] `python-dotenv` is declared as a core dependency with an exact pin.
- [ ] `_parse_env_value` is removed; it has no other caller.

Q00:
## Summary

`_parse_env_value` handed quoted `.env` values to `ast.literal_eval`, applying **Python** string grammar to a POSIX-shaped file. Parsing is now delegated to `python-dotenv`, which owns that grammar. All trust policy stays in `config/loader.py`.

Closes #1783.

## The bug

```
_parse_env_value(r"'C:\new\table'")  ->  10 chars, contains control bytes
```

The value is 12 characters. `\n` became a newline and `\t` a tab, because single quotes are literal everywhere except in Python. **Any single-quoted Windows path in a `.env` was silently corrupted**, surfacing later as a path error far from its cause.

Measured against python-dotenv 1.2.2:

| Line | before | after |
|---|---|---|
| `X='C:\new\table'` | 10 chars, control bytes | 12 chars, literal |
| `X='a\nb'` | 3 chars, real newline | 4 chars, literal |
| `X="abc" "def"` | `abcdef` | line rejected |
| `X="tail" junk` | `"tail" junk` | line rejected |
| `X='it''s'` | `its` | line rejected |
| `X="a\nb"` | real newline | real newline (unchanged) |

The `"abc" "def"` row is the sharpest — Python's implicit string concatenation invented a value that appears nowhere in the file.

## Security posture is unchanged, and pinned down by tests

`.env` is a documented trust boundary here. The bug was value-side only; `_UNTRUSTED_ENV_DENYLIST` is key-based and was never weakened. Delegating the grammar must not delegate the policy, so:

- the untrusted-source key denylist still runs per key — **all 110 pre-existing tests pass untouched**, including the full denylist suite;
- placeholder-key skipping and the "never override an already-set real environment value" precedence are unchanged;
- **`interpolate=False`** is deliberate. python-dotenv expands `${VAR}` from the environment by default. Leaving it on would change today's behaviour (`$HOME` is currently literal) *and* hand an untrusted project `.env` a way to read the real process environment into a value it controls. `test_untrusted_env_cannot_interpolate_a_denied_key` pins that a cloned repo's `.env` cannot exfiltrate `PATH` via `${PATH}`.

## Dependency: core, and exact-pinned

`python-dotenv` previously reached installs only as a transitive dependency of the `mcp` / `claude` / `all` extras:

```
python-dotenv v1.2.2
└── pydantic-settings v2.13.1
    └── mcp v1.28.1
        └── ouroboros-ai (extra: mcp)
```

`config/loader.py` runs on **every** install, including a bare `pip install ouroboros-ai`, so it is declared as a core dependency rather than relying on an extra.

It is **exact-pinned** (`==1.2.2`) rather than range-pinned like its neighbours, because this package parses a documented trust boundary and an unreviewed upgrade is a supply-chain path — the same rationale as the AI-stack pins from #1347. The comment in `pyproject.toml` records that reasoning so the deviation from the surrounding style is not mistaken for an accident.

Honest cost: a downstream resolver needing a different python-dotenv will conflict. `mcp` → `pydantic-settings` requires `>=0.21.0`, which `==1.2.2` satisfies today.

## Test plan

- [x] `uv run pytest tests/unit/config/test_loader_env.py -q` → **129 passed** (110 pre-existing, unmodified + 19 new)
- [x] `uv run pytest tests/unit/config/ -q` → 421 passed
- [x] New `TestEnvValueGrammar` covers: the Windows-path regression (asserting length 12 and no control bytes), single-quote literalness, unchanged double-quote escapes, malformed-line rejection, `$VAR`/`${VAR}` non-interpolation, the `PATH` exfiltration guard, seven established forms that must not change, and a denylist re-assertion through the new parser
- [x] `_parse_env_value` removed — it had no other caller
- [x] `uv run ruff check` / `ruff format --check` clean
- [x] `uv run mypy src/ouroboros/config/loader.py` → Success
- [x] `uv lock` regenerated

## R-run comparison


Q00:
Blocker confirmed and fixed in 25569af. Two commits on this branch since the review:

## Blocker — a hostile `.env` could stop every command from starting

Reproduced exactly as described:

```
>>> dotenv_values(stream=io.StringIO("'BROKEN=KEY'=value\n"), interpolate=False)
OrderedDict({'BROKEN=KEY': 'value'})
>>> os.environ["BROKEN=KEY"] = "value"
ValueError: illegal environment variable name
```

And `_load_env_file(Path(".env"))` runs at module import, so this is reachable from a cloned repository's `.env` as a denial of service against the whole CLI. The previous hand-rolled parser could not produce such a key — it split on the first `=` — so delegating the grammar introduced this, and it is worse than the value corruption the delegation was fixing. Thank you for catching it.

Fix is `_is_assignable_env_key`: reject empty keys and any key containing `=`, NUL, or whitespace. That is a superset of what CPython refuses, and it matches the previous parser, which also skipped keys with whitespace. The assignment is additionally wrapped in `except ValueError` as defence in depth — this module is imported at startup, so a future parser change must degrade to skipping one entry, never to an unstartable process.

Coverage, including the import-path regression you asked for:

- `test_import_survives_a_hostile_project_env` launches a **fresh interpreter** with its cwd holding the hostile `.env` and asserts `import ouroboros.config.loader` still exits 0. Testing `_load_env_file` alone would not have covered the module-level call that a crash actually takes down.
- `test_hostile_key_does_not_stop_later_entries` pins that one bad line costs one entry, not the rest of the file.
- `_is_assignable_env_key` is parametrized over `=`, space, tab, empty, and NUL keys, plus four ordinary keys that must still be accepted.

## Also fixed — CI failure on all three Python versions

```
AssertionError: Runtime dependency missing upper bound: python-dotenv==1.2.2
```

`test_runtime_and_optional_dependencies_have_upper_bounds` already encodes this repository's position, and its docstring is explicit: core runtime deps stay bounded *ranges* and must use `<`; `==` is reserved for the optional AI extras. python-dotenv is a core dependency, so the exact pin was the wrong instrument.

Now `>=1.2.2,<1.3.0` — a ceiling deliberately tighter than the major-version bounds on its neighbours, since this package parses a documented trust boundary and only patch releases should flow without review. `uv.lock` pins 1.2.2 exactly, so installs stay reproducible regardless.

`tests/unit/config` → 433 passed.

## Unrelated failure seen in the same run

`tests/unit/evolution/test_generation_watchdog.py::test_busy_run_without_material_progress_times_out` failed with `assert 'idle_timeout' == 'no_material_progress_timeout'`. That file passes in isolation (15/15) and this PR does not touch the watchdog — it looks like a race between two timeout kinds under `xdist` load, not a consequence of this change.

Q00:
Confirmed and fixed in f7a0354.

Reproduced in a fresh interpreter before fixing — `BROKEN=\xFF` in the cwd's `.env`:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 7: invalid start byte
```

`UnicodeDecodeError` is a `ValueError`, not an `OSError`, so the existing clause did not catch it. And `_load_env_file` runs at module import, so this took down every command — the same denial-of-service shape as the illegal key name in 25569af, reached by a different door.

## Why I did not just add `UnicodeDecodeError`

That is the fix, and it is not sufficient on its own. This is the **second** startup-killing failure mode found in this one function in two review rounds. Enumerating what `dotenv_values` can raise is not converging, and each miss costs the whole CLI rather than one variable.

So: both known cases are named for their comments, and a backstop guarantees the invariant.

```python
except (OSError, UnicodeDecodeError):
    # ... `UnicodeDecodeError` is a `ValueError`, not an `OSError` ...
    return
except Exception:  # noqa: BLE001 - startup must survive any parser failure
    # This module runs `_load_env_file` at import, so a malformed `.env`
    # must degrade to "no variables loaded", never to an unstartable
    # process. Enumerating the parser's failure modes has already missed
    # two, so the invariant is enforced here rather than predicted.
    return
```

Same reasoning I applied to the validator in #1786 after the second escaping exception type there.

## Coverage

`test_malformed_env_files_do_not_abort_import` runs the **real import path** — a fresh interpreter whose cwd holds the malformed file — parametrized over three inputs:

- `'BROKEN=KEY'=value` (illegal key name)
- `BROKEN=\xFF` (invalid UTF-8)
- 32 bytes of binary garbage

Plus `test_non_utf8_env_leaves_the_environment_untouched`, which pins that a file rejected at parse time sets nothing at all, rather than partially applying.

`tests/unit/config/test_loader_env.py` → 145 passed; `tests/unit/config` → 437 passed.

PR #1786, opened 2026-07-27
fix(providers): share response_format handling so the schema guard reaches every adapter

### LINKED ISSUE #1785: A malformed json_schema crashes four LLM adapters; only zcode got the guard
## Summary

`_validate_response_format_payload` is copy-pasted across five provider adapters. One copy was fixed; the other four were not.

`src/ouroboros/providers/zcode_cli_adapter.py:314` validates the schema before using it:

```python
Draft202012Validator.check_schema(schema_payload)
Draft202012Validator(schema_payload).validate(parsed)
except JsonSchemaError as exc:
    return f"invalid JSON schema: {exc.message}"
except JsonSchemaValidationError as exc:
    return exc.message
```

`pi_llm_adapter.py:154`, `gjc_llm_adapter.py:131`, `ourocode_llm_adapter.py:345`, and `goose_cli_adapter.py:170` have no `check_schema` call and catch only `JsonSchemaValidationError`.

## Impact

If a caller supplies a `json_schema` that is not itself a valid JSON Schema, those four adapters raise an uncaught exception instead of returning a validation message.

```python
>>> Draft202012Validator({"type": "not-a-real-type"}).validate({"a": 1})
jsonschema.exceptions.UnknownType: Unknown type 'not-a-real-type' for validator ...
```

`UnknownType` is not a `ValidationError`, so the `except JsonSchemaValidationError` clause does not catch it and it propagates to the caller. The function's contract is to *return* a string describing the problem; instead the adapter throws. `zcode` returns `"invalid JSON schema: ..."` for the same input.

`grep -c check_schema src/ouroboros/providers/*.py` matches exactly one file.

## Root cause

Duplication with no shared implementation. The bodies are byte-identical across copies:

| function | identical copies |
|---|---|
| `_build_response_format_directive` | **4** — pi, gjc, ourocode, zcode (plus two drifted variants in goose and copilot) |
| `_validate_response_format_payload` | **3** — pi, gjc, ourocode (goose differs only in `if/return` vs a conditional expression; zcode carries the fix) |

Nothing connects them, so a fix applied in one adapter cannot reach the rest. That is exactly what happened here.

## Proposal

Extract the shared behaviour into one module and have the adapters delegate:

- `build_response_format_directive()` — the 4-way identical version, no behaviour change.
- `validate_response_format_payload()` — zcode's corrected version, which fixes the crash for the other four.

## Out of scope

`goose` and `copilot` carry genuinely divergent **directive** builders — goose does not unwrap a nested `schema` key and omits the `json.dumps` fallback; copilot renders through module-level templates and unwraps differently. Converging those changes prompt text, which is a separate judgement. This issue leaves them alone and records the divergence.

## Acceptance criteria

- [ ] One implementation backs the directive builder for the four adapters that share it today.
- [ ] One implementation backs the validator, including `check_schema`.
- [ ] A malformed schema returns `"invalid JSON schema: ..."` from every adapter that validates, instead of raising.
- [ ] Directive output is byte-identical to today for all four adapters.
- [ ] Regression tests cover the malformed-schema path per adapter.

Q00:
## Summary

`_build_response_format_directive` and `_validate_response_format_payload` were copy-pasted across five provider adapters. One copy got a fix; the other four did not, and the gap is a crash. Both halves now live in `providers/response_format.py`.

Closes #1785.

## The bug

`zcode_cli_adapter` validates the schema before using it; pi, gjc, ourocode, and goose do not. Given a `json_schema` that is not itself a valid schema:

```python
>>> Draft202012Validator({"type": "not-a-real-type"}).validate({"a": 1})
jsonschema.exceptions.UnknownType: Unknown type 'not-a-real-type' for validator ...
```

`UnknownType` is **not** a `ValidationError`, so the lone `except JsonSchemaValidationError` clause did not catch it and it propagated to the caller. The function's contract is to *return* a reason string; four of five adapters threw instead. `grep -c check_schema src/ouroboros/providers/*.py` matched exactly one file.

## Why it happened

Nothing connected the copies, so a fix could not travel:

| function | identical copies before |
|---|---|
| `_build_response_format_directive` | 4 — pi, gjc, ourocode, zcode (verified by AST body hash) |
| `_validate_response_format_payload` | 3 — pi, gjc, ourocode; goose differs only in `if/return` vs a conditional expression; zcode carries the fix |

## Changes

- **`build_response_format_directive`** — the 4-way identical version. Verified byte-identical output against the pre-change implementation across nine inputs, including the OpenAI-style `schema` wrapper, a missing `type`, an unrecognized `type`, and an unserializable schema. **No behaviour change.**
- **`validate_response_format_payload`** — zcode's corrected version, which fixes the crash for the other four.
- Each adapter keeps its method and its docstring; only the body becomes a delegation. Those docstrings carry backend-specific rationale ("Pi JSON mode does not expose a Codex-style hard `--output-schema` flag") that a shared implementation cannot express, so they stay at the call site.

Net **-231 lines**.

## Deliberately out of scope

`goose` and `copilot` carry genuinely divergent **directive** builders — goose does not unwrap a nested `schema` key and omits the `json.dumps` fallback; copilot renders through module-level templates and unwraps differently. Converging those changes prompt text, which is a separate judgement this PR does not make. The divergence is recorded in #1785 rather than silently left unmentioned.

## A note on the original scope

This started as "converge the runtime adapter property boilerplate" — `_feeds_prompt_via_stdin`, `_build_permission_args`, `working_directory`/`permission_mode`/`llm_backend`. On inspection that would have been the wrong change: those bodies are trivial (`return False`, `return []`, `return self._cwd`) but each carries a **distinct** docstring naming the backend and its reason:

```
"""Return False — Copilot CLI accepts the prompt via the -p flag."""
"""Return False — Gemini CLI accepts the prompt via the --prompt flag."""
"""Goose permission mode is provided through GOOSE_MODE env."""
```

Deduplicating those destroys per-backend documentation to save three lines each, and the accessors exist to satisfy `AgentRuntime`, which is a `Protocol` — structural typing whose point is *not* requiring a shared base. Restricting the AST duplicate scan to functions above 30 nodes surfaced this response-format cluster instead: real duplicated logic, and a real bug.

## Test plan

- [x] New `tests/unit/providers/test_response_format.py` — 26 cases
- [x] `TestMalformedSchemaNoLongerCrashes` asserts the regression per adapter (pi, gjc, ourocode, goose, zcode), normalizing over the fact that `ourocode` declares its validator `@staticmethod` while the rest take `self`
- [x] Directive equivalence verified against the pre-change implementation across nine inputs — 0 differences
- [x] `uv run pytest tests/unit/providers/ -q` → 655 passed
- [x] `uv run pytest tests/unit/providers tests/unit/orchestrator -q` → **4754 passed, 1 skipped**
- [x] `uv run ruff check` / `ruff format --check` clean
- [x] `uv run mypy src/ouroboros/providers/` → Success (25 source files)

## R-run comparison

### DIFF
--- a/src/ouroboros/providers/gjc_llm_adapter.py
+++ b/src/ouroboros/providers/gjc_llm_adapter.py
@@ -11,9 +11,6 @@ import re
 from typing import Any
 from uuid import uuid4
 
-from jsonschema import Draft202012Validator
-from jsonschema.exceptions import ValidationError as JsonSchemaValidationError
-
 from ouroboros.config import get_gjc_cli_path
 from ouroboros.core.errors import ProviderError
 from ouroboros.core.json_utils import extract_json_payload
@@ -36,6 +33,10 @@ from ouroboros.providers.gjc_rpc_protocol import (
     validate_response_ack,
 )
 from ouroboros.providers.profiles import resolve_completion_profile_result
+from ouroboros.providers.response_format import (
+    build_response_format_directive,
+    validate_response_format_payload,
+)
 
 _SAFE_MODEL_PART_PATTERN = re.compile(r"^[A-Za-z0-9_.:@-]+$")
 
@@ -94,39 +95,7 @@ class GjcLLMAdapter(CodexCliLLMAdapter):
         response_format: dict[str, object] | None,
     ) -> str | None:
         """Translate response_format into cooperative GJC prompt instructions."""
-        if not response_format:
-            return None
-        fmt_type = response_format.get("type")
-        if fmt_type == "json_object":
-            return (
-                "Respond with ONLY a valid JSON object. Do not use markdown fences, "
-                "headers, or explanatory text."
-            )
-        if fmt_type == "json_schema":
-            schema = response_format.get("json_schema")
-            if not isinstance(schema, dict):
-                return None
-            schema_payload = (
-                schema.get("schema") if isinstance(schema.get("schema"), dict) else schema
-            )
-            top_type = (
-                schema_payload.get("type", "object")
-                if isinstance(schema_payload, dict)
-                else "object"
-            )
-            type_noun = {"array": "JSON array", "object": "JSON object"}.get(
-                str(top_type), "JSON value"
-            )
-            try:
-                rendered = json.dumps(schema_payload, indent=2, sort_keys=True)
-            except (TypeError, ValueError):
-                rendered = str(schema_payload)
-            return (
-                f"Respond with ONLY a valid {type_noun} that matches this schema. "
-                "Do not use markdown fences, headers, or explanatory text.\n\n"
-                f"JSON schema:\n{rendered}"
-            )
-        return None
+        return build_response_format_directive(response_format)
 
     def _validate_response_format_payload(
         self,
@@ -134,26 +103,7 @@ class GjcLLMAdapter(CodexCliLLMAdapter):
         response_format: dict[str, object],
     ) -> str | None:
         """Validate extracted JSON against the requested response_format."""
-        try:
-            parsed = json.loads(payload)
-        except json.JSONDecodeError as exc:
-            return f"invalid JSON: {exc}"
-
-        fmt_type = response_format.get("type")
-        if fmt_type == "json_object":
-            return None if isinstance(parsed, dict) else "expected a JSON object"
-        if fmt_type == "json_schema":
-            schema = response_format.get("json_schema")
-            if not isinstance(schema, dict):
-                return "json_schema response_format is missing a schema object"
-            schema_payload = (
-                schema.get("schema") if isinstance(schema.get("schema"), dict) else schema
-            )
-            try:
-                Draft202012Validator(schema_payload).validate(parsed)
-            except JsonSchemaValidationError as exc:
-                return exc.message
-        return None
+        return validate_response_format_payload(payload, response_format)
 
     def _compose_prompt(self, messages: list[Message]) -> str:
         return "\n\n".join(f"{message.role.value}: {message.content}" for message in messages)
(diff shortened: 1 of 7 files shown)


Q00:
Confirmed and fixed in c1b0c45.

Reproduced exactly as described:

```
>>> Draft202012Validator.check_schema({"$ref": "#/$defs/missing"})     # passes
>>> Draft202012Validator({"$ref": "#/$defs/missing"}).validate({"a": 1})
jsonschema.exceptions._WrappedReferencingError
MRO: _WrappedReferencingError -> _RefResolutionError
     -> referencing.exceptions.Unresolvable -> Exception
```

Neither a `SchemaError` nor a `ValidationError`, so both clauses missed it — including in the copy that already carried the `check_schema` guard, which the shared helper then inherited.

## Why I did not catch `Unresolvable` by name

That was the obvious fix and I think it is the wrong one twice over:

1. It requires `from referencing.exceptions import Unresolvable`. `referencing` is an **undeclared transitive dependency** of `jsonschema` — nothing else in this tree imports it, and `pyproject.toml` does not list it. Reaching into it would be the same fragility that PR #1784 is currently removing from `.env` handling.
2. It leaves the next such class uncovered. This is the **second** failure mode found in this function that raises from outside jsonschema's own hierarchy, after `UnknownType`. Enumerating them one review round at a time is not converging.

The caller-visible contract is that this function *reports* problems by returning them. So the contract is now enforced rather than enumerated: the specific clauses stay for their precise messages, and a final `except Exception` returns

```
could not validate against JSON schema: <ExceptionType>: <detail>
```

which names the class, so a genuine defect is still diagnosable from the message rather than silently flattened.

## Coverage

- `test_shared_validator_reports_instead_of_raising` — the shared helper
- `test_adapter_reports_unresolvable_reference` — parametrized across pi, gjc, ourocode, goose, zcode
- `test_a_resolvable_ref_still_validates` — the guard must not swallow schemas whose `$ref` *does* resolve. Accepts a conforming payload and rejects a non-conforming one through the same `$defs` indirection, so the fix cannot degrade into "never validate `$ref` schemas".

33 tests in the file; `tests/unit/providers/` → 662 passed.

Issue #1699, opened 2026-07-21
bug(wsl2): persisted owner identity drift can interrupt live detached jobs


kigam-research:
> [!IMPORTANT]
> This is the canonical tracker retained at the maintainer's direction. PR #1790 fixed only the process-local stdio watchdog. The remaining scope is the persisted cross-process owner identity used by detached JobManager and Auto recovery on WSL2.

## Current remaining problem

Persisted Linux owner records derive an epoch start time by combining `/proc/<pid>/stat` start ticks with a newly read `/proc/stat` `btime`. WSL2 can move `btime` while a process remains alive, so a later reader can reject the live persisted owner and interrupt a detached job.

## Required narrow fix

- [ ] Define a versioned persisted Linux identity that is stable across processes and WSL2 `btime` drift.
- [ ] Preserve fail-closed compatibility for existing persisted owner records.
- [ ] Parse `/proc/<pid>/stat` as bytes, including complex or non-UTF-8 process names.
- [ ] Keep live owners valid under synthetic drift while rejecting dead and PID-recycled owners deterministically.
- [ ] Cover JobManager, Auto owner liveness, cold restart, drift, death, PID reuse, and unchanged non-Linux behavior.

The fix must remain limited to persisted owner identity and must not broaden recovery authority or terminal-event semantics. The historical reproduction is retained below as requested by the maintainer.

---

## Summary

A fresh stdio MCP server can falsely decide that its live client process has exited, cancel its serve loop, and leave a synchronous `ouroboros_auto` call hung. The background `ouroboros_start_auto` form also interrupted its detached job while the Auto session remained non-terminal.

## Environment

- WSL2 / Linux
- Ouroboros source base: v0.50.5, local version `0.50.6.dev1`
- MCP server interpreter: Python 3.12
- MCP client interpreter: Python 3.14
- `mcp==1.28.1`
- Server command: `uvx --from 'ouroboros-ai[mcp,claude]' ouroboros mcp serve`
- The local patch under test changes Seed QA repair and Codex evidence transport; it does not modify MCP parent-liveness or job ownership code.

## Reproduction

1. Launch the server above as a fresh stdio child using `mcp.client.stdio.stdio_client`.
2. Initialize a `ClientSession`.
3. Call `ouroboros_start_auto` with a long but valid goal and `skip_run=true`, then poll the returned job.
[...]

PR #1790, opened 2026-07-28
fix(orchestrator): cache /proc/stat btime so drift cannot fake process death (#1699)

chojuyeon94:
Fixes the root cause behind #1699.

> **Note on history:** the first two commits took a different approach — caching `/proc/stat`'s `btime` inside `heartbeat._get_process_start_time`. Review correctly rejected that: the same helper also defines the *persisted cross-process* identity, so a cached epoch made an older observer disagree with a later worker and terminalize live jobs. That approach is fully reverted; `heartbeat.py` is unchanged by this PR. The description below reflects the final implementation.

## Problem

The stdio orphan watchdog decides whether its MCP client is still alive by comparing the client's process start time against the value recorded at startup, with a 2-second tolerance. On Linux that start time is computed as:

```python
return btime + clock_ticks / hz      # btime re-read from /proc/stat on every call
```

`clock_ticks` (`/proc/<pid>/stat` field 22) is constant for a given pid, but on WSL2 **`btime` is not stable** — the kernel re-derives boot time from a clock that resyncs, so successive reads drift upward:

```
11:03:31  btime=1784710818
11:03:39  btime=1784710818
11:03:44  btime=1784710820   <- +2s within 13 seconds
11:03:52  btime=1784710820
```

So the computed start time of a **live, unchanged pid keeps moving**. Once the drift crosses the tolerance, a healthy client is declared gone and the serve loop is cancelled.

macOS is unaffected: the Darwin branch reads `ps -o lstart=`, which is stable. That is why this reproduces only on Linux, and routinely on WSL2.

## Evidence

A bare `ouroboros mcp serve` under a live Python parent that keeps stdin open and never sends a request:

```
mcp.server.serve_error duration_ms=55098 error_type=CancelledError transport=stdio
mcp.server.serve_error duration_ms=55100 error_type=CancelledError transport=stdio
```

with `MCP client (pid …) gone — orphan exit` on stderr while that pid was demonstrably alive. Replicating the watchdog loop against a live, non-wrapper parent:

```
t+  0s kill=True start=1785204157.29 rec=1785204157.29 stat='S' comm='python3' ALIVE=True
t+  5s kill=True start=1785204159.29 rec=1785204157.29 stat='S' comm='python3' ALIVE=True
t+ 35s kill=True start=1785204160.29 rec=1785204157.29 stat='S' comm='python3' ALIVE=False  <- exits
```

`kill(pid,0)` succeeds, `stat` is `S` (not `Z`), `ppid` never changes. Only the recomputed start time moved. Every server on that machine died 45–80s after startup — always a multiple of the 5s watchdog poll, the exact value depending on where in the drift cycle it started.

## Fix

The watchdog gets its own comparison value instead of reusing the persisted epoch identity:

```python
def _process_start_marker(pid: int) -> float | None:
    if sys.platform == "darwin":
        return process_start_time(pid)          # ps lstart does not drift
    raw = Path(f"/proc/{pid}/stat").read_bytes()
    close = raw.rfind(b")")
    fields = raw[close + 1 :].split()
    return int(fields[19]) / os.sysconf("SC_CLK_TCK")
```

`/proc/<pid>/stat` field 22 is seconds since boot. It never involves `btime`, so it cannot drift, and it reads identically from any process. It is recorded and compared inside a single server process and never persisted, so a boot-relative value is sufficient here — the epoch form remains the cross-process identity.

Two parsing details that matter:

- Field 22 is located relative to the **last** `)`, because `comm` may contain spaces and parentheses.
- The line is read as **bytes**, because `comm` is arbitrary non-NUL bytes. `read_text()` raises `UnicodeDecodeError` on a legal name like `b"bad-\xff-name"` — not an `OSError`, so it would escape `_resolve_client_identity` and abort server startup.

## Verification

| | before | after |
|---|---|---|
| bare `ouroboros mcp serve` lifetime | 55.098s / 55.100s self-exit | **still alive at 181.4s** (probe limit) |
| `orphan exit` messages | present | none |
| watchdog verdict under drifting btime | `False` by the 3rd poll | stays `True` |

- New `tests/unit/cli/test_mcp_client_watchdog_marker.py`: marker stability under drifting btime, a live client staying alive, a reaped pid still reported dead, pid-recycle detection, `comm` with spaces/parentheses, `comm` with non-UTF-8 bytes, and concurrent reads agreeing.
- `tests/unit/cli/`, `test_process_local_authority.py`, `test_job_manager.py`, `test_detached_job_worker.py` → **1480 passed**. The only two failures in that selection (`test_setup.py::test_setup_claude_removes_legacy_timeout_override`, `::test_setup_claude_skips_write_when_args_already_current`) reproduce identically on unmodified `main` here, so they are pre-existing and unrelated.
- `ruff format --check`, `ruff check`, and `mypy` clean on the changed files.
- Existing `TestResolveClientIdentity` / `TestClientIsAlive` were updated to the new seam, and `test_mcp_shutdown_drain` now patches `_client_is_alive` rather than `is_process_identity_alive`, which the watchdog no longer calls.

## Scope

`src/ouroboros/cli/commands/mcp.py` plus tests. No change to `heartbeat.py`, `JobManager` recovery, lease acquisition/ownership, PID-registry payloads, terminal-event semantics, or anything RFC #1681 reserves for Foundation B.

Known limitation left alone deliberately: the persisted epoch identity still recomputes `btime` on every read, so an observer and a worker separated by a long uptime can disagree on WSL2. Fixing that touches persisted-format compatibility, which belongs with the reserved recovery work rather than in this scoped patch.


chojuyeon94:
Noting scope explicitly, since #1701 was closed for reaching into reserved territory.

This PR is deliberately confined to **how a process start time is computed**. The full diff is one function in `src/ouroboros/orchestrator/heartbeat.py` (`_get_process_start_time` now calls a `_linux_btime()` helper that reads `/proc/stat` once) plus a new test file. It does **not** touch:

- `JobManager` recovery, `_active_start_lease_error()`, `_read_owner_identity()`, or `_job_owner_is_dead()`
- lease acquisition/ownership or release paths
- terminal-event semantics (`mcp.job.interrupted`) or Auto receipts
- anything in the MCP recovery surface reserved for Foundation B by RFC #1681

The behavioural change is strictly narrowing: `_get_process_start_time(pid)` stops returning a *moving* value for an unchanged pid. Every existing caller — including `is_owned_by_current_process()` and the lease payload written by `acquire()` — sees a stable number instead of one that drifts past its own tolerance. No comparison that previously succeeded starts failing.

#1699 is retained as the reproduction, and the two new tests fail on `main` and pass with this change. Happy to narrow further or split if any part still overlaps Foundation A/B.

chojuyeon94:
Thanks — both findings were correct, and I reworked the change rather than patching around them.

## Blocker 1 (cross-process identity) — accepted, approach reverted

You were right, and my earlier scope comment was wrong. I checked only same-process behaviour and missed that the same helper defines a persisted contract:

```
job_manager.py:583   owner_pid, owner_start_time = current_process_identity()   # process A writes
job_manager.py:601   "owner_start_time": owner_start_time                        # persisted
job_manager.py:2053  owner_pid, owner_start_time = _read_owner_identity(...)     # process B reads
job_manager.py:2184  is_process_identity_alive(owner_pid, owner_start_time)      # process B compares
```

A long-lived observer caching `B0` while a later worker persists `B1 + ticks` does exactly what you described: the observer computes `B0 + ticks`, declares a signalable worker dead, and `mcp.job.interrupted` is persisted. **`heartbeat.py` is now untouched — the commit reverts that approach entirely.**

## New approach — the watchdog owns its own comparison

Your Design Notes named the real problem: *"A process-local epoch cannot safely serve both contracts."* So the shared helper keeps serving the persisted contract unchanged, and the watchdog gets a separate marker:

```python
def _process_start_marker(pid: int) -> float | None:
    if sys.platform == "darwin":
        return process_start_time(pid)          # ps lstart does not drift
    raw = Path(f"/proc/{pid}/stat").read_text()
    close = raw.rfind(")")
    fields = raw[close + 1 :].split()
    return int(fields[19]) / os.sysconf("SC_CLK_TCK")
```

`/proc/<pid>/stat` field 22 is seconds since boot. It never involves `btime`, so it cannot drift, and it reads identically from any process. It is recorded and compared inside one server process and never persisted, so a boot-relative value is sufficient here.

## Blocker 2 (concurrent initialization) — no longer applicable

There is no cache to initialize: the marker is computed per call from a value that does not move. Concurrent callers cannot observe different values. `test_concurrent_marker_reads_agree` pins that with 64 concurrent reads across 8 threads.

## Command names with spaces

Since this parser is new, field 22 is located relative to the **last** `)`, so `comm` containing spaces or parentheses parses correctly — the naive `.split()[21]` issue you and #1701 flagged does not exist in this path. `test_marker_parses_comm_containing_spaces_and_parens` covers it with a synthetic `(my (weird) proc)`.

## Coverage for both failure orderings you asked for

`tests/unit/cli/test_mcp_client_watchdog_marker.py`:

| Test | Guards |
|---|---|
| `test_marker_does_not_move_while_btime_drifts` | marker immune to drift |
| `test_live_client_stays_alive_while_btime_drifts` | the #1699 failure itself |
| `test_dead_pid_is_still_reported_dead` | drift immunity did not weaken liveness (real forked + reaped pid) |
| `test_recycled_pid_counts_as_dead` | pid-recycle protection intact |
| `test_marker_parses_comm_containing_spaces_and_parens` | last-`)` parsing |
| `test_concurrent_marker_reads_agree` | concurrent first access |
| **`test_cross_process_epoch_identity_is_left_uncached`** | **fails the moment anyone caches btime in the shared helper — a standing guard against exactly the regression you found** |

Existing `TestResolveClientIdentity` / `TestClientIsAlive` were updated to the new seam.

## Verification

- On `main`, a drifting-btime probe flips `_client_is_alive` to `False` on the **third** poll for a live pid. With this change it stays `True`.
- End-to-end: a bare `ouroboros mcp serve` under a live parent that keeps stdin open and sends nothing self-exited at **55.1s** before; it now survives past **181s** (probe limit) with zero `orphan exit` messages.
- `tests/unit/cli/`, `test_process_local_authority.py`, `test_job_manager.py`, `test_detached_job_worker.py` → **1480 passed**. The only two failures in that selection (`test_setup.py::test_setup_claude_removes_legacy_timeout_override`, `::test_setup_claude_skips_write_when_args_already_current`) reproduce identically on unmodified `main` here, so they are pre-existing and unrelated.
- `ruff format --check` and `ruff check` clean.

## Scope

`src/ouroboros/cli/commands/mcp.py` plus tests. No change to `JobManager` recovery, lease acquisition/ownership, terminal-event semantics, or anything RFC #1681 reserves for Foundation B — this revision moves further away from that boundary than the previous one did.

chojuyeon94:
Confirmed and fixed in `a660f51`. Thanks for the probe — this one was a real startup-abort, not a degradation.

## Blocker: non-UTF-8 `comm`

Reproduced locally before changing anything:

```
read_text -> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 10
  is UnicodeDecodeError a subclass of OSError?  False
```

That is exactly why the handler missed it: `except OSError` cannot catch a `ValueError` subclass, so it propagated out of `_resolve_client_identity` and aborted startup. A caller-controlled process name stopping the server from starting is strictly worse than the drift it was meant to fix.

The line is now parsed as bytes:

```python
try:
    raw = Path(f"/proc/{pid}/stat").read_bytes()
except OSError:
    return None
close = raw.rfind(b")")
fields = raw[close + 1 :].split()
return int(fields[19]) / os.sysconf("SC_CLK_TCK")
```

`int()` accepts bytes, and locating field 22 relative to the last `b")"` is unchanged. `test_marker_survives_non_utf8_process_name` covers it with a synthetic `(bad-\xff-name)`.

## Non-blocking 1 — over-specified test removed

You are right, and this is the better catch of the two. `test_cross_process_epoch_identity_is_left_uncached` asserted that a live process identity *changes* under drift. I wrote it as a guard against reintroducing the cache, but it pins today's broken behaviour as a requirement and would reject a future correct cross-process implementation. Removed rather than reworded — a scoped patch should not be the place that freezes that contract.

## Non-blocking 2 — stale patch target

Also correct: `test_mcp_shutdown_drain` patched `is_process_identity_alive`, which the watchdog no longer calls, so those tests were passing on an incidental condition rather than the one they name. They now patch `_client_is_alive` directly (3 sites).

## Verification

- `tests/unit/cli/`, `test_process_local_authority.py`, `test_job_manager.py`, `test_detached_job_worker.py` → **1480 passed**. The two failures in that selection (`test_setup.py::test_setup_claude_removes_legacy_timeout_override`, `::test_setup_claude_skips_write_when_args_already_current`) reproduce identically on unmodified `main` here — pre-existing and unrelated.
- `ruff check` and `ruff format` clean.
- Local WSL2 end-to-end still holds: a bare `ouroboros mcp serve` under a live parent with stdin open and no requests self-exited at 55.1s before the change and survives past 181s after it, with zero `orphan exit` messages.

Scope is unchanged: `src/ouroboros/cli/commands/mcp.py` plus tests. Persistence, replay, PID registry payloads and JobManager ownership are untouched.

PR #1792, opened 2026-07-28
feat(bigbang): extract CONTEXT_REFERENCES as JSON object arrays (#1729 slice 4)

### LINKED ISSUE #1729: [Feature] Extend the CONSTRAINTS JSON-array extraction contract (#1714) to the remaining pipe-delimited seed fields

sumin220:
## Summary

Slice 4 of #1729 — the final slice: CONTEXT_REFERENCES moves to a single-line JSON object array, completing the object-array rollout across all six fields (#1714 → #1730 → #1762 → #1767 → this).

This field is where the legacy colon split hurts most: a Windows path like `C:\repos\api` loses its drive letter to the role position, and a summary like `API layer: handlers | middleware` is truncated at both delimiters.

| Lane | Behavior |
|---|---|
| Extraction (strict) | Must be a valid JSON array of `{"path", "role", "summary"}` objects — `summary` is optional and defaults to empty, and a present non-string summary poisons the claim; legacy pipe lines, blank values, non-objects, and missing or blank path/role raise so the reformat-retry path resends. `[]` is the explicit empty spelling. |
| Stored requirements (lenient) | Bracket-led valid JSON arrays parse to models; everything else keeps the historical `path:role:summary` split with the summary absorbing further colons; malformed or blank-segment entries are skipped instead of raising (the legacy loop previously raised a model ValidationError on e.g. `:primary:x`); pre-parsed model tuples pass through; never raises. |

The brownfield extraction trailer and the seed-architect agent template request the new format, and the brownfield test fixture migrates with it. Reuses the shared decode/validation helpers hardened across the previous slices' reviews.

## Rollout status

With this slice, every pipe-delimited extraction field named in #1729 is on the JSON contract, and no in-repo producer emits the legacy format anywhere. Stored legacy data remains readable indefinitely through the lenient lane.

## Test plan

- [x] 3 new regression tests: strict parsing with drive colons in paths and colons/pipes in summaries plus optional-summary handling, strict rejection of legacy/blank/malformed values, lenient legacy-split preservation and never-raise fallbacks (including blank segments the old loop raised on) with stored-JSON and pass-through coverage
- [x] `uv run pytest tests/unit/bigbang -q` — 708 passed
- [x] `uv run pytest tests/ -q -n 4` — 14547 passed, 12 skipped

Closes #1729 (slice 4 of 4)

### DIFF
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -65,7 +65,7 @@ Format: single-line JSON array of objects with "name", "description", and "crite
 ### 7. BROWNFIELD CONTEXT (if applicable)
 If the interview mentions existing codebases, extract:
 - **PROJECT_TYPE**: 'greenfield' or 'brownfield'
-- **CONTEXT_REFERENCES**: path:role:summary (pipe-separated, role is 'primary' or 'reference')
+- **CONTEXT_REFERENCES**: single-line JSON array of objects with "path", "role" (primary or reference), and optional "summary"
 - **EXISTING_PATTERNS**: Key patterns that must be followed (single-line JSON array of strings)
 - **EXISTING_DEPENDENCIES**: Key dependencies to reuse (single-line JSON array of strings)
 
@@ -85,7 +85,7 @@ ONTOLOGY_FIELDS: [{"name": "<name>", "type": "<string|number|boolean|array|objec
 EVALUATION_PRINCIPLES: [{"name": "<name>", "description": "<description>", "weight": <0.0-1.0>}, ...]
 EXIT_CONDITIONS: [{"name": "<name>", "description": "<description>", "criteria": "<criteria>"}, ...]
 PROJECT_TYPE: greenfield|brownfield
-CONTEXT_REFERENCES: <path>:<role>:<summary> | ...
+CONTEXT_REFERENCES: [{"path": "<path>", "role": "<primary|reference>", "summary": "<summary>"}, ...]
 EXISTING_PATTERNS: ["<pattern 1>", "<pattern 2>", ...]
 EXISTING_DEPENDENCIES: ["<dep 1>", "<dep 2>", ...]
 ```
--- a/src/ouroboros/bigbang/seed_generator.py
+++ b/src/ouroboros/bigbang/seed_generator.py
@@ -502,6 +502,91 @@ def _parse_ontology_fields(raw_value: object, *, strict: bool = False) -> tuple[
     return tuple(fields)
 
 
+def _parse_context_references(
+    raw_value: object, *, strict: bool = False
+) -> tuple[ContextReference, ...]:
+    """Parse CONTEXT_REFERENCES from JSON object arrays or the legacy pipe list.
+
+    Mirrors the other #1729 object-array parsers: the extraction format
+    requests a single-line JSON array of ``{"path", "role", "summary"}``
+    objects — ``summary`` is optional and defaults to empty — so colons in
+    paths (e.g. Windows drives) and colons/pipes in summaries survive as
+    data. Strict mode raises on anything else so the retry path reformats;
+    lenient mode keeps the historical ``path:role:summary`` split where the
+    summary absorbs further colons, skips malformed entries, and never
+    raises.
+    """
+    field_label = "CONTEXT_REFERENCES"
+
+    def _build(entry: object) -> ContextReference | None:
+        if isinstance(entry, ContextReference):
+            return entry
+        if not isinstance(entry, dict):
+            if strict:
+                raise ValueError(
+                    f"{field_label} entries must be JSON objects; got "
+                    f"{type(entry).__name__}: {entry!r}"
+                )
+            return None
+        try:
+            summary = entry.get("summary", "")
+            if not isinstance(summary, str):
+                raise ValueError(
+                    f"{field_label} entry field 'summary' must be a string; got {summary!r}."
+                )
+            return ContextReference(
+                path=_require_object_string(entry, "path", field_label=field_label),
+                role=_require_object_string(entry, "role", field_label=field_label),
+                summary=summary.strip(),
+            )
+        except (ValueError, TypeError, PydanticValidationError):
+            if strict:
+                raise
+            return None
+
+    def _build_all(entries: list | tuple) -> tuple[ContextReference, ...]:
+        return tuple(built for built in (_build(entry) for entry in entries) if built)
+
+    if isinstance(raw_value, list | tuple):
+        return _build_all(raw_value)
+    if not isinstance(raw_value, str):
+        return ()
+    text = raw_value.strip()
+    if not text:
+        if strict:
+            raise ValueError(
+                f"{field_label} must be a single-line JSON array of objects; "
+                "use [] for an intentionally empty list."
+            )
+        return ()
+    decoded = _decode_object_array(text, field_label=field_label, strict=strict)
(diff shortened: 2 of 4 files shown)


Q00 (approved):
I am looking forward to see the completion of #1729. 
thx for your contribution

Issue #1794, opened 2026-07-28
Two order- and load-dependent test failures turn main's Tests job red at random


Q00:
Two `Tests` failures landed on `main` during the 2026-07-28 merge batch. Neither reproduces on a rerun, and both are independent of the code they failed on — they are properties of the test suite, not of the changes under test.

## 1. Watchdog timeout tests race two thresholds on wall-clock time

`38df811` / Python 3.12:

```
FAILED tests/unit/evolution/test_generation_watchdog.py::test_no_progress_timeout_emits_retry_directive
- generation.watchdog:lin-578-unstuck:1:no_material_progress_timeout:reflecting
+ generation.watchdog:lin-578-unstuck:1:idle_timeout:reflecting
```

The same file failed on PR #1767 with a different test in the same family (`test_busy_run_without_material_progress_times_out`).

`GenerationProgressWatchdog._raise_if_threshold_exceeded` checks `idle_timeout` before `no_material_progress_timeout`. The affected tests set the no-progress threshold to 0.07s and leave idle at the helper default of 1.0s, then rely on real `asyncio.sleep` to reach the no-progress threshold first. Both thresholds measure `time.monotonic()`, and `_last_activity_at` only advances when a poll actually observes an event — so a poll loop starved for a second under a loaded parallel run reaches the idle threshold first and the generation is cancelled for the wrong reason.

`test_session_activity_resets_idle_timeout` has the inverse form of the same defect: it asserts the generation stays alive across four real 0.04s sleeps against a 0.07s idle threshold, so a single scheduling hiccup fails it.

The file already establishes a `_FakeMonotonicClock` idiom for exactly this; the affected tests just do not use it.

## 2. `reset_logging()` leaves structlog louder than any configured state

`18944f5` / Python 3.14:

```
FAILED tests/unit/cli/test_detached_auto_wait_command.py::test_cli_detached_auto_wait_invalid_handle_fails_with_actionable_error
FAILED tests/unit/cli/test_detached_auto_wait_command.py::test_cli_detached_auto_wait_failed_job_exits_nonzero_with_stable_error
FAILED tests/unit/cli/test_detached_auto_wait_command.py::test_cli_detached_auto_wait_reports_in_progress_background_work
assert not <re.Match object; span=(0, 12), match='2026-07-28 0'>
```

`mcp.tool.job_wait.*` debug lines land in the CLI's captured output.

[...]

PR #1795, opened 2026-07-28
fix(tests): make the watchdog and logging flakes deterministic

Q00:
Refs #1794

Two `Tests` failures landed on `main` during the 2026-07-28 merge batch (`38df811`, `18944f5`). Neither reproduces on a rerun, and neither is caused by the code it failed on.

## 1. Watchdog tests raced two thresholds on wall-clock time

`_raise_if_threshold_exceeded` checks `idle_timeout` before `no_material_progress_timeout`. The affected tests set the no-progress threshold to 0.07s, leave idle at the helper default of 1.0s, and rely on real `asyncio.sleep` to reach the shorter one first. But `_last_activity_at` only advances when a poll actually *observes* an event, so a poll loop starved for a second under a loaded parallel run reaches the idle threshold first — and the generation is cancelled for the wrong reason:

```
- generation.watchdog:lin-578-unstuck:1:no_material_progress_timeout:reflecting
+ generation.watchdog:lin-578-unstuck:1:idle_timeout:reflecting
```

The file already establishes a `_FakeMonotonicClock` idiom for this; these tests just did not use it. Simulated time now advances only from inside the work coroutine, so the thresholds are ordered by the scenario rather than by machine load.

`test_session_activity_resets_idle_timeout` is the inverse form of the same defect — it asserts the generation stays *alive* across four real 0.04s sleeps against a 0.07s idle threshold, so one scheduling hiccup failed it. It is now stepped deterministically, mirroring the sibling `test_parent_execution_child_events_reset_idle_timeout`.

## 2. `reset_logging()` left structlog louder than any configured state

`reset_logging()` ended with `structlog.reset_defaults()`, restoring the library defaults: every level emitted, through a `PrintLogger` bound to **stdout**. A configured Ouroboros logger filters at `LoggingConfig.log_level` and writes to **stderr**, so stdout stays reserved for command output.

Clearing `_configured` is supposed to make the next `get_logger()` reconfigure, but that never fires for the 68 `src` modules that bind a bare `structlog.get_logger` at import time — they keep logging through whatever the reset left behind. `tests/unit/observability/test_logging.py` calls `reset_logging()` from an autouse fixture, so it poisoned its xdist worker for every later test asserting on CLI output, and `mcp.tool.job_wait.*` debug lines landed in `result.output`.

The fix restores the default filter alongside the reset. `_configured` and `_current_config` still clear, so the next `get_logger()` performs a full reconfiguration.

## Verification

| Check | Result |
|---|---|
| Watchdog file, 20 runs under full CPU saturation — **before** | 1/20 failed (`test_busy_run_without_material_progress_times_out`, the test that failed on #1767) |
| Watchdog file, 20 runs under identical load — **after** | **0/20 failed** |
| Logging leak, deterministic repro (import CLI modules → `reset_logging()` → invoke `job wait`) | reproduced byte-identically to CI, then fixed |
| New regression test without the `logging.py` fix | fails |
| `tests/unit/{evolution,observability,cli,orchestrator}` | 5659 passed, 7 skipped |
| `ruff check` / `ruff format --check` / `mypy src/` | clean |

This changes no production behaviour: `reset_logging()` has no callers outside tests, and the real `ouroboros job wait` stdout was already clean because the CLI configures logging before anything logs.

### DIFF
--- a/src/ouroboros/observability/logging.py
+++ b/src/ouroboros/observability/logging.py
@@ -45,7 +45,6 @@ import logging
 from logging.handlers import TimedRotatingFileHandler
 import os
 from pathlib import Path
-import sys
 from typing import Any
 
 from pydantic import BaseModel, Field
@@ -637,6 +636,17 @@ def reset_logging() -> None:
     regression. ``_configured`` stays false and ``_current_config`` stays
     ``None``, so the next :func:`get_logger` still performs a full
     reconfiguration.
+
+    The stream is carried by reusing the project's own factory with no file
+    handler, not by handing ``sys.stderr`` to ``structlog.PrintLoggerFactory``.
+    Two reasons, both of which a captured stream gets wrong.
+    ``_FileWritingPrintLogger._log`` resolves ``sys.stderr`` at emit time, so a
+    reset that runs while pytest's ``capsys`` is active cannot pin that
+    fixture's buffer into the global configuration and raise on a write after
+    its teardown. And that logger is the only one that consults
+    ``_console_logging_enabled``, so routing through it keeps
+    :func:`set_console_logging` honored across a reset instead of silently
+    re-enabling console output.
     """
     global _configured, _current_config
     # Read before clearing: the reset must not be louder than what it replaces.
@@ -650,5 +660,7 @@ def reset_logging() -> None:
     structlog.reset_defaults()
     structlog.configure(
         wrapper_class=structlog.make_filtering_bound_logger(outgoing_level),
-        logger_factory=structlog.PrintLoggerFactory(file=sys.stderr),
+        # No file handler: the outgoing one was just closed by
+        # _close_root_handlers(), and a reset must not hold a resource.
+        logger_factory=_FileWritingPrintLoggerFactory(None),
     )
--- a/src/ouroboros/observability/logging.py
+++ b/src/ouroboros/observability/logging.py
@@ -610,6 +610,25 @@ def reset_logging() -> None:
 
     This is primarily for testing purposes. It resets the module state
     but does not reconfigure the loggers.
+
+    A reset must never leave the process *louder* than a configured one.
+    ``structlog.reset_defaults()`` restores the library defaults, which
+    emit every level — including debug — through a ``PrintLogger`` bound
+    to stdout, while a configured Ouroboros logger filters at
+    ``LoggingConfig.log_level`` and writes to stderr so stdout stays
+    reserved for command output.
+
+    That gap is not self-healing. Most modules bind their logger once at
+    import time via a bare ``structlog.get_logger``, which never routes
+    through :func:`get_logger` and so never triggers the lazy
+    reconfiguration that clearing ``_configured`` sets up. Once those
+    proxies exist, a reset silently turns every ``log.debug`` in the
+    process into stdout output, which contaminates any CLI result parsed
+    by a caller or asserted on by a later test.
+
+    So restore the default filter alongside the reset. ``_configured``
+    stays false and ``_current_config`` stays ``None``, so the next
+    :func:`get_logger` still performs a full reconfiguration.
     """
     global _configured, _current_config
     _configured = False
@@ -619,3 +638,8 @@ def reset_logging() -> None:
     structlog.contextvars.clear_contextvars()
     # Reset structlog configuration
     structlog.reset_defaults()
+    structlog.configure(
+        wrapper_class=structlog.make_filtering_bound_logger(
+            _get_log_level(LoggingConfig().log_level)
+        )
+    )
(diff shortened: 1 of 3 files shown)


Q00:
Both follow-up findings reproduced, and both are fixed in `180f6a3`.

**`reset_logging()` left the default `PrintLoggerFactory` in place** — confirmed. Filtering debug was only half the contract; an INFO record after a reset still went to stdout:

```
$ configure_logging(...) ; reset_logging() ; structlog.get_logger('probe').info('INFO_AFTER_RESET')
STDOUT: 2026-07-29 01:58:46 [info     ] INFO_AFTER_RESET
```

**Hard-coding the `LoggingConfig` default made a quieter process louder** — confirmed, with one correction to the mechanism. A logger already *used* under the ERROR configuration keeps its cached bound level (`cache_logger_on_first_use=True`), so a pre-bound `log.info` stays silent. The regression shows up on a logger first bound after the reset:

```
$ configure_logging(log_level='ERROR') ; reset_logging() ; structlog.get_logger('fresh').info(...)
STDOUT: 2026-07-29 01:58:57 [info     ] INFO_AFTER_RESET_from_ERROR_config
$ # same without the reset: silent
```

The reset now carries both halves across: `logger_factory=PrintLoggerFactory(file=sys.stderr)` and the outgoing configuration's level instead of the default. `_configured` and `_current_config` still clear, so the next `get_logger()` fully reconfigures.

Two tests pin the newly closed halves, and both fail against the previous commit:

- `test_reset_keeps_emitted_levels_off_stdout` — INFO after a reset stays off stdout and lands on stderr
- `test_reset_preserves_a_quieter_configured_level` — an ERROR-level process stays silent at INFO after a reset

The non-blocking documentation point is fixed too: the docstring no longer claims the reset "does not reconfigure the loggers", and now says it installs a neutral global configuration while clearing Ouroboros state.

Verification: `tests/unit/{observability,cli,evolution,mcp/tools}` 2473 passed / 6 skipped; the deterministic CLI-pollution repro still passes; ruff, format and mypy clean.

Q00:
## Adopted #1801's factory choice; kept this PR's level preservation

#1801 fixes the same `reset_logging()` defect and its inline comment correctly identifies a hazard in what this PR had:

> The project factory resolves sys.stderr at call time and honors set_console_logging(); **capturing the stream object here would keep a capsys/redirect buffer alive past its teardown and raise on write.**

That is right, and there was a second problem alongside it: `structlog.PrintLoggerFactory` never reaches `_FileWritingPrintLogger._log`, which is the only place `_console_logging_enabled` is consulted — so a process that had called `set_console_logging(False)` got console output silently restored by the reset.

Swapped to `_FileWritingPrintLoggerFactory(None)` (`9927519d0`). No file handler, since `_close_root_handlers()` just closed the outgoing one and a reset must not hold a resource.

**The level preservation stays.** #1801 hardcodes `logging.INFO` here, which is a volume regression for any process configured quieter — resetting an `ERROR`-level process to `INFO` is the mirror image of the bug both PRs are fixing. Filtering lives in `wrapper_class`, so carrying the outgoing level is orthogonal to the factory choice and costs nothing.

Two regression tests added, both **proven non-vacuous** — they fail against the old factory and pass against the new one:

```
$ # with structlog.PrintLoggerFactory(file=sys.stderr)
FAILED test_reset_resolves_the_stream_at_emit_time
FAILED test_reset_keeps_console_suppression
2 failed

$ # with _FileWritingPrintLoggerFactory(None)
2 passed
```

Local: `test_logging.py` + `test_generation_watchdog.py` + `test_job_command.py` → 89 passed. ruff check/format and mypy clean. The module-level `import sys` this PR had added went with the change; `_log` keeps its function-local import, as on main, so the logging.py diff is now confined to `reset_logging()`.

### On #1801

The watchdog half of the two PRs is the same fix — same `_FakeMonotonicClock`, same `monkeypatch.setattr(watchdog_module, "time", ...)`, same `clock.advance()` placement — so it is pure duplication, not a competing approach.

What #1801 has that this PR does not is the **cached-bound-logger** fix: `configure_logging` sets `cache_logger_on_first_use=True`, so a bound logger freezes its wrapper and its `_FileWritingPrintLogger` (holding `self._file_handler`), and a later reconfigure never reaches it. That is a real second defect and #1801's live-sink (`_current_min_level` / `_current_file_handler`) is the right shape for it. It should land — rebased on this PR, with the duplicated watchdog changes dropped.

Its `event_store.py` payload (+208/-34 plus 570 test lines: `_run_to_settlement`, the write registry drained by `close()`, `initialize`/`close` mutual exclusion) is EventStore lifecycle and concurrency hardening, not test determinism. Worth splitting into a third PR so the three review units stay separable: reset contract / cached-logger contract / EventStore lifecycle.

cc @sumin220

PR #1801, opened 2026-07-29
fix(observability): route cached loggers through the live sink

sumin220:
## Summary

- keep import-time and cached structlog bound loggers governed by the current logging configuration
- route those cached loggers through a stable live sink instead of leaving them bound to stale handlers
- preserve stderr-only output, active level and renderer, console suppression, and current file-handler ownership across `reset_logging()` and reconfiguration
- add focused regressions for cached proxies, cached bound loggers, level changes, renderer changes, file rotation/reconfiguration, and reset behavior

## Scope split

The watchdog portion of #1794 already landed on `main` in #1795. The EventStore cancellation/lifecycle redesign formerly carried here was split into independently reviewable PR #1820 at the maintainer’s request. This PR now contains only the remaining cached-logger delta: two observability files on top of current `main`.

## Verification

- 172 focused observability/watchdog/persistence tests passed after the split
- Ruff check and format passed
- MyPy passed for `logging.py`

Related to #1794.


sumin220:
Addressed the first review in faf9ee1b. You were right that the shield belonged in production, not the fixture: append_with_rowid and append_batch now run each transaction attempt under asyncio.shield, so a caller cancelled mid-append still observes cancellation immediately while the write commits or rolls back to completion — no abandoned transaction can poison the pooled connection and starve emit_decision's batch. The watchdog tests drop their test-only shields, so cancellation is deliberately permitted to land inside an append; that unshielded pattern is the regression pinning the store contract. 40 consecutive serial runs of the watchdog file pass clean (previously ~2/30 failed without the production fix), persistence suite 220 passed, full suite 14741 passed. Also documented reset_logging's baseline in the docstring per the note.

sumin220:
Addressed the second review in 0b486d80. The shields are replaced with settlement semantics: on cancellation the caller now waits for the inner transaction task to complete — absorbing any further cancellations — and only then re-raises, so a write can no longer outlive its caller or EventStore.close(), and the round-one cancellation-atomicity contract is preserved. Two deterministic regression tests pin exactly your probe: they gate aiosqlite.Connection.commit to hold the transaction open, cancel the caller mid-commit, assert cancellation does not surface until settlement, and replay the events before close — both verified failing against the shield-only code. Watchdog file still 40/40 clean serial runs; persistence+evolution 358 passed; full suite 14743 passed; ruff and mypy clean.

sumin220:
Addressed the third review in fbd6d42d. The drain now absorbs a settling transaction's own failure instead of letting it replace the caller's cancellation: after settlement the failure is retrieved and the saved CancelledError is re-raised with it chained as the cause, so lifecycle code awaiting a cancelled task — the watchdog included — always observes CancelledError and proceeds to its decision batch. Two deterministic regression tests cancel each append path mid-commit with a commit that raises, asserting CancelledError with the chained cause still surfaces (both verified failing against the previous settlement). Watchdog file 40/40 clean; full suite 14745 passed; ruff and mypy clean.

sumin220:
Addressed the fourth review in 4efe2304. Blocker: the store now tracks settling transactions in a registry and close() drains them before the WAL checkpoint and engine disposal, so a write that outlived a cancelled caller cannot land after shutdown — the regression tests race a cancelled append and batch against close() past checkpoint_wal's 2s busy window, assert close() waits, and verify the events are durable on reopen (both failing without the drain). Follow-up: the reset baseline swaps PrintLoggerFactory(sys.stderr) for the project's _FileWritingPrintLoggerFactory, which resolves stderr at call time and honors set_console_logging(); new tests cover logging after capture teardown and with console output disabled. Watchdog file 40/40 clean; full suite 14749 passed; ruff and mypy clean.

sumin220:
Addressed the fifth review in 2436bde9. Blocker: write admission is now synchronized with closing — the settlement helper refuses a write when the store is closing, and that check shares one synchronous event-loop block with registry registration, so a write either registers before close() drains or is refused outright; close() flips the flag before draining and initialize() clears it. The regression tests pause close() inside the WAL checkpoint, start an append and a batch mid-close, and assert both are refused and absent after reopening (failing before the fix). Follow-up: the baseline now also governs loggers cached before reset — a module-level flag makes the project print logger enforce never-below-INFO/stderr-only and skip the old file handler while active, cleared by configure_logging; the new test materializes a proxy under DEBUG+file logging before reset and asserts post-reset debug vanishes, info is stderr-only, and the old log file is not reopened. Watchdog 40/40 clean; full suite 14752 passed; ruff and mypy clean.

sumin220:
Addressed the sixth review in 109c1127. Blocker: the session start/terminal/pause CAS paths are now fenced wrappers — admission check, settlement, and registry entry around the unfenced CAS body — so every public append transaction honors the closing fence uniformly; the regression test races a session-start and a terminal append against a paused close() and asserts both are refused and absent after reopen. Follow-up: the baseline flag is replaced by live-sink routing — the print logger reads the minimum level and the file handler from module state published by configure_logging()/reset_logging(), so cached bound loggers always follow the current configuration: reconfiguration cannot revive a stale DEBUG floor or a closed file handler. New test caches a proxy under DEBUG/file logging, resets, reconfigures INFO/no-file, and asserts the stale proxy obeys the new configuration (failing before). Watchdog 40/40 clean; full suite 14754 passed; ruff and mypy clean.

sumin220:
Addressed the seventh review in 386bc453. Blocker: initialize() now serializes with close() through a close gate — close clears the gate in the same synchronous block that flips the closing flag and sets it only after dispose, so initialization during an in-flight close waits instead of reopening admission; the probe's initialize-then-append during a paused checkpoint is refused and absent after reopen (regression failing before). Follow-up: bound loggers are no longer cached — both configure_logging and the reset baseline pass cache_logger_on_first_use=False, so proxies follow the current filtering and processors in both directions; regressions cover INFO→DEBUG relaxation and DEV→PROD renderer switching (both failing before). P2: the session-pause path has its own explicit close-race pin. Suggestion: closing refusals now carry the actual operation name. Watchdog 40/40; full suite 14758 passed; ruff and mypy clean.

sumin220:
Addressed the eighth review in 6222f7ad. Blocker 1: initialize() and close() now run their complete lifecycles under one asyncio.Lock, so they are mutually exclusive in both orders — a close started during initialization waits for it instead of finishing first; the close gate is retired in favor of the lock. Blocker 2: the settling-write drain moved inside close()'s guarded block, and a cancelled close restores a recoverable state when the engine survives, so cancellation mid-drain can no longer wedge the store in a permanently-closing state — the regression re-initializes and writes after a cancelled close. P2: documented the explicit get_logger().bind() snapshot limitation next to the cache setting. Suggestion: the logger docstring now says stderr. Both new regressions verified failing before; watchdog 40/40; full suite 14760 passed; ruff and mypy clean.

Q00:
## #1795 landed on main — this PR now needs a rebase, and the split is mechanical

#1795 merged as `debb9d079`. I verified this PR no longer applies to the new main:

```
$ git apply --check <this PR's diff>   # against origin/main
error: src/ouroboros/observability/logging.py: patch does not apply
error: tests/unit/evolution/test_generation_watchdog.py: patch does not apply
error: tests/unit/observability/test_logging.py: patch does not apply
```

Exactly the three overlapping files. `event_store.py` and `test_event_store.py` are untouched by the conflict — I checked, and **that half applies cleanly to main today**:

```
$ git apply --check <event_store half only>
  ✅ (clean)
```

### Suggested split

**Drop — already on main.** The watchdog test changes are the same fix, not a competing one: same `_FakeMonotonicClock`, same `monkeypatch.setattr(watchdog_module, "time", SimpleNamespace(monotonic=clock))`, same `clock.advance()` inside the work coroutine.

**Drop — already on main, and your version was adopted.** Your `reset_logging()` critique was right and #1795 took it (`9927519d0`): it now uses `_FileWritingPrintLoggerFactory(None)` rather than `structlog.PrintLoggerFactory(file=sys.stderr)`, for both reasons your comment gave — late-bound `sys.stderr` so a `capsys` buffer cannot be pinned past teardown, and routing through the only logger that consults `_console_logging_enabled`.

One difference kept from this PR's version: the outgoing level is preserved instead of hardcoding `logging.INFO`. Resetting an `ERROR`-level process to `INFO` is a volume regression in the same family as the bug being fixed, and since filtering lives in `wrapper_class` it is orthogonal to the factory choice.

**Keep — this is the part main still lacks.** The cached-bound-logger fix. `configure_logging` sets `cache_logger_on_first_use=True`, so a bound logger freezes its wrapper *and* its `_FileWritingPrintLogger` (holding `self._file_handler`), and a later reconfigure never reaches it. The live sink (`_current_min_level` / `_current_file_handler`) read at emit time is the right shape. #1795 does not address this at all — please keep it.

**Split — separate PR.** `event_store.py` +208/-34 plus 570 test lines: `_run_to_settlement`, the write registry drained by `close()`, `initialize`/`close` mutual exclusion, the `*_unfenced` variants. That is EventStore lifecycle and concurrency hardening. It may well be load-bearing for determinism under load, but its risk profile is nothing like a logging reset, and bundling it means one review has to hold both. It applies to main unchanged, so the split costs a `git apply` and a new branch.

Result: two clean review units — cached-logger contract, and EventStore lifecycle — instead of one 1,050-line PR carrying two already-merged fixes.

Happy to do the rebase and split myself if you'd rather (`maintainer_can_modify` is on), but it is your work and rewriting your branch without asking seemed wrong. Say the word.

cc @sumin220

sumin220:
Thanks for checking the overlap file by file — and for asking rather than just rewriting the branch. Please go ahead with the rebase and split; you already have the boundaries verified against main, so it is more efficient than me redoing that work.

Your reading matches mine on all four parts. The cached-bound-logger fix is the piece I would most want preserved: cache_logger_on_first_use=True freezes both the wrapper and the _FileWritingPrintLogger instance holding self._file_handler, so a later reconfigure never reaches an already-bound logger — the live sink read at emit time is what closes that. And keeping the outgoing level instead of hardcoding INFO is right; resetting an ERROR process to INFO is the same family as the bug being fixed, and filtering living in wrapper_class makes it orthogonal to the factory choice.

Q00 (approved):
Approved at current HEAD 171dfb60d after semantic conflict resolution against #1817/current main. The merge preserves configured/read-only EventStore path authority and the #1795 watchdog/logging semantics while adding lifecycle-wide cancellation settlement, initialize/close mutual exclusion, write draining, close admission fencing, and cancelled-close recovery. Verified locally: 436 focused persistence/evolution/logging tests, Ruff, format, and Mypy; all current CI checks are green.

Q00:
@ouroboros-agent rereview exact HEAD 802f4609be3a3372f45e8e703640f7f1a6ca2c3a. Maintainer repairs are pushed; please publish an explicit exact-head verdict after required checks complete.

Q00:
@ouroboros-agent rereview exact HEAD 6731cec6a. The logging-generation concurrency repair now also preserves structlog.testing.capture_logs() by sharing each freshly published processor-list container with structlog. The three prior CI failures reproduce as the tuple .copy() incompatibility and are fixed; the full CI-equivalent suite completed locally with 15,326 passed, 19 skipped, 1 xfailed, and all deterministic configure/reset overlap regressions remain green. Please publish an explicit exact-head verdict after required checks complete.

Q00:
@ouroboros-agent rereview exact HEAD 47b0c40c6. The synchronous generation wrapper now captures the exact application frame before entering its custom processor loop, while async records retain structlog preserved calling-stack attribution; regressions verify filename, line, and function for both paths without module/file-name blanket ignores. Repeated reset now derives its neutral baseline from the active live generation when current config is absent, so ERROR cannot relax to INFO. Relevant logging/orchestrator coverage: 357 passed; all capture users: 646 passed; eight concurrency/capture/callsite/reset regressions passed across 10 repetitions; Ruff, format, MyPy, and diff checks are clean. Please publish an explicit exact-head verdict after required checks complete.

PR #1805, opened 2026-07-29
perf(project-map): share the git capability probe per session start (#1796 L1)

### LINKED ISSUE #1796: perf(project-map): cache the Git capability probe and stop double-resolving identity per session start
## Problem

`resolve_project_identity` (added in #1769, `src/ouroboros/core/project_identity.py`) is on the
**session-start critical path** and spawns six `git` subprocesses per call with no memoization.
A session start resolves it **at least twice** — once when the execution contract is built, and
again immediately before immutable publication.

Measured on a linked worktree of this repository (macOS, warm page cache, 5 runs):

| git call | n | avg ms | share |
|---|---:|---:|---:|
| `worktree list --porcelain -z` | 5 | 136.7 | 42.5% |
| `rev-parse --is-bare-repository` (common dir) | 5 | 44.6 | 13.9% |
| `rev-parse --is-bare-repository` (checkout) | 5 | 38.9 | 12.1% |
| **`git --version`** | 5 | **35.8** | **11.1%** |
| `rev-parse --show-toplevel` | 5 | 33.7 | 10.5% |
| `rev-parse --git-common-dir` | 5 | 31.9 | 9.9% |
| **total per resolve** | **6** | **321.5** | |

So a single session start pays roughly **640 ms of subprocess time** for metadata that
`docs/rfc/project-map-v1.md` explicitly classifies as attribution-only.

The RFC contains no mention of caching, latency, or cost — the performance dimension was never
considered during design.

## Call sites

Resolution entry points (`origin/main`):

- `src/ouroboros/orchestrator/runner.py:3256` — `resolve_managed_project_identity` (managed worktree)
- `src/ouroboros/orchestrator/runner.py:3320` — `resolve_project_identity` (direct)
- `src/ouroboros/orchestrator/session.py:87` / `:126` — publication-boundary revalidation

`OrchestratorRunner._project_identity()` is itself reached from four places
(`runner.py:3334`, `:5644`, `:5693`, `:6451`).

### L1 — cache the Git capability probe (small, safe, ship first)

`_require_supported_git()` (`project_identity.py:273`) runs `git --version` on **every** resolution,
including the second one inside the same session start. The installed Git binary cannot
meaningfully change within one process lifetime, and the function is a pure capability gate with no
path input.

Wrap it in a process-lifetime cache (`functools.cache`). Both raised exception types
(`ProjectIdentityGitVersionError`, `ProjectIdentityUnavailableError`) need explicit handling —
`functools.cache` does not memoize exceptions, so a transient spawn failure must not be permanently
cached, while a genuine version rejection may be.

Measured saving: **~36 ms per resolve, ~11% of subprocess time**, ~72 ms per session start.
Modest, but it is the only zero-risk item on the list.

This mirrors the fix already applied in c7e988679, where `/proc/stat` `btime` was cached for exactly
the same reason.

### L2 — investigate the double resolution (bigger win, needs design)

`session.py::_validate_project_identity_publication` re-runs the **full** resolver at the
publication boundary. Its stated purpose is to detect a workspace that changed between contract
build and publication — but `project_identity.py` already carries `_LiveDirectoryEvidence`
(device + inode) and `_revalidate_live_directories()` for precisely that class of detection.

Question to answer: can the publication check revalidate the captured live-directory evidence
without re-querying Git topology? If yes, session start drops from ~640 ms to ~320 ms.

This is a TOCTOU-sensitive path, so it needs its own analysis and must not be folded into L1.
`worktree list` alone is 42% of the remaining cost and is inherent to the current design.

## Acceptance criteria

- `_require_supported_git()` spawns `git --version` at most once per process.
- A transient (`ProjectIdentityUnavailableError`) probe failure is **not** cached; a version
  rejection (`ProjectIdentityGitVersionError`) may be.
- A regression test asserts the spawn count across repeated `resolve_project_identity` calls.
- L2 is answered in writing on this issue (either a follow-up PR or a documented "keep as is, here
  is why").

## Non-goals

- Caching resolved identities per cwd. The resolver's fail-closed and TOCTOU guarantees depend on
  fresh evidence; that trade-off is out of scope here.

Refs #1389

sumin220:
Closes #1796.

A session start resolved project identity twice — once while building the execution contract and again at the immutable publication boundary — and each resolution paid its own `git --version` capability probe plus the full set of topology queries.

This PR removes the second resolution for the common simple-checkout shape without reusing capability authority. Contract construction captures the resolver's repo-local input closure, and the publication boundary revalidates from that evidence by metadata alone; when anything is in doubt it escalates to the full re-resolution, which probes for itself exactly as on main.

How it works:

- `resolve_project_identity_for_publication` runs the unchanged resolution pipeline and captures the closure of its five fixed topology queries: every boundary `.git` marker between the effective directory and the checkout root, plus the marker directory's `HEAD`, `config`, `config.worktree`, `commondir`, `gitdir`, and the `worktrees` registry with each entry's `gitdir` link — each recorded as stat generation (device, inode, mtime_ns, size) plus a content hash. The capture is bracketed: the closure is captured before the resolver's queries and re-captured after, and the evidence is usable only when the two captures are byte-identical, which binds the recorded state to the exact resolution that produced the identity.
- At publication, `publication_evidence_is_stable` accepts only when the published workspace canonicalizes to the captured directory, the evidence carries the complete resolver-shaped closure for exactly the published identity, and every captured input is unchanged. The recheck runs no git subprocess at all, so the V1 rule that resolvers validate Git before topology queries is satisfied vacuously — there are none.
- Everything outside the enumerable closure fails closed: bare repositories, gitdir-pointer checkouts (linked worktrees), commondir constellations, config include directives (detected by an unanchored, case-insensitive byte search — a sound over-approximation of Git's config grammar rather than a reimplementation), oversized or unreadable inputs. All of these escalate to the full re-resolution.
- The earlier probe-scope design from this PR's review history is fully removed; construction probes per call exactly as main does, and no probe result crosses between resolutions.

Coverage: a mutation matrix proving each closure input individually destabilizes the evidence, a zero-git-subprocess assertion on the recheck, escalation exercised against a real `git worktree add` checkout and BOM-prefixed include bytes, binding of evidence to its own workspace and resolution (interleaved-mutation regression), refusal of incomplete or caller-assembled evidence, session-boundary tests for foreign and drifted evidence, and `prepare_session` wiring tests counting publication re-resolutions.

### DIFF
--- a/src/ouroboros/core/project_identity.py
+++ b/src/ouroboros/core/project_identity.py
@@ -15,6 +15,7 @@ from dataclasses import dataclass
 import os
 from pathlib import Path, PurePosixPath
 import re
+import shutil
 import stat
 import subprocess
 import tempfile
@@ -235,17 +236,26 @@ def _git_environment() -> dict[str, str]:
     return environment
 
 
-def _run_git_command(*arguments: str) -> bytes:
+def _run_git_command(*arguments: str, executable: str | None = None) -> bytes:
     """Run one bounded, non-interactive Git command and return complete stdout.
 
     A nonzero exit is unavailable because Git does not expose a portable
     exit-code distinction between malformed topology and transient repository
-    I/O.
+    I/O. Inside a capability probe scope every command binds to the executable
+    that scope verified; outside one, ``git`` resolves from PATH per call as
+    it always has.
     """
+    if executable is None:
+        scope_cell = _active_probe_scope()
+        executable = (
+            scope_cell.executable
+            if scope_cell is not None and scope_cell.executable is not None
+            else "git"
+        )
     try:
         with tempfile.TemporaryFile() as output:
             completed = subprocess.run(
-                ["git", *arguments],
+                [executable, *arguments],
                 check=False,
                 stdin=subprocess.DEVNULL,
                 stdout=output,
@@ -273,13 +283,40 @@ def _run_git(start: Path, *arguments: str) -> bytes:
     return _run_git_command("-C", str(start), *arguments)
 
 
-# ``None`` = no scope: every call probes, exactly as before #1796. An open
-# scope carries a single-element mutable cell — resolutions inside the scope
-# frequently run under ``asyncio.to_thread``, whose copied context shares the
-# cell object, so a verification recorded in a worker thread is visible to
-# the scope's later resolutions while still evaporating with the scope.
-# Concurrent tasks get their own cells, so they cannot observe each other.
-_PROBE_SCOPE: ContextVar[list[bool] | None] = ContextVar("git_capability_probe_scope", default=None)
+class _ProbeScopeCell:
+    """Mutable scope state shared across context copies.
+
+    ``executable`` holds the verified absolute path once the scope's single
+    probe succeeds; every scoped git command binds to it. ``closed`` is set
+    on scope exit — the cell object is shared by reference with any context
+    copies (``asyncio.to_thread`` workers, escaped child tasks), so closing
+    it revokes the scope's trust everywhere at once.
+
+    Today's only scope (``prepare_session``) resolves sequentially; two
+    concurrent cold resolutions in one scope would issue duplicate probes,
+    which is benign (never stale) and intentionally unsynchronized — no lock
+    means nothing to inherit across ``fork()``.
+    """
+
+    __slots__ = ("closed", "executable")
+
+    def __init__(self) -> None:
+        self.executable: str | None = None
+        self.closed = False
+
+
+# ``None`` = no scope: every call probes and every command resolves ``git``
+# from PATH, exactly as before #1796.
+_PROBE_SCOPE: ContextVar[_ProbeScopeCell | None] = ContextVar(
+    "git_capability_probe_scope", default=None
+)
+
+
+def _active_probe_scope() -> _ProbeScopeCell | None:
+    cell = _PROBE_SCOPE.get()
+    if cell is None or cell.closed:
+        return None
+    return cell
 
 
 @contextmanager
@@ -289,25 +326,42 @@ def git_capability_probe_scope() -> Iterator[None]:
     A session start resolves project identity at least twice — once while the
     execution contract is built and again at the publication boundary — and
     each resolution re-ran ``git --version`` (#1796). Within one scope the
-    probe runs once; the reuse window is the same class the per-resolution
-    design already accepts between its own probe and topology queries, so no
-    cross-time cache (and none of its integrity obligations: executable
-    binding, shim detection, fork-safe locking) is introduced. Outside any
-    scope, behavior is unchanged: every resolution probes.
+    probe runs once and every scoped git command binds to the executable it
+    verified, so a later resolver skipping its own probe can never run a
+    different binary than the one proven supported. The reuse window is the
+    same class the per-resolution design already accepts between its own
+    probe and its topology queries; no cross-time cache is introduced, and
+    the scope's trust is revoked on exit even for escaped tasks. Outside any
+    scope, behavior is unchanged: every resolution probes and every command
+    resolves ``git`` from PATH.
     """
-    token = _PROBE_SCOPE.set([False])
+    cell = _ProbeScopeCell()
+    token = _PROBE_SCOPE.set(cell)
     try:
         yield
     finally:
+        # Close the shared cell before resetting: context copies in worker
+        # threads or escaped child tasks hold the same object, and nothing
+        # may outlive the scope (#1796 review round six).
+        cell.closed = True
         _PROBE_SCOPE.reset(token)
 
 
 def _require_supported_git() -> None:
     """Reject Git versions that lack the unambiguous topology query grammar."""
-    scope_cell = _PROBE_SCOPE.get()
-    if scope_cell is not None and scope_cell[0]:
-        return
-    output = _run_git_command("--version")
+    scope_cell = _active_probe_scope()
+    executable: str | None = None
+    if scope_cell is not None:
+        if scope_cell.executable is not None:
+            return
+        # Resolve the binary once and probe exactly that path; scoped commands
+        # then bind to it, so a later resolver skipping its own probe cannot
+        # run a different git than the one verified here (round six).
+        resolved = shutil.which("git")
+        if resolved is None:
+            raise ProjectIdentityUnavailableError("Git executable is unavailable on PATH")
+        executable = os.path.abspath(resolved)
+    output = _run_git_command("--version", executable=executable)
     match = _GIT_VERSION_PATTERN.fullmatch(output)
     if match is None:
         raise ProjectIdentityGitVersionError(
@@ -321,8 +375,8 @@ def _require_supported_git() -> None:
             f"Git {_MINIMUM_GIT_VERSION_TEXT} or newer is required for project identity; "
             f"found {found}"
         )
-    if scope_cell is not None:
-        scope_cell[0] = True
+    if scope_cell is not None and not scope_cell.closed:
+        scope_cell.executable = executable
 
 
 def _git_path(output: bytes) -> Path:
--- a/src/ouroboros/core/project_identity.py
+++ b/src/ouroboros/core/project_identity.py
@@ -16,6 +16,7 @@ import shutil
 import stat
 import subprocess
 import tempfile
+import threading
 from uuid import NAMESPACE_URL, uuid5
 
 PROJECT_ID_PREFIX = "project_"
@@ -274,12 +275,28 @@ def _run_git(start: Path, *arguments: str) -> bytes:
 
 
 _verified_git_key: tuple[str, int, int, int, int] | None = None
+_git_probe_lock = threading.Lock()
 
 
 def _reset_git_capability_cache_for_tests() -> None:
     """Clear the capability cache (test isolation only)."""
     global _verified_git_key
     _verified_git_key = None
+
+
+def _is_capability_cacheable(executable: str) -> bool:
+    """Only a native binary's capability is pinned by its file metadata.
+
+    A version-manager shim keeps a stable path/device/inode/mtime while
+    delegating to whatever Git its external configuration currently selects,
+    so a script's metadata proves nothing about capability — scripts
+    (shebang-led files) are probed on every gate call (review round four).
+    """
+    try:
+        with open(executable, "rb") as handle:
+            return handle.read(2) != b"#!"
+    except OSError as exc:
+        raise ProjectIdentityUnavailableError("Git executable is unavailable") from exc
 
 
 def _current_git_key() -> tuple[str, int, int, int, int]:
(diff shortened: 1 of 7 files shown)


sumin220:
Addressed the first review in 839dd7db. The cache key is now the resolved executable's identity — path, device, inode, mtime, plus the PID — computed with two syscalls and no subprocess, so the ~36 ms saving stands while the fail-closed contract is restored: a git that disappears from PATH fails closed immediately (the warm cache invalidates itself through the failed lookup), a replaced or PATH-switched binary is re-probed exactly once, forked children re-verify, and failures remain uncached. The resume test drops its manual cache clear — simulating a vanished git now runs the unmasked production sequence. Two new regressions cover warm-cache missing git and replaced git (both failing against the boolean cache). Full suite 14745 passed; ruff and mypy clean.

sumin220:
Addressed the second review in 3a717212. Blocker: _run_git_command now accepts an explicit executable and the gate probes exactly the binary its key resolved, so key resolution and the probe can no longer race a mutable PATH — the regression sequences the PATH lookup to switch between the two steps and asserts the probe ran the key's executable (failing before). P2: added same-path inode/mtime replacement and forked-child PID regressions exercising those key fields directly. Suggestion: the docstring now describes key resolution as a filesystem metadata lookup without a subprocess. Full suite 14748 passed; ruff and mypy clean.

sumin220:
Addressed the third review in 3bd05a4a. The selected executable is anchored with os.path.abspath before both the stat that forms the key and the probe execution, so a relative which() result — an empty or relative PATH entry resolving against the mutable cwd — can no longer reopen the lookup race through cwd. The regression makes which() return a bare relative name from a temporary cwd and asserts the probe executed the absolute path (failing before). Full suite 14749 passed; ruff and mypy clean.

sumin220:
Addressed the fourth review in 94a623ec. Blocker: capability caching is now limited to native binaries — a shebang-led script keeps stable metadata while its delegation target changes, so shims are probed on every gate call (a two-byte magic read decides, no subprocess), and the regression shows an unchanged shim downgrading from 2.44 to 2.20 being re-probed and rejected. Native binaries, whose capability is a function of their content, stay cached (pinned by its own regression). P2: concurrent cold callers now share a single probe through a module lock, with a two-thread regression asserting one probe. Full suite 14752 passed; ruff and mypy clean.

sumin220:
Addressed the fifth review in 02cf0572 by redesigning rather than patching. Each round was finding a real obligation of the same root cause — a process-lifetime cache makes the gate vouch across time for a binary it cannot see — so that machinery (executable keying, TOCTOU binding, path anchoring, shim magic, the probe lock) is removed entirely. The probe is now shared only inside a scope prepare_session opens around one session start, where the issue's measured redundancy actually lives; a mutable cell in a ContextVar carries the verification through asyncio.to_thread while concurrent sessions stay isolated and nothing survives the scope. The reuse window matches the class the per-resolution design already accepts between its probe and its topology queries. On this round's findings concretely: no lock exists, so fork inheritance cannot deadlock; no cacheability judgment exists, so .bat/compiled launchers and shims are probed per resolution exactly as on main; no header read exists, so execute-only binaries are unaffected; and the tests are hermetic fake-probe counters. Outside a scope behavior is byte-identical to main. Saving is one probe per session start (~36 ms) — the deliberate trade against the integrity surface. Full suite 14746 passed; ruff and mypy clean.

sumin220:
Addressed the sixth review in ace0ff60. Blocker 1 is the important one and you were right that the scope was unsound as written: skipping the publication resolver's probe only holds if its topology queries cannot run a different git, so the scope now resolves the binary once, probes exactly that absolute path, and binds every scoped git command to it — the capability verified is the capability every scoped command exercises, and a mid-scope PATH switch changes nothing (regression pins this). Blocker 2: scope exit now closes the shared cell and readers treat a closed cell as no scope, so a task created inside the scope re-probes after exit rather than inheriting trust through its context copy (regression included). P2: concurrent cold resolutions inside one scope can still duplicate a probe — documented as benign and deliberately unsynchronized, since duplicates are never stale and a lock would reintroduce the fork-inheritance hazard the previous design had. Full suite 14748 passed; ruff and mypy clean.

Q00:
## The 3.14 failure was inherited; #1795 is now on main

`Test Python 3.14` was failing on `test_detached_auto_job_wait_stdout_is_deterministic_for_identical_inputs`:

```
AssertionError: assert '2026-07-29 0..._cli_stable\n' == '2026-07-29 0..._cli_stable\n'
```

Two stdout captures that should be identical, both beginning with a log timestamp — log lines leaking into stdout, not anything this PR did. The cause was `reset_logging()` dropping to structlog's library defaults (stdout, unfiltered). #1795 fixed it and merged as `debb9d079`.

**A rebase onto main should clear that check.** This PR is based on `ffc94e8a7`, which predates the fix, so a bare re-run will not pick it up.

The two review blockers are unrelated to that and still stand — both look right to me:

1. **The publication-time Git-version gate is skipped.** Once contract construction marks the cell verified, the second resolution inside the same scope trusts it. A Git downgrade between construction and publication goes unnoticed, which breaks the V1 contract that every public resolver validates ≥2.36 before topology queries.
2. **The cell escapes the scope via `asyncio.create_task()`.** A child task inherits the same list object; resetting the parent `ContextVar` does not touch the child's copy, so the "nothing survives the scope" contract does not hold for escaped tasks.

For what it's worth, the scoped-`ContextVar` design is a better answer to #1796 L1 than what I wrote in the issue — I had suggested `functools.cache`, which would need explicit handling to avoid caching a transient `ProjectIdentityUnavailableError` and would introduce a cross-time cache with its own binding and shim-detection obligations. Bounding reuse to a session start avoids that class entirely. Worth fixing the two escapes rather than changing approach.

Measured baseline for context, on a linked worktree of this repo: 6 git spawns and ~321 ms per resolve, ~640 ms per session start since it resolves twice. `git --version` is ~36 ms of that (11%). The bigger remaining lever is #1796 L2 — whether the publication-boundary revalidation can reuse the captured `_LiveDirectoryEvidence` instead of re-querying Git topology.

cc @sumin220

sumin220:
Rebased onto main (debb9d07, which carries #1795) in b986d1ac — that should clear the inherited 3.14 stdout-determinism failure you diagnosed. Thanks for the pointer; I would have spent a while looking for it inside this diff.

Both blockers were fixed in the previous push and survive the rebase. The publication gate is closed by binding rather than by re-probing: the scope resolves the binary once, probes exactly that absolute path, and every scoped git command runs it, so the capability verified is the capability every scoped command exercises — a downgrade cannot slip in between construction and publication because a switched PATH no longer changes which binary the scope runs. The escaped-task hole is closed by making scope exit close the shared cell, with readers treating a closed cell as no scope, so a child created inside the scope re-probes after exit. Regressions pin both.

Noted on L2 — reusing the captured _LiveDirectoryEvidence at the publication boundary is a much bigger lever than this 11%. Happy to look at it separately once this lands, if it is still unclaimed.

Full suite 14753 passed, 18 skipped on the rebased head; ruff and mypy clean.

sumin220:
Round seven settles the sharing question for me: a pathname pins neither content nor capability, content pinning is not portable (no fexecve or procfd equivalent on macOS), and metadata classification of delegating launchers was already refuted in rounds four and five. Of the two options the review leaves, publication-time probing reduces this PR to a no-op — preparation has exactly the two resolutions — and an actually immutable binding does not exist portably. I do not believe an eighth sharing variant survives, so I will not send one.

Instead I would like to point this PR at the lever you named on this thread: publication reusing the captured evidence. Sketch:

- a resolver variant that keeps what resolution already computes and then discards — the _ResolvedProjectIdentity with its _LiveDirectoryEvidence population — extended with marker evidence for each boundary decision the walk made (lstat generation of the accepted or rejected .git marker, plus a content hash when the marker is a gitdir-pointer file, so an in-place repoint is observable).
- the publication boundary runs a stat-only precheck: _revalidate_live_directories over the captured population, re-walk the boundary candidates by lstat, recheck marker generations and hashes. Any delta — or anything it cannot classify — escalates to today's full re-resolution and comparison. The fast path can only accept when nothing it can observe has changed, so it is fail-closed by construction.
- the honest limit, stated up front: mutations inside a .git directory (core.bare flipped, commondir rewritten, a worktree link repointed) change what a full re-resolution would conclude while staying invisible to any stat set — enumerating git's actual inputs would mean reimplementing git. So the real question is the one V1 has to answer: given the RFC classifies this identity as attribution-only, is directory-generation plus marker stability sufficient at publication, or should specific shapes (linked worktrees, gitdir-pointer checkouts) always take the full re-resolution path?

That is an owner call about the contract, same shape as the #1681 sequencing ruling. If you draw the boundary I will implement it in this PR right away. If you would rather keep publication re-resolving wholesale, I will close this as 'L1 is inseparable from the per-resolver validation invariant' — after seven rounds I am confident that is simply true.

For scale: the probe is 36 ms of a 321 ms resolve; evidence reuse addresses the entire second resolve.

sumin220:
Reworked in d3848d87, taking the remedy from the last review literally: the second resolution is now avoided without reusing capability authority, and the probe scope from rounds six through eight is deleted entirely.

Contract construction now captures the resolver's repo-local input closure: every boundary `.git` marker between the effective directory and the checkout root, plus `HEAD`, `config`, `commondir`, and the `worktrees` registry with its per-entry `gitdir` links, each recorded as stat generation plus content hash. At the publication boundary the identity is accepted only when that closure is byte-for-byte unchanged and the published workspace canonicalizes to the exact directory the evidence was captured from. The recheck runs no git subprocess at all, so there is no topology query for the version invariant to guard and no probe result crossing between resolutions.

Anything the capture cannot enumerate fails closed: bare repositories, gitdir-pointer checkouts, config include directives, a commondir file, unreadable or oversized inputs all mark the evidence as escalating, and publication falls back to the full re-resolution, which probes for itself exactly as on main. The round-seven concern cannot recur in this shape, because the fast path never executes git and the escalation path carries no state from the first resolution.

On tests: a mutation matrix checks that each closure input individually destabilizes the evidence (config edit, commondir appearing, a new intermediate marker, HEAD rewrite, worktrees registry appearing); the happy path asserts zero subprocess spawns during the recheck; escalation is exercised against a real `git worktree add` checkout; and session-level tests cover foreign evidence, drifted closures, and a workspace that does not match the captured directory. The prepare_session wiring test counts publication-boundary re-resolutions instead of probes now. ruff, mypy, and the full suite pass.

If the escalation boundary from my earlier design question ends up drawn differently, `_capture_publication_closure` is the single place to adjust — the accept/escalate classification lives there and nothing else changes.

sumin220:
Addressed the tenth review in f15c494d, all four blockers.

1. Capture is now bracketed around the resolution: the closure is captured before the resolver's queries and re-captured after, and the evidence is usable only when the two captures are byte-identical. A mutation landing between the queries and the capture perturbs a stat generation or a content hash, the pair differs, and the evidence escalates. Your probe shape is the regression: a hook mutates `.git/config` right after `_resolve_canonical_project_identity` returns, and the resulting evidence must escalate.

2. `config.worktree` joined the closure — content-hashed, include-checked, its absence pinned like every other input. A pre-existing file edited in place now destabilizes the evidence (previously only its appearance was caught, via the marker directory's generation), and a `config.worktree` containing an include directive escalates at capture.

3. The include pattern no longer anchors to line starts: it is an unanchored case-insensitive search for `[include` over the raw bytes. Any valid include or includeIf section header must contain that byte sequence regardless of BOM, position, or casing, so the search over-approximates Git's grammar instead of reimplementing it; false positives merely escalate, which is the safe direction. Your exact BOM bytes are a regression, plus a mid-file `includeIf`.

4. Acceptance now rests on what the evidence proves rather than where the object came from: `publication_evidence_is_stable` refuses any evidence that does not carry the complete resolver-shaped closure for exactly the published identity — marker directory at the project root, every administrative file, a `gitdir` link per currently present worktrees entry, a pinned-missing marker for each boundary candidate below the root, and live directory anchors covering the effective directory and the root. Your empty-evidence probe is a regression, and a parametrized test drops each closure entry in turn and requires refusal.

The same commit also updates the PR body, which you rightly called stale, and gives the linked-worktree test an explicit committer identity — that test was failing on CI runners with no global git config, which is what broke the checks on the previous head. Full suite (14,759), ruff, formatting, and mypy pass locally; CI should be green on this head.

On the two "Not met" requirement rows tied to the earlier scoped-probe description: those described the previous design, which this head deliberately replaces along the lines your round-eight remedy suggested — the PR body now says so explicitly. If keeping L2 untouched is a hard sequencing requirement rather than a description of the old approach, I can split the evidence work into its own PR on top of a probe-only change instead; happy to go either way.

sumin220:
Addressed the eleventh review in 7037616e, taking the first remedy you named: reusable evidence now lives behind runner-owned provenance instead of a public call argument.

The `project_identity_evidence` parameter is gone from `create_session` — the signature reverts to its pre-evidence form, so nothing about the fast path is suppliable through the repository API. The resolver deposits the evidence it issues into the publication scope that `prepare_session` owns (the same ContextVar-with-shared-cell transport from this PR's earlier rounds, carrying observations rather than capability), and the publication boundary reads that scope itself.

Acceptance additionally no longer trusts shape at all: `publication_evidence_is_stable` first requires the evidence to be the exact object the resolver issued — an identity-keyed weak-value registry compared with `is`, so an equal-valued reconstruction fails and entries die with the evidence they describe. Only after issuance is proven do the identity, workspace anchor, completeness, and stability checks run. Your probe is the regression, at both levels: a closure assembled with the capture helpers over an empty `.git` directory is refused by the acceptance check, and the session-level test plants that same forged object directly into the scope and requires `create_session` to fall through to the full re-resolution, which fails closed on the invalid marker shape with `ProjectIdentityUnavailableError` and publishes nothing.

As defense in depth the completeness check now also requires a `HEAD` regular file, so the empty-`.git` shape is refused structurally even before considering where the object came from.

Full suite (14,761), ruff, formatting, and mypy pass; CI was green on the previous head after the committer-identity fix and this head changes no CI-visible environment assumptions.

Issue #1797, opened 2026-07-29
refactor(orchestrator): extract OrchestratorRunner collaborators behind the module-size ratchet


Q00:
> [!IMPORTANT]
> This Q00-authored issue is the single extraction epic. PR granularity does not require permanently open phase issues: each pure extraction PR is activated sequentially, just in time, and references this epic.

## Current state and gate

The module-size ratchet shipped in #1806. No collaborator extraction has merged. Q00-authored PRs #1851 and #1854 currently modify `runner.py`; begin the first extraction only after overlapping runner work has merged or closed and the baseline has been remeasured.

## Ordered checklist

- [x] Module-size ratchet - #1806.
- [ ] **Phase 1: ProjectAnchor.** Move project identity resolution, legacy representation, and error translation; preserve the public runner surface and behavior.
- [ ] **Phase 2: execution-contract ownership.** Activate only after Phase 1 merges; move v9 construction, restoration, validation, and normalization into a data-oriented collaborator.
- [ ] **Phase 3: cancellation lifecycle.** Activate only after Phase 2 merges; preserve ordering, competing-terminal behavior, and compatibility entry points.
- [ ] **Phase 4: process-local authority lifecycle.** Activate only after Phase 3 merges; preserve generation, lease, stale-owner, unavailable, and retry-retention semantics.
- [ ] **Phase 5: session persistence.** Activate only after Phase 4 merges; preserve EventStore ordering, idempotency, terminal-winner behavior, schemas, and cold replay.
- [ ] **Phase 6: resume design.** Only after Phases 1-5 merge and the new baseline is measured, reopen a dedicated RFC for the finite resume state machine and implementation sequence.

Each Phase 1-5 PR must be a pure move plus delegation, keep existing test expectations, avoid bug or policy changes, and reseed the ratchet when required. Only one phase is active at a time. No extraction is being marked complete by this reconciliation.

---

## Problem

`OrchestratorRunner` is a single class of **11,100 lines with 180 methods**, inside a
**11,980-line** module. It is the largest object in the codebase and it keeps growing — #1769 alone
added ~324 lines and nine private methods to it.

Measured on `origin/main` (`ffc94e8a7`):

```
CLASS OrchestratorRunner    lines=11100  methods=180
```

Four methods account for **3,647 lines** — a third of the class:

| lines | method |
|---:|---|
| 1238 | `_resume_session_impl` |
| 1194 | `execute_precreated_session` |
|  634 | `_restore_execution_contract` |
[...]

PR #1769, opened 2026-07-27
feat(project-map): anchor sessions to canonical projects

Q00:
## Summary

- Introduces Project Map V1 slice A: one canonical, durable project identity for every session start.
- Anchors linked worktrees, managed task workspaces, and ordinary checkouts without granting the project map dispatch or acceptance authority.
- Makes resume reject project-anchor drift while preserving resumability for historical pre-anchor v9 sessions.

### Canonical project identity

- Derives `project_id` deterministically from the canonical symlink-resolved project root.
- Records a repository-relative POSIX `workspace_path` and supports normal, nested, and provably linked Git worktrees.
- Keeps submodules as separate projects and fails closed for malformed, outside-root, oversized, NUL-containing, or file paths.
- Uses durable `repo_root` / `original_cwd` anchors for managed `TaskWorkspace` instances instead of generated worktree paths.

### Atomic session anchor and resume validation

- Resolves identity once in the same worker call that builds the execution contract.
- Emits matching top-level `project_id`, `project_root`, and `workspace_path` fields on `orchestrator.session.started` and nested frugality-proof scope.
- Reconstructs the immutable session-start identity and validates top-level, nested-contract, and current-workspace equality on resume.
- Retains the historical direct-cwd representation for pre-anchor v9 sessions.

### Documentation and regression coverage

- Adds the Project Map V1 RFC and event-contract documentation.
- Covers Git/worktree/submodule/local-directory identity, fail-closed inputs, event emission, session reconstruction, resume drift, cancellation, and historical compatibility.

## Notes

- Part of #1389; this PR is slice A only. It does not close the issue.
- #1744 already completed bounded escalation; this work is independent from the open Verified-MECE PR.
- Project identity is attribution/read-model metadata only. Advisor, Admission Kernel, and Final Gate authority remain unchanged.
- Follow-up slices will add deterministic EventStore-backed `ProjectRunSummary` / `ProjectRecord` projections, then read-only MCP/CLI surfaces with JSON parity.

## Verification

- `UV_OFFLINE=1 uv run pytest -q` — 14,221 passed, 12 skipped
- Focused broad regression — 694 passed, 1 skipped
- Identity/resume/cancellation follow-up — 165 passed
- `uv run mypy src` — passed (477 source files)
- `uv run ruff check src tests` — passed
- `uv run ruff format --check src tests` — passed (1,131 files)
- Offline wheel build and wheel-packaging test — passed
- `git diff --check` — passed

PR #1709, opened 2026-07-22
fix(auto): reject non-runnable acceptance contracts

sergiobuilds:
## Summary
- define `expected_artifacts` as literal workspace-relative paths across schema and Seed guidance
- reject descriptive labels, POSIX traversal, Windows rooted, drive-relative, drive-absolute, and UNC paths before Grade A
- reject `output_assertion` unless a `verify_command` exists to produce output
- preserve concrete relative files and directories, including paths containing spaces

Closes #1691.

## Verification
1. Focused adversarial contract cases: 4 passed
2. Affected GradeGate, Seed, and SeedGenerator suites: 233 passed
3. Direct POSIX and Windows path probes covered rooted, drive-relative, drive-absolute, UNC, traversal, and valid relative forms
4. Ruff lint, format check, and `git diff --check` passed

The live canonical `cli-todo` scenario reached the harness but was blocked before round 1 because the isolated Claude SDK runtime was not logged in. Nine structural canonical checks passed; no runtime metrics were produced.

## R-run comparison

| Metric | Baseline | This PR | Ratio |
|---|---:|---:|---:|
| Rounds completed | 0, OAuth blocked | 0, OAuth blocked | n/a |
| Per-round wall-clock | n/a, no round | n/a, no round | n/a |
| Terminal reason | Claude SDK not logged in | Claude SDK not logged in | n/a |
| EventStore event count | n/a, pre-round block | n/a, pre-round block | n/a |

## Credit
Retains the original bot-authored implementation commit from #1693 and adds the independent correctness fixes required by its maintainer review.

PR #1806, opened 2026-07-29
ci(module-size): ratchet src module sizes so god-modules can only shrink

Q00:
## Summary

Adds a CI ratchet on `src/ouroboros` module sizes: a module already over the cap may shrink but never
grow, and a module under the cap may not cross it. This is the mechanical floor under the #1797
extraction work — the extractions are worth little if the modules re-inflate between PRs.

No source module is touched. The gate is green on `main` as-is.

## Why now

Measured on `main` @ `ffc94e8a7` (480 Python files, 258,614 lines):

| threshold | files | share of files | lines | share of src |
|---|---:|---:|---:|---:|
| > 3,000 | 9 | 1.9% | 50,541 | 19.5% |
| > 2,000 | 25 | 5.2% | 89,036 | **34.4%** |
| median module | | | 292 | |

The distribution is not gradual. 5% of files hold a third of the source, because they are the only
obvious home for a new concern. Nothing prevented that: #1769 added ~324 lines and nine private
methods to `runner.py` while under review, and no gate noticed.

## The rule

1. Any module **not** in `GRANDFATHERED` must stay at or under `SOFT_CAP` (2,000 lines).
2. A module **in** `GRANDFATHERED` must stay at or under its recorded budget, and must **re-seed**
   once it drops more than `RESEED_SLACK` (200) below it.
3. A grandfathered module that reaches the cap must be **deleted** from the table, not re-seeded. That
   deletion is the unit of progress on #1797.

Rule 2's re-seed half is the part that makes this a ratchet rather than a static cap. Without it, a
PR that removes 500 lines leaves the headroom available for the next contributor to spend. The
failure message prints the exact replacement line, so satisfying it is a one-line edit in the same PR.

## Choosing the cap

2,000 sits in a **natural gap**: no module in the repository is currently between 1,800 and 2,000
lines, and p95 is 2,088. It cannot bite an ordinary module by accident — the median is 292.

Physical lines are counted, blanks and comments included. That is gameable in principle by writing
longer lines, but `line-length = 100` (pyproject.toml:170) already bounds it, and physical length is
what actually costs a reviewer.

## Verification

Live-proven against the real repository, both directions:

```
$ printf '\n# probe\n' >> src/ouroboros/orchestrator/runner.py
$ python3 scripts/check-module-size.py
module-size: FAILED
  src/ouroboros/orchestrator/runner.py: 11982 lines, budget 11980 (+2)     exit=1

$ head -11700 runner.py.bak > src/ouroboros/orchestrator/runner.py
$ python3 scripts/check-module-size.py
module-size: FAILED
    "src/ouroboros/orchestrator/runner.py": 11700,   # was 11980            exit=1

$ # restored
module-size: OK (480 modules, cap 2000, 25 grandfathered)                   exit=0
```

- `tests/unit/scripts/test_check_module_size.py` — 13 passed. Covers all four transitions (growth,
  new module over cap, shrink-to-reseed, reach-cap-to-retire), the vanished-entry fail-loud path,
  the inclusive cap boundary, the `RESEED_SLACK` boundary, a missing source root, and the real
  repository staying green.
- `uv run ruff check` / `ruff format --check` — passed.
- `python3 scripts/check-module-size.py` on a clean tree — OK, exit 0.

## Notes

- No RFC. `CONTRIBUTING.md` has no RFC process, all 32 docs in `docs/rfc/` are product/architecture
  semantics, and all four existing CI gates (`check-auto-boundary.py`, `check-pr-issue-link.py`,
  `check-auto-perf-budget.py`, `check-max-turns-envelope.py`) landed with no `docs/rfc/` change.
  `a0cc33944` is the pattern: a CI script *enforces* a product RFC rather than getting one. The
  rationale lives in the script header, following `pr-hygiene.yml` (#1778).
- A vanished grandfathered path is a **hard error**, mirroring `ANCHOR_FILES` in
  `check-auto-boundary.py` — a rename must not silently drop the cap on whatever the module became.
- `GRANDFATHERED` is a closed set frozen at 2026-07-29. Entries may only be removed.

## Deliberate non-goals

- **No per-file rule against relocating mass into a new module.** An extraction PR that moves 3,000
  lines out of `runner.py` into one new 3,000-line file would be caught by rule 1 (it exceeds the
  cap and cannot be grandfathered), so the escape hatch is already closed without a special rule.
- **`parallel_executor.py` (13,111 lines) is only frozen here, not addressed.** It is a separate
  god-module and needs its own issue; #1797 explicitly scopes it out.
- **No class-level or function-level limits.** `OrchestratorRunner` is 11,100 lines of the 11,980,
  so the module cap is a faithful proxy today. If extraction ever produces a slim module holding one
  fat class, revisit then.

Refs #1797


Q00:
## `Test Python 3.13` failure is inherited from main, not introduced here

```
FAILED tests/unit/evolution/test_generation_watchdog.py::test_no_material_progress_timeout_emits_cancellation_mode
  - AssertionError: expected at least one watchdog_decision event
1 failed, 14822 passed, 8 skipped in 449.89s
```

Evidence it is not this PR:

1. **main fails the same test at this PR's exact base.** Run [30414435950](https://github.com/Q00/ouroboros/actions/runs/30414435950) on `ffc94e8a` (the #1769 merge commit) fails `test_no_material_progress_timeout_emits_cancellation_mode` with the identical assertion, plus `test_session_activity_resets_idle_timeout`.
2. **This PR touches no source.** Only `scripts/`, `.github/workflows/`, and one new test file. 3.12 and 3.14 pass on the same commit.
3. **The root cause is already diagnosed.** The watchdog timeout tests race two thresholds on wall-clock time: `_raise_if_threshold_exceeded` checks `idle_timeout` before `no_material_progress_timeout`, and these tests set no-progress to 0.07s while leaving idle at the 1.0s helper default. Under a loaded parallel run a starved poll loop hits the idle threshold first and cancels for the wrong reason. The second main failure above shows exactly that — `timeout_kind: idle_timeout` where the test expects a no-progress decision.

The fix (drive the tests from the existing `_FakeMonotonicClock` instead of real sleeps) is written but not yet on main. Landing it unblocks main generally, not just this PR.

`enforce-module-size` — the gate this PR adds — passes.

Q00:
## Blocker fixed — the policy is now validated against the base branch

The finding was correct: `GRANDFATHERED` was both the policy and its own baseline, so the anti-growth guarantee was unenforced. `dabb3e7a0`.

Reproduced on the real repository first, so the fix is measured rather than assumed:

```
# grow runner.py by 2 lines AND raise its budget by 2, in one commit
$ python3 scripts/check-module-size.py
module-size: OK (480 modules, cap 2000, 25 grandfathered)          exit=0    ← the bypass

$ python3 scripts/check-module-size.py --baseline-ref <base>
module-size: FAILED
  src/ouroboros/orchestrator/runner.py: 11980 -> 11982 (+2)        exit=1    ← closed
```

### What changed

The proposed policy is now compared with the base branch's:

- **No key may be added** to `GRANDFATHERED`. This also closes re-grandfathering — a retired module is simply absent from the baseline, so bringing it back reads as an addition.
- **No budget may increase.**
- **`SOFT_CAP` may not rise.** Not in the original finding, but the same class of bypass: raising the cap from 2,000 to 20,000 would admit anything, and it is editable in the same commit.

Removals and decreases stay unrestricted — that direction is the point of the gate.

### Trust and failure modes

- The baseline is parsed out of an earlier copy of this script with `ast.literal_eval`, **not imported**. A gate must not execute the code it is judging.
- A ref that does not resolve **fails closed**, matching `check-auto-perf-budget.py`'s contract that indeterminate inputs never produce a silent pass. A skipped comparison is precisely the hole being closed.
- A ref that resolves but carries no copy of the script is this gate's own introducing commit — no predecessor policy exists to tighten, so the check announces itself and is skipped. That is the state of this PR right now, which the CI log will show.
- The workflow now checks out with `fetch-depth: 0` and passes `--baseline-ref origin/main`.

### Tests

13 → **21 passed**. New `TestPolicyTightening` covers every case called out as untested: budget increase, table addition, re-grandfathering, cap raise, plus the permitted directions (lowering, removal) and both baseline failure modes (unresolvable ref, absent baseline script).

`test_clean_tree_passes` kept its behaviour but gained a docstring saying why it only asserts the snapshot layer — that an arbitrary oversized module passes *once listed* is exactly why the table cannot be its own baseline.

### Non-blocking suggestion

Adopted. `E501` is in the `ignore` list at pyproject.toml:187 and the formatter cannot split a long string literal or URL, so `line-length = 100` is awkwardness rather than a bound. The docstring now says so and points at the base-branch comparison as the thing that actually makes the gate hard to walk around.

### Note on `Test Python 3.13`

Still failing, still inherited from main — `test_generation_watchdog.py` fails identically on `ffc94e8a` (this PR's base) with no source change here. #1795 carries the fix. Detail in the [earlier comment](https://github.com/Q00/ouroboros/pull/1806#issuecomment-5115610778).

Q00:
Rebased onto `debb9d079` (post-#1795 main). **All 12 checks green**, including `Test Python 3.13` — that failure was the inherited watchdog flake and #1795 resolved it; main is green again too.

The blocker is addressed in `96d3f9fe3`: the proposed policy is now compared against the base branch's, so a budget raise, a table addition, a re-grandfather, or a `SOFT_CAP` increase all fail regardless of what the source snapshot says. Details and the live before/after reproduction are in [the previous comment](https://github.com/Q00/ouroboros/pull/1806#issuecomment-5118555261). Ready for re-review.

Q00:
@ouroboros-agent rereview exact HEAD d9c5a42e2. Maintainer repairs are pushed; please publish an explicit exact-head verdict after required checks complete.

Q00:
@ouroboros-agent rereview exact HEAD ecf2e8650. The significant-shrinkage transition now compares the predecessor budget with measured current size and requires the proposed replacement budget to equal that size exactly. Regression proves 5000 → current 4500 → proposed 4700 fails while exact 4500 passes. The vanished-path remediation now matches the closed grandfather set. Full script suite: 394 passed; Ruff, format, MyPy, and live gate passed. Please publish an explicit exact-head verdict after required checks complete.

Q00:
@ouroboros-agent Please review exact HEAD `c414d8855`. This normally merges final main `db9a19d05` after #1709 and #1823, then exact-reseeds the introducing module-size baseline: all 486 Python modules were remeasured, the 26 modules above the 2,000-line cap exactly equal the closed GRANDFATHERED key set, and every budget exactly equals current physical lines. The table now captures `seed_generator.py` at 2,637 rather than silently exempting it later. Policy test suite passes 33; Ruff, format, MyPy, diff-check, and `--baseline-ref origin/main` pass. Please publish an explicit exact-head APPROVED or current blocker verdict.

Q00:
@ouroboros-agent Please review final exact HEAD `e0784b90fa8e3c771c432d3d7c2a05ec63eb7c49`. It contains the independently audited exact baseline from `c414d8855` plus only the audit-requested workflow comment correction from the stale `25 / 34.4%` statistic to the measured `26 / 35.3%`. Independent audit found no blocker: exact 26-key/count equality, no <=2,000 grandfather, immutable baseline SHAs, fail-closed policy/baseline bypass probes, 33 tests, Ruff, and format all pass.

PR #1811, opened 2026-07-30
refactor(seed): weigh acceptance criteria by category, not by count

### LINKED ISSUE #1810: Replace the 3-7 acceptance-criteria count rule with a judgment criterion
## Problem

Seed generation controls acceptance-criteria granularity with a **count**: "produce 3-7 acceptance criteria". The number is repeated in four independent places and pinned by tests:

| Location | Form |
|---|---|
| `src/ouroboros/bigbang/seed_generator.py:1103` | user prompt: "produce 3-7 outcome-level criteria" |
| `src/ouroboros/bigbang/seed_generator.py:1210` | system prompt: same + "if you would list more than 7, merge" |
| `src/ouroboros/agents/seed-architect.md:44-47` | "Produce **3-7** acceptance criteria" + "if you draft more than 7, merge" |
| `skills/seed/SKILL.md:130` | quality bar: "3-7 outcome-level items"; "more than 7 criteria" = a difference as severe as a missing piece |
| `src/ouroboros/auto/grading.py:294` | advisory finding fires at `len(acceptance_criteria) > 9`, hint says "aim for 3-7" |
| `tests/unit/bigbang/test_seed_generator.py:1385,1407` | `assert "3-7" in prompt` / `in system_prompt` |

A count is a **proxy** for the property we actually want: *each criterion is one independently valuable, user-visible outcome, and no criterion is a sub-step of a sibling.* That property is already written down right next to the number — but the number is the part that is cheap to satisfy, so it is the part that gets optimized. A model that has 9 genuinely distinct outcomes is instructed to merge two of them; a model that has 2 is nudged to split one. Both are the rule producing exactly the defect it was written to prevent.

## The rule is already breeding rules

The clearest evidence that the number is wrong is that the codebase had to invent a **second, different number** to keep the first one from doing damage. `grading.py:284-292` fires at `>9`, not `>7`, and carries a comment explaining why the grader deliberately disagrees with the prompt:

> The trigger is >9 (not the prompt-level 3-7 target) to leave generous room for genuinely multi-outcome goals and only flag clear over-decomposition.

So we now maintain two thresholds, a rationale for their divergence, and tests that freeze the literal string `"3-7"` into the prompt text. Removing or tuning the rule means touching six sites. This is what a hard rule costs: it does not stay one rule, and it makes the structure rigid rather than the output good.

## Why this direction is wrong in general

Numeric hard rules hand the model an instruction it can comply with **without reasoning**. They replace the judgment we are paying for ("is this criterion an outcome or a step?") with arithmetic ("is this list between 3 and 7?"). The model's ability to tell an outcome from an implementation step is the entire value of having a model in this position — a count rule routes around it.

This is the same failure shape as a growing prohibition list: when the list has to keep growing, or keeps needing exceptions, that is the signal the boundary was drawn in the wrong place.

## Proposed direction

Replace the count with the property and its failure mode, and let the model judge:

- **State the property, not the quantity.** "Each acceptance criterion is one independently valuable, user-visible outcome. An AC that is a sub-step of a sibling AC is a defect, as severe as a missing requirement." (This text already exists — it just needs to stand alone without the number carrying it.)
- **Make the check relational, not numeric.** The grader should ask "is any criterion a sub-step of a sibling?" — a question about the relationship between criteria — rather than counting them. A seed with 11 genuinely orthogonal outcomes should pass; a seed with 4 where two are "create the model" / "add a field to the model" should be flagged.
- **Drop the `>9` threshold and the divergence comment** once the relational check exists; there is nothing left for it to backstop.
- **Unpin the literal string in tests.** Assert the behavior (that over-fragmented seeds get flagged, that orthogonal multi-outcome seeds do not) rather than that a specific substring appears in a prompt.

## Scope

- `src/ouroboros/bigbang/seed_generator.py` (user + system prompt)
- `src/ouroboros/agents/seed-architect.md` (granularity contract)
- `skills/seed/SKILL.md` (quality bar)
- `src/ouroboros/auto/grading.py` (over-fragmentation finding)
- `tests/unit/bigbang/test_seed_generator.py` (string-pinned assertions)

Out of scope for this issue, but worth noting as the same pattern: `src/ouroboros/agents/simplifier.md:61` ("removes at least 50% of components") and `skills/unstuck/SKILL.md:222` ("start with exactly 2 tables").

cohemm:
Closes #1810

## What this changes

The acceptance-criteria granularity contract was carried by a **number** — "produce 3-7 acceptance criteria" — repeated in the extraction prompts, the seed-architect contract, and the QA quality bar, backed by a second and deliberately different threshold (`>9`) in the deterministic grade gate, with the literal string `"3-7"` pinned by tests.

This replaces the quantity with the distinction the quantity was standing in for.

## Why

A count is a proxy for the property we actually want: **a criterion names a state of the finished work a user can see is true; an implementation step names a means of reaching that state, and only the first belongs in the list.** That property was already written next to the number — but the number is the part that can be satisfied without reasoning, so it is the part that gets optimized. A goal with nine genuinely distinct outcomes gets two merged away; a goal with two gets one split. The rule produced the defect it was written to prevent.

The rule had also already started breeding rules: `grading.py` fired at `>9` rather than `>7`, and carried a comment explaining why the gate deliberately disagreed with the prompt. Two thresholds, a rationale for their divergence, and string-pinned tests — six sites to touch before the rule could be tuned at all.

## How

- **Extraction prompts** (`seed_generator.py`) and the **seed-architect contract** now state the outcome/means category difference and why it matters — a means left in the criteria list commits the seed to a path no one has verified — and say that how many criteria a goal has is *discovered by* that judgment. The prose states the perspective and stops there — it does not argue against the number it replaced, since a prohibition aimed at a deleted rule is the same pattern in mirror form; the structural guarantee lives in the tests instead.
- **QA quality bar** (`skills/seed/SKILL.md`) flags a criterion that is intelligible only as a move toward a sibling, weighing each criterion against its siblings rather than against a number.
- **Grade gate** (`grading.py`) drops the `>9` advisory and its plumbing. Over-fragmentation is a judgment about the *relationship* between criteria; a deterministic gate cannot make that judgment, and the threshold only ever proxied for it. The judgment now lives only where judgment exists.
- **Tests** assert the shape of the contract — the distinction is stated, and the rule contains no digit at all — instead of pinning a literal substring.

The guidance is written as a perspective to reason from rather than a list of prohibitions or worked examples, so it stays applicable to goals none of us have seen yet.

## What is deliberately not weakened

Every check the grade gate can actually make deterministically is untouched — observability, success contracts, vagueness, goal match, provenance. A new test covers the case the removal could plausibly have broken: an unobservable criterion inside a long list is still caught.

## Verification

- `uv run pytest tests/unit/auto tests/unit/bigbang -q` → 2024 passed
- `uv run ruff check` / `ruff format --check` on all touched files → clean

This PR's only change under `src/ouroboros/auto/` is a deletion: `GradeGate`
loses the `len(seed.acceptance_criteria) > 9` advisory and the
`advisory_findings` plumbing that carried it. No code path gains an
operation, no per-round work is added, and the gate is not on a per-round
loop — it runs once per seed grade. There is therefore no per-round
wall-clock quantity that this change could move, so the rows are filled
`N/A` explicitly rather than left blank, per the template's substrate-only
guidance.

[x] N/A — PR touches `src/ouroboros/auto/` but adds no per-round work (pure
deletion in a once-per-grade path); no canonical R-run was captured.

## Follow-up (not in this PR)

The same pattern appears in `src/ouroboros/agents/simplifier.md:61` ("removes at least 50% of components") and `skills/unstuck/SKILL.md:222` ("start with exactly 2 tables"), noted in #1810.


Q00:
@ouroboros-agent Please re-review current HEAD 5fc85e13e. Production Auto Seed QA now carries the semantic outcome-vs-implementation-means granularity contract, explicitly preserves independent observable outcomes without a count proxy, and has focused payload coverage. Targeted suites: 250 passed; Ruff passed.

Issue #1239, opened 2026-05-26
RFC: Reference-aware interview adapters with domain glossary packs and provenance prefixes


Q00:
## Problem

Ouroboros interviews should reduce beginner drop-off when domain terms are unfamiliar, but they must not let domain references or preset personas contaminate the user's ontology before the Socratic pass has extracted the user's own problem frame.

The current design tension is:

- Too little domain context: users can stall on vocabulary, especially in UI/UX, game design, AI infra, legal, medical, and other specialized domains.
- Too much domain context: the interview stops surfacing unknown unknowns and starts filling blanks inside a reference-derived frame.
- Persona-first expansion is risky: a strong persona can steer interpretation before the user has stated their actual model.
- Reference material is useful, but only if treated as vocabulary, contrast, or hypothesis rather than requirement source of truth.

In short: reference should inform vocabulary, not ontology.

## Desired behavior

Introduce a reference-aware interview mechanism that preserves the default pure interview path, then injects only the minimum domain vocabulary or reference contrast needed when the session shows evidence that the user is blocked by terms or domain examples.

This should behave more like an interview adapter than a persona system.

The default flow should remain:

1. Blind Socratic pass: extract problem, goals, constraints, success criteria, and open assumptions without domain pack injection.
2. Domain glossary injection only when triggered: attach a small glossary pack when user language or explicit confusion indicates it is needed.
3. Reference as contrast, not context: when a user provides examples or references, use them to ask what assumptions differ, what tradeoffs they imply, and whether the user values surface, structure, or outcome.
4. Distilled provenance before seed contract: separate user-stated requirements from reference-derived candidates and model-inferred hypotheses before generating ACs.

## Proposed solution

Define small `domain glossary pack` or `interview adapter` manifests instead of expanding persona files first.

Suggested manifest shape:

```yaml
name: ui_ux_basics
domain: ui_ux
glossary_terms:
  - term: information architecture
    plain_language: How screens, pages, and content are organized so users can find things.
  - term: affordance
    plain_language: A visual or behavioral cue that suggests what the user can do.
disambiguation_prompts:
  - "When you say this should feel like the reference, do you mean layout, workflow, interaction speed, or outcome quality?"
  - "Which part of the reference is essential, and which part is only an example?"
applies_when:
  - user uses repeated domain vocabulary without defining it
  - user explicitly says they do not understand a domain term
  - interview detects a term-level stall rather than a requirement-level gap
does_not_apply_when:
[...]

PR #1823, opened 2026-07-31
feat(interview): withhold observation answers from requirement extraction

### LINKED ISSUE #1755: Observation-marked answers reach requirement extraction and become Seed requirements

cohemm:
## Summary

Closes #1755. An answer prefixed `[from-code]`, `[from-repo]`, or `[from-research]` is a fact the caller adopted, not a decision the caller made — but every consumer walked `state.rounds` and appended answers verbatim, so an adopted fact could be paraphrased into a Seed acceptance criterion nobody decided. Provenance is now settled once where the answer enters, carried as a typed field, and read by every requirement-producing consumer through one projection.

The design was settled in [this comment](https://github.com/Q00/ouroboros/issues/1755#issuecomment-5134079777); the issue body was amended to match. Two acceptance criteria were struck there and are **not** implemented here on purpose — see "Struck criteria" below.

## What changed

**Decided once, carried as a type**

- `InterviewRound.provenance: Literal["user", "observation"]`, settled in `InterviewState.record_answer` — the one method every transport now funnels through. It fills a round persisted question-only or appends a new one, so the "a round is not always born whole" MCP path is covered by the same decision point. No answer-carrying `InterviewRound(...)` construction remains in the tree.
- `bigbang/answer_provenance.py` holds the single classifier and the projection. `[from-auto]` / `[from-safe-default]` classify as `user`: they are not *human* decisions, but they are decisions, and classifying them as observations would withhold the whole of `ooo auto`'s contribution.

**Withheld at the input, on the walk every consumer shares**

`extraction_rounds()` renders an observation's answer as a fixed note. The content is absent from the input, so a paraphrase is impossible by construction rather than something a later check has to detect — once an extractor rewrites `[from-code] 3 attempts, 2s/4s/8s backoff` into "retry three times with exponential backoff", the marker is gone and no downstream scan can tell it from a decision.

| Consumer | Role | Withholds |
|---|---|---|
| `bigbang/seed_generator.py:1182` | dev Seed extraction | yes |
| `bigbang/pm_interview.py:1229` | PM Seed extraction | yes |
| `mcp/tools/authoring_handlers.py:1489` | plugin Seed generation | yes |
| `bigbang/pm_document.py:290` | PRD generation | yes |
| `bigbang/requirement_distillation.py:129` | deterministic promotion, **no LLM** | yes |
| `bigbang/interview.py:1505` | question generation | no |
| `bigbang/ambiguity.py:579` | ambiguity scoring | no |
| `mcp/tools/authoring_handlers.py:2443` | interview subagent history | no |
| `bigbang/pm_interview.py:565` | PM question classifier | no |

`requirement_distillation` is the one that matters most for the shape of the fix: it promotes answers to `RequirementCandidate`s and `build_promoted_reference_seed` emits Seed YAML from them with no prompt assembled at all, so filtering at prompt assembly would never have reached it. The last two "no" rows share a function with a "yes" row — the projection is applied at the call site, not inside the shared formatter.

**Withholding governs authority, not content**

Question lines render unchanged, including questions that restate an observation. Sharpening the next question is what an observation is *collected for*; the answer slot is the only place requirement authority is conferred. This is pinned by test so a later reader does not "finish the job" — redacting questions would also make the user's own decision uninterpretable ("make it five" of what?).

**Prompt**

`agents/seed-architect.md` gains a `READING THE TRANSCRIPT` section. It previously contained no occurrence of `from-` at all: the convention was explained to the interviewer model every turn and never to the model that produces requirements. This is interpretation support, not a boundary — nothing depends on it being obeyed.

## Upgrade paths that would otherwise stay exposed

Found while implementing, not in the original design:

- **Rounds persisted before this field existed** deserialize without it, so defaulting to `"user"` would leave every in-flight interview's observations holding requirement authority. A `model_validator` settles provenance from the answer — one classifier, still no consumer reading the marker.
- **A distillation cached before this change** would be judged current and reused, because `requirement_input_fingerprint` did not include provenance. It does now.

## Struck criteria (deliberately not implemented)

- ~~"A question generated after a withheld answer does not carry that answer's content into extraction."~~ Read literally this asks *where does the content leak* — the question that failed to converge in #1703 and #1749 — where the governing question is *where is authority conferred*. Satisfying it would suppress intended behavior. Replaced by its inverse, asserted by test.
- ~~"An oversized-context summary standing in for an observation is treated as one."~~ Presupposes an `initial_context` that is an observation; nothing in tree establishes one. It is the user's goal and classifies as a user decision, so the criterion has no subject. `prompt_safe_initial_context` is unchanged.

## Behavior change: a fact no longer confirms a reference contrast

Found by tracing the `ooo publish` egress path after the fact, and worth calling out because it changes when generation is refused.

A reference cue (#1239) — "make it like Stripe" — holds a required `contrast-required` candidate at `resolution=UNKNOWN, confirmation_authority=NONE` until the user says what to copy and what to drop. Answering that contrast question with `[from-code] Stripe uses a hosted payment page` used to resolve it. A fact was conferring a confirmation, and the emitted Seed recorded a confirmation nobody gave — this issue's defect, expressed in the reference lane.

It no longer resolves it. An interview whose only contrast answer is an adopted fact is now refused with the pre-existing `reference_confirmation_required` blocker, as a typed `Result.err` (the guard at `seed_generator.py:802` runs before `build_promoted_reference_seed`; verified, not a raise). The user reopens and answers the contrast.

This looks at first like the "an interview that also carries a user decision stays generatable" non-goal, since the refusal fires even when a real decision is present elsewhere. It is not:

- **No gate is added here.** Generation was already conditional on the reference gate. What changed is that a fact stopped satisfying it falsely.
- **The user's decision is unaffected** — it still renders in the extraction input and still promotes; a separate required confirmation is what is missing.
- The alternative — resolve the contrast but suppress its requirement text — withholds *content* while letting the observation keep conferring *authority*. That inverts the rule this PR implements, so it was rejected.

The issue's non-goal has been reworded to separate "this rule adds a refusal" (it does not) from "an existing confirmation is no longer satisfied by a fact" (it is not). Pinned by four tests, including one asserting a user-decided contrast still resolves, so the gate is not simply always closed.

## Two holes this had, found in review

Both fixed on the branch by @Q00 (0144e06, e3e123e); recorded here because the failure shapes generalize.

**1. A decorated answer defeats prefix classification.** `PMInterviewEngine.record_response` bundled the PM's answer as `PM answer: {user_response}` before handing it to the inner engine. Classification reads the *leading* marker, so `[from-code] ...` arrived as `PM answer: [from-code] ...` and classified as `user` — the reframed PM path defeated the mechanism silently. Fixed by leaving the answer byte-for-byte intact and bundling only the question. The rule for future readers: **anything that decorates an answer before it reaches `record_answer` breaks provenance.** The two remaining PM wrappers (`decide_later`, `skip_as_deferred`) synthesize a whole answer rather than wrapping the user's, and classify as `user` correctly.

**2. One call site can still serve two roles.** I excluded `pm_handler.py:238 _format_pm_transcript` from the surface map because its only caller feeds `build_pm_interview_subagent` — question generation. But that handler serves several actions, and `action == "generate"` is PM Seed generation, which makes the same transcript a requirement-producing input. Fixed by keying the flag off the action rather than the call site. I had already found this "one path, two roles" shape twice (`_format_interview_transcript`, `pm_interview._build_interview_context`) and still missed the variant where the split is by action, not by caller.

`PMInterviewEngine.record_response` bundled the PM's answer as `PM answer: {user_response}` before handing it to the inner engine. Prefix classification reads the *leading* marker, so `[from-code] ...` arrived as `PM answer: [from-code] ...` and classified as `user` — the reframed PM path defeated the whole mechanism silently.

Fixed in 0144e06 by leaving the answer byte-for-byte intact and bundling only the question. The general shape is worth naming for future readers: **anything that decorates an answer before it reaches `record_answer` breaks provenance.** The two remaining PM wrappers (`decide_later`, `skip_as_deferred`) synthesize a whole answer rather than wrapping the user's, and classify as `user` correctly.

## Side effect

The PM plugin path previously skipped `record_adapter_answer`, confusion-term detection, and distillation invalidation. Funnelling it through `record_answer` adds all three, which matches the non-plugin PM path — a behavior change beyond the issue's scope, flagged rather than buried.

## Follow-ups (not in this PR)

- `requirement_distillation.py:90-108` promotes `initial_context` to a required GOAL candidate with USER authority unconditionally. Same class of defect, different door — the problem there is unconditional promotion, not classification.
- `codebase_context` (`seed_generator.py:1170`) is an observation in a designated slot; constraining it needs a position rule on the extractor's output fields.
- `resolve_project_identity` fails with a bare `Cannot resolve project identity` when `git worktree list --porcelain` exceeds `_MAX_GIT_OUTPUT_LENGTH`. The bound worked as designed, but the diagnostic does not name the cause (a local checkout had ~4,977 stale worktree registrations from pytest temp dirs).

## Test plan

- [x] `uv run --no-sync pytest tests/ -n 4 --dist worksteal` — **15,544 passed, 19 skipped, 1 xfailed, 0 failed**
- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check` clean
- [x] `uv run mypy` clean on the changed modules
- [x] 42 new tests in `tests/unit/bigbang/test_answer_provenance.py`

Every assertion is a decidable postcondition on an **input** — the observation's content absent as a string, or the user's decision present as a string. No test scans generated Seed output for paraphrases; that is the undecidable place, and absence from the input is the proof.

Coverage: classification (including compound `[auto-confirmed]` and `[refined]` forms, and the `[from-auto]` → `user` mapping); provenance settled on the fill path, not just the append path; observation content absent + withheld note present, per consumer; the user decision unaffected, per consumer; question lines rendered in full, per consumer; an interview with no observation unchanged, per consumer; the non-LLM distillation path; both non-withholding boundaries; legacy rounds; and the fingerprint.

Two notes on the local run: the venv must be the CI profile (`uv sync --dev --group mcp-test`) since `claude` and `mcp` extras are declared conflicting, and a checkout with thousands of stale git worktrees fails 20 e2e tests until `git worktree prune` — neither is related to this change.


Q00:
@ouroboros-agent Please perform an exact-head zero-trust review of `917cf5de654cc9770e2dd49fbf862b5dcf394d30`. Focus on provenance settlement at every answer-ingress path, legacy persisted rounds, all requirement-producing consumers (including deterministic distillation), retry/resume and cached fingerprint behavior, while preserving raw observations for question generation. Publish an explicit APPROVED or REQUEST_CHANGES verdict tied to this SHA.

Q00:
@ouroboros-agent The PM reframe provenance loss found after the prior approval is fixed at new HEAD `0144e06f215081c52df1fecddb3ad730b16c245b`. Reframed questions remain bundled, while the raw answer is preserved byte-for-byte so `[from-code]`/`[from-research]` stays leading across record, persistence reload, and extraction withholding. Focused 236 tests plus Ruff, format, full MyPy, and diff-check pass. Please perform a fresh exact-head review.

Q00:
@ouroboros-agent The PM plugin generate bypass from exact review `4831563776` is fixed at new HEAD `e3e123e570e4e578f555780986910ecf38c105e3`. `_format_pm_transcript` now applies the shared `extraction_rounds` projection only for `action=generate`, while resume/question-generation retains raw observations. Regression proves the raw fact is present on resume and absent/replaced in the generate subagent prompt. The prior PM reframe persistence fix remains included. Focused 283 tests plus Ruff, format, full MyPy, and diff-check pass. Please perform a fresh exact-head review.

Q00:
@ouroboros-agent New exact HEAD `27d9d7e2ba8a9dec977da7f7369a5fc84d0d205c` additionally closes legacy/public-input gaps: persisted historical reframed PM answers shaped as `PM answer: [from-code] ...` recover observation provenance only under the canonical paired-question envelope, and public raw PRD `qa_pairs` classify the caller-facing marker at ingress before prompt assembly. The concurrent reference-contrast regression commit was preserved through a normal merge. Focused 289 tests plus Ruff, format, full MyPy, and diff-check pass. Please perform a fresh exact-head review across all requirement-producing consumers.

Q00:
@ouroboros-agent Please review exact HEAD `266d4a91e`. This normally merges current main `514f440cc` (#1709) without force-push. The combined boundary preserves strict JSON acceptance-contract parsing while withholding code/research observations from requirement extraction, including PM reframe/generate and historical canonical-envelope recovery. Combined provenance/PM/MCP tests pass 471; SeedGenerator tests pass 361; Ruff, format, full-source MyPy, diff-check, and semantic merge audit are green. Please publish an explicit exact-head APPROVED or current blocker verdict.

Q00 (approved):
Approved exact HEAD 266d4a91e32c9472faa1913fd496ead7ec3b3c03 after normally merging current main/#1709. Independent combined-semantics audit found the merge tree byte-identical to a clean merge, preserved strict one-line JSON acceptance contracts, withheld observations from initial and retry extraction prompts, and retained explicit user decisions. Independent suites passed 922 tests; local combined suites passed 832 tests plus Ruff, format, MyPy, and diff-check. Merge remains conditional on fresh required CI.

Issue #1827, opened 2026-08-01
Ralph loop stops as "completed" on a model-authored `verdict: "pass"` that its own QA numbers contradict


MyoungSoo7:
## Summary

`RalphLoopRunner` stops the whole multi-generation loop and reports it as a
**success** whenever the model writes `"verdict": "pass"`, even when the QA gate's
own numbers in the same payload say the generation failed.

`_qa_passed` (`src/ouroboros/ralph_loop.py:353`) reads only `meta["qa"]["verdict"]`.
That string is kept **verbatim from the model** whenever it is one of
`("pass", "revise", "fail")` — `src/ouroboros/mcp/tools/qa.py:308-314` only
re-derives it from the score when the model's verdict is _invalid_. The numeric
fields published beside it (`qa.py:632-645`) are computed by the gate itself:

```python
"passed": verdict.score >= pass_threshold,
```

So `verdict` is untrusted model text and `passed` / `score` / `pass_threshold`
are the gate's own result, and nothing reconciles them.

## Reproduction

A completion of:

```json
{
  "score": 0.1,
  "verdict": "pass",
  "differences": ["nothing was implemented"],
  "reasoning": "The artifact does not satisfy any acceptance criterion."
}
```

produces this QA meta — note the fields contradict each other:

```
score=0.1  verdict='pass'  loop_action='escalate'  pass_threshold=0.8  passed=False
```

Driving the **public** `RalphLoopRunner.run()` with an evolve handler that returns
evolve_step's real meta shape (`action="continue"`, `is_error=False`):

```
_qa_passed(...) -> True
status='completed'  stop_reason='qa passed'  iterations=1/10  is_error=False
```

The loop terminates as a success after one generation that its own QA gate scored
0.10 against a 0.80 bar and classified `loop_action="escalate"`.

Reproduces identically on the non-JSON prose path
(`_parse_non_json_qa_response`) with:

```
Score: 0.1
Verdict: pass
```

## Why this is authoritative, not cosmetic

`RalphLoopResult.to_tool_result()` sets `is_error = (status == "failed")`, so
JobManager records `JobStatus.COMPLETED` (`src/ouroboros/mcp/job_manager.py:723`).
`src/ouroboros/auto/pipeline.py:2656` only branches to BLOCKED/FAILED when
`terminal_status == "failed"`; a `completed` Ralph therefore falls through to
`_evaluate_or_complete` and, on the no-evaluator / plugin path, reaches
[...]

PR #1828, opened 2026-08-01
fix(orchestrator): gate Ralph's stop decision on the QA result, not the verdict string

MyoungSoo7:
Refs #1827

## What

`_qa_passed` (`src/ouroboros/ralph_loop.py`) decided whether to stop the Ralph loop
from `meta["qa"]["verdict"]` alone. That string is kept verbatim from the model
whenever it is one of the valid tokens (`mcp/tools/qa.py:308-314` only re-derives it
when the model's verdict is _invalid_), while `passed` / `score` / `pass_threshold`
in the same payload are computed by the QA gate itself:

```python
"passed": verdict.score >= pass_threshold,
```

So a completion of `{"score": 0.10, "verdict": "pass"}` stopped the entire
multi-generation loop with `status="completed"`, `stop_reason="qa passed"` — while
that same payload carried `passed=False`, `loop_action="escalate"` against a 0.80 bar.

`completed` is terminal downstream: `RalphLoopResult.to_tool_result()` sets
`is_error = (status == "failed")`, JobManager records `JobStatus.COMPLETED`
(`mcp/job_manager.py:723`), and `auto/pipeline.py:2656` only diverts on `"failed"` —
so a failed generation falls through to `_emit_product_emitted_terminal` +
`AutoPhase.COMPLETE`. One model-authored string turned a failed generation into a
terminal "product emitted".

## Why this is a bug and not a deliberate policy

Every other reader of this payload already gates on the computed field:

| reader                               | gates on                                 |
| ------------------------------------ | ---------------------------------------- |
| `auto/adapters.py:805`, `:879`       | `passed=bool(meta.get("passed", False))` |
| `auto/pipeline.py:2859-2862`         | `qa_result.passed`                       |
| `ralph_loop.py:353` (before this PR) | `verdict` string only                    |

The loop's stop decision was the sole exception. This brings it in line.

## The fix

Treat a `pass` verdict as a pass only when the gate's own numbers do not contradict
it — `qa["passed"] is False`, or `score < pass_threshold` when both are numeric.

A NaN score compares `False` and so lands on the fail-closed side, without needing a
special case.

Payloads that carry a verdict but no numbers keep their current behaviour, which is
what the existing fixtures use — no existing test changed.

## Tests

Two new tests in `tests/unit/test_ralph_loop_progress.py`:

1. **`test_sub_threshold_score_does_not_stop_the_loop_on_a_pass_verdict`** — the
   regression proof. Builds `meta["qa"]` exactly as `QAHandler.handle` publishes it
   and drives the public `RalphLoopRunner.run()`. Verified to **FAIL against the
   reviewed baseline**:

   ```
   >       assert result.stop_reason != "qa passed"
   E       AssertionError: assert 'qa passed' != 'qa passed'
   ```

2. **`test_uncontradicted_pass_verdict_still_stops_the_loop`** — an honest
   `0.95 / "pass"` must still stop immediately. To be explicit: **this one passes both
   before and after the fix.** It is not a regression proof; it is a guard against
   over-correcting the first one into "never stop on a pass verdict".

Existing coverage only ever exercised the _safe_ divergence (high score +
`verdict: "fail"` → keep going, `test_ralph_loop_progress.py:106-161`) and a
`verdict="pass"` with no score at all (`test_ralph_handler.py:189-206`). The unsafe
direction was untested.

## Verification

- `pytest tests/unit/test_ralph_loop_progress.py tests/unit/mcp/tools/test_ralph_handler.py` → **37 passed** (35 before + 2 new)
- `pytest tests/unit/auto` (the downstream consumers) + the two above → **1392 passed**
- `pytest tests/unit/mcp/tools/test_qa_parser.py tests/unit/mcp/tools/test_qa_allowed_tools_wiring.py` → **10 passed**
- Full suite as CI runs it: `pytest tests/ -n 4 --dist worksteal` → **16438 passed, 19 skipped, 1 xfailed** (7m52s)
- `ruff check`, `ruff format --check`, `mypy` on the changed files → clean
- `scripts/check-module-size.py` → OK; no file under `src/ouroboros/auto/` is touched

### DIFF
--- a/src/ouroboros/mcp/tools/evolution_handlers.py
+++ b/src/ouroboros/mcp/tools/evolution_handlers.py
@@ -603,7 +603,8 @@ class EvolveStepHandler(BridgeAwareMixin):
         # Post-execution QA
         qa_meta = None
         skip_qa = arguments.get("skip_qa", False)
-        if step.action.value in ("continue", "converged") and execute and not skip_qa:
+        qa_attempted = step.action.value in ("continue", "converged") and execute and not skip_qa
+        if qa_attempted:
             from ouroboros.mcp.tools.qa import QAHandler
 
             qa_handler = QAHandler()
@@ -650,6 +651,7 @@ class EvolveStepHandler(BridgeAwareMixin):
             "converged": sig.converged,
             "next_generation": step.next_generation,
             "executed": execute,
+            "qa_attempted": qa_attempted,
             "has_execution_output": gen.execution_output is not None,
             "checkpoint_commits": checkpoint_commits,
             "checkpoint_attempted_ac_ids": checkpoint_attempted_ac_ids,
--- a/src/ouroboros/ralph_loop.py
+++ b/src/ouroboros/ralph_loop.py
@@ -298,8 +298,17 @@ class RalphLoopRunner:
                 stop_reason = "qa passed"
                 break
             if action in _TERMINAL_SUCCESS_ACTIONS:
-                status = "completed"
-                stop_reason = action
+                # evolve_step runs post-execution QA for `converged` too and
+                # publishes the pre-QA action unchanged, so a converged step can
+                # arrive carrying an authoritative QA failure. Guarding only the
+                # `qa passed` branch above would leave this exit reporting a
+                # terminal success for the very payload that branch rejected.
+                if _qa_authoritative_failure(final_result.meta):
+                    status = "failed"
+                    stop_reason = "qa_failed"
+                else:
+                    status = "completed"
+                    stop_reason = action
                 break
             if action in _TERMINAL_FAILURE_ACTIONS or final_result.is_error:
                 status = "failed"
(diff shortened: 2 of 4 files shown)


MyoungSoo7:
@ouroboros-agent please re-review exact HEAD `333bf21791bb5f1b7f2acbd7bdb549145a1192f2` and supersede review `4834013168`.

Two follow-up commits. `1f61c5b` extends the authoritative QA gate to the sibling terminal-success exits the blocker named — the `converged` action and the generation-cap fallthrough — instead of guarding only the `qa passed` branch; the generation-cap path keeps `stop_reason="max_generations reached"`, which `auto/pipeline.py` already maps to a recoverable BLOCKED phase, so only the status changes. `333bf21` closes a hole that keying on *presence of a failure* opened: `evolution_handlers.py:639-661` drops the whole `qa` block whenever `QAHandler` errors, so an unparseable QA completion was safer for the run than an honest failing one — absence of the block now fails closed wherever the producer actually ran QA, with `skip_qa` and payloads predating `executed` keeping current behaviour. Tests now assert the full terminal contract — `status`, `stop_reason`, and `to_tool_result().is_error` — rather than `stop_reason` alone, which is exactly why round 1's coverage missed the sibling exits. The non-blocking fixture note is also addressed: `_qa_meta` now emits `loop_action="done"` for a passing score, matching `_determine_loop_action`.

Exact-head local evidence: 6 new test functions, all verified to FAIL at the reviewed baseline `da5a1f3` and to pass at this head; targeted suites 1896 passed; full suite as CI runs it (`pytest tests/ -n 4 --dist worksteal`) 16444 passed, 19 skipped, 1 xfailed; ruff check, ruff format and mypy clean on both changed files. All required checks are green.

Q00 (approved):
Exact-HEAD re-review: producer now emits authoritative qa_attempted state and Ralph consumes it with legacy-only fallback. Focused and combined integration suites pass.

PR #1834, opened 2026-08-01
fix(cli/plugin): stage URL cache refresh so failed clones keep the old cache

### LINKED ISSUE #1826: bug(cli/plugin): failed URL refresh deletes the last-known-good plugin cache
# Plugin URL refresh deletes the last-known-good cache before clone succeeds

## Upstream baseline

- Repository: `Q00/ouroboros`
- Commit: `90ac429a3a3563a2f4ace7240ec87254e400c720`
- Observed: 2026-08-01

## Problem

Both URL-backed plugin entry points remove the existing clone destination before attempting a replacement clone:

- `ooo plugin add <repo-url>` in `src/ouroboros/cli/commands/plugin.py` (`add_command`)
- `ooo plugin install <name> --from <repo-url>` in the same module (`_install_named_from_url`)

If `git clone` then fails because of a transient network, authentication, or remote error, the command exits cleanly but the last-known-good cached repository has already been deleted. This turns a failed refresh into irreversible local cache loss and leaves no cache available for inspection or recovery.

## Reproduction

Run the current CLI through `typer.testing.CliRunner`, create the exact sanitized cache destination with a sentinel file, and patch `_shallow_clone` to raise `subprocess.CalledProcessError(128, ..., stderr="simulated network failure")`.

The result on the baseline commit is identical for both URL paths:

```text
mode=add
exit_code=1
sentinel_before=True
sentinel_after=False
cache_dir_after=False
output=git clone failed: simulated network failure

mode=install
exit_code=1
sentinel_before=True
sentinel_after=False
cache_dir_after=False
output=git clone failed: simulated network failure
```

## Expected behavior

A failed URL refresh preserves the previous cache directory byte-for-byte. A successful refresh may replace it atomically only after the new clone and `git rev-parse HEAD` both succeed.

- Clone into a unique staging sibling instead of deleting the live cache first.
- Atomically replace the live cache only after clone and SHA resolution succeed.
- Clean up staging on failure while preserving the prior cache.
- Cover both `add <repo-url>` and `install <name> --from <repo-url>` with failing regression tests.
- Preserve the current controlled CLI error and exit code.

## Duplicate check

Checked all issue and PR titles returned by the repository-wide lists and searched the most recent 100 issue bodies plus 100 PR bodies for plugin/cache-loss, last-known-good, clone-failure, and cache-deletion terms. No matching report or implementation was found as of 2026-08-01.

sumin220:
Closes #1826.

## Problem

Both URL-backed plugin entry points — `ooo plugin add <repo-url>` and `ooo plugin install <name> --from <repo-url>` — removed the existing cache directory before attempting the replacement clone. A transient network, authentication, or remote error after that point exited cleanly but the last-known-good cached repository was already gone, turning a failed refresh into irreversible local cache loss.

## Fix

The refresh clones into a unique staging sibling (`<dest>.staging-<rand>`) and promotes it over the live cache only after the clone and its HEAD resolution both succeed. The promotion follows the same sibling-rename swap discipline the plugin_home installs already use in `_atomic_replace_dir`: rename the previous cache aside, rename staging into place, restore the backup if the promote fails, and only then drop the backup. A failure at any earlier point removes the staging tree and leaves the previous cache byte-for-byte untouched. The controlled `git clone failed: ...` error and exit code 1 are preserved.

The shared URL→destination sanitization (previously duplicated in both entry points) and the staged refresh live in a new focused module, `cli/commands/plugin_cache.py`. That keeps `plugin.py` under its module-size budget — it shrinks from 3,053 to 3,042 lines against a ratcheted budget of 3,053.

## Tests

- `test_failed_url_refresh_preserves_last_known_good_cache` (parametrized over both entry points) reproduces the issue exactly as reported: sentinel cache in place, `_shallow_clone` patched to raise `CalledProcessError(128, ..., stderr="simulated network failure")`. Verified failing on the baseline (sentinel deleted) and passing with the fix; also asserts no staging/backup litter remains.
- `test_successful_url_refresh_replaces_stale_cache` (both entry points) verifies a successful refresh fully replaces stale cache content, leaves no litter, and still records `source_kind="git"` in the lockfile.
- `test_stage_url_cache_refresh_restores_backup_when_promote_fails` fails the staging→dest rename mid-swap and requires the previous cache to be restored with no litter.

All five plugin CLI test files pass (173 tests), `scripts/check-module-size.py --baseline-ref origin/main` passes, ruff and mypy clean.

### DIFF
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -28,6 +28,10 @@ from typing import Annotated
 
 import typer
 
+from ouroboros.cli.commands.plugin_cache import (
+    stage_url_cache_refresh,
+    url_cache_destination,
+)
 from ouroboros.cli.formatters import console
 from ouroboros.cli.formatters.panels import (
     print_error,
@@ -1864,20 +1868,13 @@ def add_command(
     trust = TrustStore(root=trust_root or DEFAULT_TRUST_ROOT)
 
     if _looks_like_url(target):
-        # Shallow clone into cache_root/<sanitized-host-path>.
-        sanitized = (
-            target.replace("https://", "")
-            .replace("http://", "")
-            .replace("git@", "")
-            .replace(":", "_")
-            .replace("/", "_")
-            .strip("_")
-        )
-        clone_dest = cache_root / sanitized
-        if clone_dest.exists():
-            shutil.rmtree(clone_dest)
+        # Staged shallow clone into cache_root/<sanitized-host-path>; a
+        # failed refresh preserves the last-known-good cache (#1826).
+        clone_dest = url_cache_destination(cache_root, target)
         try:
-            git_sha = _shallow_clone(target, clone_dest)
+            git_sha = stage_url_cache_refresh(
+                lambda staging: _shallow_clone(target, staging), clone_dest
+            )
         except subprocess.CalledProcessError as exc:
             print_error(f"git clone failed: {exc.stderr.strip() if exc.stderr else exc}")
             raise typer.Exit(code=1) from exc
@@ -2464,19 +2461,11 @@ def _install_named_from_url(
 ) -> None:
     """`install <name> --from <repo-url>` qualified form for plugin_home sources."""
     _reject_subdirectory_form(repo_url)
-    sanitized = (
-        repo_url.replace("https://", "")
-        .replace("http://", "")
-        .replace("git@", "")
-        .replace(":", "_")
-        .replace("/", "_")
-        .strip("_")
-    )
-    clone_dest = cache_root / sanitized
-    if clone_dest.exists():
-        shutil.rmtree(clone_dest)
+    clone_dest = url_cache_destination(cache_root, repo_url)
     try:
-        git_sha = _shallow_clone(repo_url, clone_dest)
+        git_sha = stage_url_cache_refresh(
+            lambda staging: _shallow_clone(repo_url, staging), clone_dest
+        )
     except subprocess.CalledProcessError as exc:
         print_error(f"git clone failed: {exc.stderr.strip() if exc.stderr else exc}")
         raise typer.Exit(code=1) from exc
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -28,10 +28,7 @@ from typing import Annotated
 
 import typer
 
-from ouroboros.cli.commands.plugin_cache import (
-    stage_url_cache_refresh,
-    url_cache_destination,
-)
+from ouroboros.cli.commands import plugin_cache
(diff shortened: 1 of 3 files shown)


Q00 (approved):
Exact-HEAD re-review: filesystem promotion errors now follow the CLI contract, with formatting and module-size ratchets green locally. Focused and combined integration suites pass.

sumin220:
Addressed the re-review in e36f9877 (on top of the maintainer's error-reporting commits).

The double-failure branch is now locked down: the new regression fails the staging promotion and then the backup restoration as well, and requires the renamed-aside .bak-* directory to survive with the original bytes intact as the sole remaining copy, with no staging debris. It also asserts both renames were actually attempted, so the test cannot silently degrade into the single-failure case.

The low-priority note is addressed too: the clone-stage cleanup now catches BaseException, so a user interrupt mid-clone drops the partial staging tree while leaving the live cache untouched, with its own regression.

Plugin CLI suites pass (143), ruff, mypy, and the module-size gate are clean.

Q00 (approved):
Exact-HEAD re-review: promotion now catches user interruption inside the same rollback transaction as filesystem errors; the new regression interrupts after the live cache is parked and verifies restoration plus debris cleanup. 107 focused tests, full formatting, MyPy, and module-size ratchet pass.

sumin220:
Addressed the re-review in 0814bf4c. backup_used is now recorded before the parking rename is attempted, so an interrupt landing between the rename's side effect and the flag can no longer skip restoration — and the handler's existing backup.exists() guard keeps the early flag harmless when the rename never actually happened. The regression is your probe shape exactly: the first rename performs its real side effect and then raises KeyboardInterrupt, and the test requires the cache restored at its public path with no staging or backup litter. Plugin CLI suites pass (145), ruff, mypy, and the module-size gate are clean.

Q00 (approved):
Exact-HEAD re-review at 0814bf4c3: rollback ownership is now recorded before the parking rename, and the regression performs the rename side effect before interrupting. The live cache is restored with no staging/backup litter. 108 focused tests, full format, Ruff, MyPy, and module-size checks pass.

Q00 (approved):
Exact-HEAD re-review at 64da31f4b: staging and backup sibling components now have bounded hash-based names independent of the destination basename. A 235-byte destination refreshes successfully and leaves no transaction litter. 109 focused tests, full format, Ruff, MyPy, and module-size checks pass.

Q00 (approved):
Exact-HEAD re-review at 58b6af39a: terminal promotion+restore failure now raises a typed recovery error carrying the exact retained backup path, and both public URL commands render that path with restore guidance. Ordinary promotion failure still re-raises OSError after successful rollback. Focused plugin suite: 111 passed; Ruff, full format, MyPy, module-size, and diff checks pass.

PR #1836, opened 2026-08-01
fix(persistence): stage checkpoint writes so a failed save keeps the old chain

### LINKED ISSUE #1830: CheckpointStore.save() truncates the current checkpoint when a write fails
## Problem

`CheckpointStore.save()` rotates the current checkpoint before opening the destination directly with mode `"w"`. If serialization or the underlying write fails, `save()` returns an error but the previously valid current checkpoint has already moved to `.1`, while level 0 is left as partial JSON.

This violates atomic save semantics: a failed save mutates the last committed state. A subsequent load reports corruption at level 0 and silently returns the older rollback entry. On the first save there is no rollback at all, and repeated failed saves can also consume rollback history.

## Reproduction

Against `origin/main` at `90ac429a3a3563a2f4ace7240ec87254e400c720`:

```python
from pathlib import Path
from tempfile import TemporaryDirectory
from unittest.mock import patch

from ouroboros.persistence.checkpoint import CheckpointData, CheckpointStore

with TemporaryDirectory() as td:
    root = Path(td)
    store = CheckpointStore(root)
    store.initialize()
    first = CheckpointData.create("atomic-write", "execution", {"step": 1})
    assert store.save(first).is_ok
    current = root / "checkpoint_atomic-write.json"
    original = current.read_bytes()

    def partial_dump(obj, fp, *args, **kwargs):
        fp.write('{"seed_id":"atomic-write","phase":')
        fp.flush()
        raise OSError("simulated disk write failure")

    second = CheckpointData.create("atomic-write", "execution", {"step": 2})
    with patch("ouroboros.persistence.checkpoint.json.dump", side_effect=partial_dump):
        result = store.save(second)

    assert result.is_err
    assert current.read_bytes() == original  # fails: current is partial JSON
```

Observed:

```text
save_is_err=True
current_changed=True
current_bytes=b'{"seed_id":"atomic-write","phase":'
rollback_exists=True
Checkpoint corruption at level 0 ... Failed to parse checkpoint JSON
Recovered checkpoint ... from rollback level 1
loaded_step=1
```

## Expected behavior

A failed checkpoint save must leave the previously committed current checkpoint and rollback history unchanged. Readers must observe either the complete old checkpoint or the complete new checkpoint, never a partial destination.

- Serialize the new checkpoint to a uniquely named temporary file in the checkpoint directory.
- Flush and sync the temporary file before atomically replacing the destination.
- Rotate history and publish the new current checkpoint as one failure-safe transaction, with cleanup of temporary files.
- If serialization, flush, sync, rotation, or replace fails, return a controlled error and preserve the last committed current checkpoint and usable rollback history.
- Add regression tests with injected failures during serialization/write and before final replacement, including the first-save and existing-history cases.
- Preserve cross-platform behavior and existing file-lock synchronization.

## Duplicate check

Searched all issue and PR titles and bodies, open and closed, for checkpoint plus atomic, partial write, truncation, crash, replace, temporary-file, and durability terms on 2026-08-01. Related items cover graceful evolution interruption, checkpoint filename compatibility, and broader durability work, but none reports this failed-write replacement behavior.

sumin220:
Closes #1830.

## Problem

`CheckpointStore.save()` rotated the committed checkpoint to level 1 and then opened the destination with mode `"w"` before serializing. A serialization or write failure therefore destroyed the last committed state: level 0 was left as partial JSON, the previous checkpoint had already moved into the rollback chain, and a subsequent `load()` silently recovered an older generation. On a first save there was no rollback at all, and repeated failed saves consumed rollback history.

## Fix

`save()` now serializes the new checkpoint to a staged sibling file, flushes and fsyncs it, and only then rotates and publishes — as one reversible transaction inside the existing file lock:

- The oldest checkpoint is retired aside rather than deleted, and every rotation shift is recorded.
- The staged file is promoted over the current checkpoint with `os.replace`; all moves are same-directory renames.
- If the promotion fails, the recorded shifts are undone in reverse order and the retired oldest checkpoint is restored, reproducing the exact pre-save chain. A rename that fails mid-undo aborts the walk but destroys nothing — every checkpoint still exists at some level, where `load()`'s rollback walk finds it.
- The staged file is consumed by the successful rename; anything left is removed in a `finally`, so failed saves leave no temporary files.
- The staged and retired names are constant-length and seed-free, so they cannot breach the 255-byte filename budget the seed sanitizer enforces for real checkpoint files.

`_rotate_checkpoints` is left in place unchanged — two orchestrator tests drive it directly to simulate post-rotate states.

## Tests

- `test_failed_write_preserves_current_checkpoint_and_history` — the reproduction from the issue: a partial `json.dump` failure must leave the committed bytes, the rotation state, and the directory listing untouched, and `load()` must still return the committed generation. Verified failing on the baseline.
- `test_failed_first_save_leaves_no_partial_checkpoint` — with no prior save, a failed write leaves no level-0 file and no litter. Verified failing on the baseline.
- `test_failed_publish_restores_rotated_history` — fails the staged-file promotion itself after a full rotation history exists; the chain must be restored byte-for-byte and `load()` must return the previous generation.
- `test_successful_save_rotates_and_leaves_no_temp_files` — the staged path preserves the existing rotation contract with no staging or retirement debris.

All 49 checkpoint tests pass, plus the orchestrator agent-process suite that drives rotation directly (319 total). Ruff, mypy, and the module-size gate pass.

### DIFF
--- a/src/ouroboros/persistence/checkpoint.py
+++ b/src/ouroboros/persistence/checkpoint.py
@@ -286,8 +286,12 @@ class CheckpointStore:
                     self._publish_staged_checkpoint(checkpoint.seed_id, staged, checkpoint_path)
                 finally:
                     # A successful publish consumed the staged file via
-                    # rename; anything still here is failure debris.
-                    staged.unlink(missing_ok=True)
+                    # rename; anything still here is failure debris. The
+                    # cleanup itself must never fail a committed save.
+                    try:
+                        staged.unlink(missing_ok=True)
+                    except OSError:
+                        pass
 
             return Result.ok(None)
         except Exception as e:
@@ -324,20 +328,34 @@ class CheckpointStore:
         except Exception:
             # Undo in reverse order; a rename that fails mid-undo aborts the
             # walk but destroys nothing — every checkpoint still exists at
-            # some level, and load()'s rollback walk will find it.
+            # some level or in the retired file, and load()'s rollback walk
+            # finds the levels.
+            undo_complete = True
             for source, target in reversed(shifted):
                 try:
                     os.replace(target, source)
                 except OSError:
+                    undo_complete = False
                     break
-            if retired is not None and retired.exists():
+            # The retired checkpoint may be restored only into a provably
+            # vacated oldest slot. After a partial undo that slot can still
+            # hold an unshifted generation, and restoring over it would
+            # destroy that generation — the retired file then stays on disk
+            # as recoverable debris instead.
+            if retired is not None and retired.exists() and undo_complete and not oldest.exists():
                 try:
                     os.replace(retired, oldest)
                 except OSError:
                     pass
             raise
         if retired is not None:
-            retired.unlink(missing_ok=True)
+            try:
+                retired.unlink(missing_ok=True)
+            except OSError:
+                # Promotion is the commit point; cleanup debris after a
+                # committed save is preferable to reporting failure and
+                # inviting a retry of a transaction that already happened.
+                pass
 
     def load(self, seed_id: str) -> Result[CheckpointData, PersistenceError]:
         """Load latest valid checkpoint with automatic rollback on corruption.
--- a/src/ouroboros/persistence/checkpoint.py
+++ b/src/ouroboros/persistence/checkpoint.py
@@ -258,10 +258,12 @@ class CheckpointStore:
         for rollback support (max 3 levels per NFR11).
 
         Uses file locking to prevent race conditions during concurrent access.
-        The new checkpoint is serialized to a staged sibling file and made
-        durable before any committed file moves, and the rotate-and-publish
-        step is reversible — a failed save leaves the current checkpoint and
-        the rollback chain exactly as they were (#1830).
+        The new checkpoint is serialized to a staged sibling file, flushed,
+        and fsynced — so any process-visible write failure surfaces before a
+        committed file moves (directory entries are not fsynced; power-loss
+        ordering is out of scope) — and the rotate-and-publish step is
+        reversible: a failed save leaves the current checkpoint and the
+        rollback chain exactly as they were (#1830).
 
         Args:
             checkpoint: Checkpoint data to save.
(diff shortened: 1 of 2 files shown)


sumin220:
Addressed the first review in 16662537, both blockers.

1. The retired oldest checkpoint is now restored only when the undo walk completed and the oldest slot is provably vacant — after a partial undo it stays on disk as recoverable debris instead of overwriting whatever generation is still sitting at that level. Your fault-injection scenario is the regression, parametrized over a failure at each of the three undo steps: the test collects every generation surviving anywhere in the directory (levels and debris alike) and requires the full pre-save set {2,3,4,5} to be present.

2. Promotion is now explicitly the commit point. A failure while unlinking the retired checkpoint — or the staged file — after a successful promotion is swallowed as cleanup debris rather than surfacing the committed save as Result.err, so callers are not invited to retry a transaction that already happened. The regression fails the retired unlink after a full-chain save and requires is_ok, the new generation at level 0, and the retirement debris left behind.

Checkpoint suite 53 tests plus the orchestrator agent-process suite (323 total), ruff, formatting, mypy, and the module-size gate pass.

sumin220:
Done in 29320289.

Every save now compacts the chain before rotating: existing generations slide down to fill any holes a partially undone save left behind, so rotation only ever retires the oldest generation when the window below it is genuinely full. Compaction moves exclusively into previously empty slots and never overwrites or deletes, so a failure inside it cannot lose a generation either — the outer error handling then reports the save as failed with the chain still intact.

The regression you asked for is in, parametrized over all three undo steps: fail save six with a partial undo at step N, retry it cleanly, and require the exact chain [6, 5, 4, 3] at levels 0 through 3 with no holes. Your sparse-chain outcomes ([6,None,5,4] and friends) are what the baseline produced before the compaction pass.

Also scoped the durability wording per your non-blocking note: the staged write's flush+fsync is described as surfacing process-visible write failures before any committed file moves, with directory-entry fsync and power-loss ordering explicitly out of scope.

Checkpoint suite 56 tests, agent-process suite included (326 total), ruff, mypy, and the module-size gate pass.

Issue #1256, opened 2026-05-27
RFC: reconcile ooo auto terminal-liveness policy with the narrowed invariant


shaun0927:
> [!IMPORTANT]
> The original global Product-or-Die proposal is not the accepted runtime contract. This RFC remains open only to reconcile repository policy and documentation with the narrowed invariant below.

## Accepted invariant

`ooo auto` must not terminate as `BLOCKED` when the only cause is recoverable process or interview non-convergence. Safety, authority, explicit human confirmation, missing credentials or production permission, runtime cancellation, and irrecoverable execution failure remain legitimate `BLOCKED` or `FAILED` terminals.

## Runtime status

The current runtime and focused regression evidence support the narrowed invariant. No additional runtime change is presently justified.

## Remaining close work

- [ ] Update every authoritative repository document and public operator-facing description that still states or implies the global prohibition on `BLOCKED` (tracked in #1860).
- [ ] Mark the original I1-I6 requirements as historical/rejected wherever they are retained for context.
- [ ] Verify that issue text, tests, and public documentation describe one terminal contract.

This issue may close as completed after that reconciliation lands. Closure will not mean that the original prohibition on every `BLOCKED` terminal was implemented.

Last reconciled: 2026-08-03, `main@cb658aa`.

---

## Original global Product-or-Die proposal (historical, not accepted)

# RFC: `ooo auto` Product-or-Die Invariant — Closure Ladder Redefinition

> **Status:** DRAFT · `needs-design` · proposed L4 anchor for #1157
> **Scope:** Vision-level contract change for `ouroboros auto`. Subordinates
> #1170, #1248, #1249, #1252, #1254 under a single product invariant.
> **Single source of truth:** This document supersedes the implicit "graceful
> BLOCKED" paradigm distributed across those issues. Read this RFC first.

---

## TL;DR

`ooo auto`'s stated vision is:
> *"Accept any task, autonomously fill the interview to capture tacit
> knowledge, generate a seed, and run end-to-end without breaking — so a
> product is emitted fast enough to be inspected as a working artifact."*

The current substrate optimizes for a different vision:
> *"Accept any task, reach an explicit `BLOCKED(reason_code)` terminal state
> with named failure semantics."*

These two visions are not compatible. R3 evidence (HEAD `2f6247c6`, post
PR-A/B/β/γ merge train) confirms the gap: every graceful-BLOCKED fix
landed in the merge train was preempted by a higher-priority BLOCKED
[...]

PR #1840, opened 2026-08-02
fix(providers): let the claude backend run where the SDK cannot (#1839)

### LINKED ISSUE #1839: [Bug] MCP server launcher dropped the claude extra, so llm.backend=claude cannot generate interview questions
## Summary

`03b437fa` (#1807) changed the shipped MCP launcher from `ouroboros-ai[mcp,claude]`
to `ouroboros-ai[mcp]`. The `[mcp]` extra does not contain `claude-agent-sdk`, so
the MCP server process no longer has it — and every user whose config says
`llm.backend: claude` gets `Question generation failed: Claude Agent SDK is not
installed` on the first interview turn.

This is not about MCP 2. It reproduces on the current published release with
`mcp 1.28.1` either way; the only variable is the extra.

## Impact

Interviews stop working for `claude`-backend users as soon as their plugin picks
up the new launcher. It looks fine right now only because installed marketplace
copies lag — mine still read `[mcp,claude]` this morning and updated to `[mcp]`
at 14:08 today, at which point interviews on the *published* build stopped
working on this machine.

The failure surfaces as `Question generation failed: ...` inside an otherwise
normal interview response, which reads like a transient model error rather than a
configuration one, so users are unlikely to connect it to a plugin update.

There is also no backend to move them to. Every other provider has a CLI adapter
that needs no SDK (`codex_cli_adapter.py`, `gemini_cli_adapter.py`,
`copilot_cli_adapter.py`, `goose_cli_adapter.py`); `claude` exists only as an SDK
binding, and `anthropic_adapter.py` reads `ANTHROPIC_API_KEY`, which a Max/Pro
subscriber does not have.

## Environment

- Ouroboros: `0.50.6` (published) — reproduces here; also on a dev build from `main`
- OS: macOS 15 (Darwin 25.5.0, arm64), Python 3.14.0 / 3.12 (uvx env)
- Install: `uvx --from 'ouroboros-ai[mcp]' ouroboros mcp serve`, from
  `.claude-plugin/.mcp.json`
- Config: `llm.backend: claude` in `~/.ouroboros/config.yaml`
- `claude-agent-sdk`: not present in `[mcp]`; `0.2.110` present in `[mcp,claude]`

## Steps to reproduce

With `llm.backend: claude` configured, start each server over stdio and call
`ouroboros_interview` with any `initial_context`:

```bash
# 1. current launcher
uvx --from 'ouroboros-ai[mcp]==0.50.6' ouroboros mcp serve

# 2. pre-#1807 launcher
uvx --from 'ouroboros-ai[mcp,claude]==0.50.6' ouroboros mcp serve
```

## Expected behavior

The launcher shipped in `.claude-plugin/.mcp.json` provides what the configured
default backend needs, so an interview turn produces a question.

## Actual behavior

```
[mcp]         -> Question generation failed: Claude Agent SDK is not installed.
                 Run: pip install claude-agent-sdk
                 (details: {'import_error': "No module named 'claude_agent_sdk'"})

[mcp,claude]  -> Interview started. Session ID: interview_20260802_051550
```

Both report `ouroboros-mcp 1.28.1`, so the `mcp` version is not the variable.

- [ ] A user with `llm.backend: claude` either keeps working across the plugin
      update, or is told at the boundary what to change. A generation failure
      caused by a missing optional dependency should not be reported as an
      interview-content error.
- [ ] A Claude Code subscription remains a usable LLM backend for the MCP server
      **after** the `mcp==2.0.0` pin ships — see the constraint below, which makes
      "put `claude` back in the launcher" a one-release fix at best. The `claude`
      CLI is already located and carried through the adapter
      (`cli_path=.../claude-code/bin/claude.exe`) and then handed to the SDK
      rather than invoked, so driving it as a subprocess is the smallest version
      of this.
- [ ] A test asserts that the shipped `.mcp.json` provides what the configured
      default backend requires, so launcher and backend cannot drift apart
      silently again.

## Constraint that shapes the fix

`mcp==2.0.0` is genuinely what the server code requires, not a conservative pin:
`mcp/server/adapter.py` imports `mcp.server.MCPServer`, which exists only in
MCP 2, and `serve` raises `ImportError` without it. `claude-agent-sdk` requires
`mcp<2.0.0` at every published version, `0.2.128` included.

So once a release carries that pin, `[mcp,claude]` cannot resolve at all, and the
Claude backend cannot live in the server process in any configuration. Restoring
the extra fixes today and not the next release.

Found while verifying PR #1825 against a locally built dev wheel. That PR touches
neither `pyproject.toml` nor any MCP config, and `main` alone reproduces this.

cohemm:
Closes #1839.

## Summary

`llm.backend: claude` cannot generate anything inside an MCP server process. The
adapter's only transport is `claude-agent-sdk`; the SDK requires `mcp<2.0.0`, the
protocol server requires `mcp==2.0.0`, and the shipped launcher asks for `[mcp]`
— so there is no SDK to import. The user sees this inside an interview response:

```
Question generation failed: Claude Agent SDK is not installed.
Run: pip install claude-agent-sdk
```

which reads like a model failure rather than a missing dependency.

This adds a CLI fallback. The `claude` CLI is a separate process carrying the same
subscription, so it is the transport that survives a boundary the SDK cannot
cross.

## What changed

`ClaudeCodeAdapter.complete()` no longer fails when the SDK import fails. If a CLI
path is known it runs `claude -p --output-format json` as a subprocess and returns
the same `CompletionResponse`.

Everything above the transport is untouched — profile resolution, system-message
extraction, the JSON-format reinforcement and `_enforce_json` all run exactly as
before, because only the call at the bottom is swapped.

**This is a fallback, not a replacement.** An environment that has the SDK takes
the path it always did. Streaming, multi-turn tool use and per-turn events belong
to the SDK and are unchanged; print mode has none of them, which is why this is
reached only when the alternative is failing outright. Single-response calls —
question generation, scoring, the advisory lanes — do not use those features.

Four details decide whether it works rather than merely runs:

| | |
|---|---|
| System message | travels as `--append-system-prompt`, matching what the SDK path does with it rather than folding it into the prompt |
| `--permission-mode` | forwarded only for values the CLI accepts. This adapter's vocabulary includes `default`, which the CLI expresses by omitting the flag — passing it through would fail argument parsing on every call and look like a refusal |
| Non-JSON body | means the CLI failed before producing a result envelope (auth prompt, usage limit, rejected flag), so its stderr is carried as the only diagnosis there is |
| Neither transport | the error names both, instead of naming only `pip install claude-agent-sdk` in a process where that cannot help |

## Launcher guard

`test_shipped_mcp_launcher_extras_can_co_resolve` reads the extras out of
`.claude-plugin/.mcp.json` and checks them against `pyproject.toml`, failing if
two of them constrain one distribution incompatibly. `[mcp,claude]` was that
shape: it produces an *unsatisfiable install* rather than a degraded one, so the
server never starts at all. Checked statically, so it needs no network and names
the conflicting distribution when it fails.

## Verification

The reported failure, reproduced and then re-run against this change with
`claude_agent_sdk` forcibly absent from `sys.modules`:

```
before  Question generation failed: Claude Agent SDK is not installed
after   Interview started. Session ID: interview_20260802_052613

        What is the primary goal driving this gating decision — capturing
        revenue from larger accounts, or reducing the security/support burden
        of provisioning SSO — and would any customers currently using SSO on a
        non-enterprise plan lose access, or be grandfathered in?
```

Also confirmed the underlying claim with one variable changed, on one published
version, so the diagnosis does not rest on the resolver:

```
uvx --from 'ouroboros-ai[mcp]==0.50.6'         -> Question generation failed
uvx --from 'ouroboros-ai[mcp,claude]==0.50.6'  -> Interview started
```

Both report `ouroboros-mcp 1.28.1`, so `mcp` version is not the variable — the
presence of the SDK in the process is.

- `tests/unit`: 16072 passed, 13 skipped, 1 failed — `TestWheelPackaging::test_wheel_contains_profile_yamls`, which shells out to `uv build`; it passes on its own run and this PR touches neither `pyproject.toml` nor `profiles/`.
- `tests/unit/providers`: 667 passed, 5 skipped
- `mypy` clean on the changed module; `ruff check` / `ruff format --check` clean
- `scripts/check-module-size.py --baseline-ref origin/main`: OK

## Not in this change

A CLI transport for the *agent runtime* (subagent dispatch), which has the same
SDK dependency and the same boundary. This PR covers the LLM completion path,
which is what the reported failure sits on.

## R-run comparison

`src/ouroboros/auto/` is untouched, so there is no per-round behavior to compare.
Per RFC #1256 §I5 substrate-only guidance every cell is `N/A`.

| Metric | Baseline | This PR | Ratio |
| --- | --- | --- | --- |
| Rounds completed | N/A | N/A | n/a |
| Per-round wall-clock | N/A | N/A | n/a |

### DIFF
--- a/src/ouroboros/cli/commands/setup.py
+++ b/src/ouroboros/cli/commands/setup.py
@@ -3975,12 +3975,12 @@ def _setup_claude(claude_path: str) -> None:
         yaml.dump(config_dict, f, default_flow_style=False, sort_keys=False)
 
     # Do not register the isolated MCP 2 server with this Claude SDK profile.
-    # The persisted runtime and LLM backend below both require
-    # ``claude-agent-sdk`` (and therefore MCP 1.x), while the protocol server
-    # runs in an environment that intentionally excludes that SDK. Registering
-    # it here would create a server that boots but fails lazily on authoring and
-    # execution tools. A future CLI-only Claude LLM adapter can make this path
-    # available without weakening the dependency boundary.
+    # The protocol server excludes ``claude-agent-sdk`` (it pins MCP 1.x), and
+    # the persisted *runtime* backend below still requires it — a registration
+    # here would boot and then fail lazily on execution tools. The *LLM* half of
+    # that dependency is gone as of Q00/ouroboros#1839 (``ClaudeCodeAdapter``
+    # falls back to the ``claude`` CLI), so lifting this skip now needs the same
+    # fallback on ``ClaudeAgentAdapter``, not a change here.
     print_warning(
         "Skipped Ouroboros MCP registration for the standalone Claude SDK profile: "
         "its MCP 1.x runtime cannot share the isolated MCP 2 server process. "
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -1,8 +1,17 @@
-"""Claude Code adapter for LLM completion using Claude Agent SDK.
+"""Claude Code adapter for LLM completion, over the SDK or the CLI.
 
-This adapter uses the Claude Agent SDK to make completion requests,
-leveraging the user's Claude Code Max Plan authentication instead of
-requiring separate API keys.
+This adapter makes completion requests against the user's Claude Code Max Plan
+authentication instead of requiring separate API keys.
+
+It has two transports and prefers the Claude Agent SDK. When the SDK cannot be
+imported it falls back to the ``claude`` CLI in print mode, because the SDK
+requires ``mcp<2.0.0`` while the MCP protocol server requires ``mcp==2.0.0`` --
+so in a server process built from the ``[mcp]`` extra there is no SDK to import
+and the backend could previously generate nothing at all (Q00/ouroboros#1839).
+Policy is shared rather than duplicated: model normalization, turn budget,
+permission vocabulary, retry behavior and empty-response validation are decided
+once and read by both. What differs is the streaming surface -- print mode
+reports once at the end, with no per-turn callbacks or incremental events.
 
 Usage:
     adapter = ClaudeCodeAdapter()
@@ -140,15 +149,19 @@ def _claude_options_field_names() -> frozenset[str]:
 
 
 class ClaudeCodeAdapter:
-    """LLM adapter using Claude Agent SDK (Claude Code Max Plan).
+    """LLM adapter for Claude Code Max Plan, over the SDK or the CLI.
 
-    This adapter provides the same interface as LiteLLMAdapter but uses
-    the Claude Agent SDK under the hood. This allows users to leverage
-    their Claude Code Max Plan subscription without needing separate API keys.
+    This adapter provides the same interface as LiteLLMAdapter but uses the
+    user's Claude Code subscription under the hood, so no separate API key is
+    needed. It prefers the Claude Agent SDK and falls back to the ``claude`` CLI
+    when the SDK is not importable -- see the module docstring for why that
+    happens and what the fallback does and does not carry.
 
     Attributes:
         cli_path: Path to the Claude CLI binary. If not set, the SDK will
             use its bundled CLI. Set this to use a custom/instrumented CLI.
+            Also the binary the fallback transport runs; with no SDK and no
+            configured path, completion fails naming both.
 
     Example:
         adapter = ClaudeCodeAdapter()
@@ -866,6 +879,10 @@ class ClaudeCodeAdapter:
         the first time one of them appears.
         """
         run_once = single_attempt or self._execute_single_request_positional
+        # Name the transport that actually failed. Reporting an SDK failure from
+        # a process that has no SDK sends the reader to the wrong diagnosis --
+        # and that misdirection is the exact shape of the bug this PR fixes.
+        transport_name = "Claude Agent SDK" if single_attempt is None else "claude CLI"
         last_error: ProviderError | None = None
         effective_system_prompt = system_prompt
         phantom_recovery_used = False
@@ -968,7 +985,7 @@ class ClaudeCodeAdapter:
                         backoff_seconds=backoff,
                     )
                     last_error = ProviderError(
-                        message=f"Claude Agent SDK request failed: {e}",
+                        message=f"{transport_name} request failed: {e}",
                         details={"error_type": error_type, "attempt": attempt + 1},
                     )
                     await asyncio.sleep(backoff)
@@ -981,7 +998,7 @@ class ClaudeCodeAdapter:
                 )
                 return Result.err(
                     ProviderError(
-                        message=f"Claude Agent SDK request failed: {e}",
+                        message=f"{transport_name} request failed: {e}",
                         details={"error_type": error_type},
                     )
                 )
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -44,6 +44,7 @@ import inspect
 import json
 import os
 from pathlib import Path
+import shutil
 import traceback
 
 import structlog
@@ -61,6 +62,7 @@ from ouroboros.providers.base import (
 )
 from ouroboros.providers.profiles import resolve_completion_profile_result
 from ouroboros.providers.tool_use_diagnostics import diagnose_tool_use_turn
+from ouroboros.runtime.child_env import DEFAULT_OUROBOROS_STRIP_KEYS, build_child_env
 
 log = structlog.get_logger(__name__)
 
@@ -427,6 +429,10 @@ class ClaudeCodeAdapter:
             # environment the SDK cannot enter. See ``_complete_via_cli``.
             sdk_available = False
             if self._cli_path is None:
+                discovered = shutil.which("claude")
+                if discovered:
+                    self._cli_path = Path(discovered).expanduser().resolve()
+            if self._cli_path is None:
                 log.error("claude_code_adapter.sdk_and_cli_unavailable", error=str(e))
                 return Result.err(
                     ProviderError(
@@ -536,6 +542,20 @@ class ClaudeCodeAdapter:
     _CLI_PERMISSION_MODES = frozenset(
         {"acceptEdits", "auto", "bypassPermissions", "manual", "dontAsk", "plan"}
     )
+
+    @staticmethod
+    def _build_cli_child_env() -> dict[str, str]:
+        """Isolate the direct CLI child and enforce the shared recursion limit."""
+        return build_child_env(
+            strip_keys=(*DEFAULT_OUROBOROS_STRIP_KEYS, "CLAUDECODE"),
+            depth_error_factory=lambda depth, max_depth: ProviderError(
+                message=(
+                    f"Maximum ouroboros nesting depth ({max_depth}) exceeded "
+                    f"while starting claude CLI (depth={depth})"
+                ),
+                details={"depth": depth, "max_depth": max_depth},
+            ),
+        )
 
     def _empty_content_error(
         self,
@@ -673,11 +693,12 @@ class ClaudeCodeAdapter:
         if self._permission_mode in self._CLI_PERMISSION_MODES:
             argv += ["--permission-mode", self._permission_mode]
         if self._allowed_tools is not None:
-            # Unlike the SDK -- which drops an empty list before it reaches the
-            # CLI, hence the ``extra_args`` workaround in
-            # ``_execute_single_request`` -- the CLI honors ``--allowedTools ""``
-            # literally, so the empty envelope seals itself with no enumerate.
-            argv += ["--allowedTools", " ".join(self._allowed_tools)]
+            # ``--tools`` defines the available catalog; ``--allowedTools`` only
+            # suppresses permission prompts for names that are already available.
+            # Both are required for SDK parity, especially ``--tools ""`` for the
+            # no-tools interview/PM/QA envelope.
+            tool_names = " ".join(self._allowed_tools)
+            argv += ["--tools", tool_names, "--allowedTools", tool_names]
         argv += ["--max-turns", str(self._effective_max_turns(config))]
         if self._strict_mcp_config:
             # ``--strict-mcp-config`` blocks MCP discovery and nothing else, but
@@ -708,13 +729,17 @@ class ClaudeCodeAdapter:
 
         timeout = self._timeout or _CLI_DEFAULT_TIMEOUT_SECONDS
         try:
+            child_env = self._build_cli_child_env()
             proc = await asyncio.create_subprocess_exec(
                 *argv,
                 stdin=asyncio.subprocess.PIPE,
                 stdout=asyncio.subprocess.PIPE,
                 stderr=asyncio.subprocess.PIPE,
                 cwd=self._cwd or None,
+                env=child_env,
             )
+        except ProviderError as exc:
+            return Result.err(exc)
         except OSError as exc:
             return Result.err(
                 ProviderError(
--- a/src/ouroboros/providers/claude_code_adapter.py
+++ b/src/ouroboros/providers/claude_code_adapter.py
@@ -519,6 +519,54 @@ class ClaudeCodeAdapter:
         {"acceptEdits", "auto", "bypassPermissions", "manual", "dontAsk", "plan"}
     )
 
+    def _empty_content_error(
+        self,
+        content: str,
+        *,
+        session_id: str | None,
+        stderr_tail: str,
+    ) -> ProviderError | None:
+        """The adapter's contract for an empty body: infrastructure, not an answer.
+
+        A transport that returned ``content=""`` as a success would hand an
+        interview or QA caller a false answer, and would do it past the retry
+        loop -- the loop returns on the first `is_ok`, so nothing downstream gets
+        a second chance. Both transports read this one copy so a body that came
+        back empty means the same thing however it travelled.
+
+        Returns ``None`` when the content is fine, so the caller can treat it as
+        "no objection" rather than having to catch anything.
+        """
+        if content:
+            return None
+        log.warning(
+            "claude_code_adapter.empty_response",
+            content_length=0,
+            session_id=session_id,
+            hint=(
+                "CLI started but produced no content"
+                if session_id
+                else "CLI may still be starting (custom CLI sync, etc.)"
+            ),
+        )
+        message = (
+            "Empty response from CLI - session started but no content produced"
+            if session_id
+            # Phrased to match ``_CLAUDE_CLI_BOOTSTRAP_PATTERNS`` so the retry
+            # loop treats a bootstrap stumble as transient rather than final.
+            else "Empty response from CLI - may need retry (timeout/startup)"
+        )
+        return ProviderError(
+            message=message,
+            details={
+                "session_id": session_id,
+                "content_length": 0,
+                "stderr": stderr_tail,
+                "configured_cli_path": (str(self._cli_path) if self._cli_path else None),
+                "cwd": self._cwd,
+            },
+        )
+
     def _effective_max_turns(self, config: CompletionConfig) -> int:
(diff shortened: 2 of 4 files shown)


cohemm:
Addressed the follow-up findings in `ffcc4d45` — 4 fixed, 2 positions.

The five findings were one defect seen five times: the fallback re-declared the transport configuration instead of sharing it, so it diverged on every field it did not repeat. Each fix below routes through the code that already decides rather than adding a second copy, so the "the CLI forgot X" class shrinks instead of moving.

**Structured-output retries re-enter the missing SDK (`:461`) — fixed.** `complete()` now binds the transport once and threads it into `_enforce_json`, so retries return to whichever transport produced the response. Your probe — one CLI invocation followed by `import of claude_agent_sdk halted` — is pinned as `test_a_json_retry_stays_on_the_transport_that_answered`: prose first, valid JSON second, asserting two CLI spawns and an `is_ok` result. Under the old code that test fails with the ImportError you reproduced.

**`allowed_tools` / `strict_mcp_config` / `max_turns` ignored (`:510`) — fixed.** `--allowedTools` and `--strict-mcp-config` are now forwarded natively. Worth noting the CLI is stricter than the SDK here: it honors `--allowedTools ""` literally, which is exactly what the `extra_args` workaround in `_execute_single_request` exists to force, so the empty envelope seals with no enumerate to keep in sync. `max_turns` has no CLI flag, but under an empty tool catalog there is nothing to call and print mode returns on turn one (`num_turns: 1`, verified); that is documented in the docstring rather than faked. Pinned both directions: `test_the_no_tools_envelope_reaches_the_cli` (sealed) and `test_a_permissive_caller_is_not_sealed_by_the_fallback` (`allowed_tools=None` emits neither flag).

**Model normalization diverges (`:511`) — fixed.** Moved out of `_execute_single_request` into `_normalize_model`, read by both transports. Confirmed the harm first: `claude --model anthropic/claude-sonnet-4-5` returns `is_error: true` with `duration_api_ms: 0`. Pinned across `anthropic/`, `openrouter/anthropic/`, bare, `default`, and non-Anthropic `openrouter/` ids. `CompletionResponse.model` still reports the configured id on both paths, so this changed what is sent, not what is reported.

**Unbounded wait and unreaped child (`:526`) — fixed.** The CLI path takes a finite default instead of inheriting `timeout=None`, and timeout and cancellation both go through `_terminate_process`, which kills and then awaits with its own bounded reap so cleanup cannot become the new hang. Pinned as `test_a_nonresponsive_cli_is_bounded_and_reaped` and `test_cancellation_does_not_leave_the_child_running`, both asserting `kill` called and `wait` awaited.

**`bypassPermissions` through the wrong native flag (`:515`) — position.** The probe does not reproduce. On CLI 2.1.220 `--permission-mode` documents exactly the six values this code forwards, and the invocation succeeds:

```
$ claude -p --output-format json --permission-mode bypassPermissions "say OK"
{"is_error":false,...,"stop_reason":"end_turn"}
```

`ClaudeWorkerTransport._permission_args` is the stale side of that comparison, not the canonical one: its `_CLAUDE_PERMISSION_MODES` predates `auto`/`manual`/`dontAsk`, includes `default` (not a CLI choice), and maps bypass onto `--dangerously-skip-permissions`. Adopting it would send a strictly more dangerous flag and drop three valid modes — it would widen this envelope, not tighten it. If you can show a supported CLI version where `--permission-mode bypassPermissions` fails argument parsing, that reverses the direction and we will share the translation.

**PATH discovery for `claude` (`:378`) — position, taking your documentation alternative.** `_resolve_cli_path` has an explicit trust boundary: `OUROBOROS_CLI_PATH` and its aliases are stripped from a cloned repo's `.env` upstream in `config.loader`, so every path reaching the adapter came from a trusted source. A bare `PATH` lookup would admit a `claude` this adapter never validated, which is a new trust decision rather than a gap in this one. Took the alternative you offered instead.

**Stale setup.py guidance (`:3977`) — fixed.** Updated to say the LLM half of the SDK dependency is gone and that lifting the skip now needs the same fallback on `ClaudeAgentAdapter`, which still has no transport but the SDK.

One correction to my own PR description: I suspected the `CLAUDECODE` nested-session guard would bite the fallback the way it does the SDK path. It does not — with the prompt on stdin (which is how this invokes the CLI) `CLAUDECODE=1` completes normally. No env override added.

Next address for this same concern, so it is on the record rather than found later: the CLI path has no equivalent of `_MAX_RETRIES` / phantom-tool-use recovery. It is out of this change, but it is where "the fallback does not preserve the adapter's contracts" lives next.

Verification: `tests/unit` 16079 passed, 13 skipped, 0 failed. The Python 3.14 CI failure is `test_job_manager_drain.py::test_drain_does_not_recount_terminal_job_while_task_unwinds` (TimeoutError) — it passes locally and this change touches neither the job manager nor MCP. `ruff check` / `ruff format --check` and `mypy` clean on the changed modules; `scripts/check-module-size.py --baseline-ref origin/main`: OK.

cohemm:
Addressed round 2 in `56b76cae` — all four findings fixed, no positions. All four reproduce, and one of them corrects me.

**B1 `--max-turns` dropped (`:580`) — fixed, and you were right where I was wrong.** Last round I wrote in the docstring that `max_turns` has no CLI flag and was moot under an empty tool catalog. I had read `claude --help`, which does not list it. That reasoning was invalid, and the discriminating probe is the control I should have run first:

```
$ echo "say OK" | claude -p --output-format json --bogus-flag-xyz 20
error: unknown option '--bogus-flag-xyz'

$ echo "say OK" | claude -p --output-format json --max-turns 20
{"is_error":false,...,"num_turns":1,"stop_reason":"end_turn"}
```

The CLI rejects genuinely unknown options, so accepting `--max-turns` means it is a real hidden flag and the budget really was being discarded. Forwarded now. The effective value moved into `_effective_max_turns`, read by the CLI argv, the SDK options and the request log — three inline copies became one. Pinned across constructor value, per-request override, and default: `test_the_turn_budget_is_forwarded_to_the_cli`.

**B2 strict isolation incomplete (`:594`) — fixed at the root rather than flag by flag.** You are right that `--strict-mcp-config` closes MCP discovery and nothing else, while the SDK path also empties `setting_sources`, `skills`, `agents`, `plugins`, `hooks`, `include_hook_events`. Rather than translate six SDK fields into six CLI flags, the CLI has one place they all come from: `--agents` and `--plugin-dir` are opt-in and never passed, and everything else arrives through setting sources. So the strict envelope now sends `--setting-sources ""`, which closes the set at its root instead of enumerating members that can rot as the CLI grows new ones. Verified accepted alongside the existing flags. Pinned both directions — the strict envelope asserts it, and `allowed_tools=None` asserts none of the three flags appear.

**F1 fallback bypasses transient retry (`:495`) — fixed.** `_complete_with_transient_retry` now takes the one-shot transport to retry around, so which transport is available decides how the request travels, never how many transient failures it survives. This is the loop the module's own `_CLAUDE_CLI_BOOTSTRAP_PATTERNS` were written for, so the fallback gets them rather than being the one transport that ignores them. Pinned both ways: a 529 overload followed by success retries and returns `recovered` (2 spawns), and a non-transient error still fails on the first attempt (1 spawn).

**F2 launcher guard does not catch `[mcp,claude]` (`tests/.../test_dependencies_configured.py:329`) — fixed, and your diagnosis is exact.** The conflict is transitive: `claude` lists `claude-agent-sdk==0.2.123` and the `mcp<2.0.0` pin is one level down, so comparing direct requirement names sees `mcp` and `claude-agent-sdk`, two different names, and finds nothing. The guard asserted a postcondition it could not decide.

Rather than reach for the dependency graph, the test now asserts the thing that actually gates the failure: the launcher must request exactly `['mcp']`. Proving co-resolution statically needs the full transitive graph and a resolver; asserting the one-line literal in `.mcp.json` needs nothing and cannot silently pass. Confirmed against all three shapes:

```
ouroboros-ai[mcp]         extras=['mcp']            passes=True
ouroboros-ai[mcp,claude]  extras=['mcp', 'claude']  passes=False
ouroboros-ai              extras=[]                 passes=False
```

What this gives up, stated plainly: a genuinely new and compatible extra added to the launcher will fail this test and has to be added deliberately. For the file that decides whether the MCP server starts at all, that is the intended cost rather than a gap.

On the design note — agreed, and that is what these four changes are structurally. Policy (`_effective_max_turns`, `_normalize_model`) and reliability (the retry loop) are now read from one place by both transports rather than re-declared per path. What remains transport-specific is streaming and per-turn events, which print mode structurally does not have and which no single-response caller uses.

Verification: `tests/unit` 16082 passed, 13 skipped, 0 failed; CLI fallback suite 14 passed. `ruff check` / `ruff format --check` and `mypy` clean; `scripts/check-module-size.py --baseline-ref origin/main`: OK. Branch is level with `origin/main`.

cohemm:
Addressed round 3 in `624e97c3` — blocker fixed, suggestion taken.

**B1 empty CLI envelope is a false success (`:693`) — fixed.** Reproduced exactly as you describe: `{"is_error":false,"result":"","stop_reason":"end_turn"}` returned `is_ok` with empty content. The consequence is worse than a missed retry, and worth stating precisely: `_complete_with_transient_retry` returns on the first `is_ok`, so the empty body did not merely skip the retry — it left the loop entirely and arrived at interview and QA callers as an answer indistinguishable from a real one.

Fixed by moving the SDK path's contract (previously inline at 1514–1562) into `_empty_content_error`, which both transports now call. Deliberately not a second emptiness check written beside the CLI response builder: this is the same move as `_normalize_model` and `_effective_max_turns` from the previous rounds, and the useful signal is that neither of those has come back in a later round. Sharing the decision removes the seam; copying it relocates the next divergence. The no-session message is still phrased to match `_CLAUDE_CLI_BOOTSTRAP_PATTERNS`, so a bootstrap stumble stays transient on either transport rather than becoming final on one of them.

Pinned three ways: an empty envelope followed by a real one retries and returns the real answer (2 spawns); a persistently empty CLI fails explicitly with `Empty response` and `content_length: 0` rather than returning the emptiness; and a payload with no `result` key at all takes the same path, since that is the same absence rather than a different one.

**S1 stale docstring (`:577`) — fixed, and the criticism is precise.** The docstring's claim predates round 2: it was written when the fallback forwarded neither `--allowedTools` nor `--max-turns`, and it stopped being true the moment those went in. Semantic evaluation's 20-turn read-only envelope does execute tools on this path. Reworded to say what print mode actually gives up — per-turn callbacks and incremental message events, because it reports once at the end — rather than implying it cannot run tools at all.

On your design note, agreed on where the boundary now sits. Across three rounds the shared surface has grown to model normalization, turn budget, permission vocabulary, retry policy and now response validation; what remains genuinely CLI-specific is argv assembly and process lifecycle, which are the two things this transport exists to do.

One note on your test-coverage remark: no live `claude` binary was available on your side, so for the record the flag-level claims in this PR were checked against a real CLI (2.1.220) rather than inferred — including the control that established `--max-turns` is a real hidden flag (`--bogus-flag-xyz` → `error: unknown option`, `--max-turns 20` → accepted), and that `--setting-sources ""` is accepted alongside `--strict-mcp-config` and `--allowedTools ""`.

Verification: `tests/unit` 16085 passed, 13 skipped, 0 failed; `tests/unit/providers/test_claude_code_adapter.py` 94 passed, 1 skipped. `ruff check` / `ruff format --check` and `mypy` clean; `scripts/check-module-size.py --baseline-ref origin/main`: OK. Branch is level with `origin/main`.

cohemm:
Thanks for the approval. Both follow-ups are addressed in `9d3fe926` rather than deferred — they are low priority but they are the same mistake the original issue was about, so leaving them felt wrong.

**Transport-aware failure message (`:984`) — fixed.** The shared exception handler reported `Claude Agent SDK request failed` for every unexpected error, including those raised on the CLI path. In a process that has no SDK that message points the reader at `pip install claude-agent-sdk`, which is exactly the misdiagnosis #1839 was filed over — the fix would have reproduced the bug's shape in its own error text. The retry loop already knows which transport it is driving, so it names it; the SDK path's message is byte-identical to before, so only the fallback's differs. Pinned by `test_a_failure_names_the_transport_that_actually_failed`, using a parseable-but-malformed envelope that raises inside the CLI path and lands in the shared handler.

**Module and class docstrings (`:1`) — fixed.** They introduced the adapter as SDK-only, so the fallback was discoverable only by reading to the bottom of `complete()`. They now state both transports, why the fallback exists, what is shared across them (model normalization, turn budget, permission vocabulary, retries, empty-response validation) and what is not — the streaming surface, since print mode reports once at the end.

Verification: `tests/unit` 16086 passed, 13 skipped, 0 failed; adapter suite 95 passed, 1 skipped. `ruff check` / `ruff format --check` and `mypy` clean; `scripts/check-module-size.py --baseline-ref origin/main`: OK. Branch level with `origin/main`.

Not merging — leaving that to the maintainer.

cohemm:
Addressed round 5 in `ab0f5d13` — blocker fixed. Your probe reproduces exactly: `prompt.encode("utf-8")` sat inside `communicate()`'s argument list, so it raised after the spawn, and `UnicodeEncodeError` was in none of the handlers there (`TimeoutError`, `CancelledError`, `OSError`) — child left waiting on a stdin that would never be written, one per malformed input.

I did not add `UnicodeEncodeError` to that list, because the list is the actual defect. It is an enumerate, and this probe is precisely the complement of what it enumerated; the next one would be some class nobody wrote down. You offered both remedies in the finding and this takes both, so the question stops being askable rather than getting one more answer:

**Encode before spawning.** Both the prompt and every argv element — system text travels as `--append-system-prompt`, so it carries the same exposure and I would not have caught that by patching the stdin path alone. Unencodable input now fails with no child in existence, so it is not a cleanup path at all. Pinned by two tests that assert `create_subprocess_exec` is **never awaited**, rather than asserting that something spawned gets cleaned up — the stronger postcondition of the two.

**Unconditional reaping after a successful spawn.** The per-type handlers became `except BaseException -> terminate -> re-raise`. `_terminate_process` no-ops on an exited child, so covering every path costs nothing and leaves no path to forget. Scope-wise this is exact: past `communicate()` the child has already exited, so the region where it can leak is precisely the region this block owns. Pinned by `test_any_post_spawn_failure_reaps_the_child`, which raises a `RuntimeError` nobody enumerated and asserts kill + wait.

Worth flagging one thing I got wrong while writing that test, since it would have made the coverage look real without being real: my first version used the existing `_proc()` helper, which reports `returncode=0` — an already-exited child, which `_terminate_process` correctly declines to kill. The test passed the code path and asserted nothing. The lifecycle tests now use a child that is still running, which is the only state in which reaping is observable.

On the design note — agreed, and that is what changed here. Process ownership is now a property of the scope rather than of a list of exception types, so "which exceptions reap?" no longer has an answer that can be gotten wrong.

Verification: `tests/unit` 16089 passed, 13 skipped, 0 failed; adapter suite 98 passed, 1 skipped. `ruff check` / `ruff format --check` and `mypy` clean; `scripts/check-module-size.py --baseline-ref origin/main`: OK. Branch level with `origin/main`.

cohemm:
Thanks. The remaining documentation suggestion is addressed in `911818c8`.

**`complete()` docstring (`:402`) — fixed.** It described itself as working "via Claude Agent SDK" while it has selected between two transports since the first commit on this branch, and it is the one docstring a reader reaches before the module's. It now states that it prefers the SDK and falls back to the CLI, and that the choice is made once at that point — profile resolution, system-message extraction, JSON enforcement and its retries are identical either way. That last part is the property worth stating at this level, because it is what makes the fallback a transport swap rather than a second code path.

Verification: `tests/unit/providers` + `test_dependencies_configured` 710 passed, 5 skipped; `ruff check` / `ruff format --check` and `mypy` clean; `scripts/check-module-size.py --baseline-ref origin/main`: OK. Branch level with `origin/main`.

No open findings on my side. Leaving the merge to the maintainer.

Q00 (approved):
Exact-HEAD re-review: CLI fallback now constrains the available tool catalog, uses the shared isolated child environment, and discovers a PATH-installed CLI only when the SDK is absent. Focused and combined integration suites pass.

PR #1841, opened 2026-08-02
fix(config): pin UTF-8 on config and credentials YAML I/O

### LINKED ISSUE #1831: Config and credentials YAML I/O fails on non-UTF-8 locales
# U06 — Config files depend on the platform default encoding

## Upstream baseline

- Repository: `Q00/ouroboros`
- Ref: `origin/main`
- Exact SHA: `90ac429a3a3563a2f4ace7240ec87254e400c720`
- Verified: 2026-08-01 (Asia/Seoul)

## Reproduction

The working-tree copies of the affected files are byte-identical to `origin/main` (`git diff --quiet origin/main -- src/ouroboros/config/loader.py src/ouroboros/config/models.py` returned 0). The following diagnostic was run with Python UTF-8 mode disabled and the C locale:

```text
PYTHONUTF8=0 LC_ALL=C uv run python <reproduction script>
preferred_encoding=ANSI_X3.4-1968
load_config=FAIL UnicodeDecodeError: 'ascii' codec can't decode byte 0xed in position 17: ordinal not in range(128)
resolve_event_store_path=FAIL UnicodeDecodeError: 'ascii' codec can't decode byte 0xed in position 17: ordinal not in range(128)
```

The fixture was valid UTF-8 YAML containing Korean text in `project.name` and `persistence.database_path`.

## Root cause and affected surfaces

At the exact baseline:

- `src/ouroboros/config/loader.py:503` creates `config.yaml` with `config_path.open("w")`.
- `src/ouroboros/config/loader.py:515` creates `credentials.yaml` with `credentials_path.open("w")`.
- `src/ouroboros/config/loader.py:556` loads `config.yaml` with `config_path.open()`.
- `src/ouroboros/config/loader.py:629` loads `credentials.yaml` with `credentials_path.open()`.
- `src/ouroboros/config/models.py:890` independently reads `config.yaml` with `config_path.read_text()`.

Each operation uses the process locale instead of the UTF-8 encoding expected by YAML and by the rest of the project. A valid config written on a UTF-8 machine can therefore become unreadable on a non-UTF-8 locale. The independent EventStore resolver also leaks raw `UnicodeDecodeError`, because it catches `OSError` and `yaml.YAMLError` but not the decoding failure.

## Duplicate check

All issue and PR titles plus targeted full-text searches for `UTF-8 config`, `UnicodeDecodeError config`, and `locale config.yaml` were checked before filing. No existing item covers config/credentials YAML file encoding. Issue #598 and PR #602 concern Windows hook stdout/stderr, while PR #1819 concerns console output encoding; neither changes these configuration file operations. Recent commits matching UTF-8, encoding, or locale were also inspected by subject and contained no config-file fix.

## Proposed acceptance criteria

- All canonical `config.yaml` and `credentials.yaml` reads and writes explicitly use UTF-8, including the independent EventStore-path reader.
- A regression test runs the relevant operations with a mocked or otherwise controlled non-UTF-8 preferred encoding and non-ASCII YAML content.
- Valid UTF-8 config and credentials round-trip without depending on the host locale.
- Invalidly encoded input fails through the documented domain error boundary rather than leaking an unhandled codec exception.

## Upstream result

Issue URL: pending creation.

sumin220:
Closes #1831.

## Problem

The canonical config/credentials operations opened their files with the platform default encoding. With Python UTF-8 mode disabled under a non-UTF-8 locale (`PYTHONUTF8=0 LC_ALL=C`), loading a valid UTF-8 config containing non-ASCII values raised `UnicodeDecodeError` from `load_config`, `load_credentials`, and the independent EventStore-path reader — the exact failure in the issue's diagnostic — so the application could not start at all on such hosts. The EventStore-path reader also leaked the raw codec exception because its error boundary caught only `OSError` and `yaml.YAMLError`.

## Fix

All five operations named in the issue now pass `encoding="utf-8"` explicitly: both writers in `create_default_config`, both loaders, and `resolve_event_store_path`'s `read_text`. The load boundaries treat a decoding failure the same way as malformed YAML: the loaders raise their documented `ConfigError`, and the EventStore-path reader raises its documented `ValueError("invalid EventStore configuration")`, so an invalidly encoded file can no longer leak an unhandled codec exception.

## Tests

- `test_utf8_config_round_trips_under_non_utf8_locale` runs the full round trip in a subprocess with `PYTHONUTF8=0 LC_ALL=C LANG=C`: the script refuses to run if the preferred encoding still reports UTF-8 (so a broken environment cannot produce a vacuous pass), creates the default config through the real writers, rewrites it with Korean YAML content, and loads everything back through `load_config`, `load_credentials`, and `resolve_event_store_path`. Verified failing on the baseline with the issue's `UnicodeDecodeError`.
- `test_invalidly_encoded_files_fail_through_domain_errors` feeds non-UTF-8 bytes to all three readers and requires the documented domain errors; the EventStore case matches on the documented message, which a leaked `UnicodeDecodeError` (itself a `ValueError`) cannot satisfy. Verified failing on the baseline.

Config suite (489) and persistence suite pass; ruff, mypy, and the module-size gate are clean.

### DIFF
--- a/src/ouroboros/config/loader.py
+++ b/src/ouroboros/config/loader.py
@@ -500,7 +500,7 @@ def create_default_config(
     # Create config.yaml
     default_config = get_default_config()
     config_dict = _model_to_yaml_dict(default_config)
-    with config_path.open("w") as f:
+    with config_path.open("w", encoding="utf-8") as f:
         yaml.dump(
             config_dict,
             f,
@@ -512,7 +512,7 @@ def create_default_config(
     # Create credentials.yaml with secure permissions
     default_credentials = get_default_credentials()
     credentials_dict = _model_to_yaml_dict(default_credentials)
-    with credentials_path.open("w") as f:
+    with credentials_path.open("w", encoding="utf-8") as f:
         yaml.dump(
             credentials_dict,
             f,
@@ -553,9 +553,9 @@ def load_config(config_path: Path | None = None) -> OuroborosConfig:
         )
 
     try:
-        with config_path.open() as f:
+        with config_path.open(encoding="utf-8") as f:
             config_dict = yaml.safe_load(f)
-    except yaml.YAMLError as e:
+    except (UnicodeDecodeError, yaml.YAMLError) as e:
         raise ConfigError(
             f"Failed to parse configuration file: {e}",
             config_file=str(config_path),
@@ -626,9 +626,9 @@ def load_credentials(credentials_path: Path | None = None) -> CredentialsConfig:
         pass
 
     try:
-        with credentials_path.open() as f:
+        with credentials_path.open(encoding="utf-8") as f:
             credentials_dict = yaml.safe_load(f)
-    except yaml.YAMLError as e:
+    except (UnicodeDecodeError, yaml.YAMLError) as e:
         raise ConfigError(
             f"Failed to parse credentials file: {e}",
             config_file=str(credentials_path),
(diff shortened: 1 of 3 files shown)


sumin220:
Addressed the first review in cf3b83fe. The child program now travels as a UTF-8 script file and only ASCII crosses the command line — under the C locale the interpreter cannot decode a non-ASCII argv at all, which is exactly what your run hit on 3.14, while PEP 3120 makes source files UTF-8 regardless of locale. The BADENV guard and the assertions are unchanged, so the test still refuses to pass vacuously when the environment fails to produce a non-UTF-8 preferred encoding. Full config suite passes (489).

PR #1825, opened 2026-08-01
feat(interview): add the data_context advisory lane (#1754)

cohemm:
## Acceptance criteria this PR is measured against

Quoted verbatim from RFC #1754 so the bar is read rather than restated. The
lenses these criteria come from are named in that RFC's **Perspective** section;
each defect below was found by one of them. The RFC
was revised on 2026-08-02 — the lane executes now — and this list moved with it.
A second same-day revision removed `observed_at` (see "Decision 5" below); this
list reflects it.
Where a review's wording and this list differ, this list is the one the RFC
settled; a requirement outside it belongs in the RFC first.

- A data-driven question yields evidence beside it; a non-data question yields a
  no-op finding and completes the fan-out.
- The child takes the measurement. It reads what data stores are described to
  it, runs the read, and returns the aggregate in `values`. It calls tools on
  the same standing consent as every sibling lane, and no transport strips its
  permissions. There is no timestamp field: the measurement's temporal envelope
  is the interview itself, and a past observation is still an observation.

  **The child still does not classify the tool it names**, because whoever holds
  a tool is who classifies it. `operation: "read"` is a label on the request,
  not a promise about the tool, and MCP carries no cost or mutation metadata, so
  there is nothing true for anyone to disclose from a name. That risk is
  accepted rather than dressed in ceremony: the gate it used to justify could
  not price what it was gating, and a disclaimer on every read is one users
  learn to click through.
- Every `no_evidence_reason` is a claim the lane can make about itself, never
  about the host. `no_data_tool_available` is removed for asserting the second.
  `no_data_store_described` says nothing described holds the answer;
  `store_described_but_not_callable` says one does and the lane could not reach
  it, and reaches the parent — which sees the environment — as something to
  handle rather than as a missing data path shown to the user.
- An unreachability verdict requires an attempted call. An empty tool index is
  where to start looking, never where to stop.
- Every honest outcome is spellable, because the lane is `required: true` and a
  contract violation is popped from the aggregation. An empty result is a
  measurement; every bound is named in the field's own description.
- A measurement is rendered only for a round that carries no answer yet. Once
  the round has an answer, it is dropped rather than shown, so evidence can
  inform a decision but never revisit one.
- Lane output is never accepted as an interview answer; a `[from-data]`-prefixed
  answer is withheld from requirement extraction on every generation path. This
  is now the whole of the boundary between a measurement and the Seed.
- The durable record carries no child-authored prose, value, or identifier — and
[...]

PR #1844, opened 2026-08-02
fix(interview): gate the branches the echo actually arrives on (#1754)

cohemm:
Stacked on #1825 (base: `feat/data-context-lane`) so that PR's approval is preserved. Merge #1825 first; GitHub will retarget this to `main`.

## What this fixes

#1825 added a gatekeeper to the pending-round branch of both interview handlers and reported the plugin-passive echo leak closed. **It is not closed.** The branch it gates is dead on the transport that produces the leak.

Plugin mode never persists a question-only round: the child asks the questions and the server only records answers, so `record_answer` appends a complete round and `has_pending` is `False` on every turn. Control always falls to the ungated `else`.

Driving the real handler `start → answer → answer → answer` against a real state directory, reading each round back off disk at `2f2e96d9`:

| turn | `has_pending` | fan-out id in the persisted question |
|---|---|---|
| 1 | False | **yes** |
| 2 | False | **yes** |
| 3 | False | **yes** |

The marker, the bridge banner and `question_advisory_fanout_id` all persist, every turn. After this PR, all three are absent and the round records the issued question.

The regression in #1825 that was supposed to prove otherwise seeded a question-only pending round **by hand** — the one state shape plugin mode cannot reach — so it passed while the live path leaked. That fixture, not the code, is what the previous round verified.

## Two more writers with the same shape

Both on the subprocess path, carrying the *server's own* directive rather than the bridge's. Probed with a directive built by the real renderer, not hand-written; both recorded it whole:

- a session with no rounds yet (`last_question` is the only question there is);
- the Seed-ready-challenge reopen, which is a normal user action.

## Why these cut rather than prefer the record

None of the three has a stored question, so the gate used in #1825 does not transfer. Refusing the echo there records a placeholder instead of what the user was asked — trading a leak for exactly the damage `last_question` exists to repair.

So they cut the producer's declared append. Cutting is warranted by the same thing that makes the echo path work: the producer declared where its own text begins, which is a statement rather than an inference from prose. That is what separates it from the shape-matching that kept truncating questions in earlier rounds. `strip_declared_append` carries the residual in its docstring — a question reproducing a *whole* declared append, marker and opening together, is cut at it, which is narrower than the text this branch otherwise writes into the transcript verbatim.

## Also here, because the same round introduced or exposed them

- `stampBridge` branched on `original === undefined`, so a metadata-only response whose text is `""` gained two leading blank lines. Falsy, not undefined.
- A source-level test pins that `stampBridge` is the only writer appending to a response. The behavioural tests cannot see a fourth call site, and the comment claiming they could was false while `stamp` stays exported.
- Comments this branch falsified: *"the last marker is ours"* (true only while we were the only producer), *"PLUGIN_PASSIVE leaves content unchanged"* in three places (this module appends nothing there — the response a host sees changes anyway), and `_HOST_DIRECTIVE_OPENING` described as *the* opening rather than one of two.

## Verification

`tests/unit/mcp` + `tests/unit/auto` 2923 passed; bridge `bun test` 118 pass; `bunx tsc --noEmit` clean; `mypy src/ouroboros` clean (493 files); `ruff check` / `ruff format --check` clean; `check-module-size.py --baseline-ref origin/main` OK; `check-auto-boundary.py` OK.

New regressions: the driven plugin flow (builds no state by hand), the two subprocess branches, and the bridge source invariant. The #1825 plugin test is kept and its docstring corrected to say what its fixture actually reaches, so it cannot stand in for the transport again.

Refs #1754

### DIFF
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -1247,10 +1247,13 @@ def _extract_interview_question(
     ``directive_was_appended`` so the gate and the renderer share one condition.
     Cutting on shape alone truncated
     a question that quoted the sentinel whenever no directive was there to
-    find — on ``PLUGIN_PASSIVE``, where the response is deliberately left
-    unchanged, that is every turn. Auto would then answer, and persist, a
-    question the server never asked. The producer knows whether it appended;
-    asking it is cheaper and surer than inferring from the text.
+    find — on ``PLUGIN_PASSIVE``, where the server appends none, that is every
+    turn. Auto would then answer, and persist, a question the server never
+    asked. The producer knows whether it appended; asking it is cheaper and
+    surer than inferring from the text. It is also the only honest question to
+    ask, since the server is no longer the only producer that writes into that
+    text: the OpenCode bridge appends to it too, and this gate speaks only for
+    ours.
     """
     if dispatch_appended:
         text = split_appended_dispatch(text)
--- a/src/ouroboros/mcp/tools/advisory_dispatch.py
+++ b/src/ouroboros/mcp/tools/advisory_dispatch.py
@@ -57,10 +57,11 @@ _SEQUENTIAL_HOST_ACTION = "process_payloads_sequentially"
 #: where it is parsed.
 QUESTION_ADVISORY_DISPATCH_MARKER = "<!-- ouroboros-question-advisory-dispatch-v1 -->"
 
-#: The first words of the directive that always follows the marker. Recognising
-#: the pair is what lets the strip tell its own output from a question that
-#: merely quotes the marker; written once here so the emitter and the stripper
-#: cannot drift into disagreeing about what a directive looks like.
+#: The first words of the directive this module appends after the marker.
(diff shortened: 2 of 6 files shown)


cohemm:
Closing. The work here is rolled back to #1825's approved head (2f2e96d9); this branch is not the shape the fix should take.

Recorded for whoever picks it up: the gate #1825 added sits on the pending-round branch, which plugin mode never reaches — it persists no question-only round, so `has_pending` is false every turn and control falls to an ungated `else`. Driving the real handler start → answer → answer → answer, the marker, the bridge banner and `question_advisory_fanout_id` persist on all three turns.

What this branch tried — cutting the producer's declared append on the branches that hold no stored question — does not close it either. The grammar is three byte-literals, so an echo that drops the HTML comment, re-indents the blockquote, re-quotes the append, or normalises the em-dash is not recognised at all; and aiming the cut destroys a question that legitimately quotes a directive. Both directions fail, which is why this is not another aiming rule but a contract question for the RFC.

PR #1858, opened 2026-08-03
fix(cli/mcp): stop test isolation leak of proxy/token vars into os.environ

### LINKED ISSUE #1793: test isolation: shell-env test leaks HTTPS_PROXY / GH_TOKEN / ANTHROPIC_BASE_URL into os.environ for the rest of the worker
`tests/unit/cli/test_mcp_shell_env.py::test_shell_env_whitelist_admits_proxy_and_gateway_vars` leaves four environment variables set in `os.environ` after it finishes. Every later test in the same xdist worker inherits them, and any test that shells out to the network fails.

### Deterministic repro (two tests, ~6s)

```
uv run pytest \
  tests/unit/cli/test_mcp_shell_env.py::test_shell_env_whitelist_admits_proxy_and_gateway_vars \
  tests/unit/orchestrator/test_profile_loader.py::TestWheelPackaging::test_wheel_contains_profile_yamls
```

```
E       AssertionError: uv build failed (rc=2):
E         error: Failed to build `/…/ouroboros`
E           Caused by: Failed to resolve requirements from `build-system.requires`
E           Caused by: Failed to fetch: `https://pypi.org/simple/hatchling/`
E           Caused by: tunnel error: failed to create underlying connection
E           Caused by: dns error
1 failed, 1 passed
```

Either test passes on its own.

### What leaks

Appending a probe test that asserts the four keys are absent:

```
leaked into os.environ: {'HTTPS_PROXY': 'http://proxy:8080',
                         'NO_PROXY': 'localhost',
                         'ANTHROPIC_BASE_URL': 'https://gw.example.com',
                         'GH_TOKEN': 'gho_x'}
```

Running each test in the file individually against that probe confirms `test_shell_env_whitelist_admits_proxy_and_gateway_vars` is the only source.

### Cause

`MonkeyPatch.delitem` only records an undo entry when the key is **present**:

```python
def delitem(self, dic, name, raising=True):
    if name not in dic:
        if raising:
            raise KeyError(name)
    else:
        self._setitem.append((dic, name, dic[name]))
        del dic[name]
```

In this test the sequence is:

1. line 143 — `monkeypatch.delenv(key, raising=False)` while the key is **absent** → records nothing, so there is no undo entry saying "this was unset".
2. line 152 — the code under test (`mcp._ensure_shell_env`) writes the four vars **directly into `os.environ`**, outside monkeypatch's knowledge.
3. line 168 — the cleanup loop `monkeypatch.delenv(key, raising=False)` now sees the key **present**, so it records the *polluted* value as the original and deletes it.
4. teardown — monkeypatch faithfully restores that recorded "original", putting the polluted value back.

So the test's own cleanup loop is what creates the leak. Reproducible in isolation:

```python
def test_double_delenv_semantics(monkeypatch):
    os.environ.pop("ZZ_PROBE", None)
    monkeypatch.delenv("ZZ_PROBE", raising=False)   # absent -> records nothing
    os.environ["ZZ_PROBE"] = "polluted"             # production code writes directly
    monkeypatch.delenv("ZZ_PROBE", raising=False)   # records "polluted" as original

def test_zz_after(monkeypatch):
    assert os.environ.get("ZZ_PROBE") is None       # fails: "polluted"
```

### Impact

- `HTTPS_PROXY=http://proxy:8080` — any later test spawning a subprocess that reaches the network fails. `TestWheelPackaging::test_wheel_contains_profile_yamls` is the one I hit, because it shells out to `uv build`.
- `GH_TOKEN=gho_x` — a later subprocess invoking `gh` would authenticate with a bogus token rather than the real credential.
- `ANTHROPIC_BASE_URL=https://gw.example.com` — a later client would target a non-existent gateway.

It presents as flaky rather than broken because it only bites when both tests land on the same xdist worker.

### Suggested fix

Replace the cleanup loop with `os.environ.pop(key, None)`, or register the undo before the production call by using `monkeypatch.setenv(key, <sentinel>)` first so a real undo entry exists. Adding an autouse fixture in that module that snapshots and restores `os.environ` would also close it for the whole file.

Happy to open a PR.

AmirF194:
`_ensure_shell_env()` merges allowed login-shell vars (proxy config, `GH_TOKEN`,
`*_API_KEY`/`*_BASE_URL`) straight into `os.environ`, bypassing monkeypatch's
tracking. `test_shell_env_whitelist_admits_proxy_and_gateway_vars` tried to clean
up after itself with `monkeypatch.delenv(key, raising=False)` once its assertions
ran, but `delenv` snapshots the *current* value as the state to restore on
teardown. Since the value it captured was the one the test itself had just leaked
into the real process environment, monkeypatch's teardown put it right back
afterwards. The apparent cleanup made the leak permanent instead of preventing it.

Same shape, three more places in this file: `test_shell_env_loader_preserves_mcp_stdin`
(`OUROBOROS_TEST_ENV`), `test_shell_env_merge_is_whitelisted` (`SOME_VENDOR_API_KEY`),
`test_shell_env_cache_written_with_whitelisted_subset` (`FOO_API_KEY`), and the two
cache tests (`CACHED_TEST_API_KEY`, `OPENAI_API_KEY`) — found while tracing the
reported one, since they all call the same `_ensure_shell_env` merge path.

Replaced the per-test `monkeypatch.delenv(...)` cleanup attempts with one autouse
fixture that snapshots `os.environ` before each test and restores it afterward,
independent of how a given test mutated it.

## Verification

- New test `test_environ_mutations_do_not_survive_past_the_test`: fails on
  unmodified `main` (`HTTPS_PROXY`/`NO_PROXY`/`ANTHROPIC_BASE_URL`/`GH_TOKEN`/
  `SOME_VENDOR_API_KEY` all present in `os.environ`), passes on this branch.
  Ran both ways in a clean `python:3.13-slim` container.
- Full `tests/unit/cli/` suite (1470 tests): same 22 pre-existing failures on
  both `main` and this branch (unrelated Codex-setup tests that need a real
  `uv`/executable on `PATH`, not present in this container); no new failures.
- `ruff check`, `ruff format --check`, and `mypy src/ouroboros` all clean.
- Not verified: behavior under `pytest-xdist` (`-n auto`); the fix is process-local
  environment restoration, which xdist workers already isolate by process, but I
  did not run the suite under `-n auto` here.

Closes #1793

### DIFF
--- a/tests/unit/cli/test_mcp_shell_env.py
+++ b/tests/unit/cli/test_mcp_shell_env.py
@@ -8,7 +8,18 @@ import os
 import subprocess
 import sys
 
+import pytest
+
 from ouroboros.cli.commands import mcp
+
+
+@pytest.fixture(autouse=True)
+def _restore_environ():
+    """Restore os.environ after every test; _ensure_shell_env writes it directly."""
+    saved = dict(os.environ)
+    yield
+    os.environ.clear()
+    os.environ.update(saved)
 
 
 def test_shell_env_loader_preserves_mcp_stdin(monkeypatch, tmp_path) -> None:
@@ -70,7 +81,6 @@ def test_shell_env_merge_is_whitelisted(monkeypatch, tmp_path) -> None:
     assert "/extra/bin" in os.environ["PATH"]
     assert "PYTHONPATH" not in os.environ
     assert "VIRTUAL_ENV" not in os.environ
-    monkeypatch.delenv("SOME_VENDOR_API_KEY", raising=False)
 
 
 def test_shell_env_cache_skips_login_shell(monkeypatch, tmp_path) -> None:
@@ -90,7 +100,6 @@ def test_shell_env_cache_skips_login_shell(monkeypatch, tmp_path) -> None:
     mcp._ensure_shell_env(timeout=1.25)
 
     assert os.environ["CACHED_TEST_API_KEY"] == "cached"
-    monkeypatch.delenv("CACHED_TEST_API_KEY", raising=False)
 
 
 def test_shell_env_cache_merges_other_provider_key_when_anthropic_is_present(
@@ -112,7 +121,6 @@ def test_shell_env_cache_merges_other_provider_key_when_anthropic_is_present(
     mcp._ensure_shell_env(timeout=1.25)
 
     assert os.environ["OPENAI_API_KEY"] == "openai-cached"
-    monkeypatch.delenv("OPENAI_API_KEY", raising=False)
 
 
 def test_shell_env_cache_written_with_whitelisted_subset(monkeypatch, tmp_path) -> None:
@@ -134,7 +142,6 @@ def test_shell_env_cache_written_with_whitelisted_subset(monkeypatch, tmp_path) 
     cached = json.loads(cache_file.read_text(encoding="utf-8"))
     assert "FOO_API_KEY" in cached
     assert "RANDOM_VAR" not in cached
-    monkeypatch.delenv("FOO_API_KEY", raising=False)
 
 
 def test_shell_env_whitelist_admits_proxy_and_gateway_vars(monkeypatch, tmp_path) -> None:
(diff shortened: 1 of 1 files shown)


shaun0927:
### Directionality and merge-readiness note

Point-in-time review at HEAD **85eea2d9feeb0cafd2a8ee034dc9cdb716434452** (2026-08-03 KST).

- **Directional fit:** Aligned. Isolating process-global environment mutations in tests supports reliable and reproducible verification.
- **Current recommendation:** Make one small correction before merge.

The autouse fixture that snapshots and restores os.environ is appropriately scoped and fixes the actual leak. The final regression test, however, asserts that HTTPS_PROXY, NO_PROXY, ANTHROPIC_BASE_URL, GH_TOKEN, and SOME_VENDOR_API_KEY are unconditionally absent. If any of those variables legitimately existed before pytest started, the fixture correctly restores them and the test then fails.

Please compare against a captured baseline, or use sentinel values that distinguish ambient state from values leaked by earlier tests. The current green CI run does not exercise legitimate proxy/gateway-configured environments.

AmirF194:
Fixed in e4f78918: the regression test now captures each variable's value at module import, before any fixture runs, and compares against that baseline instead of asserting absence. Verified in Docker with HTTPS_PROXY/NO_PROXY/GH_TOKEN set ambient before pytest: 7/7 pass, including the fixed test. Also ran under -n 4 --dist worksteal and ruff, both clean.

AmirF194 (commented):
Addressed the ambient-environment blocker, see comment above.

Q00 (approved):
Verified current head e4f789188886a69f24cf44c67b63b75d4bafff92. The latest commit resolves the prior stale review by restoring each environment key to its module-import baseline rather than assuming it was absent. Serial, hostile pre-existing environment, and xdist isolation paths are covered. The previously cancelled Python 3.13 job was rerun and passed; all required checks are now green.

Issue #1766, opened 2026-07-27
Lenient weight parsing: JSON -Infinity token saturates to 0.0 instead of the documented 1.0 default


Q00:
Follow-up from the final review round on #1762 (#1729 slice 2).

`_JsonNonFiniteToken` documents that non-standard JSON constants (NaN/Infinity/-Infinity) should fall back to the default weight (1.0) in lenient mode, unlike genuine numeric overflow which saturates by sign. In practice the lenient branch of `_clamp_weight` converts the token via `float(str(raw_weight))`, so a stored `-Infinity` literal resolves to `-inf` and saturates to 0.0 instead of defaulting to 1.0 — the token is not distinguished from real overflow on that path.

- No crash, strict mode is unaffected (tokens correctly trigger the extraction retry there).
- Only reachable via stored legacy requirements containing a literal `-Infinity` JSON constant, so severity is low.
- Fix: handle `_JsonNonFiniteToken` explicitly in the lenient branch before the string-conversion fallback, plus a regression test for the `-Infinity` token in lenient mode (NaN/Infinity are already covered).

Taking this as a maintainer follow-up so it doesn't block #1762.

Refs #1762, #1729.

sumin220:
This is a follow-up on the weight parsing I wrote for #1762, so I'd like to take it. The lenient branch needs to recognize the token marker before the float(str(...)) conversion so the documented 1.0 default applies to all three constants, while genuine numeric overflow keeps saturating by sign. PR shortly.

PR #1734, opened 2026-07-24
fix(parsing): tolerate markdown fences and prose in LLM JSON responses

MyoungSoo7:
## Problem

`WonderEngine`, `ReflectEngine` and `AssertionExtractor` each parse an LLM's JSON response with the same fragile heuristic:

```python
cleaned = content.strip()
if cleaned.startswith("```"):
    lines = cleaned.split("\n")
    cleaned = "\n".join(lines[1:-1])
data = json.loads(cleaned)
```

This only strips a fence when the **entire** response starts with one, and blindly drops the first and last lines. It therefore fails whenever a model:

- adds prose **before** the fence — e.g. Gemini-family models routinely emit `Here is the analysis:\n```json\n{...}\n````, or
- adds trailing text **after** the closing fence.

In both cases `json.loads` raises, and the failure is swallowed: Wonder silently falls back to its parse-error `WonderOutput`, while Reflect returns `None` and `AssertionExtractor` returns `()`. The evolutionary loop and spec verification quietly degrade instead of using the model's actual output.

This is ironic because the repo already ships a robust helper — `ouroboros.core.json_utils.extract_json_payload` — which strips fences *anywhere* and bracket-extracts the first valid JSON payload. It is already used by the semantic and consensus evaluators; these three call sites just never adopted it.

## Fix

Replace the ad-hoc fence stripping in all three parsers with `extract_json_payload`, matching the existing `evaluation/semantic.py` idiom.

## Reproduction

| completion shape | old ad-hoc | `extract_json_payload` |
| --- | --- | --- |
| `Here is the analysis:\n\`\`\`json\n{...}\n\`\`\`` | ❌ `JSONDecodeError` | ✅ |
| `\`\`\`json\n{...}\n\`\`\`\nLet me know...` | ❌ `JSONDecodeError` | ✅ |
| `\`\`\`json\n{...}\n\`\`\`` (bare fence) | ✅ | ✅ |
| `{...}` (no fence) | ✅ | ✅ |

## Tests

Adds `tests/unit/test_llm_response_fence_robustness.py` (parametrized over the four shapes above) for both `WonderEngine._parse_response` and `AssertionExtractor._parse_response`.

- Before the fix: the two prose-wrapped variants fail for each parser (4 failures).
- After the fix: all pass.
- Existing suites for the touched modules stay green: `tests/unit/evolution/`, `tests/unit/test_verification.py`, `tests/unit/evaluation/test_semantic.py` → 183 passed.
[...]

PR #1876, opened 2026-08-03
fix(bigbang): route interview JSON parsing through the hardened extractor

### LINKED ISSUE #1838: Interview parsers never adopted the hardened JSON extractor: an echoed schema example wins over the model's real answer
## Summary

The interview layer never adopted the hardened JSON extractor. `bigbang/ambiguity.py`
and `bigbang/pm_interview.py` still take the **first** fenced object out of a model
response, so when the model restates the requested schema before answering — routine
behaviour — the echoed example becomes the authoritative answer and the real one is
discarded.

`ouroboros.core.json_utils.extract_json_payload` treats two supported fences as
ambiguous and returns `None` (fail closed). Every provider adapter,
`verification/extractor.py`, `evolution/wonder.py` and `mcp/tools/qa.py` use it.
`grep -rn "extract_json_payload" src/ouroboros/bigbang/` returns nothing.

The five sites still on the pre-#1734 heuristic
(`re.search(r"```(?:json)?\s*(\{.*?\})\s*```", …)` with a greedy `\{.*\}` fallback):

| file:line                            | consumer                                                  |
| ------------------------------------ | --------------------------------------------------------- |
| `bigbang/ambiguity.py:703`           | `_parse_scoring_response` — the interview-completion gate |
| `bigbang/ambiguity.py:847`           | `_parse_dimension_response` — the per-dimension fan-out   |
| `bigbang/pm_interview.py:1328`       | `_parse_pm_seed` — the PMSeed document                    |
| `bigbang/question_classifier.py:317` | `_parse_response`                                         |
| `bigbang/interview.py:658`           | `_parse_question_candidate`                               |

## Reproduction

Model output — a restated schema, then the real assessment:

````
Understood. I will respond in exactly this shape:

```json
{"goal_clarity_score": 0.9, "goal_clarity_justification": "<why>", ...}
```

Now the actual assessment of these requirements:

```json
{"goal_clarity_score": 0.15, "goal_clarity_justification": "Goal is a one-liner with no deliverable.",
 "constraint_clarity_score": 0.1, ..., "success_criteria_clarity_score": 0.05, ...}
```
````

Observed against `main` @ 90ac429:

```
extract_json_payload(text)          -> None          # the hardened path, fail-closed
AmbiguityScorer._parse_scoring_response(text)
    goal_clarity.clarity_score      -> 0.9
    goal_clarity.justification      -> '<why>'       # a literal placeholder, stored as the reason
```

The model scored 0.15 / 0.1 / 0.05 — ambiguous, and not seed-ready. The system records
0.9 and a justification of `<why>`.

`pm_interview._parse_pm_seed` fails the same way and is more visible in the artefact:
the saved PMSeed carries `product_name='<product name>'`, `goal='<one sentence goal>'`,
zero user stories and no success criteria.

Both paths already have correct fail-closed behaviour they never reach — `score()`
retries on `ValueError` and then returns `Result.err`, and `generate_pm_seed()` returns
`Result.err`.

## Second defect in the same function: non-finite scores clamp _up_

`ambiguity.py:729`:

```python
def clamp_score(value: Any) -> float:
    score = float(value)
    return max(0.0, min(1.0, score))
```

`json.loads` accepts the non-standard `NaN` / `Infinity` literals by default, and
`min(1.0, nan)` returns `nan` because `nan < 1.0` is False, so `max(0.0, nan)` is `nan`
→ clamped to perfect clarity. Observed:

| model emits | clarity score |
| ----------- | ------------- |
| `NaN`       | **1.0**       |
| `-Infinity` | 0.0           |

Opposite outcomes for equally malformed input is the tell. This is the class merged
PRs #1766 / #1800 closed in `seed_generator.py`; `ambiguity.py` has no `NaN` test at
all (`grep -rn "NaN" tests/unit/bigbang/` hits only `test_seed_generator.py`).

## Suggested fix

Replace the ad-hoc regex block with `extract_json_payload`, raising `ValueError` when
it returns `None` so the existing retry-then-`Result.err` path takes over, and decode
with a `parse_constant` that rejects the non-finite literals.

Scope suggestion: the two authority-bearing files (`ambiguity.py`, `pm_interview.py`)
carry the completion gate and the saved artefact. `question_classifier.py` and
`interview.py` are the same one-line swap and can follow, so the change stays reviewable.

Existing coverage is unaffected: `test_parse_response_with_markdown_code_block` (prose
plus a single fence) still passes under `extract_json_payload`, and
`tests/unit/bigbang/test_ambiguity.py` is green on `main` (87 passed).

sumin220:
Closes #1838.

## Problem

The interview layer never adopted the hardened JSON extractor. All five consumers still ran the pre-#1734 first-fence heuristic, so when a model restated the requested schema before answering — routine behaviour — the echoed example became the authoritative payload: the interview-completion gate recorded placeholder clarity scores (`0.9` with a justification of literal `<why>`), and the saved PMSeed carried `product_name='<product name>'` with zero user stories, while the real assessment in the second fence was discarded. Both paths already had correct fail-closed behaviour they never reached.

The same parser also clamped non-finite scores in opposite directions: `NaN` clamped **up** to perfect clarity (because `min(1.0, nan)` is `nan` and `max(0.0, nan)` is `nan`), while `-Infinity` clamped to zero.

## Fix

- All five sites — `_parse_scoring_response`, `_parse_dimension_response`, `_parse_pm_seed`, `question_classifier._parse_response`, `interview._parse_question_candidate` — now extract through `ouroboros.core.json_utils.extract_json_payload`. An ambiguous response (two supported fences) yields `None`, which raises `ValueError` at the four raising sites so the existing retry-then-`Result.err` paths engage, and returns `None` at the candidate parser so a bad candidate drops out of the panel as designed.
- Ambiguity scoring rejects non-finite values at two layers: `json.loads(..., parse_constant=...)` refuses the `NaN`/`Infinity`/`-Infinity` literals at decode time, and both clamp paths verify `math.isfinite` after conversion so a string-smuggled `"nan"` cannot slip through either. This is the same defect class #1766/#1800 closed in `seed_generator.py`.
- The scoring-parser error message keeps its existing `Invalid JSON response` prefix, so the established message contract (and the test pinning it) is unchanged.

## Tests

New regression module `tests/unit/bigbang/test_interview_json_extraction.py` (12 tests), all verified failing on the baseline except the positive-path guard:

- The issue's echoed-schema reproduction against each of the five consumers — the four raising parsers must raise rather than adopt the echo, and the candidate parser must return `None`.
- A prose-plus-single-fence response still parses (the shape the previous heuristic handled), guarding against over-tightening.
- Non-finite literals (`NaN`, `Infinity`, `-Infinity`) rejected in full scoring and per-dimension scoring, plus the string-smuggled `"nan"` case.

Full Big Bang, auto, and MCP-tools suites pass (3,396 tests); ruff, mypy, and the module-size gate are clean.

### DIFF
--- a/src/ouroboros/bigbang/ambiguity.py
+++ b/src/ouroboros/bigbang/ambiguity.py
@@ -712,6 +712,8 @@ Additional context (intentional deferrals — do not penalise):
             data = json.loads(text, parse_constant=_reject_non_finite_score)
         except json.JSONDecodeError as e:
             raise ValueError(f"Invalid JSON response: {e}") from e
+        if not isinstance(data, dict):
+            raise ValueError("Invalid JSON response: payload must be an object")
 
         # Numeric score fields must be present. Missing justifications are recoverable.
         required_score_fields = [
@@ -852,6 +854,8 @@ Additional context (intentional deferrals — do not penalise):
             data = json.loads(text, parse_constant=_reject_non_finite_score)
         except json.JSONDecodeError as exc:
             raise ValueError(f"Invalid JSON response: {exc}") from exc
+        if not isinstance(data, dict):
+            raise ValueError("Invalid JSON response: payload must be an object")
 
         if "clarity_score" not in data:
             raise ValueError("Missing required field: clarity_score")
--- a/src/ouroboros/bigbang/ambiguity.py
+++ b/src/ouroboros/bigbang/ambiguity.py
@@ -13,7 +13,7 @@ import asyncio
 from dataclasses import dataclass, field
 from enum import StrEnum
 import json
-import re
+import math
 from typing import Any
 
 from pydantic import BaseModel, Field
@@ -27,6 +27,7 @@ from ouroboros.bigbang.interview import (
 )
 from ouroboros.config import get_llm_model_for_role
 from ouroboros.core.errors import ProviderError
+from ouroboros.core.json_utils import extract_json_payload
 from ouroboros.core.types import Result
 from ouroboros.providers.base import CompletionConfig, LLMAdapter, Message, MessageRole
 
@@ -330,6 +331,11 @@ def qualifies_for_seed_completion(
         score,
         is_brownfield=is_brownfield,
     )
+
+
+def _reject_non_finite_score(token: str) -> float:
+    """json.loads hook: NaN/Infinity literals are malformed scores, not values."""
+    raise ValueError(f"non-finite score literal: {token}")
 
 
 @dataclass
@@ -696,21 +702,14 @@ Additional context (intentional deferrals — do not penalise):
         Raises:
             ValueError: If response cannot be parsed.
         """
-        # Extract JSON from response (handle markdown code blocks)
-        text = response.strip()
-
-        # Try to find JSON in markdown code block
-        json_match = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
-        if json_match:
-            text = json_match.group(1)
-        else:
-            # Try to find raw JSON object
-            json_match = re.search(r"\{.*\}", text, re.DOTALL)
-            if json_match:
-                text = json_match.group(0)
+        # One authoritative payload or nothing: an echoed schema example
+        # must never outrank the real answer (#1838).
+        text = extract_json_payload(response.strip())
+        if text is None:
+            raise ValueError("Invalid JSON response: no unambiguous JSON payload")
 
         try:
-            data = json.loads(text)
+            data = json.loads(text, parse_constant=_reject_non_finite_score)
         except json.JSONDecodeError as e:
             raise ValueError(f"Invalid JSON response: {e}") from e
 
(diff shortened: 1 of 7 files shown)


sumin220:
Addressed the first review in cbe6b99f, both blockers.

All four raising consumers now validate that the decoded payload is an object before any field access — a fenced top-level array raises ValueError inside the documented boundaries instead of letting AttributeError escape them. That covers the two sites you flagged and the two ambiguity parsers, which had the same latent shape hole reachable through an array whose elements happen to contain the required field names; the candidate parser keeps its existing isinstance guard.

Regressions are at the public boundaries you asked for: classify over a fenced empty array must return the safe passthrough fallback, and generate_pm_seed over the same input must return Result.err rather than aborting — both verified failing before the guard. Unit cases cover the array shape against both ambiguity parsers as well.

Big Bang, auto, and MCP-tools suites pass (3,400); ruff, mypy, and the module-size gate are clean.

PR #1891, opened 2026-08-04
fix(evolution): make evolve_step generation ownership a durable lease

### LINKED ISSUE #1889: [Bug] Concurrent evolve_step calls can execute and record one generation twice
## Problem

Two concurrent `evolve_step` calls can replay the same lineage state, both select the same generation number, execute external work twice, and append duplicate completion and terminal events.

This was surfaced during the exact-head adversarial review of #1846. The ownership path at `EvolutionaryLoop.evolve_step` is unchanged from `main`, so it is tracked separately rather than expanding #1846's focused-convergence slice.

## Reproduction evidence

A deterministic barrier probe ran two calls against one SQLite-backed `EventStore` and one lineage:

```text
both_ok= [True, True]
created_count= 1
completed_count= 2
generation_numbers= [1, 1]
terminal= ['lineage.converged', 'lineage.converged']
```

The race is between replay/generation selection and durable generation ownership:

- `src/ouroboros/evolution/loop.py`: replay and generation selection in `evolve_step`
- `src/ouroboros/evolution/loop.py`: completion and terminal event append after execution

## Required behavior

- Exactly one caller owns a lineage generation at a time.
- A losing concurrent caller must fail closed before invoking the external executor.
- Ownership must work across separate loop instances/processes sharing the same durable EventStore, not only through an in-process lock.
- Crash recovery must permit a later caller to resume or reclaim according to an explicit lease/CAS contract.
- Replay must not project duplicate completion or terminal events for one generation.

## Acceptance criteria

- [ ] Deterministic two-caller barrier test proves one executor invocation and one completion/terminal transition.
- [ ] Ownership is an atomic durable claim, lease, unique transition, or version CAS at the EventStore boundary.
- [ ] Loser behavior is explicit and tested.
- [ ] Crash/reclaim behavior is explicit and tested.
- [ ] Existing interruption/resume and cross-session `evolve_step` tests remain green.

sumin220:
Closes #1889.

## Problem

Two concurrent `evolve_step` calls could replay the same lineage state, both select the same generation number, run the external executor twice, and append duplicate completion and terminal events — the deterministic barrier probe in the issue showed both callers succeeding with `generation_numbers=[1, 1]` and two terminal transitions.

## Fix

Ownership of one `(lineage_id, generation_number)` attempt is now a durable lease decided at the database boundary, in `evolution/generation_claims.py`:

- **Atomic claim.** The winner inserts the lease inside a single write transaction (`BEGIN IMMEDIATE` on SQLite; the composite primary key closes the same absent-row race on every backend). The losing concurrent caller fails closed with an explicit ownership error before the executor is invoked.
- **Heartbeat and release.** The owner refreshes the lease on a heartbeat while the generation runs, and the claim is always released on exit — completion, failure, and interruption alike — so a resuming call takes ownership immediately with no waiting.
- **Crash reclaim.** A lease that stops refreshing for its duration (default 300s) is presumed crashed and may be reclaimed. The steal is a compare-and-set on the observed token, so two concurrent reclaimers cannot both win; a live owner whose refresh subsequently fails observes the loss, logs it, and stops claiming ownership.
- **Store fallback.** Stores that expose no real database URL — unit-test fakes, and in-memory databases that are private to one process by construction — use a per-process claim table with identical semantics, so cross-process guarantees are never silently pretended where they cannot hold.

The claim wraps exactly the external-work boundary in `evolve_step` (the `AgentProcess` spawn through result collection). To keep `loop.py` inside its ratcheted module-size budget, the seed-reconstruction ladder moves unchanged into `evolution/step_seed.py`; `loop.py` ends 71 lines under its budget.

## Tests

- `test_concurrent_evolve_step_executes_exactly_one_generation` — the issue's barrier reproduction against a real SQLite store and two loop instances: exactly one winner, exactly one executor invocation, at most one completion event, and a loser error that states the ownership conflict. Verified failing on the baseline with two winners.
- `test_sequential_evolve_steps_reacquire_released_claims` — completion releases the claim and the next call proceeds.
- Lease-contract unit tests: a fresh claim cannot be stolen; an expired claim is reclaimable and the presumed-crashed owner observes the loss; heartbeats keep a lease from expiring; release hands ownership over; a foreign-token release is a no-op; concurrent reclaim of one expired lease has exactly one winner.

Evolution suite (157), persistence suite, and the evolve/lineage/ralph-related MCP and integration tests pass; ruff, mypy, and the module-size gate are clean.


sumin220:
Addressed the first review in 919b7154, taking the remedy from blocker four literally: ownership now serializes the lineage-level replay/selection boundary, and the other three findings fall out of that move plus a fencing rework.

The lease is per lineage and is acquired before the caller replays any state. That closes all three state-consistency blockers at once: a released lease is now an invitation to replay fresh state, never a reusable generation slot, so a delayed contender advances the lineage instead of re-running the completed generation (regression: the second caller must execute generation 2, not 1); gen-1 lineage.created is written inside the lease, so a losing caller mutates nothing durable (regression asserts exactly one creation event after a contended start); and a caller arriving mid-generation is denied at the lineage boundary rather than selecting last_gen + 1 off a live phase (the concurrent regression holds the winner mid-executor while the loser is denied).

Lease loss now fences the owner rather than just logging: owned_lineage_step yields a lease handle whose loss event fires when a refresh reports the claim was reclaimed, and evolve_step runs a fence task that aborts the in-flight AgentProcess work at that moment. A transient refresh exception is retried on the next beat instead of being treated as a loss, and the heartbeat's outcome — return, loss, or exception — can no longer bypass the release, which sits behind the await in its own guard (regressions: an outage-then-steal scenario must end with the fenced owner reporting an error, and a raising refresh must still release on exit).

Per the follow-up, fallback claims are namespaced per store object (WeakKeyDictionary), so unrelated fakes or private in-memory stores sharing a lineage ID cannot deny each other, with a regression for both identity directions.

Evolution suite (161), persistence, and the evolve/lineage/ralph-related MCP and integration tests pass; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the second review in 2cf57d02, all three blockers.

1. Fencing no longer depends on the stale owner happening to observe the replacement token. The lease discipline is now two-sided: an owner whose refreshes cannot be confirmed fences itself once half the lease has elapsed since its last confirmed refresh — strictly before a reclaimer is permitted to steal at the full lease — and the default heartbeat tightens to lease/6 so the check cannot overshoot the steal threshold. The stale attempt has therefore always stopped before a successor can begin, rather than up to a heartbeat later. The regression is deterministic and involves no manual steal at all: under a refresh outage the blocked owner must fail on its own, and only then does the successor acquire and complete.

2. Lease-loss cancellation is now a distinct fenced path that performs no lineage writes. The fence marks the lineage before aborting, and the cancellation handler in _run_generation observes the mark and appends nothing — the successor owns the generation's terminal record. The regression stubs one layer beneath the watchdog so the real cancellation handler remains on the aborted path, and asserts the stale owner leaves zero lineage.generation.failed events behind.

3. Backend classification now parses the SQLAlchemy URL instead of substring-matching: every SQLite in-memory form — pathless and explicit :memory: alike — routes to the per-process claims. Your exact reproduction is the regression: EventStore("sqlite+aiosqlite://") resolves to local claims and evolve_step completes on it.

Evolution and persistence suites pass (415), plus the ownership module's fifteen tests; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the third review in cf21a357, both blockers.

1. The post-executor persistence boundary is token-fenced. Every lineage transition after the executor returns — completed, ontology evolved, exhausted, stagnated, converged — now goes through a conditional append whose token check shares one database transaction with the event insert on the durable backend. The single-writer database serializes that transaction against any steal, which gives the exact property the finding asked for: a completion that commits necessarily precedes the successor's post-steal replay (the steal's CAS reads the claim row after the commit), and a steal that serializes first refuses the stale write, which surfaces as an explicit lease-lost error instead of a journaled generation. The guarded write still runs to settlement inside the store's registry, so close() drains it and cancellation cannot abandon it mid-transaction. The regression is deterministic: ownership transfers while the executor result is pending, and the completion append must be refused with zero completed events persisted.

2. Lease expiry is decided in database time. Refresh stamps and the expiry/CAS comparison both read the clock through the database connection — the one time authority every contender already shares — so a skewed or forward-stepped client clock can no longer steal a fresh lease from a live owner before its monotonic half-lease fence fires. Your probe shape is the regression: with the contender's wall clock stepped 400 seconds forward against a 300-second lease, the steal must be denied. The owner-side self-fence stays on the monotonic clock deliberately, since it measures the owner's own elapsed time and must be immune to wall-clock steps in the other direction.

To keep loop.py inside its module-size budget the generation record and completion event builders moved unchanged into step_seed.py. Evolution and persistence suites pass (417) with the ownership module at seventeen tests; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the fourth review in 5f56cabf, all three blockers.

1. Reclaim is now two-phase, which gives the provable fencing acknowledgement the finding asked for. A reclaimer first marks the expired lease revoked and may take ownership only after a grace period of one heartbeat interval; a revoked lease refuses the owner's refresh, so a scheduling owner observes the transfer on its next beat and is fenced and aborted inside the grace window — strictly before the takeover CAS can succeed. The overlap your probe produced (successor completing during the stale owner's heartbeat gap) is now structurally impossible for a scheduling owner: the regression revokes a blocked owner's lease and requires the owner to have failed before the takeover is even permitted. Both phases are CAS on observed state, so concurrent reclaimers still resolve to one winner; a reclaimer that itself goes silent for a full lease can be superseded. For an owner whose process cannot schedule at all, the journal remains protected by the token-fenced appends and the grace bounds the window to one heartbeat; a truly stopped-world process with an independent external effector is beyond any lock service without resource-side fencing, and the module documents that boundary.

2. Fence identity is attempt-scoped. The lease handle travels down to the cancellation handler, and the lineage-wide mutable set is gone entirely, so there is no state a denied caller's finally can clear. Your add/discard interleaving is the regression: a second caller is denied and completes while the fenced owner's cancellation is deliberately held open, and the stale attempt must still append zero lineage events.

3. Crash reclaim resumes abandoned work. Any nonterminal last phase — EXECUTING included, not only FAILED and INTERRUPTED — retries its generation rather than advancing past it, restarting from the last completed generation's seed; a lineage with nothing completed asks for an explicit seed with the existing clear error rather than guessing. Your probe is the regression: an expired Gen-1 EXECUTING journal with an explicit seed must execute generation 1 again, not generation 2.

Evolution and persistence suites pass (420) with the ownership module at twenty tests; the evolve/lineage/ralph MCP and integration tests pass; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the fifth review in 486e62ce, all three blockers.

1. Phase two of a reclaim is open to any caller. The revocation carries fencing intent, not reclaimer identity, so the realistic retry path — a fresh token on every evolve_step invocation — completes a reclaim a previous call started. The takeover CAS matches the full observed row state, so concurrent reclaimers still resolve to exactly one winner. Your probe is the regression: token A marks the revocation and a later token-B retry must win the takeover after the grace.

2. Every heartbeat refresh attempt is bounded by the self-fence deadline. A refresh that hangs is cancelled at the moment the fence must fire, so the owner is fenced at half the lease even when the claims backend stops answering entirely, and a reclaimer can no longer take over while a running owner remains unaware. The regression drives owned_lineage_step with a never-returning refresh and requires the lost event by the deadline with the release still performed.

3. The conductor-preservation failure append goes through the same token-fenced path as every other post-executor lineage transition — that was the one remaining unconditional lineage write in the post-executor section. The regression transfers ownership while the executor result is pending, forces the preservation branch, and requires zero failed events with the stale owner reporting an error.

Evolution and persistence suites pass (423) with the ownership module at twenty-three tests; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the sixth review in c0a1da57, both blockers.

1. The creation write is fenced and the lost-lease check spans the entire owned section. The gen-1 lineage.created append now goes through the same token-conditional path as every other lineage transition, so a caller whose lease was reclaimed during a delayed replay is refused at the write and returns the lease-lost error instead of seeding durable state; and a pre-spawn lease.lost check means external work can never start after authority has moved, whichever pre-spawn step was slow.

2. The local fallback's guarded append is atomic. Ownership validation and the store append now share one critical section under the claims lock, so an expiry-and-reclaim serializes behind an in-flight owner append — the same commit-before-steal ordering the durable backend gets from its single transaction — rather than interleaving between the check and the write. The dispatch prefers this path for every LocalStepClaims (subclasses included), and the documented identical-semantics claim is true again.

To stay inside loop.py's size budget the projection/terminated/resume-point block moved unchanged into step_seed.project_step_state. Evolution and persistence suites pass (423) with the ownership module at twenty-three tests; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the seventh review in 53766266, all three blockers.

1. The attempt's lost state now gates every persistence and success path. append_lineage_event_if_owner takes the lease handle itself and refuses before touching the store once lost has fired — self-fencing is process-local, so the stored token being technically valid no longer matters. Your probe is the regression: a cancellation-resistant executor swallows the fence's abort and returns a completed result, and the attempt must surface an error with zero completion events persisted.

2. Takeover requires an acknowledged stop. After the fence fires, the heartbeat switches to a fenced-lingering liveness beat that keeps the claim row's refresh stamp current without touching the revocation marker. Phase two requires an expired lease, so a stale attempt that is provably alive blocks the reclaim outright; the row's deletion on exit is the termination acknowledgement that lets the successor proceed, and a crashed process stops beating and is reclaimed through the unchanged expiry path. The regression holds a resistant executor alive past the abort, proves a lingering beat refreshed the row, ages the revocation past the grace, and requires the takeover to stay denied until the stale attempt exits and releases.

3. The two-phase regressions are deterministic. Wall-clock sleeps are gone: the lease is too large for scheduler delay to cross, and both transitions are driven by explicit row aging — one assertion proves no takeover before the grace, a separate one proves exactly one winner after it. The suite holds under repetition.

Evolution and persistence suites pass (425) with the ownership module at twenty-five tests; ruff, mypy, and the module-size gate are clean.

sumin220:
Addressed the eighth review in 4a60908f, all three blockers.

1. The lost state is coupled to the commit. The guarded transaction consults the attempt's lost state immediately before committing, not only on entry, so a self-fence that fires after the append was admitted rolls the write back. The regression drives the inner transaction directly with a lost lease over a durably valid token and requires refusal with nothing persisted.

2. The ownership read takes a row lock. On lock-based backends the claim row is read FOR UPDATE inside the guarded transaction, so a reclaimer's token update serializes against the stale insert instead of slipping between the read and the write; SQLite's immediate write transaction already closed that window, and a compile-level regression pins the FOR UPDATE clause on the PostgreSQL dialect.

3. The fence now reaches the effectful boundary. The attempt's lease is published to its task tree via current_step_fence(), giving executors the fencing token your finding asked for — the one channel that still works when the owner's database path has failed entirely. The ordering makes it provable: the owner self-fences at half the lease on unconfirmable refreshes, while a reclaimer needs the full lease plus a grace, so fencing-aware work always stops strictly before a successor can begin. The regression partitions every database interaction after acquisition and shows the fencing-aware executor stopping on the self-fence with the reclaimer still denied, then acquiring normally once the owner exits. For external work that ignores both cancellation and the fence on a fully stopped world, the module documents that only resource-side fencing can help — no lock service can reach it.

Evolution and persistence suites pass (428) with the ownership module at twenty-eight tests, held under repetition; ruff, mypy, and the module-size gate are clean.

Q00:
## Latest-main reconciliation and exact-head evidence

Authoritative HEAD: `8a9fed3c0da866b621d8e5d12f422a76ff2908e3`
Ancestry: formatting-only `8a9fed3c` → merge `19680e10` with parents prior approved HEAD `4a60908f0e2945b14d65c242c1478c55a421cba8` + current main `efa8dbcd4cd2cb8f426dff0eecc4ef4494414650`.

A dedicated worker and independent verifier re-read all review rounds and #1889. Every blocker remains mapped to current exact-head behavior:

- lease acquired before replay/selection and guarded lineage creation;
- atomic database claim, database-time expiry, two-phase CAS reclaim, one winning reclaimer;
- bounded heartbeat, half-lease self-fence, token-checked release and acknowledged-stop liveness;
- attempt-scoped lost state, task-tree fencing, cancellation-resistant stale-owner rejection;
- conditional event commits with pre-commit lost recheck, SQLite serialization, `FOR UPDATE` on locking backends, atomic local fallback;
- abandoned nonterminal generation recovery without skipping;
- current main single-flight/durable receipt/hidden-checklist/hard-crash policy preserved around the lease.

Fresh exact-head verification:
- ownership suite: `28 passed`;
- actual real-SQLite two-caller single-executor regression: `30/30` repeated passes;
- evolution + persistence: `730 passed, 71 skipped`;
- evolve/convergence/checklist: `168 passed`;
- durable single-flight receipts: `14 passed`;
- adjacent evolve/MCP/Ralph: `132 passed`;
- Ruff, full format, MyPy (`528` source files), module-size: pass.

Broader optional MCP adapter selection had only documented environment import failures because the optional `mcp` package is absent; relevant MCP unit coverage passed. #1919 paths are byte-identical to current main.

@ouroboros-agent Please formally review PR #1891 against current `main` at exact HEAD `8a9fed3c0da866b621d8e5d12f422a76ff2908e3`. This supersedes prior-head evidence. Approval must record `commit_id=8a9fed3c0da866b621d8e5d12f422a76ff2908e3`.

Q00:
## CI determinism remediation — new authoritative HEAD

Exact HEAD: `7e19f36997139ff26074adbc7d569a10cc63525c`.

The prior `8a9fed3c` Python matrix exposed the exact timing blocker called out in earlier review rounds: short live leases could self-fence before executor entry under full-suite xdist load, and a first barrier fix could block pre-executor guarded appends.

The new regressions are event-ordered instead of scheduler-ordered:
- refresh remains healthy until the executor `entered` barrier, then explicit loss fences cancellation-resistant work;
- the outage begins only after executor entry, then still requires self-fencing before successor execution;
- revocation uses a long-live lease, explicit DB row aging, fast heartbeat, and release as durable stop acknowledgement, with no grace sleep.

Independent verifier evidence at this exact head:
- the prior failing Python 3.14 xdist shape passed `5/5` rounds (`2,750 passed, 355 skipped` aggregate);
- ownership `28 passed`;
- full Ruff/format, MyPy (`528` source files), module-size: pass;
- #1919 paths unchanged.

@ouroboros-agent Please formally re-review PR #1891 against current `main` at exact HEAD `7e19f36997139ff26074adbc7d569a10cc63525c`. This supersedes all prior-head evidence. Approval must record `commit_id=7e19f36997139ff26074adbc7d569a10cc63525c`.

Q00:
@ouroboros-agent CI is now fully green at authoritative exact HEAD `7e19f36997139ff26074adbc7d569a10cc63525c`, including Python 3.12, 3.13, and the previously failing 3.14 full suite. Please complete the requested formal review against current `main`; approval must record `commit_id=7e19f36997139ff26074adbc7d569a10cc63525c`.

Q00:
@ouroboros-agent please re-review this exact HEAD. Authoritative head is `7e19f36997139ff26074adbc7d569a10cc63525c` against current `main`; all exact-head CI checks and independent verification are green, including the repaired Python 3.14 xdist ownership regressions. Approval must record this commit ID.

PR #1895, opened 2026-08-04
fix(providers): stop reporting a degraded ourocode turn as a clean one

### LINKED ISSUE #1894: ourocode adapter returns an empty or truncated turn as a clean completion
## Summary

`OurocodeLLMAdapter` returns a degraded ACP turn as a clean completion, in two independent ways. Both are on the plain (non-`response_format`) path, which is the one the orchestrator uses.

### 1. A turn that produced no text returns `Result.ok`

`src/ouroboros/providers/ourocode_llm_adapter.py:201-204` on `000f9580`:

```python
if not config.response_format:
    return Result.ok(
        await self._run(client, prompt_text, config, response_model=response_model)
    )
```

Nothing inspects the text. A turn that produced none comes back as a success carrying `content=''`, and that includes `stopReason: "refusal"` — a refusal arrives indistinguishable from an ordinary answer that happened to be empty.

This adapter is alone in this. `gjc_llm_adapter.py:447`, `codex_cli_adapter.py:1023`, `gemini_cli_adapter.py:469`, and `claude_code_adapter.py:1252` all fail closed on empty content.

### 2. Truncation never surfaces as `length`

`src/ouroboros/providers/ourocode_llm_adapter.py:322-323`:

```python
return {"end_turn": "stop", "cancelled": "cancelled"}.get(stop_reason, stop_reason)
```

ACP's `max_tokens` passes through verbatim. `finish_reason == "length"` is what this codebase tests for — `bigbang/ambiguity.py:527` and `:921` both key on it to detect a truncated response — so under ourocode that check never fires and a cut-off answer is consumed as a complete one.

## Reproduction

Through the public `complete()` with `OurocodeAcpClient.run_turn` patched to return the given `AcpTurnResult`:

| `text` | `stopReason` | On `000f9580` | Expected |
|---|---|---|---|
| `""` | `end_turn` | `Result.ok`, `content=''` | `Result.err` |
| `""` | `refusal` | `Result.ok`, `content=''` | `Result.err` |
| `""` | `max_tokens` | `Result.ok`, `content=''` | `Result.err` |
| `"  \n"` | `end_turn` | `Result.ok`, `content='  \n'` | `Result.err` |
| `"hi"` | `max_tokens` | `Result.ok`, `finish_reason='max_tokens'` | `finish_reason='length'` |
| *control* `"hi"` | `end_turn` | `Result.ok`, `finish_reason='stop'` | unchanged |

## Impact

`complete()` is where the orchestrator decides whether a provider call succeeded. An `ok` with empty content is consumed as a real completion instead of a retry or a failed step, and downstream parsers see `""` rather than an error — so the failure resurfaces later as a parse error attributed to the wrong cause. The truncation hole is quieter but worse-shaped: the guard that exists specifically to catch a cut-off response is simply never reached for this provider.

## Proposed fix

Reject an empty turn in `_run_result` before it becomes `Result.ok`, matching the sibling adapters. Translate `max_tokens` (and `max_turn_requests`, which `claude_code_adapter.py:1104` already treats as `length` for the turn-budget case) to `length`, leaving unrecognized reasons passing through rather than recoded as a clean stop.

MyoungSoo7:
Fixes #1894.

## The defect

Two independent ways `OurocodeLLMAdapter` made a degraded turn indistinguishable from a good one, both on the plain (non-`response_format`) path:

- **An empty turn returned `Result.ok`.** `_run_result` handed back whatever `_run` produced without inspecting it, so a turn with no text became a success carrying `content=''` — including on `stopReason: "refusal"`, which arrived looking like an ordinary answer that happened to be empty. Every sibling adapter fails closed here: `gjc_llm_adapter.py:447`, `codex_cli_adapter.py:1023`, `gemini_cli_adapter.py:469`, `claude_code_adapter.py:1252`.
- **Truncation never surfaced as `length`.** `_finish_reason` translated only `end_turn` and `cancelled`, so ACP's `max_tokens` reached callers verbatim. `finish_reason == "length"` is the marker this codebase tests for — `bigbang/ambiguity.py:527` and `:921` both key on it — so under ourocode that guard was never reached and a cut-off answer was consumed as a complete one.

## The change

Reject an empty turn in `_run_result` before it becomes `Result.ok`, carrying the stop reason in `details` so the failure is attributable. Translate `max_tokens` and `max_turn_requests` to `length`; `claude_code_adapter.py:1104` already treats the turn-budget case as `length`, so this follows the established reading rather than inventing one.

Unrecognized stop reasons still pass through unchanged. Recoding them to `stop` would hide a new ACP value behind a clean-completion verdict, which is the same failure shape being fixed here.

Emptiness is tested with `.strip()`, so a turn of pure whitespace counts as no answer.

Scope note: the structured `response_format` path is untouched — it already fails closed, because an empty response yields no extractable JSON and falls through to its existing non-conforming-output error.

## Verification

Through the public `complete()` with `OurocodeAcpClient.run_turn` patched:

| `text` | `stopReason` | Before | After |
|---|---|---|---|
| `""` | `end_turn` | `ok`, `content=''` | `err` |
| `""` | `refusal` | `ok`, `content=''` | `err` |
| `""` | `max_tokens` | `ok`, `content=''` | `err` |
| `"  \n\t"` | `end_turn` | `ok`, `content='  \n\t'` | `err` |
| `"hi"` | `max_tokens` | `ok`, `finish_reason='max_tokens'` | `ok`, `finish_reason='length'` |
| `"hi"` | `max_turn_requests` | `ok`, `finish_reason='max_turn_requests'` | `ok`, `finish_reason='length'` |
| *control* `"hi"` | `end_turn` | `ok`, `finish_reason='stop'` | unchanged |
| *control* `"hi"` | `cancelled` / `refusal` | `ok`, passthrough | unchanged |

Both are parametrized regression tests on the public path: `test_complete_rejects_a_turn_that_produced_no_text` and `test_complete_reports_truncation_as_length`. The second covers the passthrough reasons too, so a future translation cannot quietly become `stop`.

## Gates

- `tests/unit/providers` — 631 passed, 5 skipped
- `ruff check` / `ruff format --check` / `mypy` — clean
- `scripts/check-module-size.py --baseline-ref upstream/main` — OK (493 modules, cap 2000)
- `scripts/check-auto-boundary.py` — OK (39 files scanned, 0 findings); no `auto/` file is touched

Rebased on `000f9580`.

### DIFF
--- a/src/ouroboros/providers/ourocode_llm_adapter.py
+++ b/src/ouroboros/providers/ourocode_llm_adapter.py
@@ -199,9 +199,20 @@ class OurocodeLLMAdapter:
         response_model: str,
     ) -> Result[CompletionResponse, ProviderError]:
         if not config.response_format:
-            return Result.ok(
-                await self._run(client, prompt_text, config, response_model=response_model)
-            )
+            parsed = await self._run(client, prompt_text, config, response_model=response_model)
+            if not parsed.content.strip():
+                # A turn that produced no text is a failed turn, whatever it
+                # stopped for. Returning it as `ok` hands the caller an empty
+                # string where an answer belongs, and every sibling adapter
+                # fails closed here instead.
+                return Result.err(
+                    ProviderError(
+                        message="Empty response from ourocode",
+                        provider="ourocode",
+                        details={"stop_reason": parsed.finish_reason},
+                    )
+                )
+            return Result.ok(parsed)
 
         last_response_preview = ""
         for _attempt in range(_STRUCTURED_RESPONSE_ATTEMPTS):
@@ -320,7 +331,19 @@ class OurocodeLLMAdapter:
 
     @staticmethod
     def _finish_reason(stop_reason: str) -> str:
-        return {"end_turn": "stop", "cancelled": "cancelled"}.get(stop_reason, stop_reason)
+        """Translate an ACP ``stopReason`` into this codebase's ``finish_reason``.
+
+        ``max_tokens`` has to become ``"length"``: that is the marker callers
+        test for, and an untranslated one reads as an unremarkable stop rather
+        than a truncation. Unknown reasons pass through unchanged so a new ACP
+        value is visible rather than silently recoded as a clean stop.
+        """
+        return {
+            "end_turn": "stop",
+            "cancelled": "cancelled",
+            "max_tokens": "length",
+            "max_turn_requests": "length",
+        }.get(stop_reason, stop_reason)
 
     @staticmethod
     def _compose_prompt(messages: list[Message]) -> str:
--- a/src/ouroboros/providers/ourocode_llm_adapter.py
+++ b/src/ouroboros/providers/ourocode_llm_adapter.py
@@ -209,7 +209,7 @@ class OurocodeLLMAdapter:
                     ProviderError(
                         message="Empty response from ourocode",
                         provider="ourocode",
-                        details={"stop_reason": parsed.finish_reason},
+                        details={"stop_reason": parsed.raw_response["stop_reason"]},
                     )
                 )
             return Result.ok(parsed)
@@ -259,7 +259,10 @@ class OurocodeLLMAdapter:
             model=response_model,
             usage=_ZERO_USAGE,
             finish_reason=self._finish_reason(result.stop_reason),
-            raw_response={"ourocode_model": response_model},
+            raw_response={
+                "ourocode_model": response_model,
+                "stop_reason": result.stop_reason,
+            },
         )
 
     def _resolve_ourocode_model(
--- a/tests/unit/providers/test_ourocode_llm_adapter.py
+++ b/tests/unit/providers/test_ourocode_llm_adapter.py
@@ -33,6 +33,7 @@ async def test_complete_returns_completion_response() -> None:
     assert response.content == "hi there"
     assert response.model == "claude"
     assert response.raw_response["ourocode_model"] == "claude"
+    assert response.raw_response["stop_reason"] == "end_turn"
     # ACP carries no token usage — honestly zero, not fabricated.
     assert response.usage.total_tokens == 0
     assert response.finish_reason == "stop"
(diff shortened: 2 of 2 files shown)


Q00 (changes requested):
Blocking correctness issue found during merge verification: the adapter stores the normalized finish category in `details["stop_reason"]` instead of preserving the raw ACP `stopReason`. This collapses distinct causes such as `max_tokens` and `max_turn_requests` into `length`, so replay and diagnosis lose the exact terminal cause promised by the PR. Please preserve the raw stop reason in details and add an exact-value regression while retaining the normalized public finish reason.

Q00:
Implemented the requested raw-vs-normalized boundary on exact head `29ca54acbf5da8786a23d725a7374c9ca3dec777` (non-force push).

Finding mapping:

- Reproduction on prior head `cb49305ec6fe482f15efc9c9b659fc04fb9ff884`: empty turns with raw `max_tokens` and `max_turn_requests` both exposed `details["stop_reason"] == "length"`.
- Fix: keep public `CompletionResponse.finish_reason` normalized (`max_tokens` / `max_turn_requests` → `length`) while preserving the verbatim ACP `stopReason` in `raw_response["stop_reason"]`; empty-turn diagnostics copy that raw value into `ProviderError.details["stop_reason"]`.
- Regression: exact-value assertions now cover empty/whitespace fail-closed behavior, raw `max_tokens` vs `max_turn_requests` diagnostics, and the normalized-public/raw-provider pair for recognized and passthrough reasons.
- Local verification on the new head: adapter tests **23 passed**; provider suite **696 passed, 5 skipped**; Ruff check/format clean; targeted mypy clean; auto-boundary clean; module-size clean against the PR base.

An independent verifier is checking this exact SHA. CI and formal approval are still pending; no merge action was taken.

Q00:
@ouroboros-agent please re-review exact HEAD `29ca54acbf5da8786a23d725a7374c9ca3dec777` and supersede the prior-head review. Independent verification on this exact SHA confirms the raw ACP `stopReason` is preserved verbatim in diagnostics/raw response while public `finish_reason` remains normalized, with exact-value regressions and the provider suite passing. Please evaluate the new head rather than CI status alone.

Q00 (approved):
Approved on exact HEAD `29ca54acbf5da8786a23d725a7374c9ca3dec777` after independent verification. The prior raw-stop blocker on `cb49305ec6fe482f15efc9c9b659fc04fb9ff884` was reproduced; the new head preserves verbatim ACP `stopReason` in diagnostics/raw response while keeping public `finish_reason` normalized, and exact-value regressions distinguish `max_tokens` from `max_turn_requests`. Adapter/provider suites and all exact-head CI checks pass. This approval supersedes my prior-head change request. Merge remains HOLD pending the separate #1919 disposition.

Q00:
@MyoungSoo7 thank you for your contribution! we can use opencode with ourboros more better!!

PR #1896, opened 2026-08-05
fix(persistence): snapshot checkpoint state and validate integrity on save

### LINKED ISSUE #1829: bug(persistence): mutable checkpoint state can be saved with a stale integrity hash
# Mutable checkpoint state can be saved with a stale integrity hash

## Upstream baseline

- Repository: `Q00/ouroboros`
- Commit: `90ac429a3a3563a2f4ace7240ec87254e400c720`
- Observed: 2026-08-01

## Problem

`CheckpointData.create()` computes its integrity hash from `state` but stores the caller's mutable dictionary by reference. If the caller mutates that dictionary, or a nested list or dictionary, after construction and before `CheckpointStore.save()`, the checkpoint payload changes while its hash remains stale. `save()` does not validate integrity before writing and returns success for a checkpoint that `load()` immediately rejects as corrupt.

The frozen dataclass does not prevent mutation of objects nested inside `state`.

## Reproduction

Run against the current checkpoint implementation:

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from ouroboros.persistence.checkpoint import CheckpointData, CheckpointStore

with TemporaryDirectory() as td:
    state = {"steps": [1]}
    checkpoint = CheckpointData.create("mutable-state", "execution", state)
    state["steps"].append(2)

    store = CheckpointStore(Path(td))
    store.initialize()
    saved = store.save(checkpoint)
    loaded = store.load("mutable-state")

    print(f"save_is_ok={saved.is_ok}")
    print(f"load_is_ok={loaded.is_ok}")
```

Observed:

```text
save_is_ok=True
file_exists=True
load_is_ok=False
load_error=No valid checkpoint found for seed mutable-state (tried 4 levels)
```

## Expected behavior

A checkpoint accepted by `save()` must remain internally consistent and loadable unless the file is changed or damaged after the save. Caller-owned mutable state must not silently invalidate a checkpoint after construction.

- Snapshot or otherwise isolate the state used to create a checkpoint from later caller mutation.
- Validate integrity before rotating or writing any existing checkpoint history.
- Return a controlled error rather than persisting a checkpoint whose payload does not match its stored hash.
- Add regression coverage for both top-level and nested mutation after `CheckpointData.create()`.
- Preserve normal JSON-serializable state round trips.

## Duplicate check

Searched all issue and PR titles and bodies for `CheckpointData`, mutable state, state mutation, hash mismatch, and checkpoint integrity terms on 2026-08-01. The returned checkpoint issues covered the original checkpoint feature, path traversal, and unrelated harness/interview work; no matching report or implementation was found.

sumin220:
Closes #1829.

## Problem

`CheckpointData.create()` computed its integrity hash from the caller's state dictionary but stored that mutable dictionary by reference. Mutating the dictionary — or a nested list or dict inside it — after construction and before `save()` changed the payload while the hash stayed stale: `save()` reported success for a checkpoint that `load()` immediately rejected as corrupt, exactly as in the issue's reproduction (`save_is_ok=True`, `load_is_ok=False`, all four levels exhausted). The frozen dataclass protects the fields, not the objects nested inside `state`.

## Fix

- `create()` deep-copies the state before hashing, so the hash binds exactly the payload the checkpoint carries and later caller mutation is invisible. Every `create()` call site passes JSON-serializable state by construction — the hash serialization already `json.dumps`s it — so every valid state is deep-copyable; a survey of all nine production call sites confirmed plain literal dictionaries throughout.
- `save()` validates integrity before taking the file lock or rotating anything, and returns the documented `PersistenceError` (with an `integrity_mismatch` detail) for a checkpoint whose payload does not match its hash, leaving the committed checkpoint and the rollback chain untouched.

## Tests

Per the acceptance criteria, all verified failing on the baseline:

- Top-level mutation after `create()` no longer invalidates the saved checkpoint, and the loaded state is the pre-mutation snapshot.
- Nested mutation (a list append and a nested-dict write) is likewise invisible.
- A hand-constructed checkpoint whose payload does not match its stored hash is refused by `save()` with no file touched and no history rotated.

Persistence suite plus the agent-process and parallel-executor suites that build checkpoints pass (834); ruff, mypy, and the module-size gate are clean.

### DIFF
--- a/src/ouroboros/persistence/checkpoint.py
+++ b/src/ouroboros/persistence/checkpoint.py
@@ -27,6 +27,20 @@ from ouroboros.core.file_lock import file_lock as _file_lock
 from ouroboros.core.types import Result
 
 
+def _checkpoint_hash(seed_id: str, phase: str, state: dict[str, Any], timestamp_iso: str) -> str:
+    """Integrity hash over one exact checkpoint payload."""
+    serialized = json.dumps(
+        {
+            "seed_id": seed_id,
+            "phase": phase,
+            "state": state,
+            "timestamp": timestamp_iso,
+        },
+        sort_keys=True,
+    )
+    return hashlib.sha256(serialized.encode()).hexdigest()
+
+
 @dataclass(frozen=True, slots=True)
 class CheckpointData:
     """Immutable checkpoint data for workflow state.
@@ -64,15 +78,7 @@ class CheckpointData:
         # construction cannot desync the two (#1829). State is JSON-bound by
         # the hash below, so every valid state is deep-copyable.
         state = copy.deepcopy(state)
-        # Create temporary instance without hash to compute it
-        temp_data = {
-            "seed_id": seed_id,
-            "phase": phase,
-            "state": state,
-            "timestamp": timestamp.isoformat(),
-        }
-        serialized = json.dumps(temp_data, sort_keys=True)
-        hash_value = hashlib.sha256(serialized.encode()).hexdigest()
+        hash_value = _checkpoint_hash(seed_id, phase, state, timestamp.isoformat())
 
         return cls(
             seed_id=seed_id,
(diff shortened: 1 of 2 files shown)


sumin220:
Addressed the first review in da05b3c8. save() now serializes exactly once, validates that snapshot's hash, and writes that same snapshot byte for byte — the live object is never read again after validation, so a mutation landing between validation and serialization can no longer desync the persisted payload from its hash. The hash computation is a single module helper shared by create(), validate_integrity(), and the save gate, so the three can never drift. Your interleaving is the regression: the lock acquisition mutates the live object mid-save, and the persisted file must be the validated pre-mutation snapshot that load() accepts. Persistence and agent-process suites pass (330); ruff, mypy, and the module-size gate are clean.

PR #1897, opened 2026-08-05
fix(persistence): make the EventStore's SQLite-only contract explicit

### LINKED ISSUE #1832: EventStore accepts non-SQLite backends but cursor APIs require SQLite rowid
## Summary

`EventStore` accepts non-SQLite SQLAlchemy URLs and reports them as cross-process capable, but its cursor APIs emit SQLite-only `rowid` queries. On PostgreSQL those queries reference a column that does not exist in the declared `events` schema.

Verified against `origin/main` at `90ac429a3a3563a2f4ace7240ec87254e400c720`.

## Reproduction

```python
from sqlalchemy import func, select, text
from sqlalchemy.dialects import postgresql

from ouroboros.persistence.event_store import EventStore
from ouroboros.persistence.schema import events_table

store = EventStore("postgresql+asyncpg://user:pass@localhost/ouroboros")
assert store.sqlite_path() is None
assert store.supports_cross_process_workers is True

queries = [
    select(text("rowid"))
    .select_from(events_table)
    .where(events_table.c.id == "event-id"),
    select(func.coalesce(func.max(text("rowid")), 0)).select_from(events_table),
    select(events_table, text("rowid"))
    .where(text("rowid > :last_id").bindparams(last_id=0))
    .order_by(text("rowid"))
    .limit(10),
]

for query in queries:
    print(query.compile(
        dialect=postgresql.dialect(),
        compile_kwargs={"literal_binds": True},
    ))
```

The compiled statements include:

```sql
SELECT rowid FROM events WHERE events.id = 'event-id';
SELECT coalesce(max(rowid), 0) AS coalesce_1 FROM events;
SELECT ..., rowid FROM events WHERE rowid > 0 ORDER BY rowid LIMIT 10;
```

But `events_table` declares only:

```text
id, aggregate_type, aggregate_id, event_type, payload, timestamp, consensus_id
```

`rowid` is SQLite-specific and is not a PostgreSQL system column, so `append_with_rowid()`, `get_current_rowid()`, `get_events_after()`, and the other rowid-backed query paths cannot satisfy the accepted backend contract.

## Expected behavior

The backend boundary should be coherent. Either:

- reject non-SQLite URLs up front and make the EventStore contract explicitly SQLite-only, including capability reporting, or
- introduce a declared monotonic cursor column plus backend-specific initialization/query support and test each supported backend.

## Duplicate check

I searched all issues and pull requests for `rowid`, `non-SQLite`, `PostgreSQL EventStore`, and related event-store backend terms, and checked recent commits. PR #1274 documents the intended SQLite cursor semantics but does not address non-SQLite acceptance or portability.

sumin220:
Closes #1832.

## Problem

`EventStore` accepted non-SQLite SQLAlchemy URLs and reported them as cross-process capable, but every cursor API — `append_with_rowid`, `get_current_rowid`, `get_events_after` and the replay paths — emits SQLite-only `rowid` queries. On PostgreSQL those statements reference a column that does not exist in the declared `events` schema, so the accepted backend failed at first cursor use while `supports_cross_process_workers` claimed `True` the whole time.

## Fix

Of the two remedies the issue offers, this takes the coherent-boundary option: the store is explicitly SQLite-only.

- Construction refuses a well-formed non-SQLite URL with an error naming the rowid contract. The check lives in a small `persistence/backend_contract.py` module so the boundary is documented in one place — if a declared portable cursor column ever becomes worth its migration, this is the single spot that relaxes, as a deliberate change rather than a silent one. A malformed URL is still left for engine creation to report with its own, more specific error.
- Capability reporting agrees with the contract: `supports_cross_process_workers` is now simply whether a real SQLite file exists for another process to open. This also corrects the pathless `sqlite+aiosqlite://` form, which previously fell through the prefix check and misreported `True` for a database that is private to one process by construction.
- Two tests that constructed non-SQLite stores only to pin unrelated helpers (`sqlite_path`, the dashboard's daemon-target suppression) now pin those helpers directly, preserving their intent under the new boundary.

## Tests

- A `postgresql+asyncpg` URL is refused at construction with a message matching the contract (verified failing on the baseline, which accepted it).
- Every accepted SQLite form — pathless, `:memory:`, file path, and the read-only URI rewrite — still constructs.
- The capability report matches the contract: file-backed stores report `True`; pathless and `:memory:` stores report `False` (the pathless case verified failing on the baseline).

Persistence, evolution, and MCP-tools suites pass; ruff, mypy, and the module-size gate are clean — `event_store.py` stays exactly at its ratcheted budget, with the new boundary in its own module.

### DIFF
--- a/src/ouroboros/persistence/backend_contract.py
+++ b/src/ouroboros/persistence/backend_contract.py
@@ -1,0 +1,36 @@
+"""The EventStore's storage-backend contract (#1832).
+
+Every cursor API is built on SQLite's implicit ``rowid`` ordering — the
+semantics documented in #1274: ``append_with_rowid``, ``get_current_rowid``,
+``get_events_after`` and the replay paths all query ``rowid`` directly, and
+the declared ``events`` schema carries no portable substitute. A non-SQLite
+backend cannot run those statements (PostgreSQL has no ``rowid`` system
+column), so accepting such a URL only defers the failure to first cursor
+use while misreporting capabilities in the meantime. The boundary is
+therefore explicit: the EventStore is SQLite-only, checked at construction.
+Introducing a declared monotonic cursor column later can relax this in one
+place, as a deliberate change rather than a silent one.
+"""
+
+from __future__ import annotations
+
+from sqlalchemy.engine import make_url
+
+
+def require_sqlite_event_store_url(database_url: str) -> None:
+    """Reject database URLs the EventStore's cursor contract cannot serve.
+
+    A malformed URL is left for engine creation to report with its own,
+    more specific error; only a well-formed URL naming a non-SQLite
+    backend is refused here.
+    """
+    try:
+        backend = make_url(database_url).get_backend_name()
+    except Exception:
+        return
+    if backend != "sqlite":
+        raise ValueError(
+            "EventStore is SQLite-only: its cursor APIs are built on SQLite "
+            f"rowid semantics, which a {backend!r} backend cannot serve. "
+            "Use a sqlite+aiosqlite:// URL."
+        )
--- a/src/ouroboros/persistence/event_store.py
+++ b/src/ouroboros/persistence/event_store.py
@@ -682,9 +682,8 @@ class EventStore:
             parsed = make_url(database_url)
         except Exception:
             return None
-        if parsed.get_backend_name() != "sqlite":
-            return None
-        if str(parsed.query.get("mode", "")).lower() == "memory":
+        is_memory_mode = str(parsed.query.get("mode", "")).lower() == "memory"
+        if parsed.get_backend_name() != "sqlite" or is_memory_mode:
             return None
         database = parsed.database or ""
         if database.startswith("file:"):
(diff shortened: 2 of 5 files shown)


sumin220:
Addressed the first review in ba685250. _sqlite_path_from_url now recognizes mode=memory in the SQLite URI query — a named in-memory database is process-local no matter what its name looks like, so it reports no local file, and supports_cross_process_workers and the dashboard scoping inherit the correct answer. Your URI shape is the regression: the named shared-cache memory form must report the capability False and a None path, alongside the existing pathless and :memory: cases. Persistence suite passes (253); ruff, mypy, and the module-size gate are clean with event_store.py back under its budget.

sumin220:
Addressed the second review in 8bf92551. The path helper now parses the URL structurally with make_url — backend name, database component, and query are separated by the parser rather than by prefix tests on the raw suffix, mode=memory is recognized from the parsed query, and the file: prefix is decoded from the parsed database component. A query string on any accepted form can no longer masquerade as part of a filesystem path. Both of your query-bearing variants are regressions: the pathless ?cache=shared form and :memory:?cache=shared must report no local file and no cross-process capability. Persistence and dashboard-scoping suites pass (265); ruff, mypy, and the module-size gate are clean at exactly the ratcheted budget.

PR #1900, opened 2026-08-05
fix(config): deny the suffix-less OUROBOROS_CLI alias from an untrusted .env

### LINKED ISSUE #1902: Security: block untrusted OUROBOROS_CLI bridge executable overrides
## Problem

A repository-controlled project `.env` can supply the suffix-less `OUROBOROS_CLI` alias consumed by the GJC and Pi `ooo` bridges as an executable command. The untrusted environment denylist covers `*_CLI_PATH` variants but must also cover this bridge alias without removing trusted operator overrides.

## Acceptance criteria

- [x] Reject `OUROBOROS_CLI` when it originates from an untrusted project `.env`.
- [x] Preserve trusted home-configuration and real process-environment overrides.
- [x] Cover every packaged bridge dispatch source through automatic discovery rather than a test-local fixed roster.
- [x] Require successful executable-key extraction for each discovered source so one bridge cannot hide drift in another.
- [x] Keep the trust policy outside the grandfathered oversized loader module.

## Verification

Exercise config loading, GJC and Pi runtime forwarding, generated bridge installation, dot/bracket environment access, new-source discovery, and per-source parser drift.

Q00:
## Summary

The project-directory `.env` is untrusted (it travels with whatever repo the user cloned — the boundary established by CVE-2026-47211 / GHSA-c4m7-2gwp-vw76). `_UNTRUSTED_ENV_DENYLIST` carried `OUROBOROS_CLI_PATH` and every other `*_CLI_PATH` variant, but **not** `OUROBOROS_CLI` — and the suffix-less alias is the one the `ooo` frontdoor bridges actually read and execute.

Chain, verified at `v0.50.7` and at current `main`:

| Step | Location |
|---|---|
| Project `.env` applied untrusted at import | `src/ouroboros/config/loader.py:412` |
| `OUROBOROS_CLI` not denylisted (prefixes are only `DYLD_`/`LD_`) | `loader.py:194-321` |
| Whole env forwarded to the spawned vendor CLI | `orchestrator/gjc_runtime.py:165-169`, `orchestrator/pi_runtime.py:201-205` |
| Bridge uses it verbatim as the exec command | `gjc_bridge/index.ts:25,41`; `cli/commands/setup.py:3565` (Pi template) |

A repo shipping `OUROBOROS_CLI=./.tools/helper` therefore chooses the binary that runs when the victim types any `ooo ...` inside that checkout.

**Scope is wider than the `*_CLI_PATH` naming suggests:** the identical sink exists in the Pi extension emitted by `ouroboros setup --runtime pi`, not just gjc. Other runtimes forward the env too, but have no consumer of this key.

## Fix

One line of policy: `"OUROBOROS_CLI"` joins the denylist next to `OUROBOROS_CLI_PATH`, with a comment stating why a suffix-less key belongs there.

Ouroboros never sets this key itself — it is an operator escape hatch for installs where `ouroboros` is not on the child's PATH (the gjc bridge's `DEFAULT_COMMAND` is the bare name `"ouroboros"`). Denying it from an *untrusted* source costs nothing; a trusted `~/.ouroboros/.env` or a real shell export still works, and both directions are now pinned by tests.

### Deliberately not done: popping the key in `_build_child_env`

The reporter suggested also popping `OUROBOROS_CLI` in `gjc_runtime._build_child_env` alongside `OUROBOROS_AGENT_RUNTIME`/`OUROBOROS_LLM_BACKEND`. That was rejected: an unconditional pop cannot distinguish provenance, so it would break the legitimate operator escape hatch above (the child falls back to a bare-name PATH lookup that either fails or resolves a *different* `ouroboros`). After this patch, anything still in `os.environ` at spawn time is trusted by construction, so the pop would buy defence only against a future bypass of the loader while costing a working feature today. Raising it here rather than deciding silently.

## Why there are two commits

`enforce-module-size` refuses any growth of `config/loader.py` (2432-line grandfathered budget, zero headroom) and instructs you to put new code in a new module. So commit 2 is a **pure move**: the denylist, the denied prefixes, and the membership predicate go to `src/ouroboros/config/untrusted_env.py` with their comments intact, no behaviour change, names de-underscored now that they cross a module boundary. `loader.py` drops 2445 → 2277 lines. Commit 1 is the security delta on its own, so the fix stays readable independently of the move.

## Anti-drift

The bridges are TypeScript and live outside Python's import graph, which is exactly why nothing connected them to the denylist and why the gap survived two advisories. `test_bridge_dispatch_entry_env_keys_are_denylisted` re-derives the key **from the bridge sources themselves** and asserts denylist membership, with a non-vacuity guard on the regex. A new bridge, or a rename of this one, now fails CI instead of silently reopening the hole.

## Review blocker resolution

- Removed the fixed `_BRIDGE_ENTRY_SOURCES` tuple.
- Added package-wide dispatch-source discovery and per-source non-vacuity.
- Added mutation regressions for bracket access, future bridge discovery, and one-source-only parser drift.
- Updated the stale `python-dotenv` trust-boundary comment after the policy-module extraction.

## Test plan

- `uv run pytest tests/unit/config tests/unit/orchestrator/test_gjc_runtime.py tests/unit/orchestrator/test_pi_runtime.py` — 550 passed
- `uv run pytest tests/unit/cli/test_setup.py -k bridge` — 2 passed
- `ruff check` + `ruff format --check` clean
- Discovery confirms both current bridge sources independently yield `OUROBOROS_CLI`; mutation regressions prove bracket access, new-source discovery, and per-source failure behavior

## Credit

Reported privately by **Arpit Jain** (@arpitjain099), who traced the full chain and correctly read it as a naming gap rather than a decision. A draft security advisory is being prepared separately; this is not covered by GHSA-jv2h-4p9v-wf5w (0.42.1), which named neither this key nor this sink class.

Closes #1902

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     "prompt-toolkit>=3.0.0,<4.0.0",
     # Ceiling is deliberately tighter than the major-version bounds around it:
     # this parses `.env`, a documented trust boundary (see
-    # `_UNTRUSTED_ENV_DENYLIST` in config/loader.py), so only patch releases
+    # `UNTRUSTED_ENV_DENYLIST` in config/untrusted_env.py), so only patch releases
     # flow without review. An exact pin would be tighter still, but
     # `test_runtime_and_optional_dependencies_have_upper_bounds` reserves
     # `==` for the optional extras; core runtime deps stay bounded ranges.
--- a/src/ouroboros/config/loader.py
+++ b/src/ouroboros/config/loader.py
@@ -61,6 +61,7 @@ from ouroboros.config.models import (  # noqa: E402
     get_default_config,
     get_default_credentials,
 )
+from ouroboros.config.untrusted_env import is_untrusted_env_denied_key
 from ouroboros.core.errors import ConfigError  # noqa: E402
 from ouroboros.orchestrator_stage import (  # noqa: E402
     Stage,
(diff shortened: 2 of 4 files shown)


Q00:
@ouroboros-agent Please review PR #1900 at exact HEAD `c482795210435c25075e3b6717aeb0df3f41ede5` against `main`.

This follow-up resolves review #4861955344: the anti-drift test no longer uses a fixed source tuple, discovers packaged bridge dispatch sources from the executable sink, requires successful extraction for each individual source, supports dot and bracket environment access, and has mutation regressions proving new-source discovery and that one healthy source cannot mask drift in another. The stale trust-boundary dependency comment was also corrected.

Verification: 550 config/GJC/Pi tests passed; 2 bridge setup tests passed; Ruff, format, and diff checks passed. Closes #1902.

Q00:
@ouroboros-agent Please review PR #1900 at exact HEAD `658f858a8c69794ac614500559963784e17731de` against `main`.

This follow-up resolves review #4862069214: bridge-source discovery no longer depends on the local `entry` identifier. It recognizes a JavaScript/TypeScript identifier followed by `.command`, and a two-source regression proves that renaming only one sink to `resolved.command` cannot make it disappear while an `entry.command` source remains healthy.

Verification: 551 config/GJC/Pi tests passed; 2 bridge setup tests passed; focused loader suite 179 passed; Ruff, format, and diff checks passed. Closes #1902.

PR #1622, opened 2026-07-14
fix(mcp): harden job lifecycle polling and recovery

lee3Q:
## Summary

Hardens MCP job lifecycle polling and recovery for background Ouroboros jobs.

This PR focuses on:
- bounded `job_wait` polling and timeout/render metadata
- terminal `job_result` recovery from live execution snapshots
- preserving non-empty `relay_events`
- drain behavior that does not cancel externally owned live jobs
- clearer linked execution failure reporting before generic owner-dead/orphan labels
- FastMCP/call-tool lifecycle logging around MCP server activity

This is intentionally separate from the zcode runtime backend PR (#1568). It layers on top of current `main` after upstream #1620 and preserves `relay_events` while adding timeout/render metadata needed by GUI/MCP recovery paths.

## Verification

Local verification on Python 3.13:

- `python -m pytest tests/unit/mcp -q` -> `1349 passed`
- CLI wait-output + drain regression subset -> `10 passed`
- boundary/idempotence subset -> `12 passed`
- adapter/live-snapshot/bounded-scan subset -> `7 passed`
- stale-cache subset -> `1 passed`
- adapter/stale timing pair repeated 5x -> `2 passed` each run
- reverse-order timing subset -> `6 passed`
- `ruff check` on touched files -> all checks passed
- `ruff format --check` on touched files -> 9 files already formatted
- `ouroboros mcp info` -> 32 tools
- Ouroboros QA `qa-42dde7cb` -> PASS 0.82

Post-fix full repo unit rerun:

- `python -m pytest tests/unit -q` -> `9 failed, 12071 passed, 2 skipped`
- Remaining failures are outside this MCP lifecycle slice and match local/environmental areas already isolated: config TUI local defaults, dashboard socket sandbox permission, Codex runtime/profile local config expectations, and wheel build DNS/network resolution.

Live GUI/MCP smoke:

- session40 plugin-managed ZCode smoke started `job_b7e5b0f0a0ba` / `exec_3f06d5f4e917`
- used plugin namespace only, no primary `mcp__ouroboros__*` tools
- target file `/private/tmp/ozo-gui-full-loop-smoke-session40/ozo_gui_smoke_session40.txt` verified locally as exact bytes `OZO_GUI_SMOKE_SESSION40_OK`, 26 bytes, no trailing newline
- fresh ZCode turn recovered plugin `job_result` as terminal `completed` (`Execution complete: 1/1 ACs completed`)
[...]

PR #1908, opened 2026-08-06
feat(telemetry): anonymous PostHog usage telemetry with funnel instrumentation

### LINKED ISSUE #1909: Product analytics: anonymous usage telemetry (PostHog) for funnel, retention, and reliability metrics
## Why

Ouroboros has no visibility into how it is actually used. We cannot answer:

- How many people install it, and how many of them reach their first successful run?
- Where does the funnel leak — interview → seed → run → evaluate → evolve?
- Do users enter from a terminal (`ouroboros ...`) or inside an AI agent (`ooo ...` in Claude Code / Codex)?
- Which runtime backends do promises get kept on (per-backend run success rates, per-version regressions)?
- Net user growth and retention — needed both to improve the product and, once past a disclosure threshold, as public aggregate stats.

Owner-side local data already showed why this matters: interview completion 9.3%, run success 51%, auto success 14% over 9 weeks — but local EventStore data cannot answer any cross-user question.

## What

Anonymous, opt-out usage telemetry with a strict trust contract:

- **Events**: `install_started` / `install_completed` (install.sh), `command_run` (every `ouroboros_*` MCP tool call + direct CLI commands, with funnel normalization), `mcp_serve_started` (session-attach denominator)
- **Properties**: command, source (cli/mcp), frontdoor (claude/codex), runtime/llm backends, ok, duration_ms, error_type, app_version, os, ci
- **Never collected**: code, prompts, seed content, paths, arguments, PII
- **Trust contract**: TELEMETRY.md whitelist, declared uses (product improvement + public aggregate stats), fixed counting rule, k-anonymity floor (cells <10 users unpublished), append-only changelog, 3 opt-out paths, one-time first-run notice
- **Dashboards**: PostHog Overview (lifecycle/net growth, activation funnel, retention, per-backend & per-version reliability)

## Acceptance criteria

- [ ] Telemetry never blocks or crashes a command (fire-and-forget, stdlib only)
- [ ] `DO_NOT_TRACK=1`, `OUROBOROS_TELEMETRY=0`, and `telemetry.enabled: false` each fully disable it
- [ ] cli-vs-agent ratio is not inflated by internal automation (serve boots, in-process servers, detached workers)
- [ ] Test suite can never post real events
- [ ] End-to-end verified: an event sent by the shipped module is queryable in PostHog

Q00:
Closes #1909

## Summary
- Anonymous, opt-out usage telemetry (PostHog) so the install funnel, interview → seed → run → evolve conversion, terminal-CLI vs in-agent entry ratio, and per-backend success rates become measurable across all 13 runtimes
- Single chokepoint instrumentation: `MCPServerAdapter.call_tool` (covers Claude/Codex stdio, legacy serve wrapper, and the codex `ooo` intercept), plus CLI callback and `install.sh` pings; polling tools sampled 1/50 with a `sample_rate` re-weighting property
- Trust contract shipped in the same commit: `TELEMETRY.md` (event whitelist, declared uses incl. public aggregate stats, fixed counting rule "1 active user = 1 anonymous ID with ≥1 successful run/week — installs/CI/retries don't count", k-anonymity floor, append-only changelog), one-time first-run notice, 3 opt-out paths (`DO_NOT_TRACK`, `OUROBOROS_TELEMETRY=0`, `config telemetry.enabled`)

## Design notes
- stdlib-only transport (urllib + daemon thread, bounded queue) — no new dependency, never raises, never blocks a command, drops events offline
- Identity is a random UUID in `~/.ouroboros/telemetry.json` (no PII, no fingerprinting; removed by `ouroboros uninstall`); the embedded `phc_` key is PostHog's public write-only project key, safe by design in OSS
- `ouroboros mcp serve` boots are captured as `mcp_serve_started` (session-attach denominator) and excluded from terminal-CLI counts so host-spawned serves don't inflate direct usage
- In-process servers (`ooo auto`) and detached job workers bypass `call_tool`, so internal automation cannot pollute the cli-vs-agent ratio
- `tests/conftest.py` force-disables telemetry suite-wide; telemetry unit tests neutralize the embedded key and inject a fake transport

## Checks passed
- [x] ruff lint + format (9 files)
- [x] mypy (7 source files)
- [x] pytest: telemetry 20/20, config 499, cli + mcp adapter — no new failures vs clean tree (combined-run cross-contamination failures reproduce identically on main; 2 known pre-existing failures deselected)

## Test plan
- End-to-end verified live: smoke event sent through the real module and confirmed queryable in PostHog via HogQL
- After merge/release: watch the Overview dashboard (Lifecycle/funnel/reliability) populate from real installs; first-run notice appears once per machine

## Credit
@code-yeongyu suggested this Posthog to track user data in Ouroboros funnel.


Q00:
Broad direction review after the requested-changes loop:

The bounded, asynchronous, failure-isolated telemetry foundation is a strong start, and the product-learning goal fits Ouroboros. The current trust and metric contracts do not yet fit the project direction, though:

- project-controlled `.env` can override the telemetry endpoint, key, and opt-out;
- unrelated configuration errors fall back to telemetry-on;
- installer emission can happen before notice or persisted opt-out is honored;
- queued execution is counted as verified success;
- early adapter failures are omitted, biasing reliability metrics.

Those behaviors conflict with the fail-closed local trust boundary and durable-evidence semantics. Please pause implementation and define one trusted telemetry-control resolver plus distinct submission and terminal-outcome events. Then prove opt-out, malformed-config, installer ordering, and failure-path behavior with integration tests. The existing queue and failure-isolation work should remain reusable once those contracts are explicit.

Q00:
## Remediation verification — exact HEAD `871c00c7d25016f9b6134dd5cdbd60e24efad1ae`

All prior `ouroboros-agent` findings were treated as acceptance criteria, including the earlier warning/documentation items.

Resolved boundaries:
- project `.env` can no longer set telemetry opt-out/destination/key authority;
- invalid or unreadable persisted config fails telemetry closed;
- installer honors persisted opt-out and persists/displays notice before first capture;
- asynchronous Start* results emit submission `accepted`, never execution `ok`;
- durable job terminals emit a separate deduplicated `workflow_outcome`;
- only completed evaluation with explicit `final_approved=true` is `verified=true`;
- unknown tools, security denial, SDK input validation, adapter error, output validation, and SDK conversion failures are captured exactly once;
- public lineage/project status polling tools are included in 1/50 sampling;
- `TELEMETRY.md` lists the complete trigger-site and active-user contracts.

Verification:
- focused telemetry/config/installer/adapter/job-manager suites: `543 passed`
- added FAILED-terminal and SDK exactly-once regressions: `2 passed`
- unreadable installer-config regression: `1 passed`
- full suite: `17676 passed, 93 skipped, 1 xfailed`; 3 order-dependent Copilot reasoning-effort failures occurred in unchanged code and pass independently on both this worktree and exact main `099898e8` (`5 passed` on each)
- `mypy src/ouroboros`: 532 files, clean
- Ruff format/check, module-size, auto-boundary, max-turns-envelope, `bash -n scripts/install.sh`, and `git diff --check`: clean

Direction gate: CONDITIONAL_PASS resolved to implementation-complete pending exact-head bot review and independent verification. This PR remains HOLD while aggregate review is `CHANGES_REQUESTED`; no merge/approval is requested from this worker.

Q00:
@ouroboros-agent please re-review exact HEAD 871c00c7d25016f9b6134dd5cdbd60e24efad1ae

Q00:
Resolved the two independently reproduced exact-HEAD blockers on `96cf738411fc94f0ba512a852c08cdbd965a2429` (latest `origin/main` `099898e8ab91dd7a2f7e346bb0a9b289c9a9c17b` is an ancestor).

Changes:
- `workflow_outcome.verified` now requires all three conditions: normalized terminal status is `completed`, job type is `evaluate`, and `final_approved is True` (literal bool).
- Added regressions for failed/cancelled/interrupted evaluations carrying stale `final_approved=True`, completed non-evaluate work, missing/false approval, and truthy non-bool approval.
- Replaced the installer's telemetry-block-only awk parser with full YAML + `OuroborosConfig` validation in an isolated Python interpreter. Any malformed YAML, schema-invalid unrelated field, unreadable config, or unavailable validator fails closed. A genuinely absent config retains the documented default-on behavior.
- Preserved `DO_NOT_TRACK` precedence and notice-before-first-capture ordering; explicit persisted `telemetry.enabled: false` remains a complete opt-out.

Verification:
- `bash -n scripts/install.sh`
- `shellcheck scripts/install.sh`
- `uv run --extra mcp pytest -q tests/unit/test_telemetry.py tests/unit/scripts/test_install_runtime_selection.py` → 70 passed
- `uv run --extra mcp pytest -q tests/unit/config/test_loader_env.py tests/unit/mcp/test_job_manager.py tests/unit/mcp/server/test_adapter.py` → 482 passed
- post-shell-fix installer rerun → 37 passed
- `uv run ruff check src/ tests/` → passed
- `uv run mypy src/ouroboros` → 532 source files, no issues
- `git diff --check` → clean

No merge performed.

@ouroboros-agent please re-review exact HEAD 96cf738411fc94f0ba512a852c08cdbd965a2429

Q00:
## Independent exact-head re-verification — HOLD / FAIL

Verified HEAD: `96cf738411fc94f0ba512a852c08cdbd965a2429`

The two returned blockers are substantially fixed: terminal evaluation outcomes are truthful, and the installer now validates the full configuration fail-closed while preserving genuinely absent default-on plus notice-before-capture. A remaining privacy bypass is blocking.

### Blocker: dangling config symlink fails open

Both runtime surfaces treat a dangling `~/.ouroboros/config.yaml` symlink as if the user had no configuration:

- `scripts/install.sh:104`: `[ -e "$f" ] || return 0`
- `src/ouroboros/config/loader.py`: `Path.exists() == False`

Reproduction:

1. Create `~/.ouroboros/config.yaml` as a symlink to a missing target.
2. The copied remote-installer harness displays/persists the notice and emits both `install_started` and `install_completed`.
3. Python reports `os.path.lexists=True`, `Path.exists=False`, `get_telemetry_enabled()=True`, and `telemetry.is_enabled()=True`.

Expected: this is invalid/unreadable configuration, not genuinely absent configuration, so both installer and Python collection must remain disabled. Use `lexists`/`-L`-aware existence semantics and add exact regressions on both surfaces without weakening the truly absent default-on case.

### Other gate status

- Required terminal-outcome matrix: PASS.
- Installer full-config malformed/unrelated-invalid/unreadable/explicit-false/no-validator and DNT/env precedence: PASS.
- Installed-entrypoint schema-validator fallback: PASS.
- Ruff lint: PASS, but format check fails on `src/ouroboros/telemetry.py` and `tests/unit/scripts/test_install_runtime_selection.py`; exact Ruff CI is red.
- Python 3.14 CI also failed the separate #1933 revocation test flake; do not mask it by rerun.
- No exact-head formal bot review exists; aggregate remains `CHANGES_REQUESTED`.

No approval or merge was performed.

Q00:
Resolved the dangling-config-symlink privacy bypass on exact HEAD `28f2eee3e47c49e5eb477841158a024737ba37f9` (latest `origin/main` `099898e8ab91dd7a2f7e346bb0a9b289c9a9c17b` is an ancestor).

Changes:
- `scripts/install.sh` now distinguishes a genuinely absent config from a dangling symlink using `-L`; dangling config state continues through the unreadable/fail-closed path.
- `get_telemetry_enabled()` likewise distinguishes `Path.exists() == false` for absence from `Path.is_symlink() == true`; a missing symlink target reaches `load_config()` and returns disabled on `ConfigError`.
- Added a Python regression proving both `get_telemetry_enabled()` and `telemetry.is_enabled()` are false for a dangling `~/.ouroboros/config.yaml`.
- Added a copied standalone installer regression with no adjacent source tree: a dangling config symlink emits no notice, creates no telemetry identity, and performs no capture.
- The copied-installer genuine-absence test still proves disclosed default-on and notice-before-capture ordering.
- The curl test probe now separates the installer's pre-notice PyPI metadata lookup from PostHog `/capture/` calls, so ordering assertions cover only collection.
- Ran Ruff formatting on the two files called out by the verifier; exact format check is now green locally.

Verification:
- `uv run pytest -q tests/unit/test_telemetry.py tests/unit/scripts/test_install_runtime_selection.py` → 72 passed
- `uv run --extra mcp pytest -q tests/unit/config/test_loader.py tests/unit/config/test_loader_env.py tests/unit/mcp/test_job_manager.py` → 532 passed
- `uv run ruff format --check src/ tests/` → 1,229 files formatted
- `uv run ruff check src/ tests/` → passed
- `uv run mypy src/ouroboros` → 532 source files, no issues
- `bash -n scripts/install.sh`, `shellcheck scripts/install.sh`, and `git diff --check` → passed

I did not manually rerun or mask the separate #1933 lease-test flake. No approval or merge performed.

@ouroboros-agent please re-review exact HEAD 28f2eee3e47c49e5eb477841158a024737ba37f9

Q00:
@ouroboros-agent Both blockers from review `ebc3c15c` are addressed in `d072f5c6a` (merged latest main first — the `job_manager.py` conflict with #1940's reservation sets resolved by keeping both cleanup paths).

**Blocker 1 — explicit enable overriding persisted opt-out** (`loader.py`, `install.sh`): `get_telemetry_enabled()` now resolves every disabling source before anything else — `DO_NOT_TRACK`, `OUROBOROS_TELEMETRY=0`, persisted `telemetry.enabled: false`, and invalid/unreadable config all win unconditionally. `OUROBOROS_TELEMETRY=1` is never an override; with no disabling source and a genuinely absent config it is simply the default-on path. The installer's `_telemetry_enabled` drops its `=1` short-circuit and always runs `_telemetry_config_allows`. Regressions: `test_explicit_enable_cannot_override_{persisted_opt_out,malformed_config}` (Python resolver, both `get_telemetry_enabled` and `telemetry.is_enabled`) and `test_installer_explicit_enable_cannot_override_{persisted_opt_out,malformed_config}` (installer, notice + capture both suppressed). TELEMETRY.md now states this precedence explicitly.

**Blocker 2 — installer ignoring trusted `~/.ouroboros/.env`**: `install.sh` now parses exactly the four allowlisted telemetry keys (`DO_NOT_TRACK`, `OUROBOROS_TELEMETRY`, `OUROBOROS_POSTHOG_API_KEY`, `OUROBOROS_POSTHOG_HOST`) from `~/.ouroboros/.env` before any notice or capture, with real-process-environment precedence mirroring `config/loader.py` (an already-set process value — including set-but-empty — is never overridden; inline `# comments` on unquoted values are stripped so `OUROBOROS_TELEMETRY=0 # off` still opts out). Copied-installer regressions: `test_installer_honors_user_env_opt_out`, `test_installer_honors_user_env_destination_override`, `test_installer_process_env_wins_over_user_env_file`.

**Non-blocking suggestion**: both first-run notices now point at the canonical GitHub URL for TELEMETRY.md instead of the bare filename, since the wheel does not package the file.

Verification: telemetry + installer + loader-env suites 276 passed; MCP adapter + job-manager suites 298 passed; ruff lint/format clean; mypy clean on touched sources; `bash -n scripts/install.sh` clean.

Q00:
@ouroboros-agent Both blockers from round req_1786217700_1032 are addressed in 8d418f597 (plus a main merge for the conflict):

**1. Installer dotenv parser (scripts/install.sh) — fixed.** Quoted values now follow python-dotenv semantics: the value runs to the first matching closing quote, and only whitespace and/or a `# comment` may follow it. `OUROBOROS_TELEMETRY="0" # persisted opt-out` and `DO_NOT_TRACK="1" # off` now parse to bare `0`/`1` and disable collection before any notice or event. An unclosed quote or trailing garbage after the closing quote is a parse error and skips the binding, matching python-dotenv's error-skip behavior. Regressions added for double-quoted, single-quoted, uncommented-quoted, unclosed-quote, and trailing-garbage forms of both opt-out variables (tests/unit/scripts/test_install_runtime_selection.py, 420 passed).

**2. Blocking exit flush (src/ouroboros/telemetry.py) — removed.** `atexit.register(flush)` is gone; the worker remains a daemon thread and process termination never waits on telemetry, even against a stalled destination. Events still queued at exit are dropped, which the module docstring, `flush()` docstring, and TELEMETRY.md now state explicitly. `flush()` stays available for explicit callers only (tests; no production call sites). Added a stalled-transport regression: a subprocess pointed at a local accept-but-never-respond socket queues an event and must exit in well under the old ~2s blocking bound (tests/unit/test_telemetry.py::TestExitDoesNotBlock, 38 passed).

**3. Non-blocking suggestion — comment corrected.** The embedded-key comment now says an empty constant (e.g. stripped in a fork) disables telemetry, while a blank/unset `OUROBOROS_POSTHOG_API_KEY` env var falls back to the embedded key.

Verification: `uv run --extra mcp pytest` telemetry + loader-env + adapter + job-manager suites all green (532 passed), installer suite 420 passed, `ruff check`/`format --check` clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent Both findings from round req_1786309679_1218 are addressed in a8bb826d9:

**1. Installer opt-out grammar/casing parity (scripts/install.sh) — fixed.** The trusted-env parser now accepts `export` followed by any run of spaces/tabs (`export[[:space:]]*` + re-trim), and `_telemetry_enabled` lowercases and trims both flag values before matching, using exactly the application resolver's sets (`1|true|on|yes` for DO_NOT_TRACK, `0|false|off|no` for OUROBOROS_TELEMETRY). All four reproduced forms — multi-space export, tab export, `OUROBOROS_TELEMETRY=OFF`, `DO_NOT_TRACK=YES` — now disable installer telemetry before any notice or event, plus a combined `export  DO_NOT_TRACK="1" # off` regression. 6 new installer regressions; suite now 58 passed.

**2. Concurrent first-use identity (src/ouroboros/telemetry.py) — made atomic on both surfaces.** Python: `_publish_new_state` writes the candidate to a same-directory temp file and publishes with `os.link` (atomic create-if-not-exists, content complete before the link), falling back to `O_CREAT|O_EXCL` where hard links are unsupported, then re-reads the file (bounded 3×20ms retry for the fallback's partial-write window) and caches whichever `distinct_id` actually won — never its local candidate. Installer: `_telemetry_distinct_id` uses the same tmp + `ln` publish and adopts the winner from the file on re-read. New regression: 16 subprocesses synchronized on a sentinel file against one fresh HOME converge on a single UUID that matches the file (tests/unit/test_telemetry.py::TestConcurrentIdentityCreation).

**3. Test-hygiene suggestion — fixed.** The stalled-transport acceptor now catches shutdown `OSError` and is joined before the server socket closes; the telemetry suite runs clean under `-W error::pytest.PytestUnhandledThreadExceptionWarning`.

Verification: telemetry + loader-env + adapter + job-manager suites 533 passed, installer suite 58 passed, ruff check/format clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent Both blockers from round req_1786311128_1225 are addressed in 2b786fc84:

**1. Event/property whitelist enforced at serialization (src/ouroboros/telemetry.py) — fixed.** `capture()` now drops any event not in `_EVENT_ALLOWLIST` (command_run / workflow_outcome / mcp_serve_started) before it can touch the queue, and sanitizes properties against that event's allowlisted keys plus the base (`app_version`, `os`, `python_version`, `frontdoor`, `ci`) and context sets: unknown keys, non-scalar values, and strings longer than 200 chars are dropped (dropped, not truncated). `set_context()` accepts only the four resolved-backend keys with the same scalar/length gate. Your probe shapes are regressions now: `capture("not_whitelisted", {"prompt": ...})` never reaches the transport, and `prompt`/`file_path` on a valid event are stripped while allowlisted keys survive (tests/unit/test_telemetry.py::TestPropertyAllowlist, 5 tests).

**2. Malformed identity state atomically repaired — fixed on both surfaces.** `_load_state()` distinguishes absent (FileNotFoundError → unchanged create-if-absent `_publish_new_state`) from invalid (exists but unparseable / no distinct_id → new `_repair_state`): the repairer takes an `O_EXCL` repair lock, re-validates under the lock, atomically replaces the corrupt file via tmp + `os.replace`, and re-reads/adopts the survivor; losers wait bounded (10×20ms) and adopt, with a stale-lock takeover fallback. `_write_state()` itself is now tmp + `os.replace` (the non-atomic truncate-and-write you flagged as the internal corruption source is gone). `scripts/install.sh` mirrors the distinction: `ln` create-if-absent when the file is missing, `mv` atomic repair when it exists but has no extractable distinct_id, re-read-and-adopt either way. Regressions: sequential two-process convergence after `not-json{{{` corruption, 8-process synchronized concurrent convergence (all ids equal the persisted file), and an installer test proving install events are sent under the repaired persisted id.

**3. Medium follow-up (repeated notices) — fixed.** `show_first_run_notice()` claims an `O_EXCL` marker (`~/.ouroboros/telemetry.notice`) so exactly one of N simultaneous first-use processes prints; `notice_shown` is still persisted in telemetry.json via the now-atomic write. Regression: 8 synchronized subprocesses → notice appears exactly once.

Verification: telemetry + loader-env + adapter + job-manager suites 541 passed under `-W error::PytestUnhandledThreadExceptionWarning`, installer suite 59 passed, ruff clean, mypy clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent All three blockers from round req_1786312333_1233 are addressed in d78f6fe90:

**1. Installer concurrent corrupt-state repair (scripts/install.sh) — now one exclusive transaction.** The corrupt branch no longer does an unconditional `mv`: repairers race for `$f.repair.lock` via `mkdir` (atomic create-if-not-exists; a path created by either runtime's primitive — shell `mkdir` or Python's `O_CREAT|O_EXCL` file open at the same name — makes the other fail with EEXIST, so installer and Python repairers exclude each other cross-runtime, verified on macOS). The winner re-validates under the lock before replacing; losers wait bounded (10×50ms) and adopt the winner's id, with a stale-lock takeover fallback. Every process ends by adopting whatever distinct_id survives in the file, so no process emits under an id different from the persisted one. Regression: `test_concurrent_corrupt_telemetry_json_repair_converges_on_one_id` runs 4 sentinel-synchronized processes of the extracted `_telemetry_distinct_id` against `not-json{{{` and asserts all four printed ids equal the persisted id (installer suite 60 passed).

**2. Notice claim crash recovery (src/ouroboros/telemetry.py) — fixed.** A successful marker claim now prints the notice BEFORE persisting `notice_shown=true` (a crash between the two re-discloses later — benign duplicate — instead of persisting silence). On `FileExistsError`: flag true → done; flag false with marker mtime younger than 10s → live concurrent claimer, stay silent; older → crash residue, unlink and reclaim once, reprint. Your reproduced state (existing marker + `notice_shown=false`) is the new regression `test_recovers_from_stale_crash_marker`, plus `test_fresh_marker_assumes_live_claimer`; the 8-process exactly-once race test still passes.

**3. Detached worker terminal-outcome delivery (src/ouroboros/mcp/detached_worker.py) — bounded exit flush added.** `main()` wraps the entire `asyncio.run(run_worker(...))` in try/finally calling `_flush_telemetry_before_exit()` → `telemetry.flush(timeout=5.0)`, covering success, failure, and compensation paths; a single stalled batch is capped by the 4s HTTP timeout, and this is a background worker, not an interactive command, so the never-blocks-a-command contract holds (TELEMETRY.md wording updated to carve out this one deliberate exception). Tests: flush-wiring proven for successful runs, raising runs, and argv-validation early-returns (no flush), and an end-to-end delivery test where a subprocess queues `workflow_outcome` (command=evaluate, verified=true) and the exact exit-flush helper delivers it to a real local HTTP sink before the process exits (tests/unit/mcp/test_detached_worker.py, new file).

Verification: telemetry + detached-worker + loader-env + adapter + job-manager suites 547 passed under `-W error::PytestUnhandledThreadExceptionWarning`; broader rerun including tests/integration/mcp/test_detached_job_worker.py and tools/test_start_evaluate.py 574 passed; installer suite 60 passed; ruff, mypy, `bash -n` all clean.

Q00:
@ouroboros-agent All three findings from round req_1786313694_1250 are addressed in 3246360cc:

**1. Exact per-event property contract (src/ouroboros/telemetry.py + TELEMETRY.md) — fixed in both directions.** The single base∪context union is gone; serialization now uses four literal per-variant key sets — `command_run` source=mcp, `command_run` source=cli (no tool/phase/ok/duration_ms/backend-context keys), `workflow_outcome`, and `mcp_serve_started` (no python_version, no backend context) — selected by event name and the caller-provided `source`. TELEMETRY.md's "What is sent" table was rewritten to declare each exact set (mcp row now lists source/is_funnel/sample_rate/ci; cli row source/is_funnel/frontdoor/ci; workflow_outcome frontdoor/ci; serve row ci; both install rows the shell's `source` property), with an explicit "exact, not subset" note. Regressions: `TestExactPropertySets` drives every variant through its real public helper with backend context deliberately set and asserts both subset-of-contract and exact-set-of-fixture — including that mcp_serve_started strips python_version/backends and cli strips context even when genuinely present.

**2. Verified terminal outcomes for non-job evaluation paths — the reachable path is wired, the unreachable one is honestly narrowed.** Direct `ouroboros_evaluate` now emits the same `workflow_outcome` terminal event as job-backed runs via a shared boundary (`record_direct_evaluation_outcome` in mcp/telemetry_boundary.py), called at every definite terminal exit of `EvaluateHandler`: single-AC success (verified iff `final_approved=True`), evaluation failure, config-error and internal-error exits, multi-AC per-AC failures, and multi-AC checklist completion (verified iff `all_passed`); argument-validation rejections that never start an evaluation deliberately emit nothing. OpenCode plugin-mode `delegated_to_plugin` completions happen outside the Ouroboros process and cannot be observed — TELEMETRY.md's counting-rule section now states they are never counted toward verified outcomes or the active-user rule (undercount, never overcount). 14 new tests in tests/unit/mcp/tools/test_evaluate_telemetry.py cover each emitting/non-emitting exit.

**3. Cross-process-unbiased 1/50 sampling — fixed.** The process-local counter (which always captured call #1 of every fresh process as a "1/50 sample") is replaced by a per-call random draw (`_poll_rng.random() < 1/50`), unbiased regardless of process lifetime. `TestUnbiasedPollSampling` proves the first call is not structurally captured, a below-threshold draw is captured with sample_rate=50, and — the multi-process regression — 10 fresh subprocesses with precomputed seeds produce exactly the expected 5-captured/5-skipped pattern, which the old counter bug could never produce (it would capture all 10).

Verification: telemetry + evaluate-telemetry + detached-worker + loader-env + adapter + job-manager suites 569 passed under `-W error::PytestUnhandledThreadExceptionWarning`; ruff and mypy clean; scripts/install.sh untouched this round (installer suite unchanged at 60 passed).

Q00:
@ouroboros-agent The blocker and the follow-up from round req_1786315016_1266 are addressed in 18b04fb21:

**1. Single-owner terminal semantics (evaluation_handlers.py / evaluation_job.py) — fixed.** `EvaluateHandler.handle()` now takes a keyword-only `emit_terminal_telemetry: bool = True`, threaded into `_handle_multi_ac`; all seven `record_direct_evaluation_outcome` call sites are conditional on it. The reusable handler cannot know whether it owns the top-level boundary, so each caller declares it: the MCP adapter's direct `ouroboros_evaluate` dispatch keeps the default (genuine direct boundary), `ChecklistVerifyHandler` keeps the default with a comment (it is itself a top-level tool with no Start*/job-backed variant), and `run_evaluation_job` passes `False` at both of its `handle()` call sites because `JobTelemetryBoundary` already owns that job's durable terminal event with its `$insert_id` dedup. One StartEvaluate execution now emits exactly one `workflow_outcome`. The composition regression you asked for wires a REAL EvaluateHandler into a REAL StartEvaluateHandler + JobManager + EventStore, runs the job to terminal, and asserts capture is invoked exactly once (patched at `ouroboros.telemetry.capture`, the definition site both producers funnel through) with `$insert_id` present; a sibling test asserts the direct path emits exactly once with no `$insert_id` (TestJobBackedEvaluationSuppressesDirectEmission).

**2. TELEMETRY.md producer disclosure — fixed.** The `workflow_outcome` row and a new Notes bullet now name both producers — the durable background-job terminal boundary (job-derived `$insert_id` for replay dedup) and direct non-job `ouroboros_evaluate` completions (no `$insert_id`; suppressed inside background jobs so exactly one event fires per evaluation) — and the stale "comes from the durable job-terminal boundary" sentence was reworded to match. "Where the code lives" now lists `src/ouroboros/mcp/telemetry_boundary.py` (adapter wrapper, job observer, direct-evaluation boundary) and `src/ouroboros/mcp/tools/evaluation_handlers.py` as audited call sites.

Also fixed collaterally: the new keyword initially made narrow-signature EvaluateHandler test doubles raise TypeError before their body ran, which turned one cancellation test into a silent hang — every stub standing in for EvaluateHandler across the chaining/start-evaluate suites gained `**kwargs` (verified against the pre-change baseline; stubs for StartEvaluateHandler/StartRalphHandler were deliberately left untouched).

Verification: evaluate-telemetry + start-evaluate + job-manager + run/ralph chaining + telemetry suites 301 passed; wider blast radius (multi-AC, definitions, checklist-verify, adapter) 394 passed; ruff, mypy clean.

Q00:
@ouroboros-agent The identity-validation blocker from round req_1786317354_1287 is addressed in 9f9639966, with the same semantics on both surfaces:

**Python (src/ouroboros/telemetry.py).** `distinct_id` is valid only when it fullmatches the canonical hyphenated UUID pattern (case-insensitive); every load canonicalizes to lowercase, and a stored non-canonical (uppercase) form is atomically rewritten on first read, so `distinct_id()` — the value that flows to the transport outside the property sanitizer — can never return anything but a canonical UUID. Valid JSON carrying a non-UUID id (your `person@example.com` probe) is treated as an identity-contract violation: the stored text is never transmitted, a fresh uuid4 replaces it through the existing atomic `_repair_state` protocol. Malformed JSON with a regex-salvageable UUID keeps that UUID through repair (canonicalized) instead of minting a new one — that is what makes a Python repair of installer-accepted state converge on the same identity instead of fragmenting it; malformed state with nothing salvageable mints fresh as before.

**Installer (scripts/install.sh).** Extraction is now two-tier: a STRICT check (pattern fullmatch + already-lowercase + JSON well-formedness via python3 when available, best-effort skipped without it) decides whether the on-disk state is already fully valid; anything less falls into the repair path, where a LENIENT salvage extraction preserves a UUID-shaped id from broken or mixed-case state as the repair candidate, and a non-UUID id (email) yields a fresh mint. Repairs rewrite the canonical single-line JSON under the existing cross-runtime `$f.repair.lock` protocol, and events are only ever emitted under the adopted canonical id.

**Regressions** (both surfaces): valid-JSON-with-email → fresh pattern-valid UUID transmitted, never the email, file repaired; malformed JSON containing `3F2504E0-...` → salvaged and lowercased with the file rewritten canonically; uppercase-UUID valid JSON → lowercase canonical returned and persisted; malformed-without-UUID → fresh mint (salvage does not false-positive); cross-runtime convergence — Python adopts the installer's exact single-line shape byte-unchanged, Python-repaired ids fullmatch an independently transcribed copy of the shell regex, and the installer's repaired file matches the fresh-install shape the Python loader accepts verbatim. Existing concurrency/repair/notice suites unchanged and green.

Verification: telemetry suite 63 passed under `-W error::PytestUnhandledThreadExceptionWarning` (3× no-flake), installer suite 65 passed, ruff/format/mypy clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent The fail-open identity blocker from round req_1786318959_1309 is addressed in 3c1e7a472, fail-closed on both surfaces:

**Rule implemented:** an event may only be emitted under an identity that was (a) read from disk as already pattern-valid — in-memory canonicalization of an uppercase UUID counts, since the durable value exists — or (b) freshly persisted AND reread back as valid. When neither holds, telemetry is dropped for that capture; a process-local candidate UUID is never transmitted.

**Python (src/ouroboros/telemetry.py).** Every `return candidate` fallback after a failed publish/reread in `_publish_new_state` and `_repair_state` now returns None (including an `or candidate` fallback hiding inside the repair winner branch); `_load_state` propagates None without caching the failure, so a later capture retries and self-heals if the directory becomes writable; `distinct_id()` returns "" for no-identity; `capture()` resolves identity before anything is queued and drops on empty; `show_first_run_notice` no-ops on a None state rather than claiming notice_shown against never-durable state.

**Installer (scripts/install.sh).** After any publish/repair attempt, the final adopt requires a pattern-valid reread from disk; an empty reread returns nonzero, which `_telemetry_notice`/`_telemetry_ping` already treat as skip-telemetry. The already-fully-valid fast path (strict check) remains the only way to return without a post-write reread. The state write is grouped (`{ printf > tmp; } 2>/dev/null`) so a redirection open-failure in an unwritable dir stays silent.

**Regressions.** Python: your probe inverted — read-only `~/.ouroboros` (0o500, non-root-guarded, chmod restored in finally) with an invalid file → two fresh processes transmit nothing and report empty identity; valid canonical file in a read-only dir → still emits; valid-but-uppercase file in a read-only dir → emits the lowercased id despite the failed rewrite; absent file under an unwritable parent → drops; and failure-is-not-cached — the same process drops while read-only, then emits under a persisted UUID after the dir becomes writable again (TestIdentityFailClosed, 5 tests). Installer: unwritable identity storage → the install itself still completes with zero capture events; valid identity in an unwritable dir → events sent under exactly that id (67 passed).

Verification: telemetry suite 68 passed under `-W error::PytestUnhandledThreadExceptionWarning` (3× no-flake), installer suite 67 passed, ruff/format/mypy clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent The identity-schema parity blocker from round req_1786320200_1324 is addressed in f32e9ed1f:

**Installer structural validation (scripts/install.sh).** The independent sed text-match is no longer the authority for parseable documents. A python3 structural reader (`_telemetry_read_state`) json.loads the file and inspects ONLY the top-level `distinct_id`, classifying the state as: CANONICAL (top-level, pattern-valid, lowercase → adopt without repair), REPAIR (top-level but wrong case → salvage that UUID canonicalized), INVALID (valid JSON without a usable top-level id — missing, non-string, non-UUID, or only nested, like your `{"wrapper":{"distinct_id":...}}` probe → fresh mint, the lenient text search is never consulted, mirroring telemetry.py which salvages only from unparseable text), or MALFORMED (json.load fails → the sed text salvage legitimately applies). Because extraction now goes through json.load, duplicate top-level keys resolve to the LAST occurrence exactly like the Python loader — the old sed first-match divergence is gone too. Without python3 on PATH the prior text-only checks remain as an explicitly documented best-effort fallback (pure shell cannot distinguish nested from top-level), and the fail-closed reread from the previous round still applies to every repair outcome.

**Cross-runtime regressions.** Installer: your exact nested document → install events are emitted under a fresh pattern-valid UUID (never the nested one) and the file is repaired to the canonical top-level shape; duplicate top-level keys (uuidA, uuidB) → events emitted under uuidB; unparseable text containing a nested-looking `"distinct_id": "<uuid>"` → still salvaged. Python (pinned this round, no production change needed — both invariants already held): the same nested document mints fresh and is never text-salvaged (the valid-JSON gate short-circuits before the salvage regex), and the duplicate-key document adopts uuidB with the file left byte-for-byte untouched (no repair write).

Verification: installer suite 70 passed, telemetry suite 70 passed under `-W error::PytestUnhandledThreadExceptionWarning` (3× no-flake), ruff clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent All three blockers from round req_1786321774_1341 are addressed in f52836fd4:

**1. DO_NOT_TRACK project-env authority (src/ouroboros/config/untrusted_env.py) — denied.** `DO_NOT_TRACK` joins the untrusted-project denylist alongside the existing `OUROBOROS_TELEMETRY`/`OUROBOROS_POSTHOG_*` cluster, with a comment documenting the load-order race you demonstrated (project `.env` loads first, existing values are never overridden, so an undenied project value wins against the trusted home file). The single generic `is_untrusted_env_denied_key` gate covers it — no loader change needed. TELEMETRY.md's denied-key sentence now lists it. Regressions: your exact probe as a real import-order subprocess (project `DO_NOT_TRACK=0` + persisted user `DO_NOT_TRACK=1` → env ends `1`, `get_telemetry_enabled() is False`), the direct never-reaches-os.environ case, and denylist-membership pinning so a future edit can't silently drop the key.

**2. Structural notice_shown validation (src/ouroboros/telemetry.py) — fixed, fail-toward-disclosure.** `_validate_state` now coerces `notice_shown`: missing or any non-bool (including the truthy string `"false"` from your probe and even `"true"` — string content is deliberately not inspected, since trusting it would be the same silent-suppression bug) → literal False, persisted through the existing atomic repair rewrite (comparison generalized to whole-dict so either identity or notice repairs trigger it). Regressions: string-"false" → notice prints, state repaired, capture still transmits; string-"true" → still prints; nested-only `{"wrapper":{"notice_shown":true}}` → treated as missing, prints; literal true → no re-print AND the file stays byte-identical (no gratuitous rewrite).

**3. Installer structural notice check (scripts/install.sh) — fixed.** The grep-anywhere skip-check is replaced by a python3 structural read: only a valid JSON dict whose TOP-LEVEL `notice_shown` is the literal boolean true skips the notice; nested, string, non-bool, or malformed states all show it (fail toward disclosure, matching Python), with the prior grep retained solely as the documented no-python3 best-effort fallback. Regressions: your exact nested document → notice shown once, events still emitted under the top-level UUID, top-level flag persisted true afterward; string-"true" → shown; literal true → still skipped.

Verification: telemetry + loader-env suites 274 passed under `-W error::PytestUnhandledThreadExceptionWarning` (3× no-flake), installer suite 72 passed, ruff/format/mypy clean, `bash -n scripts/install.sh` OK.

Q00:
@ouroboros-agent The caller-controlled tool-name blocker from round req_1786322984_1348 is addressed in 788deb20b, at the boundary and with a backstop:

**Boundary (src/ouroboros/mcp/server/adapter.py + telemetry_boundary.py).** Both observation entry paths now emit either the canonical name of a REGISTERED tool or the fixed literal `ouroboros_unknown_tool` — never the caller's raw string. Typed adapter: `call_tool()` computes `registered = name in self._tool_handlers` (the very dict dispatch and list_tools use — no new registry import) and passes it to `observe_adapter_tool_call`, whose exception and success captures both use the sanitized name. SDK path: `call_sdk_tool` already performed a definition lookup but previously reported the raw name regardless; now the name is trusted only after `definition is not None`, so the not-found error itself and everything before the lookup report the literal, while post-lookup failure paths (schema validation, dispatch, output mismatch) keep the verified real name. These are the only two call sites (grepped). Unknown-tool failures are still captured with their existing ok=False/error semantics — just never under a caller string.

**Backstop (src/ouroboros/telemetry.py).** `capture_tool_call` now fullmatches the name against `ouroboros_[a-z0-9_]{1,64}` (verified: all 27 real tool names fit, max 32 chars) and replaces any mismatch with the fixed literal upstream of all command/tool derivation — so even a future boundary bug cannot leak a path or identifier through the one caller-controlled property value. The property-key allowlist constrains keys; this constrains that value's string semantics.

**Regressions.** Your exact hostile name `ouroboros_/home/alice/private-project` through (1) the typed adapter, (2) the SDK route, and (3) the unit backstop — each asserting `tool == ouroboros_unknown_tool`, `command == unknown_tool`, and that the hostile substrings appear NOWHERE in the full serialized event (scanned repr, not just the two fields); charset variants (uppercase, dots, spaces) and an overlong 65-char name are likewise replaced; canonical names (`ouroboros_interview`, a registered mock tool) still capture verbatim with their funnel mappings.

Verification: telemetry + adapter + evaluate-telemetry + loader-env suites 459 passed under `-W error::PytestUnhandledThreadExceptionWarning`; ruff/format/mypy clean across all touched sources.

Q00:
@ouroboros-agent The registered-extension-name blocker from round req_1786323989_1358 is addressed in e07bbf870:

**Audited canonical set, not the registry (src/ouroboros/telemetry.py).** `_CANONICAL_TOOL_NAMES` is a static, audited frozenset of the shipped `ouroboros_*` tool names (36 entries, extracted from the definitions composition and cross-checked against every internal name set) — deliberately NOT the mutable adapter registry, since `register_tool()` accepts arbitrary names. `capture_tool_call` now folds any canonical-shaped name outside that set to the fixed literal `ouroboros_extension_tool` (command `extension_tool`), keeping the distinction from `ouroboros_unknown_tool` (lookup failure vs registered-but-not-ours). Exactly three tool-name values can ever be serialized beyond the audited set's own members — the canonical names, the extension literal, and the unknown literal — and TELEMETRY.md now says so.

**Regressions on every layer.** Your probe — a registered, regex-valid `ouroboros_acme_private_project` — is asserted to fold to the extension literal on (1) the typed-adapter path and (2) the SDK path, both patched at `ouroboros.telemetry.capture` (one layer beneath `capture_tool_call`, so the real fold executes rather than a boundary mock trivially passing), with full-event scans proving `acme`/`private_project` appear nowhere; (3) the unit layer covers the same plus `ouroboros_unknown_tool` pass-through and canonical funnel mappings. Sync guards: `_TOOL_FUNNEL`/`_POLLING_TOOLS`/`_ASYNC_SUBMISSION_TOOLS` are asserted subsets of the canonical set, and a new adapter-side test asserts every tool shipped by the default `get_ouroboros_tools()` composition (32 names) is a member — a future built-in tool missing from the audited set fails CI instead of silently degrading to `extension_tool`.

Verification: telemetry + adapter + loader-env suites 449 passed under `-W error::PytestUnhandledThreadExceptionWarning`; ruff/format/mypy clean.

Q00:
@ouroboros-agent All three blockers from round req_1786324947_1363 are addressed in ee01d353e:

**1. Logical errors no longer count as success (telemetry_boundary.py).** Both observation paths now normalize success across both layers: ok requires `Result.is_ok` AND NOT `MCPToolResult.is_error` (typed path uses a failure-isolated `getattr(..., 'is_error', False)`; the SDK path applies the same check to its final success capture that previously hardcoded ok=True). Your `ouroboros_ralph`-without-lineage_id shape — `Result.ok(MCPToolResult(is_error=True))` — is the regression on both paths, asserting ok=False with error_type=None (nothing raised, so absence is the honest value); normal successes still capture ok=True.

**2. error_type restricted to an audited provenance taxonomy (telemetry_boundary.py).** A single `_safe_error_type` helper now guards every derivation site (typed exception path, typed `Result.err` path, SDK `Result.err` path, SDK exception path — all four enumerated and routed): the class name is verbatim only when its module is `builtins`, a stdlib top-level (`sys.stdlib_module_names`), or an `ouroboros*` package; every third-party/extension class folds to the fixed literal `ExtensionError`. Notably, the investigation showed raised extension exceptions were already wrapped into `MCPToolError` by the adapter — the live leak vector was handlers returning `Result.err(AcmePrivateProjectError(...))` directly, so the regressions build exactly that on both the typed and SDK paths with full-event scans proving `AcmePrivateProject`/`acme` appear nowhere, while `MCPToolError`/`ValueError` stay verbatim. One pre-existing assertion was updated accordingly: jsonschema's `ValidationError` is genuinely third-party and now correctly reads `ExtensionError`.

**3. Durable job types allowlisted (telemetry.py).** `workflow_outcome.command` can only come from a static audited set: the five funnel job types plus the two shipped detached-probe diagnostics (`detached_probe`, `detached_nested_probe` — found by sweeping every `job_type=` literal in src/, included so first-party diagnostics don't mislabel as extensions). Everything else folds to the fixed `extension_job`. `verified` still keys off the raw `job_type == "evaluate"` comparison computed before folding, and a dedicated regression asserts an extension job with `final_approved=true` never earns verified. Your `acme_private_project` probe is the leakage regression (full repr scan). TELEMETRY.md's audited-allowlist note now covers job-derived commands.

Verification: telemetry + adapter + evaluate-telemetry suites 283 passed under `-W error::PytestUnhandledThreadExceptionWarning`; ruff/format/mypy clean.

Q00:
@ouroboros-agent The error-taxonomy blocker from round req_1786326905_1382 is addressed in dda0ca159:

**Closed output vocabulary, no provenance trust (src/ouroboros/mcp/telemetry_boundary.py).** `_safe_error_type` no longer reads `type(error).__module__` at all — both of your attack surfaces are gone by construction. The class NAME is serialized only when it belongs to an audited set assembled from trusted sources on OUR side: every builtin exception enumerated at import from the interpreter's own `vars(builtins)` (71 names — an enumeration CPython ships, not a claim the object makes), the nine literal `ouroboros.mcp.errors` classes (the complete hierarchy, traced against what actually reaches the boundary: not-found, security, verbatim MCPServerError re-raise, and wrap-all branches), plus `CancelledError`/`JSONDecodeError`. Everything else folds to the fixed literal `ExtensionError`. The key property: only audited strings can ever be serialized, so even a class spoofing an audited name (`__module__ = "builtins"`, or naming itself `ValueError`) leaks nothing — the output is a closed vocabulary.

**Total failure isolation.** The helper is wrapped in try/except returning `ExtensionError`, so hostile metaclasses (raising `__name__`, non-string names) can never raise into the command path; and the caller-path audit found one more gap of the same class — the round-13 `getattr(value, "is_error", False)` read only defends against a MISSING attribute, not a hostile property that raises — now wrapped in `_is_logical_error` with the same totality, used at both call sites.

**Regressions** (typed + SDK + direct-unit): spoofed `__module__="builtins"`, the `ouroboros_acme_private.errors` prefix collision that defeated the old startswith check, non-string `__module__` (int) asserting BOTH the fold and that the handler's original `Result.err` object survives untouched (your contract-violation reproduction), and a hostile-`__name__` metaclass against the helper directly. Audited names (`MCPToolError`, `ValueError`) stay verbatim; jsonschema's `ValidationError` keeps folding. Provenance note: the four `type(...).__name__` structured-logging sites in adapter.py were git-blame-verified to predate this PR (#1622, present verbatim on origin/main) — local logs only, never PostHog — and were deliberately left untouched per the PR's scope.

Verification: adapter + evaluate-telemetry + telemetry suites 296 passed under `-W error::PytestUnhandledThreadExceptionWarning`; ruff/format/mypy clean.

Q00:
@ouroboros-agent The dynamic-CLI-command blocker from round req_1786328585_1387 is addressed in 50ff06d99, completing the closed identifier vocabulary across every entry surface:

**CLI commands audited (src/ouroboros/telemetry.py).** `capture_cli_command` serializes a subcommand verbatim only when it belongs to `_CANONICAL_CLI_COMMANDS` — the 27 statically registered built-in `ooo` subcommands, enumerated from the real typer app via `typer.main.get_command(app).commands` (the same conversion the production entrypoint uses; this caught two hidden aliases — `monitor`, `dispatch` — that a source grep would have missed). Anything `_PluginAwareGroup` resolves dynamically (installed plugin names, which commonly identify organizations/projects) folds to the fixed literal `extension_command` with `is_funnel=false`. `_CLI_SKIP` remains a separate product gate, and the funnel tuple/`_CLI_FUNNEL` are asserted subsets of the audited set.

**Actual Click dynamic-dispatch regression.** A new test stages a genuine installed plugin named `acme-private-project` (manifest + lockfile + trust-store, mirroring the plugin-dispatch suite's staging) and invokes the REAL `ouroboros_app` through `CliRunner` so resolution goes through the genuine `_PluginAwareGroup.get_command()` path — the captured event reads `command == "extension_command"` with the plugin name absent from the full event repr. A control test proves a canonical built-in (`status`) still captures verbatim through the same real callback path, and a sync guard asserts every statically registered command is a member of the audited set, failing CI if a future built-in is missed. The unit layer covers the direct hostile-name, canonical pass-through, and subset assertions.

With this round, every caller-influenced identifier channel — MCP tool names (`ouroboros_extension_tool`/`ouroboros_unknown_tool`), durable job types (`extension_job`), error classes (`ExtensionError`), and CLI commands (`extension_command`) — is a closed audited vocabulary, and TELEMETRY.md's allowlist note covers all four in one place.

Verification: telemetry + new CLI-telemetry + existing plugin-dispatch suites 106 passed under `-W error::PytestUnhandledThreadExceptionWarning` (3× no-flake on the new file); ruff/format/mypy clean.

Q00:
@ouroboros-agent Both findings from round req_1786330235_1392 are addressed in f3cf009ee:

**1. Total BaseException isolation (src/ouroboros/mcp/telemetry_boundary.py) — fixed.** `_safe_error_type` and `_is_logical_error` now catch `BaseException`, closing your reproduced escapes (hostile metaclass `__name__` raising `KeyboardInterrupt`; hostile `is_error` property raising through the observer and replacing a completed `Result.ok`). Swallowing these inside sanitization-only helpers is correct — the raise originates from an extension-controlled dunder read, not a genuine interrupt — and the rationale is documented inline. A full sweep of the file's except clauses: the two observation wrappers already caught-and-reraised `BaseException` (correct as-is), and `record_direct_evaluation_outcome` deliberately keeps `except Exception` since its inputs are typed primitives with no hostile-object attack surface. Regressions: `KeyboardInterrupt`/`SystemExit` from `__name__` and `is_error` properties at the direct-unit layer, plus typed-adapter integration — a hostile `is_error` raising `SystemExit` completes the real call with the original `Result.ok` value untouched and `ok=true`, and a raised `KeyboardInterrupt`-subclassed hostile error propagates with unchanged semantics while the captured event folds to `ExtensionError` with the identifying string absent.

**2. CI classification is operator/runtime-owned (src/ouroboros/config/untrusted_env.py) — fixed.** `CI` and `GITHUB_ACTIONS` join the untrusted-project denylist alongside the telemetry-control cluster: a cloned repository shipping `CI=1` in its `.env` could silently deregister genuine local users from the published weekly-active rule's `ci!=true` exclusion. Real CI runners set these in the actual process environment, which the loader never overrides, so genuine CI detection is unchanged — asserted by regression. Tests: neither key reaches `os.environ` from untrusted input (parametrized + membership pinning), your exact integration shape as a real import-order subprocess (project `.env` `CI=1` → env clean and the captured `command_run` carries no `ci` property), and process-env `CI=true` still stamps `ci=true`. TELEMETRY.md's denied-key sentence updated.

Verification: telemetry + loader-env + adapter + evaluate-telemetry + CLI-telemetry suites 519 passed under `-W error::PytestUnhandledThreadExceptionWarning`; ruff/format/mypy clean.

Q00:
@ouroboros-agent The installer serialization blocker from round req_1786336261_1417 is addressed in 7666d9404:

**Structural encoding (scripts/install.sh `_telemetry_ping`).** Unescaped string concatenation is gone. Three layers: (1) `os`/`arch` are token-constrained at the source — fullmatch `[A-Za-z0-9._-]{1,32}` or the literal `unknown` — which defuses your hostile-`uname` probe regardless of encoder availability; (2) with python3 on PATH (the same soft dependency the identity/notice readers use), the ENTIRE payload is built by `json.dumps` with event, distinct_id, api_key, and every key=value passed via argv — never interpolated into the `-c` source — so a hostile value can only ever become the string contents of its own declared property, and keys are defensively whitelisted to `[A-Za-z0-9_]+` (a failing key is dropped, never stringified); (3) the no-python3 fallback pushes every value (including api_key/event/distinct_id) through a strict safe-token filter (`tr -cd 'A-Za-z0-9._-'`, 64-char cap, `unknown` for empty) before concatenation — every legitimate installer value fits that charset losslessly, and no quote/backslash/control character can survive to mutate structure, consistent with the documented NO_PYTHON3 best-effort posture elsewhere in the file.

**Regressions** drive the real `_telemetry_ping` via the extracted-function driver: your exact probe — a PATH-earlier `uname` emitting `Linux\",\"leaked\":\"project-secret` — produces a payload that `json.loads` cleanly with NO `leaked` key and `os == unknown`, on BOTH the python3 path and the forced no-python3 path; and a normal ping asserts the exact declared property structure (parsed JSON, exact key-set match), doubling as the installer-side structural whitelist regression. Installer suite 75 passed, `bash -n` OK, ruff clean.

CI note: all 14 checks are green on the merge-resolved head (the previous module-size failure was fixed in 3caae4aed by moving the PR's additions out of the grandfathered adapter module).

Q00:
@ouroboros-agent The duplicate-key precedence blocker from round req_1786337804_1429 is addressed in aab288446:

**Last-wins within the trusted file, process env still first (scripts/install.sh `_telemetry_load_user_env`).** The root cause was that the per-line `${key+x}` guard — which exists to preserve real process-environment precedence — also fired against the loop's own earlier export, silently flipping duplicates WITHIN the trusted file to first-wins while `dotenv_values` resolves them last-wins. The loader is now three-phase: (1) snapshot which of the four allowlisted keys are already present in the real process environment before the file is consulted; (2) the parse loop accumulates each valid occurrence into a per-key holding variable instead of exporting — overwriting on every occurrence gives dotenv's last-wins for free, with all existing grammar (export prefix, quote/comment semantics, binding skips) unchanged; (3) after the loop, each key's last parsed value is exported only if phase 1 found it absent from the real environment. Your exact reproduction — `OUROBOROS_TELEMETRY=1` followed by `=0` — now disables installer telemetry identically to the application.

**Regressions:** enable-then-disable duplicates for BOTH `OUROBOROS_TELEMETRY` and `DO_NOT_TRACK` → disabled, no notice, no events; the reverse disable-then-enable order → enabled with events (proving last-wins didn't become sticky-disable — no other disabling control is present, so the explicit-enable-is-never-an-override rule doesn't apply); and a real process-env `OUROBOROS_TELEMETRY=0` still beats a duplicated file `=1` (process-env precedence intact). Per-occurrence empty-value skips keep today's semantics — a later valid duplicate still wins — noted as a deliberate scope choice.

Verification: installer suite 79 passed, `bash -n scripts/install.sh` OK, ruff clean.

Q00:
@ouroboros-agent The dotenv-grammar parity blocker from round req_1786338908_1431 is addressed in 720a7c91c, taking the fail-closed branch you sanctioned for grammar the shell cannot faithfully interpret:

**Quoted keys — real parity.** A key wrapped in one layer of matching quotes (`'OUROBOROS_TELEMETRY'=0`, `"DO_NOT_TRACK"=1`) strips to the bare form before the four-key allowlist check, parsing identically to unquoted keys; a half-quoted key still misses the allowlist and is skipped as before.

**Escape-bearing double-quoted values — fail closed.** `dotenv_values()` decodes backslash escapes inside double-quoted values while the shell copies them literally, and a hand-rolled parser cannot faithfully chase that grammar (`\"` alone needs real state tracking). So when an ALLOWLISTED key carries a double-quoted value containing any backslash, the loader refuses to trust either interpretation: it sets an ambiguity flag that `_telemetry_enabled` treats as an opt-out for the entire run — an opt-out the installer cannot resolve is honored, never ignored. The deliberate asymmetry is documented in the code: an ambiguous value that LOOKS enabling (`"1\n"`) also disables the installer run, diverging from the application only in the safe, under-collecting direction — the contract violation is over-collection, and "any one opt-out wins" is preserved unconditionally.

**Full-installer regressions:** your exact probe `OUROBOROS_TELEMETRY="0\n"` → no notice, no events (fail-closed agrees with the application's disabled resolution); `'OUROBOROS_TELEMETRY'=0` and `"DO_NOT_TRACK"=1` quoted keys → disabled via real parsing; `OUROBOROS_TELEMETRY="1\n"` → nothing emitted (ambiguity disables even an apparent enable); control case `OUROBOROS_TELEMETRY="1"` without backslashes → still enabled with events, proving fail-closed triggers on ambiguity only. Installer suite 84 passed, `bash -n` OK, ruff clean.

Q00:
@ouroboros-agent The multiline-binding blocker from round req_1786340056_1435 is addressed in ed7444b60:

**Unclosed quotes are multiline ambiguity, not silent skips (scripts/install.sh).** A quoted value with no closing quote on the same physical line is — to `dotenv_values()` — the start of a valid MULTILINE value the line-by-line shell reader cannot follow. Both unclosed-quote branches (double and single) now set `_TELEMETRY_USER_ENV_AMBIGUOUS` for allowlisted keys instead of skipping the binding, so `_telemetry_enabled` treats the run as opted out — the same fail-closed posture as the escape-bearing values from the previous round: a trusted-file binding the installer cannot conclusively parse is honored as an opt-out, never ignored. The one deliberate asymmetry is documented in the code: trailing garbage after a legitimate closing quote remains a plain skip because dotenv itself rejects that line as a parse error, so both resolvers already agree there; unclosed differs precisely because dotenv keeps reading and we cannot. Stray continuation lines (`"` alone) contain no `=` and fall through harmlessly.

**Full-installer regressions:** your exact probe — `OUROBOROS_TELEMETRY="0` on one line, `"` on the next — produces no notice and no events; same for the multiline `DO_NOT_TRACK="1` form; a multiline value that LOOKS enabling (`"1` / `"`) also stays silent (safe-direction under-collection); and the earlier unclosed-quote-at-EOF test was flipped from skipped-binding to the new fail-closed contract with a comment referencing both rounds. Installer suite 87 passed, `bash -n` OK, ruff clean.

PR #1915, opened 2026-08-06
fix(verification): stop remembering an unreadable extraction as an answer

### LINKED ISSUE #1914: A failed extraction parse is cached, disabling spec verification for the seed permanently
## What happens

`AssertionExtractor.extract` caches by `seed_id` so a seed is extracted once. But `_parse_response` returns an empty tuple in two very different situations:

- the model's reply **could not be read at all** — prose instead of JSON, a truncated array, a JSON object where the array was expected;
- the reply **was read** and legitimately held nothing verifiable.

`extract` caches either one (`src/ouroboros/verification/extractor.py:139`). So one unreadable reply is remembered as the extraction result for that seed, permanently, and every later generation is served from the cache — the debug log even reports it as a cache hit.

## Why it matters

The consumer cannot tell those two empties apart. `_verify_spec_compliance` (`src/ouroboros/mcp/server/adapter.py:2010`) returns `None` for an empty extraction, so the evaluator falls back to the agent's own self-report. One unreadable reply therefore turns spec verification **off for the rest of that seed's life**, silently.

The transport-failure path immediately above (`result.is_err` → `Result.err`) is already retried on the next generation. Only the parse-failure path is permanent, which makes the two inconsistent for no stated reason.

## Reproduction

Generation 1 gets an unreadable reply; generations 2 and 3 get a perfectly good one:

```
generation 1: ok=True assertions=0 llm_calls=1
generation 2: ok=True assertions=0 llm_calls=1
generation 3: ok=True assertions=0 llm_calls=1
```

The LLM is never asked again, and the good response is never seen.

## Suggested remedy

Cache successes, never rejections: have `_parse_response` report "could not read this" distinctly from "read it, nothing here", and decline to cache the former. Expected behaviour:

```
generation 1: ok=True assertions=0 llm_calls=1
generation 2: ok=True assertions=1 llm_calls=2
generation 3: ok=True assertions=1 llm_calls=2   # now a real cache hit
```

A response that was read and offered nothing stays cached, so a seed with no machine-verifiable criteria still costs exactly one extraction.

MyoungSoo7:
Fixes #1914.

## The defect

`_parse_response` returns an empty tuple for two situations that are not alike: a reply that **could not be read at all**, and a reply that **was read** and held nothing verifiable. `extract` caches either one under the seed id (`extractor.py:139`).

So one unreadable reply becomes that seed's extraction result forever. The caller cannot tell the two empties apart — `_verify_spec_compliance` returns `None` for an empty extraction and the evaluator falls back to the agent's own self-report — so spec verification stays off for every later generation of the seed, and the debug log reports it as a cache hit.

The transport-failure path three lines above already retries on the next generation. Only the parse-failure path was permanent.

## The change

`_parse_response` returns `None` when the response could not be read as an extraction at all — no JSON payload, malformed JSON, or a payload that is not the expected array. `extract` declines to cache that and returns the same `Result.ok(())` it always did, so nothing downstream changes shape; the next generation is simply allowed to ask again.

A response that _was_ read and offered nothing is still cached. That boundary is deliberate and pinned: a seed whose criteria are genuinely not machine-verifiable still costs exactly one extraction, not one per generation.

## Verification

Reproduction through the public `extract()` path — generation 1 gets an unreadable reply, generations 2 and 3 get a good one:

|              | before                   | after                                     |
| ------------ | ------------------------ | ----------------------------------------- |
| generation 1 | 0 assertions, 1 LLM call | 0 assertions, 1 LLM call                  |
| generation 2 | 0 assertions, 1 LLM call | **1 assertion**, 2 LLM calls              |
| generation 3 | 0 assertions, 1 LLM call | 1 assertion, 2 LLM calls (real cache hit) |

Reverting only `extractor.py` turns the four new cases red (`prose`, `malformed-json`, `object-instead-of-array`, `nothing-at-all`). With the change, `tests/unit/test_verification.py` and `tests/unit/mcp/server/test_adapter.py` are **124 passed** together.

The existing pinned invariant `test_invalid_json_returns_empty` — malformed response → `Result.ok(())`, no crash — is unchanged and still green.

`ruff check` + `ruff format --check` clean, `mypy src/ouroboros/verification/extractor.py` clean, `module-size: OK (510 modules, cap 2000)`.

### DIFF
--- a/src/ouroboros/verification/extractor.py
+++ b/src/ouroboros/verification/extractor.py
@@ -136,6 +136,17 @@ class AssertionExtractor:
             return Result.err(f"Extraction failed: {result.error}")
 
         assertions = self._parse_response(result.value.content, acceptance_texts)
+        if assertions is None:
+            # The response could not be read at all, which is not an answer
+            # about this seed — it is the absence of one. Remembering it would
+            # answer every later generation from a reply nobody understood, and
+            # the caller cannot tell that empty apart from "nothing here needs
+            # verifying", so spec verification would stay skipped for the life
+            # of the seed. The transport-failure path above already retries;
+            # this one now does too.
+            logger.warning("AssertionExtractor response unreadable, not caching: %s", seed_id)
+            return Result.ok(())
+
         self._cache[seed_id] = assertions
         # LRU eviction: remove oldest entry if cache exceeds max size
         while len(self._cache) > self.max_cache_size:
@@ -146,8 +157,15 @@ class AssertionExtractor:
         self,
         content: str,
         acceptance_criteria: tuple[str, ...],
-    ) -> tuple[SpecAssertion, ...]:
-        """Parse LLM response into SpecAssertions."""
+    ) -> tuple[SpecAssertion, ...] | None:
+        """Parse LLM response into SpecAssertions.
+
+        Returns ``None`` when the response could not be read as an extraction
+        at all — no JSON payload, malformed JSON, or a payload that is not the
+        expected array. That is distinct from an empty tuple, which is a
+        response that *was* read and offered nothing verifiable, and which the
+        caller is right to remember.
+        """
         try:
             # Extract the JSON payload, tolerating markdown fences and prose
             # that surround it (e.g. Gemini-style ``Here is ...`` prefixes).
@@ -157,7 +175,7 @@ class AssertionExtractor:
             data = json.loads(json_str)
             if not isinstance(data, list):
                 logger.warning("Expected JSON array, got: %s", type(data))
-                return ()
+                return None
 
             assertions: list[SpecAssertion] = []
             for item in data:
@@ -247,7 +265,7 @@ class AssertionExtractor:
 
         except (ValueError, KeyError, TypeError, ValidationError) as e:
             logger.warning("Failed to parse extraction response: %s", e)
-            return ()
+            return None
 
 
 def _is_usable_regex_pattern(pattern: str) -> bool:
--- a/src/ouroboros/verification/extractor.py
+++ b/src/ouroboros/verification/extractor.py
@@ -160,11 +160,15 @@ class AssertionExtractor:
     ) -> tuple[SpecAssertion, ...] | None:
         """Parse LLM response into SpecAssertions.
 
-        Returns ``None`` when the response could not be read as an extraction
-        at all — no JSON payload, malformed JSON, or a payload that is not the
-        expected array. That is distinct from an empty tuple, which is a
-        response that *was* read and offered nothing verifiable, and which the
-        caller is right to remember.
+        Returns ``None`` when the response could not be read as an extraction:
+        no JSON payload, malformed JSON, a payload that is not the expected
+        array, or an array that offered assertions and had every one of them
+        rejected. In none of those cases did the model answer in the schema
+        this asks for.
+
+        An empty tuple means the opposite — the array was read and was empty,
+        the model saying there is nothing here to verify. That is an answer,
+        and the caller is right to remember it.
         """
         try:
             # Extract the JSON payload, tolerating markdown fences and prose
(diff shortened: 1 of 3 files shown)


MyoungSoo7:
@ouroboros-agent Round 2. Please supersede review `4873208569` — both findings are addressed at the new head.

## The blocker: readability has to include the entry schema, not just the container

You were right, and the correction is the one you named. I drew the line around the JSON container: array parsed, therefore read. But an array the model filled with entries that all fail validation is a reply in which **nothing arrived in the schema this asks for** — the same permanent silence as prose, reached one level further in.

`_parse_response` now returns `None` when a non-empty array had every entry rejected:

```python
if data and not assertions:
    logger.warning("Every assertion in the extraction response was rejected")
    return None
```

The guard is `data and not assertions`, so the two boundaries you asked for are both held:

- a **genuinely valid empty array** (`data` falsy) still returns `()` and is still cached — a seed with nothing mechanical in its criteria pays for exactly one extraction;
- a **partly valid array** still returns and caches what survived — the rule is about a reply in which _nothing_ arrived, not about every entry being perfect.

## Your probe, reproduced

Same public-path shape you ran: generation 1 gets `[{"ac_index": 0}]`, generations 2–3 get a good reply.

|              | `origin/main`        | `1b0cb93b` (round 1) | new head                              |
| ------------ | -------------------- | -------------------- | ------------------------------------- |
| generation 1 | 0 assertions, 1 call | 0 assertions, 1 call | 0 assertions, 1 call                  |
| generation 2 | 0 assertions, 1 call | 0 assertions, 1 call | **1 assertion, 2 calls**              |
| generation 3 | 0 assertions, 1 call | 0 assertions, 1 call | 1 assertion, 2 calls (real cache hit) |

Round 1 reproduces your result exactly.

## New public-`extract()` coverage for wrong inner schemas

`test_an_array_whose_every_entry_is_rejected_is_not_remembered`, parametrized over the four shapes you listed plus one more, each asserting first `()` → second **1 assertion** → third identity-cached, `await_count == 2`:

| id                                     | payload                                                     |
| -------------------------------------- | ----------------------------------------------------------- |
| `no-tier`                              | `[{"ac_index": 0}]`                                         |
| `tier-that-does-not-exist`             | `[{"ac_index": 0, "tier": "t9_imaginary", …}]`              |
| `ac_index-past-the-end`                | `[{"ac_index": 7, …}]`                                      |
| `item-that-is-not-an-object`           | `["just a string"]`                                         |
| `structural-assertion-with-no-pattern` | `[{"ac_index": 0, "tier": "t2_structural", "pattern": ""}]` |

And `test_an_array_only_partly_rejected_is_remembered_as_what_survived` pins the other side: one good entry between a `no-tier` and an out-of-range one → 1 assertion, cached, **one** LLM call.

## The follow-up: the robustness suite is green

`TestAssertionExtractorFenceRobustness` was red because 16 of its functions assert the old `()` sentinel for responses that are now `None`. All 16 updated to the new distinction — `None` for unreadable, `()` reserved for a valid empty extraction. **319 passed** in that file.

That suite going red is also the honest account of my last round: I ran `test_verification.py` and `test_adapter.py` and not the third file that touches this parser, so I shipped a contract change without the suite that pins the contract. This round the **full** unit suite ran on all three interpreters.

## Verification at the new head

- Reverting only `extractor.py` to `origin/main`: **78 failed** across `test_verification.py` + `test_llm_response_fence_robustness.py`. With the change: **449 passed** across those two plus `tests/unit/mcp/server/test_adapter.py`.
- Full `tests/unit` suite on 3.14: **16699 passed, 87 skipped**. The one failure is `test_wheel_contains_profile_yamls`, where `uv build` cannot reach `pypi.org` from this sandbox (`dns error: failed to lookup address information`) — unrelated to the change, and green in CI.
- All 13 CI checks SUCCESS at `5f1adfca`, including Test Python 3.14 / 3.13 / 3.12.
- `ruff check`, `ruff format --check`, `mypy src/ouroboros/verification/extractor.py` clean; `module-size: OK (510 modules, cap 2000)`.

PR #1916, opened 2026-08-07
feat(evolution): add hidden-checklist convergence loop

Q00:
## Summary

- Keep harness-owned verifier commands and output assertions hidden from every worker path, including direct passive-plugin execution.
- Continue evaluable runs through formal evaluation and explicitly rejected verdicts through bounded Ralph evolution.
- Make the parent own a pollable evaluation/Ralph/Evolve chain with collision-resistant Seed/run lineage identity.

## Changes

- Render worker-safe Seed projections and sanitize retry evidence with one shared contract redactor. Verifier keys are removed, then raw, quoted, and escaped occurrences of their values are redacted longest-first across every remaining worker-visible field.
- Carry `auto_evolve` and an opaque, session-bound Seed handoff through plugin transport; raw Seed paths are never forwarded to workers.
- Reuse the configured production handler graph for passive OpenCode interception while keeping public plugin Ralph delegation separate from the private parent-owned loop.
- Project complete multi-AC evaluation results into generation 1 so passing ACs freeze and failed ACs remain active in generation 2+. Preserve the conservative stage completed by every AC pipeline instead of inventing a rejection stage.
- Use 120 bits of SHA-256 over the complete `(seed_id, session_id)` tuple for a replay-safe lineage identity that fits the production 36-character aggregate-ID contract.
- Publish Gen1 with an atomic event batch behind the existing durable lineage single-writer claim; repair legacy creation-only state under the same claim.
- Elect one Ralph successor owner durably, publish its `job_id` as the claim receipt, and reconnect active or terminal jobs before enqueueing.
- Preserve fail-open run/evaluation verdicts when successor parsing or enqueueing fails, carry the source execution completion status into durable Gen1 replay, and prevent Auto run handoffs from starting a second evaluation or Ralph lineage.

## Test plan

- [x] `uv run pytest -q` — 17,140 passed, 93 skipped, 1 xfailed
- [x] Auto ownership follow-up — 197 focused and 2,550 Auto/MCP adjacent tests passed
- [x] Durability follow-up — 34 Gen1/successor single-writer tests passed after the final adversarial replay adjustment
- [x] Production-width and truthful-stage follow-up — 207 focused chain/evaluation/evolution tests passed
- [x] Overlapping hidden-contract follow-up — 575 parallel-executor tests passed
- [x] `uv run --no-sync ruff check src/ tests/`
- [x] `uv run --no-sync mypy src/ouroboros` — 518 source files
[...]

PR #1851, opened 2026-08-02
feat(orchestrator): adapt concurrency from provider feedback

Q00:
## Summary

- Replace the static AC delivery semaphore with an AIMD window that starts conservatively, halves on backend pressure, and grows after sustained success.
- Admit capacity at every executor-owned provider-call boundary so nested child, decomposition-policy, and coordinator feedback takes effect before the next provider entrance.
- Distinguish concurrency rejection from explicit quota exhaustion so 429 pressure tunes dispatch while quota keeps the existing durable pause/resume authority.
- Preserve outcome quality and completed work: the controller changes provider entrance scheduling only, never acceptance gates or achieved AC results.

## Changes

- Add a cancellation-safe adaptive concurrency controller with pressure epochs, additive growth, multiplicative decrease, and bounded `Retry-After` cooldowns.
- Wrap atomic leaf, SessionSignal follow-up, decomposition-policy, and level-coordinator streams in provider-scoped admission and observation.
- Apply observations before permit release; coordinator pressure delays the next stage, while coordinator quota reaches the runner's PAUSED owner before another stage can cross a provider boundary.
- Project nested rate-limit headers and prevent explicit concurrency-cap failures from being misclassified as quota windows.
- Treat backend `max_concurrency` and `OUROBOROS_MAX_CONCURRENCY` as compatible pre-flight estimates; retain `max_parallel_workers` as the hard adaptive ceiling.
- Persist `aimd/v2` plus `admission_scope: provider_call` in execution semantics so incompatible resume contracts fail closed.
- Align uncertain-sibling diagnostics with the actual provider-effect boundary terminology.

## Validation at exact HEAD

HEAD: `d07630ebdafe39a325304faf9f8545c5674f8e83`

- `1,251 passed, 1 skipped` across adaptive concurrency, backend limits, coordinator, parallel executor, recoverable failure, runner, and routing-resume suites on Python 3.12.12.
- Coordinator `Retry-After: 2` delays the next stage provider entrance by two seconds.
- Coordinator usage quota stops before the next stage provider effect and reaches a four-hour durable PAUSED decision in the runner.
- Full CI command reached `17,262 passed, 15 skipped, 1 xfailed`; the only two local-host failures were a transient version probe against a locally installed Copilot CLI. Both pass with the GitHub Actions-equivalent PATH where Copilot is absent.
- Full Ruff and format checks pass across `src/` and `tests/`.
- Full MyPy passes across 513 source files.
- Module-size, auto-boundary, and `git diff --check` pass.
[...]

PR #1923, opened 2026-08-07
fix(orchestrator): restore post-merge integration gates

Q00:
## Summary

- move retry classification, coaching, and manifest loading from the oversized parallel executor into the existing retry-hints collaborator
- restore the module-size ratchet after independently green #1851 and #1916 consumed the same grandfathered budget
- update #1851 resume-contract tests to satisfy #1898 explicit project-identity intent

## Verification

- `python3 scripts/check-module-size.py --baseline-ref origin/main`
- `python3 scripts/check-auto-boundary.py`
- `python3 scripts/check-max-turns-envelope.py`
- `uv run --no-sync ruff check src tests`
- `uv run --no-sync ruff format --check src tests`
- `uv run --no-sync mypy src/ouroboros`
- focused retry/routing suites: 233 passed
- full suite: 17,562 passed, 93 skipped, 1 xfailed; one local Copilot executable-identity probe failed during the parallel run and passed both isolated and in its full 13-test module

Refs #1797.

### DIFF
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -83,7 +83,12 @@ from ouroboros.harness.deliver_gate import (
 from ouroboros.harness.journal import EvidenceEntry, EvidenceManifest
 from ouroboros.harness.traceguard_validator import validate_evidence_claims
 from ouroboros.observability.logging import get_logger
-from ouroboros.orchestrator import adaptive_concurrency, provider_admission, provider_effect_scope
+from ouroboros.orchestrator import (
+    adaptive_concurrency,
+    provider_admission,
+    provider_effect_scope,
+    retry_hints,
+)
 from ouroboros.orchestrator.ac_execution_capsule import (
     UnmaterializableSuccessContractError,
     bind_capsule_to_runtime_handle,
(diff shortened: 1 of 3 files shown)


Q00 (commented):
Independent verifier verdict: **APPROVE** for exact head `37d79475f17b552d38b75767b3f4249446e98b41` against latest `origin/main` (`703273695ca731da6f0a7f98caa4a0af7f870faa`). No blocking findings.

The retry classification, retryability, assertion-safe coaching/lateral directive, and best-effort manifest loading preserve the prior `parallel_executor` behavior after delegation to `retry_hints`. Import/TYPE_CHECKING boundaries are cycle-safe; manifest failures remain warning-only and fail open; directly used private wrappers remain compatible. The two routing-resume test updates correctly satisfy #1898 by making absent project identity explicit for semantics-only contract fixtures.

Independent verification: focused retry/routing suites 404 passed; Ruff and format passed; MyPy passed (526 source files); module-size, auto-boundary, max-turns-envelope, diff-check, and import smoke passed. The reported Copilot local probe did not reproduce in the complete 13-test module twice serially or under xdist (13 passed each), supporting an unrelated local executable-environment flake. All required CI checks, including full Python 3.12/3.13/3.14 suites, are green.

Direction note: #1797 explicitly excludes `parallel_executor` from its ordered runner extraction phases, so this must not count as phase progress. As a narrowly scoped integration repair restoring the ratchet after #1851/#1916 and moving logic into an existing collaborator, it is nevertheless aligned with the repository modularity direction.

GitHub could not record this as an `APPROVED` review because the active authenticated account (`Q00`) is the PR author; the platform rejects self-approval.

PR #1956, opened 2026-08-09
fix(cli): apply the saved logging.level instead of ignoring it

### LINKED ISSUE #1955: config set logging.level has no effect: saved level is never passed to configure_logging
## Summary

`ouroboros config set logging.level warning` reports success and persists to `~/.ouroboros/config.yaml`, but has no effect on what the CLI actually prints. Every command keeps emitting INFO records to stdout.

## Reproduce

```bash
ouroboros config set logging.level warning
# ╭─────────── Success ───────────╮
# │ logging.level: info → warning │
# ╰───────────────────────────────╯

grep -A1 '^logging:' ~/.ouroboros/config.yaml
# logging:
#   level: warning

ouroboros init start --orchestrator "I want to build a task management CLI tool"
```

Still prints, on stdout, before the first question:

```
2026-08-09T02:01:20.400880Z [info     ] claude_code_adapter.initialized cli_path=... permission_mode=bypassPermissions timeout_seconds=None
2026-08-09T02:01:20.409733Z [info     ] interview.started              initial_context_length=42 interview_id=... is_brownfield=False
2026-08-09T02:01:20.410898Z [info     ] interview.state_saved          file_path=...
2026-08-09T02:01:20.430913Z [info     ] claude_code_adapter.sdk_absent_using_cli cli_path=...
2026-08-09T02:01:30.091484Z [info     ] claude_code_adapter.cli_request_completed content_length=194 session_id=... stop_reason=end_turn
2026-08-09T02:01:30.092520Z [info     ] interview.question_generated   question_length=194 round_number=1
```

## Cause

There are two different `LoggingConfig` classes, and the one the user can set is never handed to the logger.

- `src/ouroboros/config/models.py` — `LoggingConfig.level: Literal["debug","info","warning","error"]`. This is the field `config set` writes and `config show` displays.
- `src/ouroboros/observability/logging.py` — a separate `LoggingConfig` with a different field name, `log_level: str = Field(default="INFO")`, and `handler.setLevel(_get_log_level(config.log_level))`.

Nothing bridges the two. `configure_logging` is only called from the CLI when `--debug` is passed:

```python
# src/ouroboros/cli/commands/init.py
if debug:
    configure_logging(LoggingConfig(log_level="DEBUG"))
    print_info("Debug mode enabled - showing verbose logs")
```

Without `--debug`, the logger keeps the module default (INFO), so `logging.level` from the saved config is never read. The same shape appears in `cli/commands/pm.py`.

## Impact

Two things, one of them user-facing in a way that matters for first impressions:

1. A documented config key silently does nothing, and the success message says otherwise.
2. Because these records go to **stdout** rather than stderr, they interleave with the interview UI. Redirecting (`2>/dev/null`) does not clean it up. Anyone recording a terminal demo, piping output, or just reading the questions gets structured log lines mixed into the prompt text. This is what blocked me from recording a usable `ooo interview` demo.

## Suggested fix

Call `configure_logging` unconditionally during CLI startup, mapping the saved config across the field-name gap, with `--debug` still overriding:

```python
level = "DEBUG" if debug else config.logging.level.upper()
configure_logging(LoggingConfig(log_level=level))
```

Two adjacent things worth considering in the same pass:

- Route these records to stderr so stdout stays clean for piping and recording.
- Give the two `LoggingConfig` classes distinct names, or reuse one, so `level` and `log_level` cannot drift apart again.

Verified against `main` at `afa94ee` with `ouroboros` installed via `uv tool install 'ouroboros-ai[mcp]'`.

Q00:
## Summary

`ouroboros config set logging.level warning` reports success, writes to `~/.ouroboros/config.yaml`, and changes nothing about what the CLI prints. Every command keeps emitting INFO records.

## Cause

Two unrelated classes are both named `LoggingConfig`:

- `config/models.py` — `LoggingConfig.level`. This is the field `config set` writes and `config show` displays.
- `observability/logging.py` — `LoggingConfig.log_level`, which is what reaches `handler.setLevel(...)`.

Nothing bridged the two, and the CLI only called `configure_logging` behind `--debug`:

```python
if debug:
    configure_logging(LoggingConfig(log_level="DEBUG"))
```

Without `--debug` the logger kept the module default (INFO), so the saved level was never read. Same shape in `init.py` and `pm.py`.

## Changes

- New `src/ouroboros/cli/logging_setup.py` — the single seam between the two field names. `resolve_cli_log_level(debug=...)` returns `DEBUG` when the flag is set, otherwise the saved `logging.level`, upper-cased.
- `init.py` and `pm.py` now call `configure_cli_logging(debug=debug)` unconditionally instead of configuring only under `--debug`. The debug banner still prints only with the flag.
- `config` is imported lazily inside the resolver so the `--debug` path does not pay for a filesystem read.
- An unreadable config falls back to INFO rather than aborting the command. Logging setup is not what the user asked for, so it should not be able to take the run down.

## Why it goes through one helper

The two names drifted apart once already and nothing caught it. Routing every `--debug` entry point through one function means a future call site cannot reintroduce the gap by picking the wrong field name.

## Impact beyond the config key

These records go to **stdout**, not stderr, so they interleave with the interview UI and `2>/dev/null` does not clean them up. With this change, `logging.level: warning` gives a clean `ooo interview` session, which is what unblocks recording a terminal demo.

Routing the records to stderr is a larger behavioural change and is deliberately not in this PR.

## Test plan

- [x] New `tests/unit/cli/test_logging_setup.py` (8 cases): saved level is applied; every level survives the field-name gap (parametrized `debug`/`info`/`warning`/`error`); `--debug` overrides and does not read the config at all; unreadable config falls back to INFO; the resolved level reaches `configure_logging`.
- [x] `tests/unit/cli/test_pm_interactive_logging.py` updated to patch `configure_logging` inside the bridge, so it still asserts the effective level (`DEBUG`) rather than just the call.
- [x] `tests/unit/cli/test_logging_setup.py`, `test_pm_interactive_logging.py`, `tests/unit/observability/` — 160 passed.
- [x] `ruff check` and `ruff format --check` clean; `mypy` clean on the three touched modules.

Note on local runs: `tests/unit/cli/test_setup.py` has 21 failures on unmodified `main` on my machine, and `tests/unit/cli/test_status.py` is flaky here with sqlalchemy errors that vary run to run. Both reproduce without this change, so I left them alone and am relying on CI for the authoritative matrix.

Closes #1955

### DIFF
--- a/src/ouroboros/cli/commands/init.py
+++ b/src/ouroboros/cli/commands/init.py
@@ -27,6 +27,7 @@ from ouroboros.bigbang.seed_generator import SeedGenerator
 from ouroboros.cli.formatters import console
 from ouroboros.cli.formatters.panels import print_error, print_info, print_success, print_warning
 from ouroboros.cli.formatters.prompting import multiline_prompt_async
+from ouroboros.cli.logging_setup import configure_cli_logging
 from ouroboros.config import get_clarification_model, get_llm_backend
 from ouroboros.core.errors import ProviderError
 from ouroboros.core.hitl_contract import (
@@ -47,7 +48,6 @@ from ouroboros.events.hitl import (
     create_hitl_answered_event,
     create_hitl_requested_event,
 )
-from ouroboros.observability import LoggingConfig, configure_logging
 from ouroboros.package_profiles import (
     PublicAgentRuntimeBackend as AgentRuntimeBackend,
 )
@@ -982,9 +982,9 @@ def start(
         print_error("Initial context is required when not resuming.")
         raise typer.Exit(code=1)
 
-    # Configure logging based on debug flag
+    # Apply the saved logging.level, with --debug overriding it.
+    configure_cli_logging(debug=debug)
     if debug:
-        configure_logging(LoggingConfig(log_level="DEBUG"))
         print_info("Debug mode enabled - showing verbose logs")
 
     if runtime and not orchestrator:
--- a/src/ouroboros/cli/commands/pm.py
+++ b/src/ouroboros/cli/commands/pm.py
@@ -26,10 +26,10 @@ from ouroboros.bigbang.pm_interview import PM_UNCERTAINTY_GUIDANCE
 from ouroboros.cli.formatters import console
 from ouroboros.cli.formatters.panels import print_error, print_info, print_success, print_warning
 from ouroboros.cli.formatters.prompting import multiline_prompt_async
+from ouroboros.cli.logging_setup import configure_cli_logging
 from ouroboros.config import get_clarification_model, get_llm_backend
 from ouroboros.core.owner_only import secure_directory, write_owner_only
 from ouroboros.core.types import Result
-from ouroboros.observability import LoggingConfig, configure_logging
 from ouroboros.pm.handoff import build_pm_dev_handoff_command
 from ouroboros.providers.factory import (
     create_llm_adapter,
@@ -128,8 +128,8 @@ def pm_command(
     if ctx.invoked_subcommand is not None:
         return
 
+    configure_cli_logging(debug=debug)
     if debug:
-        configure_logging(LoggingConfig(log_level="DEBUG"))
         print_info("Debug mode enabled - showing verbose logs")
 
     console.print("\n[bold cyan]Ouroboros PM Generator[/] - Product Requirements Document\n")
--- a/src/ouroboros/cli/commands/run.py
+++ b/src/ouroboros/cli/commands/run.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
 
 from ouroboros.cli.formatters import console
 from ouroboros.cli.formatters.panels import print_error, print_info, print_success, print_warning
+from ouroboros.cli.logging_setup import configure_cli_logging
 from ouroboros.config.loader import (
     get_config_dir,
     get_max_parallel_workers,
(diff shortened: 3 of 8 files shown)


Q00:
Both findings reproduced and fixed in `ed5ad2c`. Thanks — the prod-mode one was a regression I introduced, and the probe that caught it was the right check.

**Blocker 1 — `OUROBOROS_LOG_MODE` was being discarded.** Correct. `configure_logging()` consults the environment only when called with no config, and my helper always handed it one whose default mode is `DEV`. So the fix for a level regressed the mode.

The seam no longer constructs a config. `observability` now exposes `default_logging_config()` — the config `configure_logging()` would have picked on its own, meaning the active config if logging is already set up, otherwise the environment-derived default — and the seam copies that and replaces `log_level` alone. Mode, `log_dir`, `max_log_days` and `enable_file_logging` all carry through.

Two tests cover it: `OUROBOROS_LOG_MODE=prod` survives applying a level (parametrized over both `--debug` values, and pinning `_current_config` to `None` so the env default is what is actually exercised), and a second asserting the emitted config differs from the base in `log_level` and nothing else.

**Follow-up — `run` bypassed the seam.** Also correct. `run` takes `--debug`, and `_run_orchestrator` imports the orchestrator lazily, so whichever of the seam and that module-level `get_logger()` runs first wins. `configure_cli_logging(debug=debug)` now runs in the `workflow` command body before that call, and the module docstring records the ordering constraint so it does not get moved later by accident.

Covered by a test asserting the seam is called before `_run_orchestrator` is entered.

**Not in this PR:** routing the records to stderr. It is the other half of what makes the interleaving painful, but it changes observable output for anyone parsing stdout today, so it belongs in its own change.

Local: `ruff check` and `ruff format --check` clean across `src/` and `tests/`; `mypy` clean on the touched modules; 164 passed across `test_logging_setup.py`, `test_pm_interactive_logging.py` and `tests/unit/observability/`.

PR #1959, opened 2026-08-09
docs: show the interview running in Quick Start

### LINKED ISSUE #1958: README shows the interview command but never shows it running
## Problem

The README explains the interview in prose and shows the command, but never shows the command running. A reader has to take the central claim on faith:

> `ooo interview "I want to build a task management CLI"`

Nothing on the page demonstrates what makes that different from typing the same sentence into any coding agent. The two things that actually distinguish it are hard to convey in text and obvious in two seconds of video:

1. The questions are specific. They ask about durability, command scope, identifier shape, and concurrency, not "tell me more about your project."
2. The ambiguity score is a real gate. It refuses to generate a Seed when the score is above 0.2, and the refusal is visible.

Point 2 is the load-bearing claim of the whole project, and right now it is only described.

## Proposal

Add a short terminal recording of a real `ooo interview` session to the Quick Start section of all three READMEs, immediately after the command it illustrates.

It should be a genuine session, not a mockup: real questions from the model, real answers, and the real ambiguity score at the end.

## Notes

The recording depends on #1955 being fixed. Before that, INFO records land on stdout between the questions and there is no way to filter them out, so a recording is unusable regardless of how it is made.

Q00:
## Summary

The README showed the interview command and described what it does, but never showed it running. Adds a real recording to the Quick Start section of all three READMEs.

## What the recording shows

`ouroboros init start --orchestrator "..."` — the terminal CLI route, not the `ooo interview` command that runs inside an agent host. The two are different surfaces: `ooo interview` goes through `ouroboros_interview` and layers closure and restate gates on top, while gating on the same `AMBIGUITY_THRESHOLD` of 0.2. The recording carries its own command block and is introduced as the terminal route so the two are not conflated.

25 frames, 76.28 seconds of playback (63.0 seconds of recorded session), 746,505 bytes.

The questions came from the model, not from a script. Four rounds: what the current workflow is and what breaks in it, whether the ordering the user keeps losing is a manual rank or derivable from fields, durability once SQLite makes ordering explicit, and what `list` gives you to name a task with. Those are the assumptions a one-line brief leaves open.

It ends on the part that is hardest to convey in prose: ambiguity is 0.31 against a threshold of 0.2, so the CLI warns and asks whether to continue interviewing, generate a Seed anyway (force), or cancel. The gate is the project's central claim, and until now the README could only assert it.

Nothing is staged. The answers are what a user would type, the score is what the run produced.

## Corrections made during review

Both were real defects in the first revision, verified against the source `asciinema` cast rather than the rendered GIF:

- The recording sat directly under the `ooo interview` Quick Start line while running a different command. Fixed by relabeling and relocating it as the terminal CLI example (`716a668`).
- The captions said 0.28 and "no Seed is generated". The actual final frame reports 0.31 and offers continue / force / cancel. Fixed in all three languages (`716a668`).
- The alt text listed "concurrency" as a topic; the interview never asks about it. Replaced with the topics it does ask about (`551dd5f`).

## Dependency

Recorded against the fix in #1955 (merged as #1956). Before that fix, INFO records land on **stdout** between the questions and cannot be redirected away, which made any recording unusable regardless of how it was produced.

Two other things had to be true and are worth writing down for whoever records the next one:

- The session must run from a directory whose `.claude/settings.json` pins the language, or a global `language` setting makes the nested `claude` CLI write the questions in that language.
- `stty -echo` on the spawned pty, or every answer is printed twice.

## Known artifacts

Two cosmetic marks are visible and were left in rather than edited out:

- Answers appear twice. `stty -echo` is applied twice and the TUI still re-enables echo. This is an artifact of driving an interactive TUI from `expect`; a human typing does not produce it.
- `WARNING: your terminal doesn't support cursor position requests (CPR).` appears a few times. `expect`'s pty does not answer the query. It comes from the nested `claude` CLI, not from Ouroboros.

Neither misleads a reader about what the tool does, and removing them would mean editing the cast into something other than what was recorded.

## Test plan

- [x] All three READMEs render the image at the same anchor, after its own terminal command block and before the runtime-support note
- [x] Relative path `./docs/images/ooo-interview.gif` resolves from the repository root, same convention as the existing hero image
- [x] `alt` text describes the actual content in each language rather than repeating the filename
- [x] 746,505 bytes, smaller than the existing `docs/images/ouroboros.png` (1.4 MB)
- [x] Score, outcome, and question topics in the captions match the source cast

Closes #1958


Q00:
Both findings confirmed and fixed in 716a668. I verified them against the source `asciinema` cast rather than the rendered GIF, so the numbers below come from the recorded stream itself.

**Blocker 1 — the recording is not the command it sat under.** Correct. The asset runs `ouroboros init start --orchestrator "..."` while the adjacent Quick Start line tells readers to type `ooo interview` in their agent host, and `skills/interview/SKILL.md` routes that through `ouroboros_interview` with additional closure and restate gates. Presenting one as the other misrepresented the workflow.

Taking the second option you offered: the recording is now relabeled and relocated as the terminal CLI example. It carries its own `$ ouroboros init start --orchestrator ...` command block introduced with "Or from a plain terminal, without an agent host", and the caption states the relationship explicitly rather than leaving it implied — both paths gate on the same `AMBIGUITY_THRESHOLD` of 0.2, and the host path adds its own closure and restate gates before Seed generation.

**Follow-up — wrong score in the captions.** Also correct, and worse than a rounding slip. The cast tail reads:

```
Ambiguity score: 0.31
│ Ambiguity score (0.31) is too high. Consider more interview rounds to
What would you like to do?
  1 - Continue interview with more questions
  2 - Generate Seed anyway (force)
  3 - Cancel
```

So `0.28` was wrong, and "no Seed is generated" was wrong too — the CLI warns and offers a three-way choice that includes forcing one. All three languages now state 0.31 and describe the actual continue / force / cancel prompt. The threshold of 0.2 was already correct and is unchanged.

On the size note: 746,505 bytes and ~76s of encoded frame delay are the accurate figures; the 63s/729KB numbers came from my harness notes and were not load-bearing in the READMEs, but I have corrected them in the demo notes as well.

Ready for re-review.

Q00:
Follow-up fixed in `551dd5f`, and the PR body is updated.

**Alt text named a topic the interview never asks about.** Confirmed — I pulled the four questions out of the source cast rather than the frames, and none of them touch concurrency:

1. what the current workflow is and what breaks in it
2. whether the ordering being lost is a manual rank or derivable from priority / due date / age
3. durability once SQLite makes ordering an explicit column
4. what `list` gives you to name a task with, since `done` and `rm` both need to name one

All three alt descriptions now name those instead. Screen-reader users get only this text, so it mattering more than the caption is the right read.

**PR body was stale.** Also correct — the description still said 63 seconds, 729 KB, score 0.28, and refusal, which contradicted the asset and my own correction comment. Body now states 25 frames / 76.28s playback / 63.0s recorded session / 746,505 bytes / score 0.31 / continue-force-cancel, describes the recording as the terminal CLI route up front, and carries a short "corrections made during review" section so the history is visible rather than quietly rewritten.

Issue #1917, opened 2026-08-07
RFC: hidden-checklist convergence — run → eval → evolve in one ooo run


Q00:
## Summary

Turn `ooo run` from "execute once and report" into a loop that converges until the seed's hidden checklist fully passes: run → formal evaluation → bounded Ralph evolution, all from a single `ooo run` invocation. Implemented in #1916; this RFC records the design and its decisions.

## Motivation

Two structural problems made single-shot runs unreliable:

1. **Answer-key exposure (reward hacking invitation).** Workers were shown the exact `verify_command` and `Expected output: <assertion>` they would be graded against (`_build_success_contract_block`), and a second leak shipped the assertion `repr()` back through retry prompts via `result.error` tails. A struggling worker's cheapest path was to satisfy the assertion string, not the acceptance criterion (see seed_2be2907edc07 postmortem).
2. **Dead-end failures.** A failed run was never chained into formal evaluation, and a rejected evaluation was never chained into evolution — each piece of the loop existed (verify gate #1591, run→eval chain, `evolve_step`, Ralph driver, `focus.select_evolution_focus`), but nothing connected them. Failures surfaced as BLOCKED instead of being digested.

Owner-confirmed principles:

1. **Unconditional hiding** — workers see only the AC description + expected artifacts. `verify_command` / `output_assertion` are never shown. No config knob for disclosure (explicitly rejected).
2. **Hint loop instead of answer disclosure** — on verification failure, the next round's instructions are corrected from what the session actually did (tool-call trace, evidence manifest, verifier outcome), with assertion strings filtered out of hints too.
3. **run → eval → evolve as one chain** — only failed ACs go to focused re-execution (Gen 2+); passing ACs stay frozen. BLOCKED is the last resort after the evolution budget is exhausted.

### 1. Per-AC answer-key hiding (both leak paths)

- Forward path: `_build_success_contract_block` now renders description + `expected_artifacts` only; harness verifies independently.
- Backward path: the verify-gate failure reason no longer embeds the assertion `repr()`; retry prompts are built from the typed `_VerifyGateOutcome` via a new assertion-safe hint builder (`orchestrator/retry_hints.py`) that reuses the evidence manifest (`deliver_gate.load_ac_evidence_manifest`, read-only) and filters the assertion string from every fragment, including the 2000-char command-output tail.

### 2. Failed runs also chain into evaluation

[...]

PR #1961, opened 2026-08-09
docs(i18n): add a Chinese explanation of hidden-checklist convergence

### LINKED ISSUE #1960: docs/ is English only, including the design people ask about most
## Problem

The repository has three READMEs (English, Korean, Simplified Chinese), but `docs/` is English only. Every design document, including the ones for the most recently landed and most-discussed design, exists in one language.

That gap is widest exactly where it costs the most. The hidden-checklist convergence work (#1917 / #1916) is the part of this project that people ask about: hiding grading inputs from the worker, chaining run into evaluation into a bounded evolution loop, and freezing acceptance criteria that already pass. A reader who arrives at `README.zh-CN.md`, gets interested, and follows the link into `docs/` hits a wall.

## Proposal

Add `docs/hidden-checklist-convergence/README.zh-CN.md`: a Chinese explanation of the same design, covering the motivation, the three confirmed principles, the design, and the guards and decisions.

Cross-link both directions so neither version is a dead end:

- `docs/README.md` lists the Chinese page next to the English entry in the Guides index
- Each of the three English documents links to the Chinese overview from the top

## Scope

One design area, as a first step rather than a translation of `docs/` wholesale. If it proves useful, the same pattern extends to the evaluation pipeline and evolution loop guides, which are the next two most-asked-about.

Q00:
## Summary

`docs/` is English only, so a reader who arrives at `README.zh-CN.md`, gets interested, and follows a link into the design docs hits a wall. This adds a Chinese explanation of the hidden-checklist convergence design and wires it into the navigation both ways.

## Why this design first

It is the one people ask about. Hiding the grading inputs from the worker, chaining run into evaluation into a bounded evolution loop, and freezing acceptance criteria that already pass are the parts that generate questions, and they are also the hardest to convey without the reasoning behind them.

Scoped to one design area on purpose. If it earns its keep, the evaluation pipeline and evolution loop guides are the next two most-asked-about.

## What the page covers

Same structure and same claims as the English trio, not a summary:

- The two structural problems behind unreliable single-shot runs: answer-key exposure through both the forward path and the assertion `repr()` riding `result.error` back into retry prompts, and failure being a dead end
- The three confirmed principles, including that the disclosure-level config knob was explicitly rejected
- The design: per-AC hiding on both leak paths, failed runs chaining into evaluation, and the Gen1 bridge that lets `evolve_step` replay a plain run as generation 1
- Guards and decisions: lineage identity and idempotency, truthful evaluation depth, the no-recursion rule, the `ooo auto` guard, degradations, the silent-evaluation-loss fix, and the `execution.auto_evolve` default change
- What was left out of scope and why

Terminology follows `README.zh-CN.md` so the vocabulary matches what a Chinese reader already saw on the front page. Identifiers, file paths, and config keys stay in their original form.

## Navigation

- `docs/README.md` lists the Chinese page beside the English entry in the Guides index
- Each of `requirements.md`, `architecture.md` and `implementation.md` links to it from directly under the title

Neither direction dead-ends.

## Test plan

- [x] Every relative link resolves: `./requirements.md`, `./architecture.md`, `./implementation.md` from the Chinese page, and `./README.zh-CN.md` from each English document
- [x] Issue and PR references (#1917, #1916, #1591, #174) match the ones the English documents cite
- [x] Technical claims taken from the English documents rather than restated from memory: the 2000-character output tail filter, the 120-bit SHA-256 lineage key, `auto_evolve_max_generations` defaulting to 3 clamped to 1..10, and `execution.auto_evolve` defaulting to true
- [x] English documents unchanged apart from the one added link line

Closes #1960

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -331,7 +331,7 @@ src/ouroboros/
 - **Agent OS runtime** —— 跨能力发现、策略、指令、事件日志、agent 进程的可重放执行契约
 - **Runtime backends** —— 可插拔抽象层（`orchestrator.runtime_backend` 配置），原生支持 Claude Code、Codex CLI、OpenCode、Hermes；同一份工作流规约，跑在不同执行引擎上
 
-完整设计文档见 [Architecture](./docs/architecture.md)。
+完整设计文档见 [Architecture](./docs/architecture.md)（英文）。中文设计说明：[隐藏清单收敛（Hidden-Checklist Convergence）](./docs/hidden-checklist-convergence/README.zh-CN.md) —— run → 评估 → 有预算的 Ralph 链，以及为什么判分用的断言对 worker 无条件隐藏。
 
 </details>
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -49,6 +49,7 @@ replayable execution contract on your choice of runtime backend.
 - [Evaluation Pipeline Guide](./guides/evaluation-pipeline.md) - Three-stage evaluation, failure modes, and configuration
 - [Execution vs. Evaluation Contract](./guides/execution-vs-evaluation.md) - Task completion, AC verdict, and drift terminology boundaries
 - [Hidden-Checklist Convergence](./hidden-checklist-convergence/architecture.md) - Run → evaluation → bounded Ralph chaining with hidden harness grading inputs
+- [Hidden-Checklist Convergence (简体中文)](./hidden-checklist-convergence/README.zh-CN.md) - 同一设计的中文说明
 - [Shared `ooo` Skill Dispatch Router](./guides/ooo-skill-dispatch-router.md) - Runtime setup boundary for Codex CLI, Hermes, and OpenCode skill dispatch
 - [MCP Best Practices](./guides/mcp-best-practices.md) - Upstream MCP server configuration, security, and workflow mapping
 - [QA Backends](./guides/qa-backends.md) - External QA backend patterns, including OpenCron-style synthetic checks
--- a/docs/hidden-checklist-convergence/README.zh-CN.md
+++ b/docs/hidden-checklist-convergence/README.zh-CN.md
@@ -1,0 +1,66 @@
+# 隐藏清单收敛（Hidden-Checklist Convergence）
+
+> 这是设计说明的中文版。完整的英文文档在同一目录下：
+> [requirements.md](./requirements.md) · [architecture.md](./architecture.md) · [implementation.md](./implementation.md)。
+> RFC 原文：[#1917](https://github.com/Q00/ouroboros/issues/1917)，实现：[#1916](https://github.com/Q00/ouroboros/pull/1916)。
+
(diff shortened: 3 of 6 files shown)


Q00:
Both blockers addressed in 9b686ec.

**Blocker 1 — confidentiality boundary overstated (`README.zh-CN.md:23`).** Correct, and it was my omission: `requirements.md:17` and `implementation.md:97` both carry the qualification and the Chinese page dropped it. A Chinese-reading operator could have taken "无条件隐藏" as holdout isolation the runtime does not provide.

Added a dedicated `保密边界的范围（重要）` section stating that the boundary covers the data Ouroboros transports into worker prompts, contexts, events, artifacts, and retry surfaces, that it is **not** a filesystem sandbox, and that a worker with Read/Bash access can still discover a raw Seed left in its workspace — so strong holdout isolation additionally requires placing raw Seeds outside the worker-visible workspace or sandboxing the worker. Principle 1 now narrows 无条件 to mean "no configuration escape hatch" and points at that section; 不在本次范围内 lists filesystem sandboxing explicitly. The section cites the two English paragraphs it derives from.

**Follow-up — discoverability (`docs/README.md:52`).** Also correct: the new page was only reachable from the English index, so the Chinese entry point this PR set out to repair was still a dead end. `README.zh-CN.md:334` now links the Chinese page directly, alongside the English `docs/architecture.md` it already pointed to.

**Suggestion 1 — `直到…全部通过` read as a guarantee.** Fixed. The opening line now says the chain aims at a fully passing checklist but is bounded, and names the five stop conditions (convergence, oscillation, grade regression, wall clock, generation budget) with a pointer to the section that details them.

Ready for re-review.

Q00:
Both follow-ups fixed in `80d1102`, plus the suggestion.

**`README.zh-CN.md:44` — "any run that produced a session".** Correct, and I checked `is_evaluable_run_result()` rather than take it on trust:

```python
status = result.meta.get("status")
return status in {"completed", "failed"} and bool(result.text_content.strip())
```

So it is a terminal `completed` or `failed` result with non-empty text, and `execution_handlers.py` excludes paused and cancelled sessions. The Chinese text now states that definition inline instead of the looser claim, matching the English requirement's "evaluable session/artifact" qualification.

**`README.zh-CN.md:63` — "every rejected run runs up to three focused generations".** Also correct. That reads as a guarantee when Seed parsing or Ralph enqueue can fail before any generation starts, and single-AC evaluation uses full-graph focus rather than focused AC generations — both already listed two bullets above, which made the guarantee wording internally inconsistent as well. It now describes an **attempted** chain, names the two degradations, and states that the preceding verdict remains authoritative when chaining is skipped or fails, which is the actual fail-open direction.

**Suggestion — stop conditions presented as the complete boundary.** Taken. Both the opening line and the Ralph bullet now mark convergence / oscillation / grade regression / wall clock / generation budget as the main exits and note that iteration timeout, cancellation, and terminal evolution actions also end the loop.

Ready for re-review.

PR #1972, opened 2026-08-09
docs: correct the evaluation pipeline guide against the code

### LINKED ISSUE #1971: The evaluation pipeline guide documents a language auto-detection scheme that no longer exists
**Severity:** critical

## Problem

`docs/guides/evaluation-pipeline.md` is the reference for Phase 4, and several parts of it no longer describe what the code does. The largest one documents a mechanism that was removed: a reader who follows the guide configures nothing, gets a Stage 1 that silently passes everything, and has no way to tell from the docs that this happened.

### 1. The "Language Auto-Detection" section describes a scheme that no longer exists

The guide claims Stage 1 commands are "automatically populated by scanning the project directory for known marker files" and gives a ten-row priority table (`uv.lock` → `python-uv`, `Cargo.toml` → `rust`, …).

`src/ouroboros/evaluation/languages.py:1-14` says the opposite in its own module docstring: "Ouroboros no longer ships hardcoded per-language presets." Detection is one AI call in `ouroboros.evaluation.detector` that writes `.ouroboros/mechanical.toml`; `build_mechanical_config()` is only the deterministic reader for that file. There is no marker-file priority list. `detect_language()` survives at `languages.py:329-372` as a deprecated shim that re-reads the toml and emits a `DeprecationWarning`.

Consequence: a user with a `pyproject.toml` expects `ruff check .` / `pytest --cov` / `mypy .` to run. Nothing runs. Every check is skipped and treated as passed (`mechanical.py:300-307`), so Stage 1 reports success while verifying nothing. `ouroboros detect`, the command that actually produces the commands, is not mentioned anywhere in the guide.

The follow-on text inherits the error: the override-priority list still has "Auto-detected language preset" as tier 3, and the failure-mode table says a malformed toml falls back to "the auto-detected preset commands", which cannot happen — every command stays `None`.

### 2. Artifact collection limits: two of three numbers are wrong

| Guide | Code |
|---|---|
| Max files: 30 | No file-count cap exists. `artifact_collector.py:19` calls the char budget the "sole budget limiter" |
| Max file size: 50 KB | `MAX_FILE_SIZE = 100 * 1024` — 100 KB (`artifact_collector.py:20`) |
| Max total: 150,000 chars | Correct (`artifact_collector.py:19`) |

The guide also states oversized files are "skipped entirely (not truncated)". That holds for the per-file limit, but content **is** truncated against the remaining total budget (`artifact_collector.py:216-218`, setting `truncated=True`), so the blanket claim is misleading.

### 3. The consensus trigger matrix is missing its highest-priority entry

The guide documents six triggers. `TriggerType` has seven: `MANUAL_REQUEST` (`trigger.py:37`) is checked **before** all six (`trigger.py:134-148`) and short-circuits the matrix. It is how `trigger_consensus=true` reaches Stage 3. Also, both threshold comparisons are strict `>` (`trigger.py:238`, `trigger.py:260`), and `semantic_result` values override pre-populated context scalars (`trigger.py:235-237`, `257-258`).

### 4. Two documented config knobs have no effect

- `SemanticConfig.satisfaction_threshold` is documented as "Minimum score to approve". The gate at `pipeline.py:248` compares against a hardcoded `0.8`; the field is never read anywhere in `src/`. Setting it to 0.9 does nothing.
- `ConsensusConfig.diversity_required` is documented as "Prefer models from different providers". Nothing reads it either — only the definitions at `consensus.py:142` and `config/models.py:380`.

Per the severity rubric, "wrong environment variable names that silently have no effect" is the same class of defect.

### 5. Stage 2 approval logic is stated unconditionally but is conditional

The guide: `if ac_compliance == False → REJECTED (Stage 3 not attempted)`. `pipeline.py:178` rejects only `if not stage2_result.ac_compliance and not context.trigger_consensus`. With `trigger_consensus=True` the pipeline continues to Stage 3 for a second opinion.

### 6. The reward-hacking veto is undocumented

`pipeline.py:286-291` applies a final gate that flips an approved result to rejected when `reward_hacking_risk >= 0.7` (`REWARD_HACKING_VETO_THRESHOLD`, `models.py:32`). It is the single gate every approval path funnels through, it has its own `failure_reason` string (`models.py:406-411`), and the guide mentions neither it nor the `reward_hacking_risk` field on `SemanticResult` (`models.py:139`). `questions_used` and `evidence` are likewise missing from the scoring-fields table.

### 7. "Stage 2 disabled → Stage 3 implicitly disabled" is backwards

The guide states the pipeline never builds a `TriggerContext` when `stage2_enabled=False`. `pipeline.py:187-194` builds it *outside* the Stage 2 block, with a comment saying this is deliberate "so that trigger_consensus=True works even when stage2_enabled=False". Stage 3 does run in that configuration.

### 8. Stale model identifiers

The Stage 3 examples list `gpt-4o` / `claude-sonnet-4-20250514` / `gemini/gemini-2.5-pro`. The real defaults are OpenRouter-routed (`config/loader.py:126-130`). Undocumented alongside this: when the models are `openrouter/*` and `OPENROUTER_API_KEY` is missing or a `YOUR_…` placeholder, `ConsensusEvaluator` silently falls back to querying one model three times with different role prompts (`consensus.py:87-94`, `301-316`) — same 2-of-3 rule, but no reviewer independence. `SemanticConfig.model` defaults to `None` and resolves by role, not to `claude-3-5-sonnet-20241022`.

## Proposal

Correct the guide against the source: replace the auto-detection section with the `ouroboros detect` / `.ouroboros/mechanical.toml` contract and call out the silently-passing Stage 1, fix the collector numbers, add `manual_request` and the reward-hacking veto, mark the two inert knobs as inert rather than deleting them, and refresh the model identifiers.

## Verification

Every claim above cites the file and line it came from in this repository at `32642276`.

Q00:
**Severity:** critical

## Summary

Brings `docs/guides/evaluation-pipeline.md` back in line with `src/ouroboros/evaluation/`. The guide had drifted far enough that following it produced a Stage 1 that verifies nothing while reporting success.

## The main one

The guide told readers that Stage 1 commands are populated by scanning for marker files, with a ten-row priority table. `languages.py` opens by saying the opposite: "Ouroboros no longer ships hardcoded per-language presets." Detection is one AI call in `evaluation/detector.py` that writes `.ouroboros/mechanical.toml`, and `build_mechanical_config()` is only that file's reader.

So a user with a `pyproject.toml` expected `ruff` / `pytest` / `mypy` to run. Nothing ran — every command was `None`, and a `None` command is skipped and treated as passed, which makes Stage 1 a no-op gate that reports success. `ouroboros detect`, the command that actually generates the commands, appeared nowhere in the guide.

That section is replaced with the real contract: what writes the toml, what validates each proposed command before it is persisted, how to run `ouroboros detect`, and an explicit note that a silently-passing Stage 1 usually means the toml is missing.

## Everything else

- **Artifact collector limits** — the "max 30 files" cap does not exist (the char budget is the sole limiter) and the per-file limit is 100 KB, not 50 KB. Also separates the two behaviors the guide had merged: over the per-file limit a file is skipped entirely, while exhausting the shared budget truncates and sets `truncated=True`.
- **Trigger matrix** — adds `manual_request`, which is checked before all six documented triggers and short-circuits them. It is how `trigger_consensus=true` reaches Stage 3. Notes that both threshold comparisons are strict `>`.
- **Reward-hacking veto** — new section. `_build_result()` flips an approved result to rejected at `reward_hacking_risk >= 0.7`; it never rescues a rejection. Adds `reward_hacking_risk`, `questions_used` and `evidence` to the scoring-fields table.
- **Two inert knobs** — `satisfaction_threshold` is documented as the approval gate but the pipeline compares against a hardcoded `0.8` and never reads the field; `diversity_required` is read by nothing at all. Both are marked inert rather than deleted, so anyone who already set them learns why nothing changed.
- **Stage 2 approval logic** — `ac_compliance=False` rejects only when `trigger_consensus` is false; otherwise the pipeline continues to Stage 3.
- **"Stage 2 disabled → Stage 3 implicitly disabled"** — the trigger context is built outside the Stage 2 block specifically so this is not true. Replaced with the combination that does disable Stage 3.
- **Model identifiers** — Stage 3 defaults are OpenRouter-routed; `SemanticConfig.model` defaults to `None` and resolves by role. Documents the single-model fallback that kicks in when `OPENROUTER_API_KEY` is absent, including the `session/<perspective>` names it leaves in the Stage 3 events.
- Adds a note under the ASCII diagram so it is not read as the complete decision graph.

## Test plan

- [x] Every claim traced to source at `32642276`, cited by file and line in #1971 — no restatement from memory
- [x] `MAX_FILE_SIZE = 100 * 1024` and the absent file-count cap confirmed by reading `artifact_collector.py`, not inferred
- [x] `satisfaction_threshold` and `diversity_required` confirmed unread by grepping all of `src/` — only definitions and schema entries match
- [x] `manual_request` priority confirmed at `trigger.py:134-148`, above the six-check list at `trigger.py:151-158`
- [x] In-page anchors `#reward-hacking-veto` and `#parallel-consensus-failure-tolerance` match the headings they point at
- [x] Docs-only change; no source, no tests, no behavior touched

Closes #1971


Q00:
All three fixed in `bfac5d25`. I checked each against the source rather than taking them on trust, and all three hold.

**Blocker — Stage 1 does not silently skip on a missing toml.** Correct, and it undercut the PR's own point. `build_verification_artifacts()` says so in its docstring:

> Stage 1 commands are read from `.ouroboros/mechanical.toml`. When the file is missing, the AI detector is invoked once to author it (...) Any failure leaves Stage 1 with no commands.

So the missing file is not the trigger; a **failed detection** is. Rewrote the callout at line 98 to name the real causes (no LLM adapter, provider error, unwritable `.ouroboros/`, or a caller driving the lower-level pipeline directly), with the three call sites cited. Split the failure-mode table row into "auto-detection failed" and "detection bypassed" so the symptom maps to an action. Line 147 now says the run path authors the file for you, and that Stage 1 is empty only when the file is absent **and** detection failed **and** no explicit commands are set.

**Follow-up — triggers are not gated on the score.** Also correct. `pipeline.py:178` returns early only for `not ac_compliance and not trigger_consensus`, so a compliant result scoring 0.7 with high drift reaches Stage 3 and can be approved there. Line 240 now reads "after a compliant Stage 2 result" and states that the `score >= 0.8` gate decides only when no trigger matched.

**Follow-up — the veto is not always the reported reason.** Correct. `_build_result()` orders the branches Stage 1 → Stage 3 → Stage 2 AC non-compliance → veto (`pipeline.py:307-326`), so when `trigger_consensus=True` carries an `ac_compliance=False` result to an approving Stage 3 and the veto rejects it, the AC branch wins. Qualified both the prose at line 205 and the lookup row at line 447, and pointed readers at `stage2_result.reward_hacking_risk`.

I did not change the runtime ordering. Surfacing the veto ahead of AC non-compliance is a behavior decision, not a docs one, and this PR is documentation only.

Note for the queue: #1974 (the Chinese translation) is stacked on this branch and is written against the corrected text, so it picks these three up automatically.

Q00:
Round two addressed in `667749ec`. All three verified against the source before changing anything.

**Blocker — the OpenRouter roster is not universal.** Confirmed, and this one had real audit consequences. `_SENTINEL_DEFAULT_BACKENDS` (`config/loader.py:101`) covers Codex, Kiro, Copilot, Hermes, Pi, GJC, Antigravity, Grok, and Zcode, and `_default_model_for_backend()` maps each slot to the literal string `"default"` for those. `_should_use_multi_model()` returns True for anything not starting with `openrouter/` (`evaluation/consensus.py:313`), so the sentinel never reaches the advocate/devil/judge fallback. Stage 3 casts three votes labeled `default`, all from the single model that backend is configured to use.

The doc now splits the section by backend and adds an auditing note stating plainly that three votes labeled `default` are not three independent reviewers, with `OUROBOROS_CONSENSUS_MODELS` / `consensus.models` as the way to get genuine cross-vendor independence there. That felt like the part worth spelling out: an operator reading consensus evidence would otherwise count three voters.

**Follow-up — validation does not prove runnability.** Correct. `_command_is_valid()` (`evaluation/detector.py:496`) checks shell-operator injection, path escape, the executable allowlist, and manifest declaration. It never consults host `PATH`, which the docstring is explicit about. Changed "commands the project can actually run" to "safe, repository-supported commands" and added the concrete failure: a manifest-declared `pytest` persists on a machine without `pytest` and surfaces as "Command not found" at Stage 1.

**Suggestion — MCP call path.** Correct, and it was my error from round one. The handler checks `has_mechanical_toml()` then calls `ensure_mechanical_toml()` and `build_mechanical_config()` directly (`mcp/tools/evaluation_handlers.py:724`); it does not go through `build_verification_artifacts()`. The callout now describes the two routes separately and says the behavior matches while the path does not.

The `CHANGES_REQUESTED` on this PR is against `bfac5d25`; head is now `667749ec`. Ready for re-review.

Note for the queue: #1974 is stacked on this branch and written against the corrected text, so it inherits all of the above.

Q00:
Round three in `827b522e`. The blocker was my own remediation advice being wrong, which is worse than the original gap.

**Blocker — the cross-vendor remediation does not work.** Confirmed. `ConsensusEvaluator` holds a single `self._llm` (`evaluation/consensus.py:281`) and every vote goes through it (`tracked_complete(self._llm, ...)` at `:476`, `:563`, `:886`, `:1030`). The roster only decides the model **string** handed to that one adapter. So telling an operator to set `OUROBOROS_CONSENSUS_MODELS` for cross-vendor independence on a local CLI backend hands that CLI an `openrouter/...` model name; it does not route anything through OpenRouter. They get an unsupported-model failure, or three votes from the same backend wearing different labels, which is the exact failure the note was supposed to warn them away from.

Removed the advice. The section now says plainly that changing the roster does not change the adapter, and that cross-vendor independence requires the active backend to be one that can reach those providers.

I also stopped enumerating which backends have a per-call model flag. You are right that several of those adapters do, and I had not verified it per adapter, so I should not have asserted it. The section now names the set from `_SENTINEL_DEFAULT_BACKENDS` directly and states that OpenCode is not in it.

**Follow-up — `ensure_mechanical_toml()` does not never-raise.** Correct. It returns `False` for the failures it handles, but `_ask_llm()` calls `tracked_complete()` outside an exception boundary and that re-raises adapter exceptions (`evolution/provider_usage.py:443`), and `cli/commands/detect.py:90` does not catch it either. The fail-closed claim is now scoped to returned errors rather than thrown ones, with both call sites cited. I did not change the implementation — whether the contract should be enforced in code rather than documented is a behavior decision and this PR is documentation only.

**Suggestion — "trusts exactly one source".** Taken. Scoped to repository-authored configuration, with a pointer to the overrides section that documents explicit `MechanicalConfig` commands and MCP overrides as the other trusted inputs.

Head is now `827b522e`. #1974 is stacked on this branch and inherits all of it.

Q00:
Round four in `7b77ec29`. You are right and I was wrong in the opposite direction this time.

**Blocker — OpenCode is a sentinel backend.** I checked `_OPENCODE_BACKENDS` at `config/loader.py:113`, saw it was not a member of the `_SENTINEL_DEFAULT_BACKENDS` union, and concluded OpenCode was excluded. I did not check the other direction:

```python
# config/loader.py:76
_CODEX_LLM_BACKENDS = frozenset({"codex", "codex_cli", "opencode", "opencode_cli"})
```

That set **is** in the union, so `get_consensus_models("opencode")` returns `("default", "default", "default")` exactly as Codex does. An OpenCode operator reading my sentence would have expected the OpenRouter branch and then been surprised by three same-backend votes labeled `default` — the precise confusion the section exists to prevent.

Fixed, and I named the trap rather than just the fact: the doc now says OpenCode arrives through the Codex frozenset instead of on its own, and that `_OPENCODE_BACKENDS` is for permission-mode resolution rather than model defaults. That is what misled me, so the next reader gets it stated.

**Follow-up — the `ConsensusConfig` example.** Also correct. Passing an explicit tuple marks the roster as configured and skips `_normalize_configured_models_for_backend`, so someone copying the OpenRouter ids onto a sentinel backend gets none of the behavior the section above describes. Switched to `models=None` with the tuple kept commented and labelled as OpenRouter-capable backends only, plus a comment explaining what `None` buys.

Head is `7b77ec29`. #1974 is stacked and inherits this.

Four rounds on a documentation-only PR is worth noting on its own: the guide had drifted far enough from `evaluation/` that each corrected claim exposed another. I would rather land it correct than fast, so thanks for the repeated passes.

Q00:
Round five in `7f08fc5e`. All four verified before changing anything.

**Blocker — the two categories were not exhaustive.** Correct, and the omission was the worst kind: `claude_code`, `gemini`, `goose`, and `ourocode` are not in `_SENTINEL_DEFAULT_BACKENDS`, so they get the shipped OpenRouter roster, but they are not routed to OpenRouter either. The ids go to their own adapter. Reading my two categories, someone on Claude Code would have matched the OpenRouter branch and concluded three vendors voted.

Added a third section that says what actually happens: without an OpenRouter key simple consensus falls back to the session model, with a key present incidentally it attempts the `openrouter/...` ids through the local adapter, and deliberative mode has no fallback. It closes with the operational rule — the roster in your config is not evidence of who voted, so read the recorded votes.

I also renamed the first category from "OpenRouter-capable" to "OpenRouter-routed". Capability was never the discriminator and that word is what let me write a two-way split in the first place.

**Follow-up — MCP-supplied overrides do not exist.** Correct. `ouroboros_evaluate` exposes no mechanical-command parameter (`evaluation_handlers.py:437`) and the handler calls `build_mechanical_config(working_dir)` with none (`:742`). I documented a public MCP contract that is not there, and implied vetting no MCP path performs. It is now described as a Python API override with both call sites cited, and the priority list says plainly that no MCP tool exposes it.

**Follow-up — "No file paths extracted" is not text-only.** Also correct. `ArtifactCollector.collect()` falls back to `_scan_directory(project_dir)` when extraction returns nothing (`artifact_collector.py:197`), so Stage 2 usually still sees repository files. The row said the opposite, which changes what a reader thinks Stage 2 was even looking at. Rewritten, with text-only scoped to the case where the scan yields no readable files.

**Suggestions.** Both taken. Validation is per command, so one bad command is dropped while the rest persist and `False` comes back only on an empty set. And `docs/config-reference.md:476` still called `diversity_required` enforced while this guide correctly calls it inert — aligned, with a link back here so the two do not drift apart again.

Head is `7f08fc5e`.

Five rounds, and every one found something real. Worth saying plainly: this guide had drifted far enough from `evaluation/` that correcting one claim kept exposing the next. That is the argument for the PR existing, not against it.

Q00:
Round six in `14b814ed`. All three verified against source.

**Blocker 1 — the MCP override claim survived in a third place.** Correct, and this one is on me: I corrected the contract at lines 84 and 124 in round five and missed the security paragraph at 130, so the document contradicted itself inside one page. An operator reading that section still came away believing a privileged MCP command path exists. Scoped to programmatic Python callers, with an explicit "there is no MCP path that does this" and a pointer back to the corrected section.

**Blocker 2 — `satisfaction_threshold` is inert.** Confirmed. It is defined in `config/models.py:359` and `evaluation/semantic.py:85`, and nothing compares against it; `EvaluationPipeline` uses the literal `>= 0.8`. The config reference presented it as the Stage 2 approval control in both the example and the field table, which is worse than silence — it tells an operator they can tune evaluation strictness by changing a value that is never read. Marked inert in both places with a link to the guide, matching how `diversity_required` was handled in round five.

**Follow-up — stale model pin.** Also correct, and the consequence is specific: an explicit roster is preserved rather than normalized on OpenRouter-routed backends, so copying the example pins you to a model that is no longer the default. Replaced **23** occurrences of `claude-opus-4-6` with the SSOT value `claude-opus-4.8` from `config/_model_defaults.py:47`, covering the consensus roster, the `advocate_model` default, `semantic_model`, and the complete-config example.

Head is `14b814ed`.

Separately: `Test Python 3.12` on the previous head reported failure with **zero** failed tests — 4,693 `PASSED`, log truncated at 24%, job continuing ~10 minutes past the last line. Same infrastructure signature as #1952, #1961, and #1984. Reran it.

Six rounds. The pattern across them is consistent: this guide had drifted from `evaluation/` in more places than any single pass could surface, and three of the six findings were errors I introduced while fixing earlier ones. Worth the passes.

Q00:
Round seven. Both blockers confirmed, and blocker 1 is my error from round six.

**Blocker 1 — I collapsed two SSOT identifiers into one.** Round six replaced 23 `claude-opus-4-6` references with the dotted OpenRouter slug. There are two constants, and they are deliberately different shapes (`config/_model_defaults.py:31,47`):

```
DEFAULT_OPUS_MODEL           = claude-opus-4-8                        (hyphenated, Anthropic direct)
DEFAULT_CONSENSUS_OPUS_MODEL = openrouter/anthropic/claude-opus-4.8   (dotted, OpenRouter)
```

`clarification.default_model`, `resilience.wonder_model`, `resilience.reflect_model`, and `evaluation.semantic_model` all take the hyphenated one (`config/models.py:221,339,340,361`). Only `consensus.models` and `consensus.advocate_model` take the dotted slug. Restored the hyphenated form at all 17 direct-provider sites.

The mechanism of the mistake is worth naming: a blanket textual replacement over an identifier that exists in two normalized forms. The old value happened to be correct in shape at every site, so nothing in the diff looked wrong. The non-blocking suggestion about a consistency test is aimed exactly at this, and it is the right instinct — a test asserting documented defaults against `DEFAULT_OPUS_MODEL` and `DEFAULT_CONSENSUS_OPUS_MODEL` would have caught it, where review did not.

**Blocker 2 — the Copilot paragraph.** Confirmed, and worse than stated. The paragraph inherited the corruption from blocker 1 and then made a claim that is false in both directions. Tracing `map_to_copilot_model()`:

- `_STATIC_NAME_MAP` covers `claude-opus-4-6` and `openrouter/anthropic/claude-opus-4-6` (`model_discovery.py:96,101`). The current default is not in it.
- Any id containing a `.` short-circuits at `:276` and returns unchanged.
- The hyphen-to-dot fallback at `:295` runs `replace("-", ".")` over the whole string, so `claude-opus-4-8` becomes `claude.opus.4.8`, which is not a Copilot ID and misses.

So the current default passes through unmapped either way. Rewrote the paragraph to state the actual resolution order and both consequences, and to say what to do instead.

That last point is a runtime gap, not only a documentation one: the fallback can only succeed for single-hyphen IDs, which is why `claude-opus-4-6` needed a hardcoded map entry. Filed separately as #1995 rather than widening this docs PR.

**Follow-up — the full-config comment.** Correct, and it contradicted lines 303-305 in the same file. Rewrote the split as non-sentinel vs sentinel, kept "OpenRouter-routed" for adapters that actually reach OpenRouter, and changed the inline hint from "OpenRouter-capable" to "OpenRouter-routed" to match.

Seven rounds. Rounds three, four, six, and seven each found a defect I introduced while fixing an earlier one. That ratio is the real signal here.

Q00:
Round eight. Blocker confirmed on all six fields. Fixed.

I traced each one rather than trusting the pattern:

| Field | Verification |
|---|---|
| `stage1_enabled` | Only production `PipelineConfig` constructors are `mcp/server/adapter.py:1893` (hardcodes it from `OUROBOROS_EVOLVE_STAGE1`) and `mcp/tools/evaluation_handlers.py:743` (passes only `mechanical` and `semantic`). Neither reads `config.evaluation`. |
| `stage2_enabled` | Same. `adapter.py` pins `True`. |
| `stage3_enabled` | Same. `adapter.py` pins `False`. |
| `uncertainty_threshold` | Stage 3 gates on `TriggerConfig.uncertainty_threshold` inside `PipelineConfig`. Nothing copies the YAML field in. |
| `consensus.min_models` | `config/models.py:378` defines it; no read site anywhere. |
| `consensus.threshold` | Approval is `majority_ratio >= self._config.majority_threshold` (`evaluation/consensus.py:508`), where that config is a different class defaulting to `0.66` (`:141`). The YAML field defaults to `0.67`. |

All six now carry `**Currently inert.**` with the real control named, matching how `satisfaction_threshold` and `diversity_required` were handled in earlier rounds.

I also added a summary note above both sections listing all eight inert fields together. Eight rounds put these markers in one at a time, scattered across two field tables, so the total was invisible — a reader could fix the one field they looked up and still not learn that most of this surface is schema-only.

**On why this kept recurring.** These fields are not inert in a way an operator can detect. They validate (`min_models` enforces `>= 2`, `threshold` enforces `[0.0, 1.0]`), they persist, and they show up in `ouroboros config show`. Every signal says the setting took effect. Someone who sets `stage3_enabled: false` to cut cost gets a clean config and unchanged behaviour, with no error and no log line.

`threshold` is the sharpest: documented `0.67` versus effective `0.66`. Close enough that even someone checking behaviour against the docs would not catch it from the number.

**On the follow-up.** The consistency test is the right instinct and I have filed it as #1998 rather than let it sit in this thread. The point I would add: it should fail in **both** directions — when a field is added and wired to nothing, which is how all eight arrived, and when a field is later wired but keeps its "Currently inert" marker, which is the opposite error and equally misleading. A narrower variant asserting documented defaults against `DEFAULT_OPUS_MODEL` and `DEFAULT_CONSENSUS_OPUS_MODEL` would have caught round six's identifier corruption independently.

Eight rounds on a documentation PR is worth stating plainly: this guide had drifted from `evaluation/` in more places than any single pass could surface, and four of the eight findings were defects I introduced while fixing earlier ones. The mechanical detector in #1998 is the actual fix; everything here is description.

**CI note.** `Test Python 3.12` on the previous head failed with **1 failed, 19,221 passed**, ran to `[100%]`. The failure is `test_setup_claude_serializes_competing_operator_generations`, a 5-second budget for a child interpreter to import `ouroboros.cli.runtime_activation`. It also masks its own diagnosis: the assertion message expression is `first.communicate(timeout=1)` on a child deliberately blocked until released, so `TimeoutExpired` propagates instead of the assertion. Added as a fourth instance on #1994 and reran.

Q00:
## Exact-head source-fidelity verification

Pushed the independently verified, current-main candidate after preserving the remote patch stack with a 10/10 `range-diff =` rebase.

- Exact HEAD: `7d4db09a2bd4dca4ee18d55329b4d1f10802edc8`
- Protected main ancestor: `8ffe9ff1417c07ac5d4fa33a18eb7acc93a218e8`
- Remote prior HEAD preserved: `214a1c6aabd5f138fa82d614386eff101105651b`
- Independent verifier: PASS
- Broad evidence inherited from the exact parent: 1,426 relevant tests + 605 supplemental tests; Ruff, format, MyPy, module-size, links, and CLI help passed
- Exact follow-up: source-fidelity wording, clean worktree, `git diff --check`, ancestry, and 10/10 range-diff passed

The final correction now states that reviewer-independence filtering uses vendor families among **detected installed, runtime-capable backends**, matching `available_runtime_backends()` rather than merely configured vendors.

Merge remains frozen until required CI is green and a fresh official review binds API `commit_id`, body `HEAD checked`, and `Review-Metadata head_sha` to this exact SHA with blocker 0 and an explicit merge recommendation.

Q00:
## Post-merge ancestry audit and exact tree verification

PR #1972 was externally squash-merged as `5fbcc62d19a8b62fc53c3000071500804858a67a` from source HEAD `7d4db09a2bd4dca4ee18d55329b4d1f10802edc8` after `main` had advanced. The source branch therefore did not satisfy the strict latest-main ancestry gate at merge time.

The resulting merge content has now been independently audited against the latest-main rebase candidate `6d5d619ac04874f92e46e708fa2d780643340efd`:

- whole-tree comparison: byte-identical (`git diff --quiet` exit 0)
- PR patch identity: 16/16 stable
- #1999 and #2000 corrections: preserved
- linked #1971 direction: satisfied
- source-fidelity PASS for Stage 1 commands/detection/overrides, trigger ordering, backend availability and reviewer-independence filtering, post-filter tasks/events, successful-vote denominator, artifact fallback/limits, failure precedence, all eight inert config fields/effective controls, distinct Opus defaults, and Copilot mapping text
- focused tests: 1,426 passed
- Markdown rendering and all 16 local links/anchors: PASS
- Ruff/format, MyPy, module-size, and `git diff --check`: PASS

Verdict: the merge process violated the literal latest-main ancestry gate, but the final merged tree is exactly the independently verified current-main candidate and no corrective content change is required. Future stacked/squash merges should still rebase and bind the verifier/review to the actual latest-main source head before merge rather than relying on post-merge tree equivalence.

Issue #1988, opened 2026-08-09
Korean runtime guides: plan for the remaining eleven


Q00:
Follow-up to #1983. #1984 added the Korean TUI guide and #1987 the Korean Claude Code runtime guide. This tracks the rest.

## Where the Korean reader still hits English

`README.ko.md` links twelve runtime guides. One of them is now Korean:

| guide | lines | Korean |
|---|---:|---|
| claude-code | 149 | #1987 |
| codex | **385** | no |
| copilot | ? | no |
| kiro | ? | no |
| opencode | ? | no |
| gemini / goose / grok / hermes / zcode / gjc / antigravity | | no |

## Why codex is next, and why it is not one PR

By mentions in `README.ko.md`, `codex` is the most-referenced runtime (15), ahead of `copilot` (14) and `kiro` (11). It is also the largest guide in the set at 385 lines across 24 sections — 2.6x the Claude Code guide.

That size matters because this is install documentation. A rushed translation of a `Prerequisites` or `Configuration` section does not produce a slightly worse page, it produces a reader with a broken install. The Codex guide in particular carries several passages where precision is load-bearing:

- **Where each setting lives.** `~/.ouroboros/config.yaml` versus `$CODEX_HOME/config.toml` versus the settings UI, and the explicit statement that `~/.codex/config.toml` is *not* where Ouroboros stage model pins belong.
- **What `ouroboros setup --runtime codex` actually touches**, an eleven-item list including which legacy anchors it retires and which user-managed entries it preserves.
- **`--mcp-mode stdio`**, which replaces a user-managed URL entry. Getting the conditional wrong here silently breaks someone's long-running MCP server.
- **Worker subprocess isolation** and the executable version attestation section.

## Proposal

Split by reader journey rather than doing one 385-line pass:

1. **Through install** — Prerequisites, Installing Codex CLI, Installing Ouroboros, Platform Notes, Configuration, Quick Start, Verify Installation. This is where a new reader actually stops.
2. **Operating** — Command Surface, `ooo` skill availability, How It Works, Runtime Differences, CLI Options, Seed File Reference.
3. **Recovery and reference** — Troubleshooting, Cost, Active Conductor and Synapse.

Each part lands as its own PR against the same file, so review stays reviewable and a partial translation never ships as if it were complete.

After codex, the order by measured mentions is copilot (14), kiro (11), then the rest.

PR #1989, opened 2026-08-09
Korean Codex guide, part 1 (through install)

Q00:
Third Korean doc, after #1984 (TUI) and #1987 (Claude Code). Refs #1988.

## Why this is a part, not the whole guide

`docs/runtime-guides/codex.md` is 385 lines across 24 sections, 2.6x the Claude Code guide. It is also install documentation. A rushed `Prerequisites` or `Configuration` section does not produce a slightly worse page; it produces a reader with a broken install.

So this covers the slice that ends where a new reader actually stops: prerequisites, installing Codex CLI, installing Ouroboros, platform notes, configuration, quick start.

## Honesty about coverage

The page says at the top that it is part one and links the English guide for the rest. It closes with a list of exactly which sections are still English, so a reader does not have to guess whether they missed something.

`README.ko.md` still points at `codex.md`, not this file. A partial translation should not be presented as the Korean version of the page.

## Where precision was load-bearing

- **Which file each setting belongs in.** `~/.ouroboros/config.yaml` for runtime settings, `$CODEX_HOME/config.toml` for the Codex MCP/env hookup, the settings UI for everyday model choice.
- **`~/.codex/config.toml` is not where Ouroboros stage model pins belong.** Kept as an explicit warning block rather than folded into prose.
- **`--mcp-mode stdio` replaces a user-managed URL entry.** The default preserves it. Reversing that conditional silently breaks someone's long-running MCP server, so the condition is spelled out.
- **Native Windows.** "Codex CLI itself does not support native Windows" is bolded; a softened rendering reads as a mild caveat.

## Test plan

- [x] `doc_metadata` block byte-identical to the English guide (`runtime_scope: [codex]`)
- [x] Every relative link resolves (checked programmatically, 0 broken)
- [x] Code blocks, YAML keys, and CLI flags unchanged
- [x] English guide links the Korean one with its partial status noted inline
- [x] Section order matches the English guide for the range covered


Q00:
Pushed a restructure before asking for review, having found the same defect the review caught on #1987.

**What was wrong.** `docs/getting-started.md` documents the Codex entry point as:

```bash
codex plugin marketplace add Q00/ouroboros
codex plugin add ouroboros@ouroboros
```

with `ouroboros setup --runtime codex` as the standalone alternative. My part one opened with `Python >= 3.12` in the prerequisites and never mentioned the plugin at all. A reader arriving from `README.ko.md` would have taken the longer path and concluded Python was required just to start.

**What changed.**

- **Recommended path first**: the two plugin commands, what happens when you type `ooo` in a new Codex session (Ouroboros offers to prepare the runtime before changing anything), and the note that **Directly configure models** opens a temporary `localhost` settings page. Prerequisites for this path are Codex CLI installed and signed in. Nothing else.
- **Standalone install is its own section** and owns the `Python >= 3.12` floor and the "base package includes the adapter, no extras" note.
- **`npm install -g @openai/codex` kept as its own section**, since that is how you obtain the CLI in the first place and it applies to both paths.
- The translation-status note at the top now lists the section names as they actually are.

Section list is now: 시작하기 (권장 경로) / 독립 설치 / Codex CLI 설치 / 플랫폼 / 설정 / 빠른 시작 / 이후 절. All relative links resolve.

I checked this one because #1987 taught me the shape of the mistake: these runtime guides each document two installation modes, and translating the page top-down reproduces whichever order the English happened to use rather than the order the reader was sent in.

Q00:
Both blockers confirmed, plus the localization follow-up. All fixed.

**Blocker 1 — `uvx` undisclosed.** Confirmed. `.mcp.codex.json:4` is `"command": "uvx"`, so a reader who satisfies the two documented prerequisites installs the plugin and cannot start the MCP server.

One detail worth stating precisely, because it differs from the Claude side: `.mcp.codex.json` does **not** pin `--python`, unlike `.claude-plugin/.mcp.json`. So the interpreter comes from the package `requires-python = ">=3.12"` (`pyproject.toml:9`) rather than from an explicit flag. The practical answer is the same — plugin users do not install Python themselves — but the mechanism is different, and I documented the Codex one rather than copying the Claude sentence.

(Minor: the review cites `skills/welcome/SKILL.md:694` and `:23`; that file is 566 lines and its only `uvx` reference is at `:535`. The blocker holds on `.mcp.codex.json` alone, so this changes nothing about the finding.)

**Blocker 2 — `codex` on `PATH`.** Confirmed, and this is a genuine self-contradiction I wrote. The path opens with two `codex plugin ...` commands run from a shell, and then the prerequisite offers the macOS ChatGPT app bundle as an alternative to `PATH`. Ouroboros setup can discover that bundle; the shell cannot resolve `codex plugin`. A bundled-only reader follows the instructions in order and fails on line one.

Made `PATH` mandatory for this path and routed bundled-only users either to putting the executable on `PATH` or to the standalone path, which does work for them.

**Follow-up — the option label.** Correct. Localized onboarding shows `직접 모델 설정하기`, and `tests/unit/skills/test_skill_artifacts.py:253` asserts that exact string. A Korean reader looking for "Directly configure models" would not find it on screen. Now shows the Korean label with the English in parentheses, since the guide also links to English documents that use the English name.

**Same two defects existed in the English guide.** `codex.md` never mentioned `uvx` at all and listed `Python >= 3.12` unscoped, and `getting-started.md` "Codex first use" had neither. So the English source advertised a requirement the plugin path does not have while omitting the one it does. Fixed both.

This is the same defect pair I just fixed on the Claude side in #1987, and the cause is the same in both: the prerequisites were written for the standalone install and never revisited when the plugin became the recommended path. Two runtimes, one stale assumption. Worth checking `copilot.md` and `kiro.md` for it before I translate those (#1988).

Q00:
Pushed a fix for a defect the review has not raised here yet, because #1987 just raised the identical one on the Claude guide.

The recommended path opened with `codex plugin marketplace add` at line 27 and disclosed both prerequisites — `codex` on `PATH` and `uvx` — at line 37. Both are needed by the command ten lines above them. The reader who lacks either fails on the first command and never reaches the section that would have told them.

That is exactly the finding on #1987, where I had fixed the ordering in `getting-started.md` and left it broken in the guide that PR exists to add. Rather than wait for the same review to arrive here, both requirements now sit above the first command with the reason and the uv install options, and the prerequisite entry below back-references them instead of repeating the block.

Head is `39504b5`.

Q00:
Confirmed, both the blocker and the follow-up. Fixed.

**Blocker — the launcher is required on the standalone path too.** I scoped `uvx` to the marketplace plugin and then wrote that standalone needs only Python and the base package with "no extras". That combination cannot finish `ouroboros setup --runtime codex`.

`_codex_release_mcp_launcher()` (`cli/commands/setup.py:613`) resolves in order: `uvx`; an `ouroboros` install whose `mcp serve --help` exits 0, which means the `[mcp]` extra; or a Python environment containing `mcp`. With none of the three, `_register_codex_mcp_server()` aborts with `Install uv, or install Ouroboros with the [mcp] extra.`

The specific damage is that I *strengthened* a false claim while fixing something else. "extras는 필요 없습니다" was true about the runtime adapter — the base package does carry it — and I let that carry over to MCP registration, which it does not cover. Both guides now state the three-way requirement with the resolution order and the exact abort message, and say plainly that `[mcp]` is optional only when `uvx` is present.

**Follow-up — the Korean standalone section was ordered backwards.** Also correct, and worse than the ordering defect I fixed two rounds ago: it showed `ouroboros setup --runtime codex` *before* the prerequisite block and before telling the reader to install Ouroboros at all. A reader routed there from line 45 precisely because bundled Codex is not on `PATH` runs the first command they see and gets `ouroboros: command not found`.

Restructured so prerequisites and install commands come first, then setup.

**Follow-up 2 — the ambiguous back-reference.** "위 권장 경로의 사전 조건 두 가지" was ambiguous in a way that mattered: the recommended pair is `codex` on `PATH` plus `uvx`, and this path exists *for readers who do not have `codex` on `PATH`*. So the back-reference imported a requirement the section was written to relax. Replaced with an explicit standalone list, per the suggestion — that also means a future edit to the recommended path cannot silently change what standalone requires.

Head is `dbbfa0b`.

Q00:
All three confirmed. Fixed.

**Blocker — one of the two documented branches did not run.** Correct. I wrote the block while thinking about *launchers* and lost track of the fact that one of the two commands also happens to install the `ouroboros` executable and the other does not. `pipx install uv` gives you `uvx` and nothing else, so the very next line — `ouroboros setup --runtime codex` — is `command not found`. Only `pip install 'ouroboros-ai[mcp]'` covered both, and only by coincidence.

Split into three numbered steps: install Ouroboros, obtain a launcher, connect. With an explicit note that the first two are independent and that the `[mcp]` form collapses them, since that coincidence is exactly what misled me.

**Follow-up — `codex --version` contradicts the path it belongs to.** Also correct, and this is the sharper of the two. The standalone path exists *for* users whose only Codex executable is the macOS ChatGPT bundle and is not on `PATH` — lines 45 and 53 say so. Those users complete setup successfully, run the verification block, and get `codex: command not found`. **The guide tells a working installation that it is broken**, which is worse than omitting the check.

Qualified it: `command not found` is not a failure on this path, and the way to verify is `ouroboros config show | grep codex_cli_path`, which shows what setup actually resolved. Also reordered so `ouroboros --help` comes first, since that one is unconditional. `codex --version` remains the right check on the plugin path and for anyone who put `codex` on `PATH`.

**Suggestion — "no extras are required".** Taken. That sentence was true about the runtime adapter and I let it read as a claim about the whole standalone install, which is the same conflation as the blocker I fixed last round. Scoped it to the adapter with a pointer to the prerequisite that governs MCP registration.

Head is `43fd8e2`'s successor.

Q00:
Both confirmed. Fixed.

**Blocker — the replacement diagnostic does not work.** Correct. Last round I replaced `codex --version` with `ouroboros config show | grep codex_cli_path` for the macOS App-only path. That grep matches nothing: the command renders the resolved executable under the label **`CLI path:`** (`cli/commands/config.py:696-701`), and the string `codex_cli_path` never appears in the output at all.

What makes this the worst version of the mistake: I proposed a diagnostic **without running it**, on the one path that exists precisely because the obvious check does not work there. So a bundled-executable user would have followed `codex --version` → nothing, then my replacement → also nothing, and concluded twice that a working install is broken. I made the failure I was fixing worse.

Now says to run `ouroboros config show` and read the `CLI path:` line, with an explicit note not to grep for `codex_cli_path`, since that is exactly what I did.

**Follow-up — the abort message.** Also correct, and in two ways. The real string is `Could not find a launchable Ouroboros MCP command. Install uv, or install Ouroboros with the [mcp] extra, then rerun setup.` (`cli/commands/setup.py:1198`) — I abbreviated it and dropped both the prefix and the `then rerun setup` suffix.

The Rich detail is the more useful half: `print_error` renders through Rich, which consumes `[mcp]` as markup, so the terminal actually shows **`with the extra`**. Someone searching the docs for the message they saw on screen would not find the version I wrote, and would not find the source string either. Both languages now quote the full source string and state that the rendered output differs.

Head is `4d9d1d8`.

Issue #1170, opened 2026-05-22
Meta SSOT slice: L0 — Canonical Test Harness for ooo auto acceptance


shaun0927:
# Meta SSOT slice: L0 — Canonical acceptance test (minimal)

> **Implementation cleanup (2026-05-23 KST).** L0 is partially implemented: #1174 merged the canonical harness skeleton/initial scenario, and #1191 merged opt-in live wiring plus L1 catalog cross-validation. Treat remaining scenario additions and #1195 evidence alignment as the open work; nightly/replay/cost-budget sections below are superseded v2 ideas unless new evidence reopens them.
>
> **Scope confirmation (2026-05-27 KST).** L0-b/c/d scenarios (`webhook-receiver`, `vertical-slice-refactor`, `2d-kart-racer`) are **removed from this issue's close conditions** and deferred as evidence-driven follow-ups. The "Initial canonical scenarios" table, the "Sub-PR breakdown" L0-b/c/d entries, and the original 4-scenario acceptance criteria are superseded by the cli-todo-only acceptance gate documented under *Acceptance criteria* and *Sub-PR breakdown* below. Trigger to close: cli-todo R2/R3 live-run reproduction on `main` HEAD. Rationale: #1157 *Freshness sync* 2026-05-27 entry — #1219 R1 confirmed cli-todo alone surfaces substrate defects across multiple TaskClasses, so additional fixtures are not load-bearing on the acceptance gate.

This issue is the **L0 design slice** of #1157. It provides the smallest possible *acceptance gate* that the SSOT itself can point at to declare close-readiness.

## Why this is small on purpose

The earlier draft of L0 proposed a nightly CI workflow, a recorded-replay layer for hermetic mode, a `divergence_count` cross-check job, a monthly LLM-cost budget, and a refresh-rotation ownership policy. **All of that is operational sludge, not `ooo auto` behavior.** L0's *only* job is: *"give the SSOT something concrete to point at when claiming the canonical test matrix passes."*

Per Ouroboros's minimal-substrate principle (*"add substrate only when evidence demands it"*), L0 v1 ships the smallest thing that satisfies the SSOT acceptance gate:

- A `tests/canonical/` directory with one self-contained scenario per canonical goal.
- A `pytest` entry point that drives `ouroboros_auto` end-to-end against a scenario.
- A maintainer runs it *manually* when assessing SSOT readiness.

[...]

PR #1992, opened 2026-08-09
Korean Codex guide, part 2 (Command Surface through Seed reference)

Q00:
Stacked on #1989. Continues the Korean Codex runtime guide from where part 1 stopped.

## Translated

Command Surface (the 11 things `ouroboros setup --runtime codex` touches, the `~/.codex/config.toml` boundary), worker subprocess isolation, the full `ooo` skill table with CLI equivalents, executable version attestation, Codex CLI strengths, runtime differences vs Claude Code, CLI options, and the seed file reference.

Kept the Ralph note prominent, since `ouroboros cancel execution` silently does not cancel a Ralph job ID.

## Corrections to the English source

Verifying each claim before translating turned up three real drifts in the seed reference table:

1. `task_type` listed `code`/`research`/`analysis`. The model accepts seven values — `artifact`, `document`, `documentation`, and `presentation` were missing (`core/seed.py:711`).
2. `metadata.ambiguity_score` was marked **Required: Yes** and **"Must be <= 0.2"**. Neither is true: `SeedMetadata.ambiguity_score` is `Field(default=0.15, ge=0.0, le=1.0)` (`core/seed.py:409`). The 0.2 threshold is enforced at **generation**, and that gate has an explicit opt-out — the CLI's "Generate Seed anyway" and the MCP `force` parameter (`mcp/tools/authoring_handlers.py:1295`). `ouroboros run workflow` does not re-enforce it; `ouroboros auto` re-checks readiness at `auto/grading.py:226`.

   This one misleads in a specific way: a reader who hand-writes a seed and sets a high score expects to be stopped, and is not. The field is provenance, not a gate. Documented that in both languages.
3. `brownfield_context` and `metadata` were absent from the table entirely.

## Remaining

Part 3 covers Troubleshooting through Synapse. Tracked in #1988.

### DIFF
--- a/docs/runtime-guides/codex.ko.md
+++ b/docs/runtime-guides/codex.ko.md
@@ -329,7 +329,10 @@ uv run ouroboros run workflow --runtime codex --resume <session_id> ~/.ouroboros
 
 > **`ambiguity_score`의 0.2 임계값이 실제로 걸리는 지점:** 이 필드 자체는 `0.0`~`1.0`을 허용합니다([`core/seed.py:409`](../../src/ouroboros/core/seed.py)). 0.2 게이트는 **seed 생성 시점**에 걸립니다. 인터뷰가 0.2 아래로 못 내려가면 seed를 안 만들어 주는 것이고, 여기에는 명시적 우회 선택지가 있습니다. CLI의 "Generate Seed anyway", MCP의 `force` 파라미터입니다. 우회해도 **실제 점수는 그대로 메타데이터에 기록되고 감사 로그에 남습니다.**
 >
-> `ouroboros auto`는 실행 중에 readiness를 다시 보긴 하지만 **조건부입니다.** 인터뷰가 ledger 근거로 닫혔거나(`closure_mode`가 `ledger_only`·`safe_default`), Seed가 `degraded`로 표시돼 있으면 `high_ambiguity_score` 블로커가 억제됩니다([`auto/grading.py:225-226`](../../src/ouroboros/auto/grading.py)). 그런 닫힘에서는 **ledger의 구조적 완결성이 수용 신호이고 LLM이 매긴 점수는 설계상 낡은 값**이라서, 0.2를 크게 웃도는 Seed도 A 등급을 받고 실행될 수 있습니다. 나머지 채점 축은 그대로 적용됩니다.
+> `ouroboros auto`는 실행 중에 readiness를 다시 보긴 하지만 **조건부이고, 억제되는 두 경우의 결과가 정반대입니다** ([`auto/grading.py:225-226`](../../src/ouroboros/auto/grading.py)):
+>
+> - **ledger 닫힘**(`closure_mode`가 `ledger_only`·`safe_default`이고 degraded가 아닐 때): **ledger의 구조적 완결성이 수용 신호이고 LLM이 매긴 점수는 설계상 낡은 값**이라, 0.2를 크게 웃도는 Seed도 A 등급을 받고 **실행됩니다.** 나머지 채점 축은 그대로 적용됩니다.
+> - **degraded Seed**: 블로커가 억제되는 건 **부분 산출물을 내보내기 위해서일 뿐입니다.** 블로커가 없는 degraded Seed는 **등급이나 `may_run`과 무관하게** 곧바로 partial-product 종단(`AutoPhase.COMPLETE`)으로 갑니다([`auto/pipeline.py:1286`](../../src/ouroboros/auto/pipeline.py)). **RUN에는 절대 도달하지 않습니다.** 남아 있는 블로커는 하드 안전 블로커라 그대로 실행을 종료시킵니다.
 >
 > 실무적으로: 손으로 쓴 seed에 높은 `ambiguity_score`를 박아도 `ouroboros run workflow`가 막지는 않습니다. 이 값은 **강제 차단 장치가 아니라 출처 기록(provenance)**입니다.
 
--- a/docs/runtime-guides/codex.ko.md
+++ b/docs/runtime-guides/codex.ko.md
@@ -7,7 +7,7 @@ doc_metadata:
 
 > English: [codex.md](./codex.md)
 >
-> **번역 진행 상황**: 이 문서는 설치까지(시작하기 · 독립 설치 · Codex CLI 설치 · 플랫폼 · 설정 · 빠른 시작)를 옮긴 1부입니다. 그 뒤 절(Command Surface, How It Works, CLI 옵션, 문제 해결, 비용, Active Conductor)은 아직 영문입니다 — [codex.md](./codex.md)를 보세요. 진행은 [#1988](https://github.com/Q00/ouroboros/issues/1988)에서 추적합니다.
+> **번역 진행 상황**: 명령 표면 · 동작 방식 · 런타임 차이 · CLI 옵션 · Seed 레퍼런스까지 옮겼습니다. 문제 해결 · 비용 · Active Conductor는 아직 영문입니다 — [codex.md](./codex.md)를 보세요. 진행은 [#1988](https://github.com/Q00/ouroboros/issues/1988)에서 추적합니다.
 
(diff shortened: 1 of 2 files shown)


Q00:
Blocker and all three follow-ups confirmed. Fixed.

**Blocker — `ouroboros auto` does not re-check uniformly.** Correct, and this is the sharpest kind of error I can make: I added that sentence last round *while correcting* an over-strict claim about the same threshold, and overshot in the opposite direction.

`auto/grading.py:225-226` suppresses `high_ambiguity_score` when `closure_mode` is `ledger_only` or `safe_default`, or when the Seed is `degraded`. The comment above it explains why — under ledger closure the structural completeness of the ledger *is* the acceptance signal, so the LLM-derived score is stale by design. It even cites the regression that motivated it (#1170 R2, `cli-todo` terminating BLOCKED at `ambiguity_score=0.467` despite `ledger_only` closure).

Stating the re-check without the exceptions was the more dangerous half of the sentence: a reader trusts Auto to catch what generation let through, and under the two most common ledger closures it does not. Both languages now name the three suppression conditions and say plainly that a Seed well above 0.2 can grade A and run.

**Follow-up 1 — the skill table was incomplete.** Correct. The table said it showed each skill and listed 16. Setup packages 21 `skills/*/SKILL.md` directories (I count 21, not the 22 in the review). Missing: `auto`, `brownfield`, `config`, `pm`, `resume-session`, plus bare `ooo`.

`ooo auto` is the material one — the managed Codex rule routes it to `ouroboros_start_auto` (`codex/ouroboros.md:16`) and instructs the agent to stop rather than emulate it if that tool is unavailable. Leaving the repo primary autonomous entry point out of the command surface was the worst single omission in the table. It now covers all 21 directories exactly; I verified by diffing the table rows against `ls .claude-plugin/skills/` in both directions, with no residue on either side.

**Follow-up 2 — global artifacts.** Correct. The list covered `~/.codex/` only. `ensure_config_dir()` creates `~/.ouroboros/data/` and `~/.ouroboros/logs/` (`setup.py:2632`), and a fresh config writes `~/.ouroboros/credentials.yaml` at `0600` (`setup.py:2771`). Added, with the mode, since a file written outside the runtime directory is exactly what someone auditing setup wants to know about.

On the count: my PR body said setup touches "11 things" while the list has nine bullets. That number was wrong in the PR body, not the document — the English list has nine and I did not add to it. The document now discloses more than the English source does.

**Follow-up 3 — authentication.** Correct, and self-contradictory within one file: the prerequisites call a signed-in Codex account the primary option with API keys as an alternative, while the comparison table said API key only. The capability matrix already had both. Fixed in both languages.

**Non-blocking — the banner.** Correct; it still described sections this PR translates. Updated here, and #1993 replaces it with a completed-translation note. I rebased #1993 onto this head and resolved that one-line conflict in favor of the completed note, so all four fixes above carry through the stack.

Q00:
## Strict merge hold — APPROVED metadata contradicts the exact review body

Do not merge current HEAD `42b0a2e4380d804fa1449d453eca0288a0861db2`.

The exact official review `4892748750` ends with **Request changes**, not a merge authorization. The source-backed findings remain:

- “full 21-skill command surface” is inaccurate; the exact snapshot has 22 including bare `ooo`;
- `ooo pm` and `ooo qa` are incorrectly marked MCP-only;
- brownfield and resume mappings overstate direct CLI equivalents;
- the cost text overstates API billing despite supported account/OAuth paths;
- the English partial-translation banner and a source anchor remain stale follow-ups.

This PR is stacked, has no required 6-check run, diverges from main, and inherits the current `CHANGES_REQUESTED` state of its parent #1989. `APPROVED` API state and an empty blocker table do not override the final review recommendation. Correct/rebase the stack, run independent exact verification and required CI, and obtain a fresh exact-head review that explicitly recommends merge.

Q00:
Both blockers confirmed. Fixed.

**Blocker 1 — I counted the wrong directory.** Correct, and the mechanism is worth stating: I verified my table against `.claude-plugin/skills/` (21 entries) while documenting the **Codex** surface, which is packaged from `skills/` (22, including bare `ooo`). Two skill trees, and I checked the one belonging to the other runtime.

What makes that especially annoying is that I *did* run a bidirectional diff last round and reported zero residue — against the wrong source. A verification pointed at the wrong reference is worse than none, because it produces a confident wrong claim.

Re-verified against `skills/` this time, both directions: 22 rows, 22 directories, nothing on either side. `resolve_packaged_codex_assets()` is now cited by name for where the number comes from.

**Blocker 2 — five wrong terminal equivalents, four of them claiming MCP-only for registered commands.** All confirmed against `cli/main.py`:

| Skill | Was | Actually |
|---|---|---|
| `ooo brownfield` | *(MCP only)* | `ouroboros setup scan` / `setup list` / `setup default` |
| `ooo config` | `ouroboros config show` | `ouroboros config` — `config show` is read-only output, a different thing |
| `ooo pm` | *(MCP only)* | `ouroboros pm` |
| `ooo qa` | *(MCP only)* | `ouroboros qa` (`main.py:120`) |
| `ooo resume-session` | `run workflow --resume <id>` | `ouroboros resume`, which lists in-flight sessions; `run workflow --resume` skips that discovery |

The "MCP only" errors are the damaging ones. This table is positioned as the terminal contract, so a terminal-only reader was told four capabilities were unreachable from their shell when each has a registered command. That is worse than an omission — it actively closes off a working path.

**Follow-up — the cost row.** Correct, and it contradicted two things at once: the authentication row directly above it (which I corrected to "account sign-in or API key" last round) and the guide's own Cost section, which says billing follows whichever path Codex CLI is configured for. Made conditional in both languages; the English table had the same defect.

**On the P1 suggestion.** A contract test deriving skill names from `resolve_packaged_codex_assets()` and checking both guide tables is exactly right, and this round is the argument for it — I hand-verified, was confident, and was wrong because I aimed the check at the wrong directory. A test cannot be aimed at the wrong directory. I have added it to #1998, which tracks the same "no mechanical relationship between docs and source" problem for config fields; the CLI-mapping half is a natural companion.

Head is `f9f6857`.

Q00:
Both confirmed. Fixed.

**Blocker 1 — I fixed the Korean table and left the English one.** Correct, and it is the same partial-fix pattern I have now hit four times this session. Worse here than usual: the two public guides ended up **contradicting each other**, so a reader comparing them could not tell which was right, and the English terminal user was still being told a registered command is unavailable (`cli/main.py:120`).

While in there I verified the five remaining `*(MCP only)*` rows rather than assuming the rest were fine — `evaluate`, `evolve`, `unstuck`, `tutorial`, `welcome` are all genuinely unregistered in `main.py`. So `qa` was the only wrong one, but I checked instead of asserting that.

**Blocker 2 — degraded and ledger-closure are opposite cases, and I merged them.** Confirmed against `auto/pipeline.py:1286`. Both suppress `high_ambiguity_score`, which is what led me to write them as one clause, but the consequences diverge completely:

- **Ledger closure, not degraded** — the score is stale by design, the Seed can grade A and **run**.
- **Degraded** — the suppression exists *so the run can emit a typed partial product*. A blocker-free degraded Seed goes straight to the partial-product terminal as `AutoPhase.COMPLETE` **regardless of grade or `may_run`**. It never reaches RUN.

So my sentence told a reader that a degraded Seed above 0.2 can run, when the code guarantees the opposite. The source comment is explicit that this is the contract — "a deadline-recovery Seed is a typed alternative-success terminal, not a runnable Seed" — and I had read that block when writing the original note and still collapsed the two branches.

Split into two bullets in both languages, with the terminal behaviour and the safety-blocker carve-out named.

**Follow-up — the contract test.** Agreed, and this round is more evidence for it: manual verification passed on this table last round *because I checked the wrong directory*, and this round the English half drifted from the Korean half within the same PR. That is two distinct failure modes on one table in two rounds. It is recorded on #1998 with both.

Head is `8ba9e04`.

Q00:
Closed by the #1989 squash-merge breaking this stack. Rebuilt on `main` with the same content plus everything this review cycle produced: #2006.

Q00:

PR #1993, opened 2026-08-09
Korean Codex guide, part 3 (Troubleshooting through Synapse) — completes the translation

Q00:
Stacked on #1992. Finishes the Korean Codex runtime guide.

## Translated

Troubleshooting (Codex CLI not found, authentication, the `Providers: warning` health check, `EventStore not initialized`), Cost, and Active Conductor and Synapse.

Both partial-translation banners are removed, since `codex.ko.md` now covers the whole English document.

## One addition, not a translation

The Synapse section says Codex is a proven `inform` and `after_turn` backend and "does not advertise live checkpoint `redirect` or hard `replace`." That is accurate but leaves a reader unable to answer the question they actually have.

`SessionSignalCapabilities` has **six** fields (`core/session_signal.py:100`). Codex declares three of them true (`orchestrator/codex_cli_runtime.py:453`):

| Capability | Codex |
|---|:---:|
| `inform_delivery` | true |
| `background_reply` | true |
| `after_turn_delivery` | true |
| `checkpoint_redirect` | false |
| `owned_turn_abort` | false |
| `replacement_resume` | false |

The prose named `inform` and `after_turn` but not `background_reply`, and named `redirect` and `replace` but not `owned_turn_abort`. So two of six were unaccounted for. The Korean guide gives the full table and states the consequence plainly: **you cannot steer a long turn mid-flight — you wait for it to finish.**

I also carried the `subagent_orchestration=INTERNAL` rationale from the source comment at `codex_cli_runtime.py:448` into the guide. `codex mcp-server` exposes only `codex` and `codex-reply`, so Codex's native multi-agent team tools are unreachable by an external driver. Users who expect Ouroboros to orchestrate Codex children hit this, and nothing in the guide explained why.

## Closes the Codex track in #1988

Next in that issue: `copilot.md` (14 mentions) and `kiro.md` (11).

### DIFF
--- a/docs/runtime-guides/codex.ko.md
+++ b/docs/runtime-guides/codex.ko.md
@@ -7,7 +7,7 @@ doc_metadata:
 
 > English: [codex.md](./codex.md)
 >
-> **번역 진행 상황**: 명령 표면 · 동작 방식 · 런타임 차이 · CLI 옵션 · Seed 레퍼런스까지 옮겼습니다. 문제 해결 · 비용 · Active Conductor는 아직 영문입니다 — [codex.md](./codex.md)를 보세요. 진행은 [#1988](https://github.com/Q00/ouroboros/issues/1988)에서 추적합니다.
+> 영문 원문 전체를 옮긴 완역본입니다. 원문이 갱신되면 이 문서도 함께 갱신해야 합니다.
 
 > 설치와 첫 실행 흐름 전체는 [Getting Started](../getting-started.md)(영문)를 보세요.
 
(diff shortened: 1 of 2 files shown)


Q00:
## Strict merge hold — current review requests a correction

Do not merge current HEAD `a65bb9a63052b8f2e3370becf821edd0ddc97cac`.

The exact official review `4892753106` explicitly requests a documentation correction before merge. The Korean guide currently generalizes the `CodexCliRuntime` / native-team-tool limitation into “Ouroboros cannot orchestrate Codex children,” but the separate `codex_mcp` runtime is `EXTERNAL_LEADER_DRIVEN` and drives addressable Codex worker sessions.

This PR is stacked on #1992, therefore inherits #1992’s unresolved command-surface and billing blockers, has no required 6-check run, and diverges from current main. Scope the limitation to `codex exec` / native Codex team tools, acknowledge the separate `codex_mcp` worker-session path, repair/rebase the parent stack, then require independent exact verification, CI, and a fresh final merge recommendation.

Q00:
Pushed three additions found by running the citation-gap check **in both directions**, not just English-to-Korean.

Last round I used citation density on #2003 to find what the Chinese translation predated. Running the same comparison here inverted: `codex.ko.md` carries **6** source citations, `codex.md` carries **2**. Four exist only in Korean, and each marks a fact an English reader could not get:

1. **Setup's global artifacts.** `ensure_config_dir()` creates `~/.ouroboros/data/` and `~/.ouroboros/logs/` (`cli/commands/setup.py:2632`), and a fresh configuration gets `~/.ouroboros/credentials.yaml` written at mode `0600` (`:2771`). The English list covered `~/.codex/` only. Files written outside the runtime directory, at a specific mode, are what someone auditing setup wants to know.

2. **The Synapse capability table.** `SessionSignalCapabilities` has six fields; Codex declares three (`orchestrator/codex_cli_runtime.py:453`). The English prose named `inform` and `after_turn` as supported and `redirect`/`replace` as not advertised — leaving `background_reply` and `owned_turn_abort` unaccounted for, so a reader could not answer the question they actually have. Added the full table plus the consequence in one line: **to change direction during a long turn, you wait for that turn to finish.**

3. **`subagent_orchestration=INTERNAL` and why** (`:448`). `codex mcp-server` exposes only `codex` and `codex-reply`, so Codex's native multi-agent team tools are unreachable by an external driver — Ouroboros can continue a Codex thread but cannot orchestrate Codex children. Users who expect Ouroboros to fan out Codex hit exactly this, and nothing in the English guide explained it.

**Why I am flagging the direction.** #1992's review caught `ooo qa` fixed in Korean and left `(MCP only)` in English. I treated that as a one-off and fixed the row. It was not a one-off — the same divergence had happened in three more places, in the same direction, because the translation work is where I read the source closely and the English file only gets updated when a review points at it.

So the rule I am taking from this: **after translating, diff the citation sets both ways.** English-to-translation finds what the translation predates; translation-to-English finds what the translation learned and the source never got. The second direction is the one I would not have looked for.

Head is `d3ffdae`.

Q00:
Closed by the #1989 squash-merge breaking this stack. Rebuilt on `main` with the same content plus everything this review cycle produced: #2007.

Q00:

Issue #1995, opened 2026-08-09
Copilot: the current default model claude-opus-4-8 has no mapping, and the hyphen-to-dot fallback replaces every hyphen


Q00:
Found while correcting the Copilot compatibility paragraph in `docs/config-reference.md` (#1972).

`map_to_copilot_model()` (`copilot/model_discovery.py:247`) cannot translate the current direct-provider default, while it could translate the previous one.

## The two SSOT identifiers

`config/_model_defaults.py`:

```python
DEFAULT_OPUS_MODEL           = "claude-opus-4-8"                       # :31
DEFAULT_CONSENSUS_OPUS_MODEL = "openrouter/anthropic/claude-opus-4.8"  # :47
```

`DEFAULT_OPUS_MODEL` is the default for `clarification.default_model`, `resilience.wonder_model`, `resilience.reflect_model`, and `evaluation.semantic_model` (`config/models.py:221,339,340,361`).

## What the resolver does with it

`_STATIC_NAME_MAP` (`:94`) contains the previous default but not the current one:

```python
"claude-opus-4-6": "claude-opus-4.6",
"openrouter/anthropic/claude-opus-4-6": "claude-opus-4.6",
```

So `claude-opus-4-8` falls through to the last-attempt fallback at `:295`:

```python
if "-" in model_clean:
    dotted_candidate = model_clean.replace("-", ".")
    if dotted_candidate in available_ids:
        return dotted_candidate
```

`replace("-", ".")` rewrites **every** hyphen, not the last one:

```
"claude-opus-4-8".replace("-", ".")  ->  "claude.opus.4.8"
```

`claude.opus.4.8` is not a Copilot ID, so the candidate misses and the function returns `claude-opus-4-8` unchanged. Copilot then receives an identifier it does not publish.

This fallback can only ever succeed for single-hyphen IDs. For any Anthropic-style ID it produces a malformed candidate, which is why `claude-opus-4-6` needed a hardcoded static-map entry to work at all.

## Impact

A user whose config carries the shipped defaults switches `llm.backend` to `copilot` and the role models are passed through unmapped. The same config worked before the default moved from `claude-opus-4-6` to `claude-opus-4-8`, so this is a regression introduced by the default bump rather than a longstanding gap.

Not affected: users who leave role models unset **and** let `ouroboros setup --runtime copilot` write a discovered ID, and users who set a dotted Copilot ID explicitly (dotted IDs short-circuit unchanged at `:276`).

## Suggested direction

[...]

PR #1997, opened 2026-08-09
Korean GitHub Copilot CLI guide, and three corrections to the English source

Q00:
Adds `docs/runtime-guides/copilot.ko.md` (complete translation) and reciprocal links.

Continues #1988. The Codex guide is done (#1989 → #1992 → #1993); this is the next runtime by mention count.

## Three corrections to `copilot.md`

Checking each claim against source before translating it turned up three defects, all of which would mislead a reader into a broken configuration.

### 1. The model-mapping promise is false for the current default

The guide said the adapter auto-maps well-known Anthropic IDs, so "existing config files keep working" and "You do not need to rewrite any per-role model overrides."

`map_to_copilot_model()` (`copilot/model_discovery.py:247`) resolves: verbatim catalog match → static name map → hyphen-to-dot fallback. The static map covers `claude-opus-4-6` and `claude-sonnet-4-5` (`:96`), any dotted ID short-circuits unchanged (`:276`), and the fallback at `:295` runs `replace("-", ".")` over the **whole string**, so `claude-opus-4-8` becomes `claude.opus.4.8` — not a Copilot ID — and passes through unchanged too.

So the current `DEFAULT_OPUS_MODEL` has no mapping, while the previous default did. The guide promised the opposite. Documented the real boundary and linked the runtime issue (#1995).

### 2. The configuration example uses a key that does not exist

```yaml
llm:
  backend: copilot
  default_model: claude-opus-4.6                # written by setup
```

`LLMConfig` has no `default_model` field. The same file already says so at line 105 — "There is no `llm.default_model` key in the config contract" — so the document contradicted itself, and the contradiction lands on the copyable example rather than the prose. Replaced with `llm.qa_model` and added the note where the example is.

### 3. Troubleshooting recommends a nonexistent environment variable

> override your default to the dotted form (`OUROBOROS_DEFAULT_MODEL=claude-opus-4.6`)

`OUROBOROS_DEFAULT_MODEL` appears nowhere in `src/`. This one is doubly unhelpful: it is offered as the fix for a real error message, so a reader hits the error, applies the suggested workaround, sees no change, and has no signal that the variable was ignored. Replaced with rerunning setup, and named the per-role variables that do exist.

## Audit result on the neighbouring guides

In #1989 I said I would check `copilot.md` and `kiro.md` for the plugin-prerequisite defect found on the Claude and Codex paths. **They are clean.** `copilot.md` has a prerequisites table naming the isolated launcher, and `kiro.md:29` and `hermes.md:24` both state the `uvx`/`pipx` requirement immediately above the setup command. The defect was specific to the two marketplace-plugin paths, which are the ones added most recently.

## Verification

All relative links in the new file resolve.

### DIFF
--- a/docs/runtime-guides/copilot.ko.md
+++ b/docs/runtime-guides/copilot.ko.md
@@ -1,0 +1,200 @@
+# GitHub Copilot CLI 런타임
+
+로컬에 설치된 [GitHub Copilot CLI](https://docs.github.com/copilot/concepts/agents/about-copilot-cli) 위에서 Ouroboros 워크플로를 돌립니다.
+
+> English: [copilot.md](./copilot.md)
+>
+> 영문 원문 전체를 옮긴 완역본입니다. 원문이 갱신되면 이 문서도 함께 갱신해야 합니다.
+
+Copilot 런타임은 Codex · Gemini · Hermes · OpenCode · Kiro 런타임의 형제입니다. Ouroboros가 오케스트레이션 루프를 소유하고, 호스팅 SDK와 대화하는 대신 작업마다 `copilot -p`를 셸로 호출합니다. 인증은 기존 `gh auth` 세션을 타므로 **따로 관리할 API 키가 없습니다.**
+
+> **이 런타임만 다른 점**: Copilot은 Ouroboros 백엔드 중 유일하게 **모델 목록을 실시간으로 발견합니다.** `ouroboros setup --runtime copilot`이 setup 시점에 GitHub Copilot models API를 조회해서, 하드코딩된 모델 ID를 외우게 하는 대신 **지금 당신 구독이 실제로 허용하는 것** 중에서 기본값을 고르게 합니다. GitHub가 새 모델을 내놓는 순간부터 쓸 수 있고, 갱신하려면 setup을 다시 돌리면 됩니다.
+
+## 사전 조건
+
+| 요구사항 | 이유 |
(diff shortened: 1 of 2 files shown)


Q00:

This PR was merged as `22232db1149224f6af5a36fdcb5be1ab1f995064` from exact HEAD `a84e201b45b12ca54ffa1dcd584b055086c738e0`, but the exact official review did **not** authorize merge.

Review `4892761119` had matching API `commit_id`, body `HEAD checked`, and `Review-Metadata head_sha`, yet its final **Merge Recommendation** was explicitly:

> Request changes. Update both guides to state that an unavailable dotted candidate is discarded and the original model ID is returned, and either enumerate the exact static mappings or avoid claiming broader OpenRouter coverage.

The findings reproduce against the merged source:

- `map_to_copilot_model()` returns the original model ID when its generated dotted candidate is unavailable; the guides describe the dotted candidate as the returned fallback.
- The static map does not provide the broad OpenRouter coverage claimed by the guides.
- The fallback is not always silent: setup emits a warning when bundled fallback discovery is used.

Required CI was 6/6 green, but CI does not establish runtime-claim truthfulness. Metadata `APPROVED` plus blocker-table filtering must not override a contradictory final recommendation.

I am not auto-reverting this documentation merge. Follow-up should coordinate the already-open runtime bug #1995 with an immediate English/Korean documentation correction, then require independent verification and a fresh exact-head review whose final recommendation actually says merge.

PR #1999, opened 2026-08-09
Korean Kiro CLI runtime guide

Q00:
Adds `docs/runtime-guides/kiro.ko.md` (complete translation) and reciprocal links. Continues #1988; last of the guides on that issue's list.

## Verification, and a clean result

Every claim was checked against source before translating it. **`kiro.md` is accurate** — no corrections needed, which is worth stating because the previous four guides in this series each needed two or three.

Checks that could have failed and did not:

- `RuntimeCapabilities(skill_dispatch=True, targeted_resume=False, structured_output=False)` matches `_KIRO_CAPABILITIES` at `orchestrator/kiro_adapter.py:55-58` exactly, and the source comments at `:47` and `:53` give the same reasons the guide gives.
- The registered `env` block uses `OUROBOROS_RUNTIME` (`cli/commands/setup.py:3068`), which differs from the `OUROBOROS_AGENT_RUNTIME` that the Codex, Copilot, and OpenCode entries use. This looked like a copy-paste bug, but both names are read: `config/loader.py:663` reads `OUROBOROS_AGENT_RUNTIME` and `:667` reads `OUROBOROS_RUNTIME`. The doc matches what setup writes, and what setup writes works.
- The `uvx`/`pipx` requirement is stated immediately above the setup command, so kiro does not have the prerequisite-ordering defect I fixed on the Claude and Codex plugin paths in #1987 and #1989.

All relative links in the new file resolve.

## Translation notes

Kept in English: `bypassPermissions`, `--trust-all-tools`, `--trust-tools`, `--resume-id`, `--list-sessions`, capability field names, and the commit hash `9d0db8a`. These are strings a reader will type or grep for.

Two places got emphasis the English carries structurally but Korean would lose:

- The permission section's real content is that a tool envelope **cannot** downgrade the native approval boundary. That is the sentence someone reasons about when auditing what Ouroboros can and cannot restrict.
- `targeted_resume=False` is a deliberate honesty choice, not a missing feature — the adapter does understand `--resume-id` when given an id. The guide says so and I kept that distinction explicit rather than flattening it to "resume is unsupported."

## Korean runtime guide status after this

Claude Code (#1987), Codex (#1989 → #1992 → #1993), Copilot (#1997), Kiro (this). That covers the runtimes named in #1988.

### DIFF
--- a/docs/runtime-guides/kiro.ko.md
+++ b/docs/runtime-guides/kiro.ko.md
@@ -1,0 +1,152 @@
+# Ouroboros 런타임 가이드: Kiro CLI
+
+[Kiro CLI](https://kiro.dev/docs/cli/)를 실행 런타임으로 삼아 Ouroboros를 쓰는 방법입니다. Kiro는 **헤드리스 모드**로 돕니다(`kiro-cli chat --no-interactive`, 문서는 <https://kiro.dev/docs/cli/headless/>).
+
+> English: [kiro.md](./kiro.md)
+>
+> 영문 원문 전체를 옮긴 완역본입니다. 원문이 갱신되면 이 문서도 함께 갱신해야 합니다.
+
+## 설치
+
+Ouroboros가 쓰기 전에 Kiro CLI가 설치되고 인증돼 있어야 합니다:
+
+```bash
+# 설치 확인 — --resume-id 지원에는 2.2.0 이상이 필요합니다
+kiro-cli --version
+kiro-cli chat --help | grep -- --resume-id
+```
+
+Kiro 배포판이 제공하는 방식(AWS Builder / IAM 로그인 등)으로 **한 번 로그인해 두세요.** 그래야 헤드리스 호출이 프롬프트 없이 동작합니다.
+
+## Setup
+
+```bash
+pipx install 'ouroboros-ai[mcp]'         # 또는: uv tool install 'ouroboros-ai[mcp]'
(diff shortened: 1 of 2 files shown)


Q00:

This PR was merged as `0b95a9789043f05e33e2671874b0c08f623e8c1b` from exact HEAD `90cc60bd2b3083bd2732466f5b3878d906d28e29`, but the exact official review explicitly recommended **Request changes**.

Review `4892774390` was SHA-exact and required CI eventually reached 6/6. Its final recommendation remained:

> Request changes until the Korean and English MCP examples include the mandatory Python floor, the custom-launcher wording matches setup behavior, and the documentation contract test is updated to enforce the real generated entry.

The findings were reproduced against the merged source:

- generated `uvx` MCP entries require the Python floor (`--python >=3.12`), but the English and Korean examples omit it;
- the guides imply setup entries are always `uvx`/`pipx`, while setup intentionally preserves supported custom launchers such as Docker/Nix forms;
- the documentation contract test does not enforce the actual generated entry, so green CI cannot detect this drift.

This is another case where `APPROVED` review state and green CI conflict with the source-backed final merge recommendation. The final recommendation must remain part of the merge gate.

I am not auto-reverting the useful Korean guide. Follow-up should correct both language examples and wording, derive the contract test from the setup-generated entry, obtain an independent exact verifier, and receive a fresh review whose final recommendation actually authorizes merge.

Issue #1998, opened 2026-08-09
Eight config fields validate and persist but are never read; nothing detects the next one


Q00:
Surfaced across eight review rounds on #1972. Each round found more inert fields in the same two config sections, because the only detection method available was a human reading the reference against the source.

## The fields

`EvaluationConfig` (`config/models.py:356-360`):

| Field | Why it does nothing |
|---|---|
| `stage1_enabled` | No production code builds `PipelineConfig` from it |
| `stage2_enabled` | Same |
| `stage3_enabled` | Same |
| `satisfaction_threshold` | `EvaluationPipeline` compares `stage2_result.score >= 0.8` literally |
| `uncertainty_threshold` | Stage 3 gates on `TriggerConfig.uncertainty_threshold` inside `PipelineConfig`; nothing copies the YAML value in |

`ConsensusConfig` (`config/models.py:378-380`):

| Field | Why it does nothing |
|---|---|
| `min_models` | Defined and never read; no minimum voter count is enforced |
| `threshold` (`0.67`) | Approval uses `evaluation.consensus.ConsensusConfig.majority_threshold` (`0.66`), a different class |
| `diversity_required` | `ConsensusEvaluator` holds one `self._llm`; the roster cannot produce cross-vendor independence |

The two production `PipelineConfig` constructors:

- `mcp/server/adapter.py:1893` hardcodes the stage flags, taking `stage1_enabled` from `OUROBOROS_EVOLVE_STAGE1` and pinning `stage2_enabled=True`, `stage3_enabled=False`.
- `mcp/tools/evaluation_handlers.py:743` passes only `mechanical` and `semantic`, so the stage flags fall to `PipelineConfig` defaults.

Neither reads `config.evaluation`.

## Why this is worse than a documentation bug

These fields are not silently ignored in a way an operator can notice. They **validate** (`min_models` enforces `>= 2`, `threshold` enforces `[0.0, 1.0]`), they **persist** to `~/.ouroboros/config.yaml`, and they **appear in `ouroboros config show`**. Every signal the system gives says the setting took effect.

So an operator who sets `stage3_enabled: false` to cut cost, or raises `consensus.threshold` to `0.9` to demand stronger agreement, gets a clean config, a clean `config show`, and unchanged behaviour. There is no error, no warning, and no log line. The failure is silent by construction.

`threshold` is the sharpest case: the documented default `0.67` and the effective value `0.66` are close enough that even someone comparing behaviour against the docs would not notice from the number alone.

[...]

PR #2000, opened 2026-08-09
OUROBOROS_SUB_RETRIES is documented but does not exist

Q00:
`OUROBOROS_SUB_RETRIES` is documented in two places with a default of `2`. It does not exist.

## Verification

The OpenCode bridge reads exactly one environment variable. `process.env` appears once in `src/ouroboros/opencode/plugin/ouroboros-bridge.ts`, at line 38:

```ts
export const CHILD_TIMEOUT_MS = num(process.env.OUROBOROS_CHILD_TIMEOUT_MS, 20 * 60 * 1000)
```

Retries are compile-time constants at `:40-41`:

```ts
const PATCH_RETRIES = 3
const RESOLVE_RETRIES = 5
```

`PATCH_RETRIES` applies to PATCH calls on a network or server blip (`:598`). `RESOLVE_RETRIES` applies when resolving the assistant message hosting a `callID` (`:613`), which fails closed and returns `null` if no exact match is found. Neither reads an environment variable, and `OUROBOROS_SUB_RETRIES` appears nowhere in the repository outside these two documentation tables.

## Why this shape of error is worth fixing rather than deleting quietly

Both tables sit under an "Environment knobs" heading next to `OUROBOROS_CHILD_TIMEOUT_MS`, which **is** real. That adjacency is what makes it convincing: a reader debugging flaky child spawns finds a table of two knobs, sets the one named for retries, restarts, sees no change, and has no signal that the variable was ignored.

So I replaced the row with what is actually true in each document, rather than removing it. Someone who came looking for a retry knob should learn that retries are fixed at 3 and 5 and where those live, not just fail to find the row.

The same defect shape appears in `copilot.md`, where troubleshooting recommends setting `OUROBOROS_DEFAULT_MODEL` — also nonexistent — as the fix for a real error message. That one is corrected in #1997.

## How these were found

A scan of every `OUROBOROS_*` token in `docs/` against every occurrence in the rest of the repository. Twelve names appeared only in documentation; ten were false positives worth naming, since they show what the check must not flag:

- **Five** are in `docs/contributing/findings-registry.md`, a `status: legacy-frozen` archive whose entry `CLM-003` *records* that `OUROBOROS_TUI_THEME`, `OUROBOROS_MAX_AGENTS`, and `OUROBOROS_EVENT_CACHE_SIZE` are unread. That file documents the defect rather than committing it.
- **Two** (`OUROBOROS_CODEX_RPM`, `OUROBOROS_OPENCODE_RPM`) are illustrative expansions of the documented `OUROBOROS_<BACKEND>_RPM` pattern, constructed at runtime rather than matched as literals.
- **Three** more resolve in test files or the TypeScript bridge rather than `src/**/*.py`.

That leaves two genuine defects: this one and the `copilot.md` one.

That the frozen registry already recorded this exact class in March, and it recurred twice since, is the argument in #1998: documentation and source have no mechanical relationship, so the only detector is a person reading both. A lint over `OUROBOROS_*` tokens would be cheap and would have caught both, provided it tolerates the three false-positive shapes above.

### DIFF
--- a/docs/guides/opencode-subagent-bridge.md
+++ b/docs/guides/opencode-subagent-bridge.md
@@ -172,7 +172,15 @@ does **not** abort the rest of a fan-out batch.
 | Variable                         | Default   | Purpose |
 |----------------------------------|-----------|---------|
 | `OUROBOROS_CHILD_TIMEOUT_MS`     | 1 200 000 | Per-child overall timeout (ms) |
-| `OUROBOROS_SUB_RETRIES`          | 2         | Extra retries after first child attempt |
+
+`OUROBOROS_CHILD_TIMEOUT_MS` is the only variable the bridge reads
+(`process.env` appears once, at `opencode/plugin/ouroboros-bridge.ts:38`).
+
+Retries are **not** configurable. The bridge holds two compile-time constants
+at `:40-41`: `PATCH_RETRIES = 3`, applied when a PATCH hits a network or server
+blip (`:598`), and `RESOLVE_RETRIES = 5`, applied when resolving the assistant
+message that hosts a `callID` (`:613`, fails closed and returns `null` if no
+exact match is found). Neither responds to an environment variable.
 
 ## Installation
 
--- a/docs/runtime-guides/opencode.md
+++ b/docs/runtime-guides/opencode.md
@@ -131,7 +131,11 @@ ouroboros_lateral_think persona="all"
 | Variable | Default | Purpose |
 |----------|---------|---------|
 | `OUROBOROS_CHILD_TIMEOUT_MS` | `1200000` (20 min) | Per-child wall clock |
-| `OUROBOROS_SUB_RETRIES` | `2` | Retry count on spawn failure |
+
+> `OUROBOROS_CHILD_TIMEOUT_MS` is the only environment knob the bridge reads
+> (`opencode/plugin/ouroboros-bridge.ts:38`). Retry counts are compile-time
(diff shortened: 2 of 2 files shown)


Q00:

This PR was merged as `8f63e81d7019ae3d0fe8d6cb05a686e137570eb9` from exact HEAD `b3709909c7bee47df7ef023b2daf6fea97fdaee3`, but the exact official review explicitly recommended **Request changes**.

Review `4892793065` had matching API `commit_id`, body `HEAD checked`, and `Review-Metadata head_sha`, and required CI was green. Its source-backed findings nevertheless remained unresolved at merge time:

- “the only variable the bridge reads” is false: the bridge also reads `HOME`, `USERPROFILE`, `OPENCODE_CONFIG_DIR`, `APPDATA`, and `XDG_CONFIG_HOME`; the accurate scope is the only `OUROBOROS_*` child-tuning setting;
- `PATCH_RETRIES = 3` and `RESOLVE_RETRIES = 5` are loop bounds including the first attempt, so runtime performs three PATCH attempts / two additional retries and five resolution polls / four additional retries;
- bare implementation line numbers are a valid non-blocking maintainability follow-up.

The final review recommendation was:

> Request changes. Keep removal of `OUROBOROS_SUB_RETRIES`, but narrow the environment claim and describe the constants as three PATCH attempts and five resolution polls—or two and four additional retries—before merging.

This is the same metadata/body contradiction already seen on other merges: `APPROVED` state and an empty blocker table do not override a source-backed final `Request changes` recommendation. I am not auto-reverting the useful removal of the nonexistent knob. A follow-up documentation correction should fix both files and add a mechanical environment-token/constant contract so the same drift does not recur.

PR #1974, opened 2026-08-09
docs(i18n): add the Chinese evaluation pipeline guide

Q00:
**Severity:** medium

> **Stacked on #1972.** Base branch is `docs/eval-pipeline-code-accuracy`, so the diff shown here includes that PR's commits until it merges. The only commit belonging to this PR is `docs(i18n): add the Chinese evaluation pipeline guide`. The stacking is deliberate: #1972 corrects the places where the English guide no longer matched the code, and this translation is written against the corrected text so the two versions do not disagree on the day they land.
>
> **No checks have run on this PR yet, and that is expected.** Every workflow in `.github/workflows/` is gated on `pull_request: branches: [main, develop]`, so a PR based on another branch triggers nothing — including the issue-link gate, even though `Closes #1973` is present below. Merging #1972 first auto-retargets this PR to `main`, at which point the full suite runs. Please merge #1972 before this one.

## Summary

Adds `docs/guides/evaluation-pipeline.zh-CN.md`, following the pattern #1961 established. #1960 named the evaluation pipeline as one of the next two most-asked-about guides; this is the first of them.

## What the page covers

A full companion, not a summary — same structure and the same claims as the English guide:

- All three stages: what Stage 1 actually runs and where its commands come from, Stage 2's scoring fields and approval logic, Stage 3's simple and deliberative consensus modes
- The complete consensus trigger matrix including `manual_request` and the strict `>` comparisons
- Every failure-mode table, which is the part users reach for when a verdict looks wrong
- The reward-hacking veto, artifact collection limits, error-vs-failure distinction, edge cases, the full config reference, and the event audit trail

Identifiers, file paths, config keys, event type names, CLI commands, and error strings stay in their original form — those are what a reader greps for. Terminology follows `README.zh-CN.md` (`验收标准`, `漂移`, `产物`, `三阶段关卡`) so the vocabulary matches what a Chinese reader already saw on the front page.

## Navigation

- `docs/README.md` lists the Chinese page beside the English entry in the Guides index
- The English guide links to it from directly under the title
- `README.zh-CN.md` promotes its design-docs line into a list and puts this guide above the existing hidden-checklist entry

Neither direction dead-ends.

## Test plan

- [x] All seven relative links resolve from `docs/guides/`: `./evaluation-pipeline.md`, `./execution-vs-evaluation.md`, `./seed-authoring.md`, `../architecture.md`, `../config-reference.md`, `../getting-started.md`, `../hidden-checklist-convergence/README.zh-CN.md`
[...]

PR #2003, opened 2026-08-09
docs(i18n): add the Chinese evaluation pipeline guide

Q00:
Reopens the work from #1974, which GitHub auto-closed when I merged #1972 and deleted the branch it was stacked on. My mistake — I used `--delete-branch` on a base that still had a child.

Same content, rebased onto `main` by cherry-picking the one commit that was actually this PR's (`82ce5532`); the other two on the old branch were #1972's, now on `main` as `5fbcc62d`.

## One correction, and it matters

The Chinese guide was translated from the English guide **before** #1972's eight review rounds. Checking it against the version now on `main` turned up one stale claim, and it happens to be the worst one in that set:

> 通过 `MechanicalConfig` 或 MCP `overrides` 字典以编程方式传入的命令绕过这道检查

That is the nonexistent MCP contract I invented and then removed from three separate places in the English guide across rounds three, five, and six. Reopening this PR unchanged would have reinstated it in Chinese — an operator reading the Chinese security section would come away believing a privileged MCP command path exists that can bypass the executable allowlist.

Corrected at both sites, scoped to trusted Python callers, with the same source citations the English guide carries: `ouroboros_evaluate` exposes no mechanical-command parameter (`mcp/tools/evaluation_handlers.py:437`) and its handler calls `build_mechanical_config(working_dir)` without `overrides` (`:742`).

## What was already correct

I checked the other corrections from those eight rounds rather than assuming. The translation already covers them, because they landed before the Chinese file was written:

- `satisfaction_threshold` is documented as inert at both the prose site (`:197`) and in the config example (`:228`), with the hardcoded `0.8` named.
- The Stage 2 gate, the trigger matrix, and the strict `>` comparisons match.

So the drift was limited to the one claim that was corrected last.

That is the useful takeaway: a translation forked mid-review inherits whatever was true at fork time, and nothing flags it when the source moves afterward. Same class as #1998, in a different direction.

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -380,7 +380,10 @@ src/ouroboros/
 - **Agent OS runtime** —— 跨能力发现、策略、指令、事件日志、agent 进程的可重放执行契约
 - **Runtime backends** —— 可插拔抽象层（`orchestrator.runtime_backend` 配置），原生支持 Claude Code、Codex CLI、OpenCode、Hermes；同一份工作流规约，跑在不同执行引擎上
 
-完整设计文档见 [Architecture](./docs/architecture.md)（英文）。中文设计说明：[隐藏清单收敛（Hidden-Checklist Convergence）](./docs/hidden-checklist-convergence/README.zh-CN.md) —— run → 评估 → 有预算的 Ralph 链，以及为什么判分用的断言对 worker 无条件隐藏。
+完整设计文档见 [Architecture](./docs/architecture.md)（英文）。中文文档：
+
+- [评估流水线指南（Evaluation Pipeline）](./docs/guides/evaluation-pipeline.zh-CN.md) —— 三阶段关卡的完整参考：每个阶段验什么、阈值和配置项、失败模式与排查、以及事件审计轨迹
+- [隐藏清单收敛（Hidden-Checklist Convergence）](./docs/hidden-checklist-convergence/README.zh-CN.md) —— run → 评估 → 有预算的 Ralph 链，以及为什么判分用的断言对 worker 无条件隐藏
 
 </details>
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -47,6 +47,7 @@ replayable execution contract on your choice of runtime backend.
 - [Seed Authoring Guide](./guides/seed-authoring.md) - YAML structure, field reference, examples
 - [Evolutionary Loop & Ralph](./guides/evolution-loop.md) - Wonder/Reflect cycle, convergence detection, persistent evolution
 - [Evaluation Pipeline Guide](./guides/evaluation-pipeline.md) - Three-stage evaluation, failure modes, and configuration
+- [Evaluation Pipeline Guide (简体中文)](./guides/evaluation-pipeline.zh-CN.md) - 同一份指南的中文版
 - [Execution vs. Evaluation Contract](./guides/execution-vs-evaluation.md) - Task completion, AC verdict, and drift terminology boundaries
 - [Hidden-Checklist Convergence](./hidden-checklist-convergence/architecture.md) - Run → evaluation → bounded Ralph chaining with hidden harness grading inputs
 - [Hidden-Checklist Convergence (简体中文)](./hidden-checklist-convergence/README.zh-CN.md) - 同一设计的中文说明
--- a/docs/guides/evaluation-pipeline.md
+++ b/docs/guides/evaluation-pipeline.md
@@ -5,6 +5,8 @@ doc_metadata:
 
 # Evaluation Pipeline Guide
 
+> 中文说明：[evaluation-pipeline.zh-CN.md](./evaluation-pipeline.zh-CN.md)
+
 Ouroboros Phase 4 runs every execution result through a **three-stage progressive evaluation pipeline** before assigning formal acceptance-criterion (AC) verdicts. Cheaper checks gate the expensive ones: Stage 1 is free, Stage 2 uses one LLM call, and Stage 3 (multi-model consensus) runs only when specifically triggered.
 
 > **Terminology boundary:** worker task completion is not a formal AC verdict, and task failure is not semantic drift. See [Execution vs. Evaluation Contract](./execution-vs-evaluation.md) for the shared `TaskResult` / `ACResult` distinction.
--- a/docs/guides/evaluation-pipeline.zh-CN.md
+++ b/docs/guides/evaluation-pipeline.zh-CN.md
@@ -98,9 +98,13 @@ ouroboros detect --force      # 重新检测并覆盖已有文件
 ouroboros detect --backend codex   # 为这次 detect 调用指定 LLM 后端
 ```
 
-`ensure_mechanical_toml()` 是幂等的：文件已存在且 `force` 为假时，它立即返回，不发起 LLM 调用。它从不抛异常——任何失败都只是返回 `False`，Stage 1 因此没有命令可用。
+`ensure_mechanical_toml()` 是幂等的：文件已存在且 `force` 为假时，它立即返回，不发起 LLM 调用。
 
-> **如果 Stage 1 永远通过，通常就是这个原因。** 既没有 `.ouroboros/mechanical.toml`，也没有显式配置 `MechanicalConfig` 命令时，五项检查全部被跳过并当作通过，Stage 1 就成了一道什么都不验的空关卡。跑 `ouroboros detect`，或者手写这个 toml。
+**失败有两种形态，要分开看。** 大多数失败被处理成返回 `False`，Stage 1 因此没有命令可用。但它**并非从不抛异常**：`_ask_llm()` 调用 `tracked_complete()` 时没有异常边界，因此适配器异常会向上传播，而直接执行的 `ouroboros detect` 命令也不捕获这一调用。诊断时请把「处理过的 `False`」和「抛出的异常」当作两件事。
+
+> **如果 Stage 1 永远通过，通常就是这个原因。** 既没有 `.ouroboros/mechanical.toml`，也没有显式配置 `MechanicalConfig` 命令时，五项检查全部被跳过并当作通过，Stage 1 就成了一道什么都不验的空关卡。
+>
+> **但请注意：缺少 TOML 本身并不足以解释它。** 常规 `run` 与 MCP 评估路径都会先自动尝试生成这个文件。所以在这些路径上出现空关卡，意味着**自动检测失败了**（返回 `False` 或抛出异常），而不是「文件恰好不存在」。跑 `ouroboros detect` 看它到底报什么，或者手写这个 toml。
 
 > **已废弃：** `detect_language()` 现在什么也检测不了。它只是一个兼容垫片，读取 `.ouroboros/mechanical.toml` 并发出 `DeprecationWarning`；请改用 `ensure_mechanical_toml()` 加 `build_mechanical_config()`。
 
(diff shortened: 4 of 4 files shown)


Q00:
All five confirmed. Ported.

The root cause is the one I named when reopening this PR, and I then underestimated it. The Chinese guide was translated from the English guide **before** its eight review rounds. When I reopened, I checked a handful of markers — `satisfaction_threshold`, the hardcoded `0.8` — found them present, and concluded the rest was fine. That was a spot-check reported as coverage. Every finding here is a section those eight rounds rewrote and the translation never saw.

**Blocker 1 — and my own fix was wrong.** Last round I collapsed the two Python mechanisms into "trusted callers bypass". They are not the same: `_apply_overrides()` calls `_parse_command()` on each value (`evaluation/languages.py:247`), so `build_mechanical_config(..., overrides=...)` **still** rejects shell operators, absolute paths, and non-allowlisted heads — it only skips the repository entry-point and argument-containment checks applied to TOML. Only a directly constructed `MechanicalConfig` bypasses both parsers.

Worth noting the English guide on `main` already states this correctly at line 133. So I had the right text available and translated an older one.

**Blocker 2 — the detector failure contract.** Confirmed: `ensure_mechanical_toml()` does not never-throw. `_ask_llm()` calls `tracked_complete()` outside an exception boundary and `ouroboros detect` does not catch it. Handled `False` and thrown exceptions are now separate. Also corrected the diagnostic advice: on `run` and MCP paths both attempt to author the TOML first, so a no-op Stage 1 there means **detection failed**, not that the file happens to be absent. That distinction is the whole value of the paragraph for someone debugging a missing verification gate.

**Blocker 3 — Stage 3 trust model.** This is the one I most did not want to leave standing, since the English version of this correction took three rounds to get right. The Chinese said the three default models are OpenRouter-routed and therefore three vendors. Added the full caveat: roster resolution is environment and backend dependent, sentinel backends resolve to literal `"default"`, independence filtering can alter the roster, and **every vote goes through the single adapter handed to `ConsensusEvaluator`** — changing the roster never changes the adapter. Model labels do not prove independent vendor review; read the recorded votes.

**Blocker 4 — `ArtifactCollector`.** Confirmed: `EvaluationPipeline.evaluate()` never instantiates it. Scoped collection to the MCP paths and stated the direct-caller responsibility explicitly, with the consequence spelled out — you can believe source files were evaluated when only the text artifact was.

**Follow-up — the veto is not always named.** Confirmed. The Stage 2 AC branch precedes the veto branch in `_build_result()`, so a `trigger_consensus=True` result carrying `ac_compliance=False` into an approving Stage 3 reports AC non-compliance rather than the veto. Corrected at both sites and pointed operators at `stage2_result.reward_hacking_risk`.

**Suggestion — collection limits.** Taken. "The character budget is the only limiter" was wrong on three counts: the 100 KB per-file limit, cumulative small files exhausting the 150,000-character budget, and `text_summary` living outside it.

**On the general risk.** I fixed what the review found, but I cannot claim the file is now current — the same fork-point problem could have left other sections stale, and a spot-check is what produced this situation. The durable fix is a translation-freshness signal, which is the same missing mechanical relationship tracked in #1998, in a third direction: config fields, CLI mappings, and now translations.

Head is `1a7bcb5`.

Q00:
Follow-up to my last comment, where I said I could not claim this file is current. I have now narrowed that from a caveat into a measurement.

**The method.** The eight review rounds on #1972 added precise `file.py:line` citations wherever a claim was corrected. So citation density is a proxy for which sections were rewritten. Comparing the two guides:

- Section headings: **40 in English, 40 in Chinese** — structure is aligned.
- Source citations: **16 in English, 2 in Chinese** — 15 missing.

That turns "something else might be stale" into a bounded list of 15 places to check.

**What the gaps contained.** Two were substantive and are now ported:

1. **The entire sentinel-model backend section was absent.** No `_SENTINEL_DEFAULT_BACKENDS`, no OpenCode riding in through `_CODEX_LLM_BACKENDS` (`config/loader.py:76`), no note that `_should_use_multi_model()` selects the multi-model branch for any roster without an `openrouter/` entry so three votes labeled `default` come from one model, and no reviewer-independence filtering.

   This is the material from rounds four and five, and it is precisely what stops a reader from reading an OpenRouter-looking roster as three vendors. I had added the shorter "roster is not independence" caveat two comments ago; that was the conclusion without the mechanism. The mechanism is what lets an operator recognise the situation in their own consensus evidence.

2. **The detector paragraph overstated what validation proves.** It ended at "only commands this project can actually run". `_command_is_valid()` deliberately does not consult `PATH` (`evaluation/detector.py:496`), so a manifest-declared `pytest` command is persisted even when `pytest` is absent, and Stage 1 reports `Command not found` at run time. The distinction matters exactly when someone is debugging a check that never passes.

**On the remaining thirteen.** They are mostly citations attached to text the translation does carry in substance — the veto branch ordering, for instance, is present in Chinese from my earlier fix but without the `evaluation/pipeline.py:307-326` reference. I have not added citations for their own sake.

I am stating the method rather than just the result because it generalises: **for any translated document, the source's citation set is a cheap first approximation of what the translation predates.** It is not a substitute for the blob-SHA binding I proposed on #1998 — it only works because these corrections happened to add citations — but it needs no tooling and it converted an open-ended risk into two real fixes here.

Head is `9c50b0e`.

Q00:
All three confirmed. Fixed.

**Blocker 1 — I fixed the paragraph and left the table rows.** Correct. Last round I corrected the detector prose ("safe and repository-declared" rather than "actually runnable") and the no-op-Stage-1 note, and did not carry the same distinction into the failure-mode rows below. The `没有 toml 文件` row still presented a missing file as a sufficient explanation.

That is the fifth time this session I have fixed a claim in one place and left it standing in a sibling location. The row now says what the prose says: on `run` and MCP paths detection is attempted first, so an empty gate there means **detection failed** and `ouroboros detect` will show why; only lower-level callers that bypass detection see a plain missing file.

**Blocker 2 — the Stage 3 contract contradicted itself in three places.** Also correct, and my previous fix made this worse rather than better: I added a caveat under the roster while leaving two later statements that assert the opposite. A caveat that a later section contradicts is not a caveat, it is a second opinion.

The three sites:
- The section opening said each model votes independently.
- The `diversity_required` note said **vendor diversity comes from the default roster itself** — which is exactly the claim the caveat was meant to remove.
- The deliberative section said each role uses its own model.

All simple votes go through `ConsensusEvaluator._llm`, and all deliberative roles go through the single adapter passed to `DeliberativeConsensus`. Rewrote all three so a roster entry reads as a **requested label** and never implies a separate model, vendor, or transport.

**Follow-up — the artifact-collection failure table.** Confirmed against `artifact_collector.py:201`. Two rows were wrong in the same direction, both making Stage 2 look blinder than it is:
- No extractable file path does **not** fall back to the text summary; `collect()` calls `_scan_directory(project_dir)` and evaluates eligible project files. Only an empty, inaccessible, or fully excluded scan leaves nothing.
- Skipping one oversized file does not reduce evaluation to summary-only; the remaining collected files are still used.

Both now say what actually happens, since the point of that table is letting someone work out what evidence Stage 2 reviewed.

**On the P2 suggestion.** Agreed, and I have already tried the obvious version and shown it fails — a commit-date freshness check clears the exact commit that introduced this file with five stale sections, because both paths change in the same commit. Recorded with the reproduction on #1998, along with the alternative: bind the translation to the source's **blob SHA** in front matter, so a mismatch flags it regardless of commit timing and the blob diff becomes the reviewer's worklist.

What I have been doing here instead — diffing the citation sets in both directions — found the sentinel-backend gap and three facts that existed only in Korean, but it only works because these corrections happened to add citations. It is a stopgap, not the mechanism.

Head is `bd0bc03`.

Q00:
Pushed two more fixes that no review asked for, found by sweeping ahead of the next round rather than after it.

I have now fixed a claim in one place and left it standing elsewhere **five times** on this PR alone. So instead of waiting for the sixth, I grepped this file for every remaining sentence in the two families I had been correcting — independence/diversity claims, and claims about configuration that turns something on or off.

**1. The YAML stage toggles are inert, and the Chinese guide never said so.** The section on `stage1_enabled` / `stage2_enabled` / `stage3_enabled` discussed them as working controls. The English guide carries a paragraph the translation predates: those are **direct-Python** controls on `PipelineConfig`, and the same-named top-level keys in `~/.ouroboros/config.yaml` are schema-validated but never copied into `PipelineConfig`. Likewise `evaluation.uncertainty_threshold` does not populate `TriggerConfig.uncertainty_threshold`.

That paragraph is the outcome of eight review rounds on #1972, and it is the highest-consequence thing in this area: a reader sets `evaluation.stage3_enabled: false` in `config.yaml` to cut cost, gets no error, no warning, and unchanged behaviour. Silent by construction. Added, with the explicit statement that changing those YAML keys does not change behaviour.

**2. Round-1 "independence" in the deliberative section.** `各自独立陈述立场` is true about the *positions* — Advocate and Devil's Advocate argue opposite sides — but it sits a few lines after the section establishing that both roles go through the single adapter. Right word, wrong neighbourhood. Qualified so `独立` reads as positional, not architectural.

**What the sweep found and did not find.** Fourteen `独立`/`多样性`/`厂商` occurrences: eleven were correct (the sentinel caveats, the audit note, the per-AC evaluation section, the collection-limit sentence). One was the round-1 wording above. The stage-toggle gap came from the second family, not the first.

That ratio is the point. Eleven correct out of fourteen is why re-reading does not catch these — the surrounding text looks right, and the wrong ones are wrong for reasons a reader cannot see without the source. It is also why "I checked" has been unreliable from me this session, and a grep over a claim family is not.

Head is `2f3fdf1`.

Q00:
All four confirmed. Fixed.

**Blocker 3 is the one that matters most**, so I will start there: the artifact table told direct callers to set `project_dir` "in the execution context". `EvaluationContext` has no such field (`evaluation/models.py:335` onward — `execution_id`, `seed_id`, `current_ac`, `artifact`, `artifact_type`, `goal`, `constraints`, `artifact_bundle`). The directory is an argument to `ArtifactCollector.collect(execution_output, project_dir)`, and the resulting bundle has to be attached as `artifact_bundle`.

So a caller who followed my advice would set a field that does not exist, collect nothing, and have Stage 2 silently evaluate the text artifact — **the precise failure that section exists to warn about.** I wrote a remedy that causes the disease.

**Blocker 1 — the trigger matrix was gated on the wrong thing.** Confirmed: `evaluate()` returns early only for `ac_compliance=False` without a manual consensus request. An AC-compliant result scoring 0.7 still reaches trigger evaluation, and a matching trigger can send it to Stage 3 and approve it. The `0.8` gate is a **later** decision, not an entry condition. My version told operators that sub-0.8 results never reach Stage 3, which is wrong about both cost and approval semantics.

**Blocker 2 — the escape-hatch note contradicted my own correction.** I fixed the failure table at line 141 to say `run` and MCP paths author the TOML first, then left the note at 158 saying a missing file plus no explicit commands is sufficient. Two statements about the same condition, in one document, disagreeing — the reader has no way to tell which is current. Now consistent: the silent skip requires **detection to have failed**, and a plain missing file only happens on callers that bypass detection.

**Blocker 4 — the full-config example pinned the roster.** Confirmed. An explicit tuple sets `models_are_explicit` and bypasses backend-aware normalization, so on a local or sentinel backend those identifiers produce a same-session fallback or unsupported labels. Switched to `models=None` with the OpenRouter roster commented as an adapter-qualified alternative, matching the English guide, and carried over the "non-sentinel does not mean OpenRouter-routed" warning.

**On the P2 suggestion — agreed, and your framing is more precise than mine.** "A timestamp or shared commit date is insufficient when both files can change together while stale text remains" is exactly the failure I reproduced: I built the commit-date check, ran it against the commit that introduced this file with five stale sections, and it reported FRESH. Recorded with the reproduction on #1998, along with the same remedy you name — bind the translation to the English document's blob hash so a mismatch flags parity review regardless of commit timing.

Four rounds here, and the pattern is consistent: every finding is a section the English guide corrected during its own eight rounds, which this translation forked before. That is the argument for the marker, not for more careful reading.

Head is `4c98c0d`.

PR #2005, opened 2026-08-10
test(memory): stabilize xdist resource assertions

### LINKED ISSUE #1994: Three tests flake on wall clock; the picker budget is 1.3% under its real structural cost
Run [31338468922](https://github.com/Q00/ouroboros/actions/runs/31338468922) failed both `Test Python 3.12` and `Test Python 3.14` on PR #1989.

**#1989 changes two markdown files and nothing else** (`docs/runtime-guides/codex.ko.md`, `docs/runtime-guides/codex.md`). No test failure on it can be caused by the diff.

Both jobs ran to `[100%]`, so this is not the log-truncation infrastructure failure seen on #1952/#1961/#1972/#1984. These are real assertion failures, and all three assert against a wall clock or a resource budget with very little headroom.

### 1. `test_store_contention_preserves_event_loop_timeout_and_cancellation` (3.12, `gw2`)

```
assert 0.2451878179999767 < 0.2
 +  where 0.2451878179999767 = max(<genexpr>)
```

`tests/unit/orchestrator/test_disposable_memory.py:420`:

```python
assert max(b - a for a, b in pairwise(heartbeat)) < 0.2
```

The test spawns a heartbeat task and asserts the largest gap between consecutive pulses stays under 200 ms, proving the event loop was never blocked. The measured gap was 245 ms — **22% over** the limit.

The assertion is testing the right property. The problem is the margin: this runs under `pytest-xdist` on a shared GitHub runner alongside several other workers. A 45 ms scheduling hiccup on a loaded 2-core runner is ordinary, and it is indistinguishable here from the loop actually being blocked.

### 2. `test_timeout_after_commit_gate_waits_for_durable_publication` (3.12, `gw2`)

`assert False`, same file, same worker, adjacent test. Same run. Both of these use `timeout=0.1` against real time.

### 3. `test_picker_projection_write_and_disk_budgets[telemetry.unrelated-1.25-1.02]` (3.14, `gw3`)

```
assert 1.2768420878270539 < 1.25
 +  where 1.2768420878270539 = median([1.1208992593630367, 1.3359825523798579, 1.2768420878270539])
```

This one is the most telling. The test is parameterized over six event types (`test_picker_indexes.py:1356-1361`) and **only the tightest-budgeted parameter failed** — `telemetry.unrelated` at `1.25 / 1.02`. The other five, all at `1.75 / 1.25` or looser, passed in the same worker in the same run. The overshoot was 2%.

Note the median of the three samples is already 1.277 while the lowest sample is 1.121. The samples spread ~19%, and the budget allows 25% over baseline. **The measurement noise is nearly as large as the budget itself**, so this parameter fails whenever the middle sample lands high.

## Why this is worth fixing rather than rerunning

A docs-only PR cannot break these tests, so every failure here is a false signal. Concretely, it costs:

- **Trust in the signal.** Combined with the separate log-truncation failures, red CI on this repo currently does not mean "your change is broken." That is the expensive part — the correct response to a red check becomes "rerun it," which is also the wrong response when a failure is real.
- **Wall clock.** Each rerun is 8-12 minutes.

## Suggested direction

I have not measured which fix is right, so this is a direction and not a recommendation:

- For the two `disposable_memory` tests, the property under test is "the event loop was not blocked." That can be asserted without a wall clock, or with a margin sized to the worst observed CI scheduling delay rather than to local timings.
- For the picker budget, the sample count is 3 and the spread is ~19%. Either take more samples, or widen the `telemetry.unrelated` budget to match the noise floor the other five parameters already sit above.

Marking as flaky (`pytest.mark.flaky` / rerun-on-failure) is a stopgap that keeps the false-red cost while hiding it, so I would not lead with that.

## Evidence

- 3.12 job: [93308075147](https://github.com/Q00/ouroboros/actions/runs/31338468922/job/93308075147) — 2 failed, 19,220 passed, 703.89s
- 3.14 job: [93308075161](https://github.com/Q00/ouroboros/actions/runs/31338468922/job/93308075161) — 1 failed, 19,081 passed, 457.91s

Found while getting the Korean Codex runtime guide (#1988) through CI.

Q00:
Closes #1994

## Summary

- replace wall-clock-sensitive disposable-memory assertions with state-synchronized deadlines
- prove pre-commit cancellation settles without publishing
- prove post-commit timeout waits for durable publication
- keep matched-event budgets unchanged and raise only the structural CPU-time picker budget from 1.25 to 1.5

## Verification

- independent verifier PASS at exact HEAD `de59fb48bfdff9299d9463ba0ed91d2fba06901b`
- parent matched GitHub main `ecd266d507c0f3cbcc52634a2ac7ef1a28ca98f7`
- 10 fresh xdist synchronization rounds: 20/20
- complete changed scope under xdist: 23/23
- picker budget: 18/18 across three additional xdist rounds
- adversarial timeout-equivalence probes: 6/6
- Ruff, format, MyPy, and `git diff --check`: PASS

### DIFF
--- a/tests/unit/orchestrator/test_disposable_memory.py
+++ b/tests/unit/orchestrator/test_disposable_memory.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-from itertools import pairwise
 import json
 import multiprocessing
 import os
@@ -21,6 +20,7 @@ from ouroboros.core.disposable_memory import (
 from ouroboros.events.base import BaseEvent
 import ouroboros.orchestrator.agent_process as agent_process_module
 from ouroboros.orchestrator.agent_process import AgentProcessHandle
+import ouroboros.orchestrator.disposable_memory as disposable_memory_module
 from ouroboros.orchestrator.disposable_memory import DisposableMemory
 from ouroboros.persistence.artifact_store import (
     ArtifactManifestError,
@@ -68,6 +68,45 @@ def _service(tmp_path: Path) -> tuple[DisposableMemory, _EventStore]:
         checkpoint_store=CheckpointStore(tmp_path / "checkpoints"),
     )
     return service, event_store
+
+
+class _ControlledDeadline:
+    def __init__(self) -> None:
+        self._task: asyncio.Task[Any] | None = None
+        self._expired = False
+
+    async def __aenter__(self) -> None:
+        self._task = asyncio.current_task()
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        _exc: BaseException | None,
+        _traceback: Any,
+    ) -> None:
+        if self._expired and exc_type is asyncio.CancelledError:
+            raise TimeoutError from None
+
+    def expire(self) -> None:
+        assert self._task is not None
+        self._expired = True
+        self._task.cancel()
+
+
+def _install_controlled_deadline(
+    monkeypatch: pytest.MonkeyPatch,
+    delay: float,
+) -> _ControlledDeadline:
+    deadline = _ControlledDeadline()
+    original_timeout = asyncio.timeout
+
+    def controlled_timeout(requested_delay: float | None) -> Any:
+        if requested_delay == delay:
+            return deadline
+        return original_timeout(requested_delay)
+
+    monkeypatch.setattr(disposable_memory_module.asyncio, "timeout", controlled_timeout)
+    return deadline
 
 
 def _run_disposable_process(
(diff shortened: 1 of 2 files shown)


Q00:
## Exact-head verification after latest-main rebase

Published exact HEAD: `244f693389ee6d79caf33058ead7759f9c73c9db`

- parent/live main at verification end: `93e66a3d20690589f2bf0d2cfb4ddad07296faf0`
- independent verifier: PASS
- stable patch-id unchanged across both rebases: `da747a4e3dcf018fe81fec157b300233e579f48f`
- diff remains test-only (2 files); no production source change
- Python 3.12 changed scope: 60 passed
- Python 3.14 changed scope: 60 passed
- independent native pre/post-commit timeout probes: PASS
- xdist cancellation boundary and all picker parameters: PASS
- only unmatched `telemetry.unrelated` CPU ceiling changes 1.25→1.5; matched CPU and all disk ceilings remain unchanged
- Ruff, format, MyPy, diff/no-production-change gates: PASS

The test-only controlled-deadline helper still differs from native `asyncio.timeout` in completed-task cancellation count; this is the official review's valid low-priority fidelity note. Independent probes exercise native timeout semantics and confirm the production boundary, so it remains non-blocking.

Merge remains gated on fresh CI and a new official bot review whose API `commit_id`, body `HEAD checked`, and `Review-Metadata head_sha` all equal this exact HEAD with blocker 0 and explicit Merge/Approve recommendation.

Q00:
Reporter of #1994 here. I have been collecting failure data on these all day across docs-only PRs, so here is what I can contribute: independent evidence that the picker budget is right, and a note that `Closes #1994` will close three tests this PR does not touch.

## The 1.25 → 1.5 picker budget is well-sized

I pooled every sample this test printed across its **three** failures today (each failure reports all three of its ratios):

```
1.0854  1.1209  1.2407  1.2557  1.2707  1.2768  1.2782  1.3360  1.3978
```

| | |
|---|---|
| old budget | 1.25 — **6 of 9 samples over** (67%) |
| pooled median | **1.2707**, i.e. above the old ceiling |
| observed max | 1.3978 |
| new budget | 1.5 — **0 of 9 over**, 7.3% headroom above the worst observed |

That independently supports the change, and it supports the framing in your summary: this is a **structural CPU-time** cost, not flakiness. `_measure_bulk_append` times with `time.process_time()`, so runner contention is largely excluded, and the cost is accounted for — the extra `picker_projection_version` column on 100,000 rows plus the 100,000-element `relevant = [...]` scan in `insert_events_with_picker_projection`. Raising the ceiling here is not hiding a signal.

Keeping the matched-event budgets untouched is also right: across every run I observed, only `telemetry.unrelated` ever failed. The other five parameters passed in the same workers on the same runs.

## `Closes #1994` will close three tests this PR does not fix

The issue accumulated **six** distinct tests. This PR covers three of them — the two in `test_disposable_memory.py` and the picker budget. The other three are still open:

1. `tests/unit/cli/test_setup.py::TestClaudeSetup::test_setup_claude_serializes_competing_operator_generations` — 5-second budget for a child interpreter to import `ouroboros.cli.runtime_activation` (`:5785-5793`). **Seen twice**, on #1972 and #2006. It also masks its own diagnosis: the assertion *message* expression is `first.communicate(timeout=1)` on a child deliberately blocked until released, so `TimeoutExpired` propagates instead of the assertion and the failure reads as a 1-second subprocess timeout.

2. `tests/unit/orchestrator/test_parallel_executor.py::TestParallelACExecutor::test_consumes_stage_batches_sequentially_within_stage_boundaries` — four `timeout=1` waits (`:16068`, `:16072`, `:16150`, `:16155`) in a file that elsewhere uses `timeout=30`.

3. `tests/integration/plugin/test_orchestration.py::TestAgentPoolIntegration::test_agent_pool_cancels_stuck_task` — `health_check_interval=0.05` / `task_timeout=0.1`; fails as `KeyError` on the task id when cancellation and result retrieval race, which reads like a logic bug rather than a timing one.

Plus one qualitatively different observation, `test_ralph_agent_process_pause_inspect_resume_preserves_event_tail`, where a replay saw `['wait']` instead of `['continue', 'wait']` — a later event visible while an earlier one from the same aggregate was not. I flagged that as needing someone who owns the subsystem rather than diagnosing it.

Not asking you to widen this PR — the three you fixed are a coherent unit and the picker analysis is exactly right. Just suggesting `Refs #1994` instead of `Closes`, or closing it with a note that the remaining four are being tracked separately, so they do not disappear silently.

Happy to open a follow-up issue with the four if that is easier.

Q00:
Opened #2010 with the four remaining tests, as offered — so `Closes #1994` here is now accurate rather than lossy. No change needed on this PR.

For the record, #2010 carries: the `test_setup_claude_serializes_...` 5-second interpreter budget (**and** the separate defect that its assertion message expression raises `TimeoutExpired`, erasing the real cause), the four `timeout=1` waits in `test_parallel_executor.py` sitting alongside `timeout=30` in the same file, the `AgentPool` cancellation race that surfaces as `KeyError`, and the event-tail observation I flagged as needing a subsystem owner rather than a timing fix.

Q00:
## Final stable-main exact-head update

Published exact HEAD: `2bd792a5fb78fd303b71c6943a16720b12bb681d`  
Exact parent/live main: `9cfe73cc095aecc8fee2068bd657cf1fc72f47b2`

Independent verifier: STRICT PASS.

- stable patch-id remains `da747a4e3dcf018fe81fec157b300233e579f48f`
- raw binary diff remains byte-identical to the prior independently verified candidate
- diff is still exactly 2 test files; no production source change
- Python 3.12 and 3.14 changed scope: 60/60 each
- independent native pre/post-commit timeout probes: PASS
- xdist cancellation boundary and all picker parameters: PASS
- unmatched `telemetry.unrelated` CPU ceiling alone changes 1.25→1.5
- matched CPU and all disk ceilings remain unchanged
- Ruff, format, MyPy, diff/no-src gates: PASS

The controlled-deadline cancellation-count fidelity difference remains the known test-only Low-priority note and is independently separated from native timeout semantics.

Prior CI/reviews are stale. Merge remains gated on this exact SHA's fully green required CI and an official review with matching API `commit_id`, body `HEAD checked`, and `Review-Metadata head_sha`, blocker 0, and explicit Merge/Approve recommendation.

Q00:
## Re-review requested — final exact HEAD

Please re-review exact HEAD `e0c6f9049cee99b722aff7bb8aa0f8cde462d4a6`, based directly on current main `a5d383b9eba0c28ce0858e78652a8d526791ef12`.

This is a pure latest-main rebase of the previously approved test-only patch:

- stable patch-id `da747a4e3dcf018fe81fec157b300233e579f48f` unchanged
- raw binary diff byte-identical
- Python 3.12 and 3.14 changed scope: 60/60 each
- native pre/post-commit timeout probes and xdist boundary/picker probes: PASS
- Ruff, format, MyPy, diff/no-production-source: PASS
- independent verifier: STRICT PASS

Prior review bodies target older SHAs and are not merge authority for this head.

Q00:
## Re-review requested — release-main exact HEAD

Please re-review exact HEAD `caddd1ca121ef9930b291c7204f0df7400fef5a1`, based directly on release main `53132a0d4f0bde815dc8ef8869040c9fba64b61c`.

This remains byte-identical to the previously approved test-only patch:

- stable patch-id `da747a4e3dcf018fe81fec157b300233e579f48f`
- raw binary diff unchanged; only 2 tests, no production source
- Python 3.12 and 3.14: 60/60 each
- native pre/post-commit timeout and xdist boundary/picker probes: PASS
- Ruff, format, MyPy, diff/no-src: PASS
- independent verifier: STRICT PASS

Prior review bodies target older SHAs and are not merge authority for this head.

Q00:
## Re-review requested — residual settlement guard fixed

Please review exact HEAD `e2c8ad5d691cd1a3e07229ba81882c7762c850b1`, based directly on current main `53132a0d4f0bde815dc8ef8869040c9fba64b61c`.

This removes the remaining unsupported two-second scheduler assumption at the durable-publication settlement boundary. The release event remains the synchronization boundary; a 30-second timeout is now only a finite deadlock guard, while the test still awaits publication and retry completion and asserts one execution, identical results, and a fetchable artifact.

Independent exact-head evidence: the target case passed 12/12 under simultaneous picker contention; the disposable suite passed 85/85; changed-file xdist scope passed 60/60; Ruff, MyPy, module-size, and diff checks passed.

PR #2007, opened 2026-08-10
Korean Codex guide, part 3 — completes the translation

Q00:
Rebuilds #1993 on top of #2006 after #1989 squash-merged and broke the stack. Same content.

Completes the Korean Codex guide: Troubleshooting, Cost, and Active Conductor / Synapse. Both partial-translation banners come off, since `codex.ko.md` now covers the whole English document.

## Three facts ported into the English guide

While finishing this I ran the citation-set comparison **in both directions**, not just English-to-Korean. `codex.ko.md` carried 6 source citations against `codex.md`'s 2, and four existed only in Korean. Three marked facts an English reader could not get, so they are now in `codex.md` as well:

1. **Setup's global artifacts.** `ensure_config_dir()` creates `~/.ouroboros/data/` and `~/.ouroboros/logs/` (`cli/commands/setup.py:2632`), and a fresh config gets `~/.ouroboros/credentials.yaml` at mode `0600` (`:2771`). The English list covered `~/.codex/` only.

2. **The Synapse capability table.** `SessionSignalCapabilities` has six fields and Codex declares three (`orchestrator/codex_cli_runtime.py:453`). The English prose named `inform` and `after_turn` as supported and `redirect`/`replace` as not advertised, leaving `background_reply` and `owned_turn_abort` unaccounted for — so a reader could not answer the question they actually have. The table plus the consequence: **to change direction during a long turn, you wait for that turn to finish.**

3. **`subagent_orchestration=INTERNAL` and why** (`:448`). `codex mcp-server` exposes only `codex` and `codex-reply`, so Codex's native multi-agent team tools are unreachable by an external driver. Ouroboros can continue a Codex thread but cannot orchestrate Codex children. Users who expect fan-out hit this and nothing explained it.

That direction of drift is worth naming: translation is where the source gets read closely, and the English file only gets updated when a review points at it. Checking English-to-translation finds what the translation predates; checking translation-to-English finds what the translation learned and the source never received. I would not have looked for the second one.

## Closes the Codex track in #1988

Next on that issue: `copilot.md` and `kiro.md` — both now merged (#1997, #1999).

### DIFF
--- a/docs/runtime-guides/codex.ko.md
+++ b/docs/runtime-guides/codex.ko.md
@@ -7,7 +7,7 @@ doc_metadata:
 
 > English: [codex.md](./codex.md)
 >
-> **번역 진행 상황**: 명령 표면 · 동작 방식 · 런타임 차이 · CLI 옵션 · Seed 레퍼런스까지 옮겼습니다. 문제 해결 · 비용 · Active Conductor는 아직 영문입니다 — [codex.md](./codex.md)를 보세요. 진행은 [#1988](https://github.com/Q00/ouroboros/issues/1988)에서 추적합니다.
+> 영문 원문 전체를 옮긴 완역본입니다. 원문이 갱신되면 이 문서도 함께 갱신해야 합니다.
 
 > 설치와 첫 실행 흐름 전체는 [Getting Started](../getting-started.md)(영문)를 보세요.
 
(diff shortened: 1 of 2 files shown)


Q00:
Closed by the #2006 squash-merge breaking the stack again. Rebuilt on `main` as #2008 with the same content plus one more English-side gap the bidirectional citation check caught on the rebuild: the `CLI path:` verification note, which the Korean guide received during #1989 review and the English guide never did.

Q00:

PR #2018, opened 2026-08-10
fix(skills): resolve compatible Python for plugin flows

### LINKED ISSUE #2001: Plugin path is documented as Python-free but the bundled skills shell out to python3
The marketplace plugin is documented and marketed as the path that needs no Python. The bundled skills do not hold up that promise: they shell out to `python3` unconditionally.

## Where

Twelve `python3` invocations across three shipped skills:

| Skill | Count | Example |
|---|---:|---|
| `.claude-plugin/skills/welcome/SKILL.md` | 7 | `:68` — readiness detection, `if python3 - "$HOME/.ouroboros/config.yaml" <<'PY'` |
| `.claude-plugin/skills/setup/SKILL.md` | 4 | `:98` — recording the first-run preference into `~/.ouroboros/prefs.json` |
| `.claude-plugin/skills/seed/SKILL.md` | 1 | |

Both `:68` and `:98` are on the **first-run** path, so this is hit immediately rather than in some later corner.

## Why `uvx` does not cover it

`.claude-plugin/.mcp.json:4` launches the MCP server with:

```
uvx --python '>=3.12' --from 'ouroboros-ai[mcp]' ouroboros mcp serve ...
```

That provisions an interpreter **for the isolated MCP process**. It does not put `python3` on the user's `PATH`, and the skill snippets above are plain shell that resolve `python3` themselves.

So the exact host configuration the onboarding documents describe — Claude Code plus `uvx`, no system Python — fails during its own first `ooo setup` / welcome flow.

## Why this is easy to miss

Every layer is individually correct. The manifest correctly pins an interpreter for the server. The skills correctly use Python for JSON manipulation that would be painful in shell. The documentation correctly says the plugin does not require a `pip install` of Ouroboros.

The gap is between the two Pythons: the one uv provisions for the MCP subprocess, and the one the skill snippets invoke from the shell. Nothing in the repository connects them, and nothing tests the combination.

## Suggested direction

Two shapes, and I have not measured which fits better:

1. **Make the skills resolve an interpreter.** Try `python3`, then `python`, then `uv run python` — the last works on any host that already satisfies the `uvx` requirement, which makes uv the single prerequisite the docs claim it is.
2. **Declare `python3` a plugin prerequisite** and stop describing this path as Python-free.

Option 1 is the one that makes the documentation true rather than making it longer, so it seems preferable — but it touches three skills and needs a regression test that exercises a host with no global Python, which is the part that does not exist today.

Whichever way it goes, the test matters more than the fix: without a no-global-Python job, this reappears the next time a skill needs to parse JSON.

## Documentation status

#1987 now states `python3` as a prerequisite in `claude-code.ko.md`, `claude-code.md`, and `getting-started.md`, with the `uvx`-does-not-cover-it note. That describes the current behaviour; it is not a fix.

Q00:
## Summary

- Restore the marketplace promise that uv is the only host runtime prerequisite.
- Resolve a compatible Python 3.12 or newer for welcome, setup, and seed flows instead of calling python3 unconditionally.
- Keep canonical root skills and the actually shipped Claude plugin copies mechanically aligned.

## Changes

- Add one quoted resolver contract to all six root and shipped skill files.
- Prefer compatible python3, then compatible python, then uv provisioning with Python 3.12 or newer.
- Preserve arguments and heredoc stdin while failing clearly when no supported executor exists.
- Add byte-alignment, direct-bypass, adversarial PATH, uv-only first-run, wheel, and marketplace source checks.
- Align English, Korean, and Chinese prerequisite documentation with the implemented uv-only contract.

## Verification

- Focused artifact suite: 46 passed.
- Full suite: 19,802 passed, 94 skipped, 1 expected xfail.
- Actual uv-only, old-global-Python, quoting, heredoc, and injection-like argument probes: PASS.
- Built wheel and marketplace shipped-source checks: PASS.
- Ruff, format, MyPy, diff-check, ancestry, bundle, and clean-tree: PASS.
- Independent exact-head verifier: PASS at d082a6b3647c7644c0ff1d0bb7905fbdb3ae0529 on main 25f958dd7938d3c383ccfd14d551467bcf6e6bd6.

Closes #2001

### DIFF
--- a/.claude-plugin/skills/seed/SKILL.md
+++ b/.claude-plugin/skills/seed/SKILL.md
@@ -33,6 +33,38 @@ ooo seed [session_id]
 
 When the user invokes this skill:
 
+### Python Runtime (Required)
+
+Before running any shell snippet below, define this resolver in the same shell.
+It accepts only Python 3.12 or newer, prefers `python3` and then `python`, and
+uses uv as the final fallback. Call `ouroboros_python` directly and quote every
+argument passed to it; the function preserves arguments and heredoc/stdin input.
+
+<!-- ouroboros-python-resolver:start -->
+```bash
+ouroboros_python() {
+  if command -v python3 >/dev/null 2>&1 &&
+    command python3 -c 'import sys; raise SystemExit(sys.version_info < (3, 12))' >/dev/null 2>&1
+  then
+    command python3 "$@"
+    return
+  fi
+  if command -v python >/dev/null 2>&1 &&
+    command python -c 'import sys; raise SystemExit(sys.version_info < (3, 12))' >/dev/null 2>&1
+  then
+    command python "$@"
+    return
+  fi
+  if command -v uv >/dev/null 2>&1; then
+    command uv run --no-project --quiet --python '>=3.12' python "$@"
+    return
+  fi
+  printf '%s\n' 'Ouroboros skills require Python >= 3.12 or uv on PATH.' >&2
+  return 127
+}
+```
+<!-- ouroboros-python-resolver:end -->
+
 ### Load MCP Tools (Required before Path A/B decision)
 
 The Ouroboros MCP tools are often registered as **deferred tools** that must be explicitly loaded before use. **You MUST perform this step before deciding between Path A and Path B.**
@@ -408,7 +440,7 @@ Then check `~/.ouroboros/prefs.json` for `star_asked`. If `star_asked` is not se
 Create `~/.ouroboros/` directory if it doesn't exist. Preserve existing keys such as `welcomeShown`, `welcomeCompleted`, and `welcomeVersion` when updating `star_asked`:
 
 ```bash
-python3 - <<'PY'
+ouroboros_python - <<'PY'
 import json, os
 path = os.path.expanduser('~/.ouroboros/prefs.json')
 os.makedirs(os.path.dirname(path), exist_ok=True)
--- a/.claude-plugin/skills/seed/SKILL.md
+++ b/.claude-plugin/skills/seed/SKILL.md
@@ -44,19 +44,19 @@ argument passed to it; the function preserves arguments and heredoc/stdin input.
 ```bash
 ouroboros_python() {
   if command -v python3 >/dev/null 2>&1 &&
-    command python3 -c 'import sys; raise SystemExit(sys.version_info < (3, 12))' >/dev/null 2>&1
+    (unset PYTHONHOME; command python3 -c 'import sys; raise SystemExit(sys.version_info < (3, 12))') >/dev/null 2>&1
   then
-    command python3 "$@"
+    (unset PYTHONHOME; command python3 "$@")
     return
   fi
   if command -v python >/dev/null 2>&1 &&
-    command python -c 'import sys; raise SystemExit(sys.version_info < (3, 12))' >/dev/null 2>&1
+    (unset PYTHONHOME; command python -c 'import sys; raise SystemExit(sys.version_info < (3, 12))') >/dev/null 2>&1
   then
-    command python "$@"
+    (unset PYTHONHOME; command python "$@")
     return
   fi
   if command -v uv >/dev/null 2>&1; then
-    command uv run --no-project --quiet --python '>=3.12' python "$@"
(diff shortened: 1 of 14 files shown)


Q00:
@ouroboros-agent please perform a fresh formal review of exact remote HEAD d082a6b3647c7644c0ff1d0bb7905fbdb3ae0529, based directly on protected main 25f958dd7938d3c383ccfd14d551467bcf6e6bd6.

This candidate closes #2001 by making uv the sole host runtime prerequisite while preserving compatible global Python as an optional fast path:
- six root and actually shipped welcome, setup, and seed resolver blocks are byte-identical
- compatible python3, then compatible python, then exact uv Python 3.12-or-newer provisioning
- old global interpreters fall through safely
- quoted arguments, heredoc stdin, spaces, globs, and injection-like payloads remain literal
- no direct python3 bypass remains outside the resolver contract
- marketplace plugin source and built wheel canonical source are both verified
- English, Korean, and Chinese docs no longer claim host Python is required

Exact-head evidence:
- focused artifact suite: 46 passed
- full suite: 19,802 passed, 94 skipped, 1 expected xfail
- uv-only and adversarial PATH first-run flows across all six sources: PASS
- wheel and marketplace payload checks: PASS
- Ruff, format, MyPy, diff, ancestry, bundle integrity, clean-tree: PASS
- independent exact-head verifier: PASS, blocker 0

Please inspect shell quoting, interpreter version selection, uv fallback recursion, shipped-source identity, and documentation truthfulness. Merge remains frozen until exact-head official approval, fully green same-head CI, no human changes requested, latest-main ancestry, and independent verification all hold.

Q00:
@ouroboros-agent please perform a fresh formal review of exact remote HEAD e5b9516bc1165a09043df0e67e7b9d6c0a67df36, based directly on current protected main 76a16fb66466aadee39e5f8c7ece296b7c78b1e3, and supersede CHANGES_REQUESTED review 4893797941 plus every stale review/body.

The official PYTHONHOME blocker was accepted and generalized only where real CPython failures were independently reproduced. The shared resolver now isolates five interpreter-shaping variables for every python3/python version probe, every selected global interpreter invocation, and the uv-managed final invocation:
- PYTHONHOME
- PYTHONPATH
- PYTHONPLATLIBDIR
- PYTHONEXECUTABLE
- __PYVENV_LAUNCHER__

Each variable independently caused unsanitized uv-managed CPython to fail in exact runtime probes. With all five poisoned simultaneously, the current uv-only path starts Python >=3.12 successfully, the child sees all five absent, and the caller shell retains every original value. Sanitization occurs inside the resolver subshell; argv, stdin, heredoc, and caller environment remain intact.

Independent exact-head evidence:
- predecessor poison reality matrix: 5/5 failures reproduced
- current poisoned selection paths: compatible python3; old python3 to compatible python; old globals to uv; uv-only first-run all PASS
- six root/shipped welcome/setup/seed resolver blocks byte-identical, SHA-256 0b77aa92d5917e72eab53ca293190b2afbbe6df7104c1430534ef0b29223d985
- all five variables are unset at all five boundaries in every block; no direct bypass remains
- adversarial argv containing command substitution, semicolon, backticks, glob, spaces plus heredoc stdin arrived byte-for-byte and did not execute
- no-python/no-uv returns clear exit 127 with no fatal interpreter output
- artifact 55 passed; resolver matrix 16 passed; broad skills/plugin 97 passed; wheel integration 2 passed; telemetry/config preservation 302 passed
- built wheel contains each shipped resolver exactly once and byte-identical to canonical source
- #1908 telemetry/install behavior is preserved; candidate has zero telemetry/config implementation diff
- Ruff, format, MyPy 569 modules, module-size, diff-check, latest-main ancestry, clean tree, and bundle integrity PASS
- independent exact-head verifier PASS, blocker 0

Please issue a review whose REST API commit_id, body HEAD checked, and Review-Metadata head_sha all match e5b9516bc1165a09043df0e67e7b9d6c0a67df36. Merge remains frozen until exact-head approval with blocker 0 and explicit Approve/Merge recommendation, no human CHANGES_REQUESTED, fully green same-head required CI, latest-main ancestry, and independent exact-head verification all hold.

PR #2019, opened 2026-08-10
fix(mcp)!: require credentials to serve MCP beyond loopback

Q00:
## Why

`ouroboros mcp serve` accepted no credentials on its network transports. An instance bound to a routable address let anyone who could reach the port call `ouroboros_execute_seed` — caller-supplied YAML run through a local agent runtime with that runtime's full file and shell authority, in a caller-supplied directory.

Reproducible with:

```bash
ouroboros mcp serve --transport streamable-http --host 0.0.0.0 --port 8080
```

The default (`stdio` on `localhost`) and every loopback bind were never exposed — the SDK enables DNS-rebinding protection for `127.0.0.1` / `localhost` / `::1` on its own. Plugin and Claude Desktop installs are unaffected.

## Root cause

Not "auth defaulted to `AuthMethod.NONE`". Authentication was **not implemented** for network transports:

- `credentials` never reached `call_tool` — nothing extracted them from HTTP headers.
- `serve()` raised `ValueError` if any auth method was configured, so turning auth on made the server refuse to start.
- No `TransportSecuritySettings` were passed, and the SDK only synthesizes them for three literal host spellings. Every other bind, including `0.0.0.0`, ran with `Host`/`Origin` validation off.

## What changed

- `serve()` refuses a non-loopback bind without credentials, naming the capability at risk rather than failing abstractly.
- Bearer-token auth is wired through the SDK's `TokenVerifier`/`AuthSettings` to Ouroboros's existing `Authenticator`, so one credential authority covers both the HTTP edge and the per-tool `SecurityLayer`. The SDK's decision crosses back as an `AuthContext`, restoring the client identity authorization and rate limiting key on — **rate limiting is no longer blanket-refused**.
- Explicit `TransportSecuritySettings` for every bind the SDK does not auto-protect.
- `execute_seed`'s `cwd` can be confined to operator-chosen roots (`--workspace-root`). Checked *before* the existence test so a refusal cannot be used to probe paths.
- CLI: `--auth-token` (env `OUROBOROS_MCP_AUTH_TOKEN`), `--allow-remote`, `--allowed-host`, `--allowed-origin`, `--workspace-root`, and a startup banner stating the actual posture.
- `docs/cli-reference.md` no longer demonstrates a bare `--host 0.0.0.0`.

## Breaking change

A remote bind now needs `--auth-token`, `--allow-remote`, and `--allowed-host` for wildcard binds. An existing `--host 0.0.0.0` command refuses to start until those are supplied. **`stdio` and loopback binds are unchanged** — no credential, no new flags.

## Verification

End to end against a real `0.0.0.0` bind (not mocks), re-run after rebasing onto current `main`:

| request | result |
|---|---|
| anonymous | `401` |
| wrong token | `401` |
| correct token | `200` + full `initialize` handshake |
| forged `Host: attacker.example.com` | `421` |

Tests: 61 new in `tests/unit/mcp/server/test_network_security.py`; `tests/unit/mcp/server` + `tests/unit/cli` green (2287 passed); `ruff` and `mypy` clean.

## Note for reviewers

The rebase onto `main` interacts with #1908 — telemetry split `call_tool` into an observer wrapper plus `_call_tool_impl`, and moved the SDK dispatch body into `telemetry_boundary.call_sdk_tool`. The auth-context bridge had to follow into both. Git's auto-merge produced a version where `call_tool` accepted `auth_context` and silently dropped it while `_call_tool_impl` referenced an undefined name; that is resolved here, and the e2e run above is the proof the live path still carries the identity.

https://claude.ai/code/session_01MzysvKLDuJW9bkKQFBnP5B

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -1136,6 +1136,11 @@ ouroboros mcp serve [OPTIONS]
 | `-h, --host TEXT` | Host to bind to (default: localhost) |
 | `-p, --port INTEGER` | Port to bind to (default: 8080) |
 | `-t, --transport TEXT` | Transport type: `stdio`, `sse`, or `streamable-http` (default: stdio). Note: `http` is only a client config alias for outbound MCP connections and is NOT a valid serve transport. |
+| `--auth-token TEXT` | Shared secret clients present as `Authorization: Bearer <token>`. Required for a network transport on a non-loopback host. Prefer the `OUROBOROS_MCP_AUTH_TOKEN` environment variable — a token on the command line is visible to every process on the machine through `ps`. |
+| `--allow-remote` | Acknowledges that a non-loopback bind exposes seed execution beyond this machine. Required alongside `--auth-token` to serve on a routable address. |
+| `--allowed-host TEXT` | `Host` header value clients will use, e.g. `ouroboros.internal:8080`. Repeatable. Required for wildcard binds (`--host 0.0.0.0`), whose reachable name cannot be inferred. A `:*` suffix allows any port. |
+| `--allowed-origin TEXT` | `Origin` header value to permit. Repeatable. Empty by default, which rejects every browser-originated request. |
+| `--workspace-root TEXT` | Confines seed execution to directories under this path. Repeatable. Strongly recommended for network binds; unset means a caller may name any existing directory on the machine as an agent working tree. |
 | `--db TEXT` | Path to the EventStore database file |
 | `--runtime TEXT` | Agent runtime backend for orchestrator-driven tools (`claude`, `claude-sdk`, `claude-cli`, `codex`, `opencode`, `hermes`, `gemini`, `copilot`, `goose`, `kiro`, `pi`, `gjc`, `antigravity`, `grok`, `zcode`). The MCP 2 server rejects SDK-backed `claude`/`claude-sdk`; use `claude-cli` for its out-of-process Claude worker. |
 | `--llm-backend TEXT` | LLM backend for interview/seed/evaluation tools (`claude_code`, `litellm`, `codex`, `copilot`, `opencode`, `gemini`, `goose`, `kiro`, `pi`, `gjc`). Affects which tool variants are instantiated |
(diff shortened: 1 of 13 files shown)


Q00:
Applied `no-issue`: this fix came from an email security report rather than a tracked issue, so there is no issue number to link. The user-facing description will ship in the v0.52.0 release notes.

Also pushed a follow-up to the module-size ratchet failure: `adapter.py` is grandfathered under #1797, so the network security policy moved into `mcp/server/auth.py` as `resolve_network_security()` rather than extending the adapter. `serve()` now just asks for the wiring and passes it to the SDK. adapter.py: 2657 → 2584 (budget 2612).

Q00:
Pushed a cleanup pass on top of the module-size fix:

- `_NETWORK_TRANSPORTS` was duplicated in `cli/commands/mcp.py` after the policy moved; the CLI now imports `NETWORK_TRANSPORTS` from `mcp/server/auth.py` so the transport list has one definition.
- The `auth.py` module docstring still pointed at `MCPServerAdapter.serve` as the place the refusal happens; it now names `resolve_network_security`, which is where it actually lives.
- Added four direct tests for `resolve_network_security` — it is a public entry point now, not just something `serve()` reaches through.

All checks were green on the previous head; this push is a re-run of the same suite over the cleanup.

Q00:
Fixed the IPv6 finding in `de6b0ac10` — the report was accurate and I reproduced it before changing anything:

```
build_auth_settings(host='::1')         -> ValidationError
build_auth_settings(host='2001:db8::1') -> ValidationError
build_transport_security(host='2001:db8::1') -> allowed_hosts ['2001:db8::1:8080', ...]
```

Root cause is the one you named: a bind address and its URL spelling differ for IPv6. Sockets take the bare literal and reject brackets; URLs and `Host` values need brackets or the trailing `:port` merges into the address.

Added `as_url_authority()` and applied it at both sites. The bare value still reaches the socket — only URL- and Host-shaped strings get bracketed:

```
::1          -> http://[::1]:8080
2001:db8::1  -> http://[2001:db8::1]:8080   allowed_hosts ['[2001:db8::1]:8080', '[2001:db8::1]:*']
127.0.0.1    -> http://127.0.0.1:8080       (unchanged)
```

Live probe against `--host ::1 --transport streamable-http` with `OUROBOROS_MCP_AUTH_TOKEN` set — the case that crashed before startup: **anonymous 401, correct token 200**.

Seven new tests in `TestIPv6Binds` cover the authority helper, `AuthSettings` construction for both IPv6 shapes, the inferred allowlist, and authenticated IPv6 loopback plus routable binds — including that a routable IPv6 bind without credentials is still refused.

Issue #2020, opened 2026-08-10
Assertion redaction is verbatim-only: transformed encodings survive into the retry hint


Q00:
Hidden-contract redaction is **verbatim-only**. It enumerates five fixed encodings of the assertion and removes them with `str.replace`. Any encoding produced by *transforming* the string — rather than quoting it — survives into the retry hint that goes back to the worker.

Raised by @hardness1020 in #1979 while reviewing the hidden-assertion work. The mechanism they proposed differs from what the code does (details below), but the conclusion holds and is reproducible.

## Where the boundary is today

`hidden_contract_variants` (`src/ouroboros/orchestrator/contract_redaction.py:10`) builds exactly five variants per hidden value:

```python
hidden, repr(hidden), shlex.quote(hidden), json_quoted, json_quoted[1:-1]
```

sorted longest-first, then `redact_hidden_contract_values` (`:32`) applies them as literal substring replacements (`:42`).

That set is applied to every fragment of the retry hint via `_sanitize_fragment` (`src/ouroboros/orchestrator/retry_hints.py:29`), including the 2 000-character harness output tail (`retry_hints.py:101-104`, tail size at `src/ouroboros/orchestrator/parallel_executor.py:1745`) and the fallback error tail (`retry_hints.py:110-115`).

The second pass, `redact_and_truncate_text` (`src/ouroboros/orchestrator/decomposition_policy.py:690`), does **not** help here: it matches bearer tokens and secret-like key/value pairs, and performs no whitespace, case, or control-character normalization before or after the contract pass.

## The gap

`verify_command` output is arbitrary program output. Several routine transformations break exact matching while leaving the assertion fully readable to a model:

| Transformation | Source | Verbatim match |
|---|---|---|
| Line wrapping at terminal width | any CLI writing to a pty | broken |
| Diff line prefixes (`+ `, `- `, `> `) | `diff`, `git diff`, snapshot tests | broken |
| ANSI colour codes injected mid-string | `pytest --color=yes`, `rg`, `jest` | broken |
| pytest assertion rewriting (`E   assert 'FOO' in ...` with continuation indents) | pytest | broken |
| Case normalisation / whitespace collapsing | log formatters | broken |
| HTML/XML entity escaping | tools emitting XML or JUnit reports | broken |

[...]

PR #2027, opened 2026-08-10
docs(zh): scope the hidden-assertion claim at the point it is made

Q00:
Relates to https://github.com/Q00/ouroboros/issues/2020.

`README.zh-CN.md` described the hidden-checklist doc as covering 为什么判分用的断言对 worker **无条件**隐藏.

The design doc is careful about that word. `docs/hidden-checklist-convergence/README.zh-CN.md:23` defines it explicitly:

> 这里的「无条件」指的是**没有配置项可以把它打开**，而不是说存在一个文件系统沙箱

That is accurate, and the distinction matters. But the qualifier lives at the destination, and the index line is what most readers see. Read alone, 无条件隐藏 says the assertion cannot reach the worker at all, which is a stronger claim than the code makes: redaction enumerates five encodings and replaces them literally (`orchestrator/contract_redaction.py:10`, `:42`), so a copy transformed by line wrapping, a diff prefix, or an ANSI code inside the string survives into the retry hint. That gap is #2020.

This moves the scope inline: 隐藏、且没有配置项能把它打开. Same meaning as the design doc, stated where the claim is made.

One line changed. No other content touched.

## Noticed while here, not fixed in this PR

`README.md` and `README.ko.md` do not link the hidden-checklist doc at all. `docs/README.md:52-53` lists both the English and Chinese versions, so only the top-level English and Korean READMEs are missing it. The Chinese README is ahead of the English one here, which is the reverse of the usual direction and worth a separate PR rather than widening this one.

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -390,7 +390,7 @@ src/ouroboros/
 完整设计文档见 [Architecture](./docs/architecture.md)（英文）。中文文档：
 
 - [评估流水线指南（Evaluation Pipeline）](./docs/guides/evaluation-pipeline.zh-CN.md) —— 三阶段关卡的完整参考：每个阶段验什么、阈值和配置项、失败模式与排查、以及事件审计轨迹
-- [隐藏清单收敛（Hidden-Checklist Convergence）](./docs/hidden-checklist-convergence/README.zh-CN.md) —— run → 评估 → 有预算的 Ralph 链，以及为什么判分用的断言对 worker 无条件隐藏
+- [隐藏清单收敛（Hidden-Checklist Convergence）](./docs/hidden-checklist-convergence/README.zh-CN.md) —— run → 评估 → 有预算的 Ralph 链，以及为什么判分用的断言对 worker 隐藏、且没有配置项能把它打开
 
 </details>
 


Q00:
Correcting the "Noticed while here" note in my own description. Please disregard it; I checked the structure afterwards and my reading was wrong.

I wrote that `README.md` and `README.ko.md` are missing the hidden-checklist doc and that the Chinese README is ahead of the English one. Neither holds.

The list that entry sits in is introduced as 中文文档 (`README.zh-CN.md:389`), right after 完整设计文档见 Architecture（英文）. It is an index of **what exists in Chinese**, not a table of contents for design docs. English readers are not missing an entry, because the English README deliberately keeps one link to `docs/architecture.md` (`README.md:432`) and lists no per-topic design docs at all. Adding hidden-checklist there would be inconsistent with how that README is organised, not a fix.

I also checked the same thing for Korean, since that is where this class of gap actually bit us earlier (#2011). It is clean in both languages:

| | translations on disk | linked from that README |
|---|---|---|
| Korean | 5 | 5 |
| Chinese | 2 | 2 |

So there is no missing-translation-link gap to open a follow-up PR for. Recording the negative result so nobody spends time re-deriving it.

The one-line change in this PR is unaffected.

Issue #2021, opened 2026-08-10
evolve: every generation fails 'Stage 2 failed: AC non-compliance' when the worker report format drifts


Q00:
Reported symptom: **every generation** of the evolve loop ends with

```
Stage 2 failed: AC non-compliance (score=<n>)
```

so the loop never converges. Traced end to end. The chain is nine steps and it starts with an exact-match regex over model-written prose.

## The chain

Every generation calls the evaluator (`evolution/loop.py:1926`), wired to `_evolution_evaluator` at `mcp/server/adapter.py:2246`.

1. `_evolution_evaluator` (`adapter.py:2039`) first tries the mechanical path: `_evaluate_mechanically` (`adapter.py:1957`), which is `_parse_legacy_execution_task_summary` (`adapter.py:547`).
2. That parser matches the execution output against one regex (`adapter.py:558-559`):

   ```python
   re.findall(r"### (?:Task|AC) (\d+): \[(COMPLETED|FAILED|PASS|FAIL)\]\s*(.*)", artifact)
   ```

   No match returns `None` (`adapter.py:561-562`). **Silently.** Nothing logs that the mechanical path was skipped.
3. `None` sends the generation to the LLM fallback (`adapter.py:2062`).
4. The fallback folds **every** acceptance criterion into a single string (`adapter.py:2064-2066`):

   ```python
   current_ac = "\n".join(f"AC {i + 1}: {ac}" for i, ac in enumerate(ac_texts(acs)))
   ```
5. Stage 2 asks one model for **one boolean** covering that whole blob: `ac_compliance` is a single `{"type": "boolean"}` field in the response schema (`evaluation/semantic.py:39`, parsed at `:282`).
6. With N criteria in one boolean, a single unmet criterion fails the entire generation, and nothing reports *which* one.
7. `trigger_consensus` is not set anywhere on the evolve path (`adapter.py:2072-2081` constructs `EvaluationContext` without it), so it defaults to `False`.
8. Stage 2 returning `False` therefore short-circuits the pipeline (`evaluation/pipeline.py:178`):

   ```python
   if not stage2_result.ac_compliance and not context.trigger_consensus:
       return self._build_result(..., final_approved=False)
   ```

   **Stage 3 never runs.** The consensus layer, including the reviewer-independence filter in `evaluation/reviewer_independence.py`, is unreachable from the evolve loop's failure path.
9. The message is built at `evaluation/pipeline.py:317`.

[...]

PR #2028, opened 2026-08-10
fix(evaluation): name the predicate in the Stage 2 failure reason

Q00:
Fixes defects 3 and 4 from https://github.com/Q00/ouroboros/issues/2021.

## The message named the wrong thing

```python
elif stage2_result and not stage2_result.ac_compliance:
    failure_reason = f"Stage 2 failed: AC non-compliance (score={stage2_result.score:.2f})"
```

The gate is `ac_compliance`, a boolean the semantic evaluator returns (`evaluation/semantic.py:39`). The message printed `score`, a float that does not decide this branch. A run can therefore fail with `score=0.92` on screen, which sends the reader looking at thresholds. The score threshold is real but lives at the approval site (`pipeline.py:248`, `>= 0.8`), a different branch entirely.

Before and after, same inputs:

```
Stage 2 failed: AC non-compliance (score=0.92)
Stage 2 failed: AC non-compliance (ac_compliance=false; semantic score 0.92 did not gate this)
```

This matters most on the evolve path, where Stage 1 is off by default and Stage 3 is hardcoded off (`mcp/server/adapter.py:1890`, `:1897`), so this line is the only explanation an operator gets for a generation.

## The ladder existed twice

The identical branch order appeared in `pipeline.py` (built before the result object exists) and in `EvaluationResult.failure_reason` (built after). Same strings, same ordering, same Stage-3-before-Stage-2 rationale in both docstrings. Fixing the message in one would have left the other, which is the shape of drift this repo keeps finding.

Both now call `build_failure_reason` in `evaluation/models.py`. It lives there because `REWARD_HACKING_VETO_THRESHOLD` already does, for the same stated reason: no `models` to `pipeline` import cycle.

Net: `pipeline.py` loses 22 lines of duplicated branching.

## Not changed

- Branch order and semantics are identical. Stage 1, then Stage 3, then Stage 2 compliance, then the reward-hacking veto.
- The `"AC non-compliance"` substring is preserved, so the two existing assertions still hold (`tests/unit/evaluation/test_checklist.py:230`, `tests/unit/evaluation/test_models.py:311`).
- The root cause in #2021 (the exact-match regex at `adapter.py:558` silently downgrading every generation to the LLM fallback) is untouched. This PR only stops the resulting message from misdirecting whoever reads it.

## Verification

- `tests/unit/evaluation/` — 611 passed.
- `ruff check`, `ruff format --check`, `mypy` clean on both files.
- Confirmed by construction that a `SemanticResult(score=0.92, ac_compliance=False)` produces the new text and still contains the asserted substring, and that an approved result still returns `None`.

### DIFF
--- a/src/ouroboros/evaluation/models.py
+++ b/src/ouroboros/evaluation/models.py
@@ -384,31 +384,51 @@ class EvaluationResult:
 
     @property
     def failure_reason(self) -> str | None:
-        """Return the reason for failure, if any.
+        """Return the reason for failure, if any."""
+        return build_failure_reason(
+            final_approved=self.final_approved,
+            stage1_result=self.stage1_result,
+            stage2_result=self.stage2_result,
+            stage3_result=self.stage3_result,
+        )
 
-        Stage 3 is checked before Stage 2 because when Stage 3 ran,
-        it is the authoritative verdict (Stage 2 may have been bypassed
-        via trigger_consensus).
-        """
-        if self.final_approved:
-            return None
-        if self.stage1_result and not self.stage1_result.passed:
-            failed = self.stage1_result.failed_checks
-            return f"Stage 1 failed: {', '.join(c.check_type for c in failed)}"
-        if self.stage3_result and not self.stage3_result.approved:
-            return (
-                f"Stage 3 failed: Consensus not reached ({self.stage3_result.majority_ratio:.0%})"
-            )
-        if self.stage2_result and not self.stage2_result.ac_compliance:
-            return f"Stage 2 failed: AC non-compliance (score={self.stage2_result.score:.2f})"
-        if (
-            self.stage2_result
-            and self.stage2_result.reward_hacking_risk >= REWARD_HACKING_VETO_THRESHOLD
-        ):
-            return (
-                "Stage 2 veto: reward-hacking risk "
-                f"{self.stage2_result.reward_hacking_risk:.2f} >= "
-                f"{REWARD_HACKING_VETO_THRESHOLD:.2f} — artifact appears optimized to game the "
-                "evaluator rather than solve the real task"
-            )
-        return "Unknown failure"
+
+def build_failure_reason(
+    *,
+    final_approved: bool,
+    stage1_result: MechanicalResult | None,
+    stage2_result: SemanticResult | None,
+    stage3_result: ConsensusResult | None,
+) -> str | None:
+    """Return why an evaluation was rejected, or ``None`` when it passed.
+
+    Stage 3 is checked before Stage 2 because when Stage 3 ran, it is the
+    authoritative verdict (Stage 2 may have been bypassed via
+    ``trigger_consensus``).
+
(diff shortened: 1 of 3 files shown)


Q00:
Both findings were correct and are fixed in `26bca688`. Rebased onto current `main` as well, since the branch predated #2022 and the diff was showing that file as removed.

## Blocker: the veto did not get named

The cause was that this function sees `final_approved` **after** `_build_result` has already applied the veto (`pipeline.py:286`), so it could not tell an approval that the veto rejected from a run that never got approved.

The veto only ever flips approve to reject, so it is the deciding gate exactly when the result would otherwise have passed. That is now reconstructed before the ladder runs:

```python
if stage3_result is not None:
    approved_before_veto = stage3_result.approved
elif stage2_result is not None:
    approved_before_veto = (
        stage2_result.ac_compliance and stage2_result.score >= SEMANTIC_APPROVAL_SCORE
    )
else:
    approved_before_veto = True
```

The veto branch then requires `approved_before_veto`, which keeps the two orderings that were already right:

- Consensus rejected plus high risk still reports Stage 3, because a rejected consensus was never approved.
- No consensus plus `ac_compliance=False` plus high risk still reports non-compliance, for the same reason.

Your reproduction case (`0.92` / compliance false / risk `0.95` / approving consensus) now returns the veto reason and no longer contains the string `AC non-compliance`.

## Follow-up: the score gate returned "Unknown failure"

Named, and the threshold is no longer a literal in two places. `SEMANTIC_APPROVAL_SCORE = 0.8` now lives in `models.py` beside `REWARD_HACKING_VETO_THRESHOLD`, for the reason that constant already gives, and `pipeline.py:249` imports it. Leaving `0.8` inline in the gate while the message that has to name it lived elsewhere would have recreated the drift this PR exists to remove.

`ac_compliance=True`, `score=0.79`, no consensus, low risk now returns:

```
Stage 2 failed: semantic score 0.79 < 0.80 (ac_compliance=true; the score gate decided this)
```

## Coverage

`tests/unit/evaluation/test_failure_reason_gates.py`, five cases:

| Case | Asserts |
|---|---|
| veto over approving consensus | starts with `Stage 2 veto:`, and `AC non-compliance` absent |
| score gate, no consensus | names the score and the threshold, is not `Unknown failure` |
| consensus rejected plus high risk | reports Stage 3, not the veto |
| no consensus, non-compliant, high risk | reports non-compliance, not the veto |
| approved runs (parametrized) | reason is `None` |

The last three are regressions for the orderings your review confirmed were already correct, so a future change to this ladder has to keep them.

`tests/unit/evaluation/` is 617 passed. `ruff check`, `ruff format --check`, `mypy` clean.

## On the event metadata

You asked that both the result and the persisted event be asserted. They are now the same value by construction: `pipeline.py` passes the output of this one function into `create_pipeline_completed_event`, and `EvaluationResult.failure_reason` calls the same function. There is no second implementation left to disagree, which was the other half of this PR. If you would still prefer an explicit end-to-end assertion through `evaluate()` on the mixed path, say so and I will add it.

PR #2036, opened 2026-08-10
docs(readme): surface the star count on the first screen

Q00:
Adds two coordinated star prompts to all three public READMEs:

- a GitHub star-count badge, placed first in each five-badge row; and
- a localized star CTA after Quick Start, where readers have enough context to decide whether the project is useful.

## Why

The existing badge row surfaced releases, CI, licensing, and sponsorship, but not the repository's adoption signal. The new CTA makes the next action explicit without pretending that GitHub exposes a direct-star URL.

## Renderer-safe behavior

`README.md` is also the PyPI long description. The CTA therefore names **Q00/ouroboros on GitHub** instead of referring to a button “at the top of this page,” so it remains true on both GitHub and PyPI.

The badge links to the verified repository root, `https://github.com/Q00/ouroboros`, because the `/stargazers` destination currently returns 404.

## Localization

The English, Korean, and Simplified Chinese CTAs preserve the same intent while using natural phrasing in each language.

## Verified

- All three READMEs contain five badges with the star badge first.
- The repository link and Shields SVG endpoint return HTTP 200.
- The three CTA fragments render with PyPI's `readme_renderer[md]` path.
- Built wheel and sdist pass `twine check`.
- Pre-commit file hooks and `git diff --check` pass.

## Scope

- Only `README.md`, `README.ko.md`, and `README.zh-CN.md` change: 20 insertions and no deletions.
- This documentation-only growth improvement intentionally has no linked issue; the `no-issue` label records that scope.
- Moving the demo or adding a star-history chart remains outside this PR.

### DIFF
--- a/README.ko.md
+++ b/README.ko.md
@@ -22,6 +22,7 @@
 </p>
 
 <p align="center">
+  <a href="https://github.com/Q00/ouroboros/stargazers"><img src="https://img.shields.io/github/stars/Q00/ouroboros?color=yellow&logo=github&label=stars" alt="GitHub stars"></a>
   <a href="https://pypi.org/project/ouroboros-ai/"><img src="https://img.shields.io/pypi/v/ouroboros-ai?color=blue" alt="PyPI"></a>
   <a href="https://github.com/Q00/ouroboros/actions/workflows/test.yml"><img src="https://img.shields.io/github/actions/workflow/status/Q00/ouroboros/test.yml?branch=main" alt="Tests"></a>
   <a href="LICENSE"><img src="https://img.shields.io/badge/license-MIT-green" alt="License"></a>
--- a/README.ko.md
+++ b/README.ko.md
@@ -269,8 +269,8 @@ ooo evaluate   →  3단계 검증: Mechanical → Semantic → Consensus
 
 
 <p align="center">
-  <sub>인터뷰가 물었는데 답 못 한 질문이 있었다면, 그게 남한테 말할 만한 부분입니다.<br/>
-  <a href="https://github.com/Q00/ouroboros"><strong>저장소에 star</strong></a>를 눌러두면, 자기가 시키지도 않은 코드를 리뷰하느라 지친 다음 사람에게 닿습니다.</sub>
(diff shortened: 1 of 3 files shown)


Q00:
Pushed a second commit: a star ask at the end of Quick Start, in all three READMEs.

There was no star ask anywhere in the README before this. The badge shows the count but never suggests adding to it.

**Placement.** Right after the Quick Start block, where a reader has either just run the interview or decided not to. That is the only moment the ask is earned. At the top it is a request from a stranger; here it follows something that either worked or did not.

**Wording.** It does not say "star us if you like it". It points at the specific thing worth telling someone about:

> If the interview asked you something you could not answer, that is the part worth telling people about. Star the repo so it finds the next person who is tired of reviewing code they did not ask for.

Two reasons for that phrasing. It names the moment people actually react to, and it gives a reason that is about the next reader rather than about our number.

**Localised, not translated.** The Korean and Chinese versions carry the same idea in their own phrasing rather than a literal rendering, since the English sentence leans on a rhythm that does not survive word-for-word.

**One thing worth knowing:** GitHub has no URL that stars a repo, so the link goes to the repo itself and the reader still clicks the real Star control. No badge or link can do it for them, and anything claiming otherwise is a redirect.

A star-history chart came up while measuring this and we decided against it for now: there is no reason to spend that vertical space on our own growth curve when the same space could carry the demo.

Q00:
Third commit, fixing two problems with the CTA I pushed an hour ago. Reviewing my own copy caught both.

**The link went nowhere.** `Star the repo` linked to `github.com/Q00/ouroboros`, which is the page the reader is already on. It promised an action and delivered a scroll to the top. GitHub has no URL that stars a repo, so any link there is decoration at best and a broken promise at worst. The link is gone; the text now points at the actual Star control.

**The ask excluded most of the people reading it.** It opened with "If the interview asked you something you could not answer", which only speaks to someone who has already installed and run it. Most readers at that point in the page have not. The condition now describes the problem rather than our product:

> Most people find out they were unclear about three files into the review. If that is a familiar feeling, the Star button at the top of this page is what puts this in front of the next person it would have saved.

Anyone who has reviewed a pull request recognises the first sentence, whether or not they ran anything.

Same change in the Korean and Chinese versions, phrased for each rather than translated.

PR #2039, opened 2026-08-10
fix(mcp): tell the default-runtime user to pass --runtime, not to reinstall

### LINKED ISSUE #2038: 'ouroboros mcp serve' on the default runtime reports a package-profile conflict instead of asking for --runtime (uvx part resolved by #2012)
`uvx --from 'ouroboros-ai[mcp]' ouroboros mcp serve` does not start. It stops on the package-profile guard, and the environment it builds does not contain what the `[mcp]` extra pins.

This is the path an MCP host would use, so it is worth separating from the general install story.

## Reproduce

```bash
uvx --refresh --from 'ouroboros-ai[mcp]' ouroboros mcp serve
```

```
Unsupported package profiles: ouroboros-ai[mcp] requires MCP 2, while
ouroboros-ai[claude] and ouroboros-ai[claude-sdk] require MCP 1.x. Use [claude]
alone for the Claude SDK runtime, or use [claude-cli] with [mcp]; run the MCP 2
server in a separate environment/process.
```

The message names a conflict with `[claude]`, but `[claude]` was never requested.

## What is actually in the environment

```bash
uvx --refresh --from 'ouroboros-ai[mcp]' python -c "
from importlib import metadata as m
for d in ['mcp','claude-agent-sdk','anthropic','ouroboros-ai']:
    print(d, m.version(d))"
```

```
mcp 1.29.0
claude-agent-sdk 0.2.123
anthropic 0.117.0
ouroboros-ai 0.51.1
```

Two things are off:

1. `mcp` resolves to **1.29.0**, but the published metadata for 0.51.1 pins `mcp==2.0.0; extra == "mcp"`. The extra does not appear to be taking effect.
2. `claude-agent-sdk==0.2.123` and `anthropic==0.117.0` are present. Those are exactly the pins the published metadata lists under `extra == "claude"` and `extra == "claude-sdk"`, neither of which was requested. `claude-agent-sdk` is not a base dependency.

So the guard is not firing on a bad user request. It is firing on an environment that contains the `claude` pins and MCP 1.x while the user asked only for `[mcp]`.

Reproduced with `--refresh`, so it is not a stale uv cache. `uvx --from 'ouroboros-ai[mcp]' ouroboros --help` and `ouroboros mcp serve --help` both work, because neither reaches the guard.

## What I have not checked

- Whether `pipx install 'ouroboros-ai[mcp]'` or a plain venv shows the same resolution. I only tested the uvx path, since that is the one an MCP client config would use.
- Whether this is a uv resolution behaviour rather than something in our metadata. The published `requires_dist` looks correct to me, which is what makes the resolved environment surprising.
- Whether `mcp==2.0.0` and `claude-agent-sdk` can coexist at all. If they genuinely cannot, then the guard is right and the packaging question is how `[mcp]` alone ends up dragging in the Claude pins.

## Why it matters beyond the error

`uvx --from ... ouroboros mcp serve` is the shape an MCP host config takes, and it is what a registry entry would point at. `docs/architecture.md:569` documents `ouroboros mcp serve` as the server mode and `docs/getting-started.md:169` documents `pip install 'ouroboros-ai[mcp]'`, so someone following either lands here.

Environment: macOS, uv installing its own interpreter for `requires-python >=3.12`, `ouroboros-ai` 0.51.1 from PyPI.

Q00:
Closes #2038

`ouroboros mcp serve` without an explicit runtime inherits the SDK-backed `claude` default. That runtime cannot execute inside the MCP 2 server process, but the command used to report the separate mixed-package-profile diagnostic and sent users toward reinstalling extras that were not the cause.

## Change

The two failure modes now have separate diagnostics:

- An actual MCP 2 + Claude SDK package mix keeps the canonical package-profile message.
- An inherited or explicit SDK runtime receives a runtime-selection message that names `--runtime claude-cli` and configuration as actionable alternatives.

The runtime-selection diagnostic intentionally makes no claim that the installation is healthy. Dependency validation happens later, so the early message stays true in no-MCP, MCP 1 + SDK, clean MCP 2, and mixed-profile environments.

## Why this remains after #2012

#2012 fixed the isolated `uvx` launcher/profile path. This PR addresses the remaining direct or bare `mcp serve` diagnosis when the selected runtime itself is not executable in the MCP server process.

## Verified

- Built-wheel console script: bare `mcp serve` returns the runtime-selection diagnostic without creating Ouroboros database, config, or session state.
- Built-wheel `--runtime claude-cli` and `--runtime codex`: real JSON-RPC `initialize` handshakes return a result and exit successfully.
- Actual no-MCP, MCP 1 + Claude SDK, clean MCP 2, and MCP 2 + Claude SDK environments exercise the intended diagnostic branches.
- Focused MCP tests: 163 passed, 6 skipped.
- Python 3.12, 3.13, and 3.14 focused regressions: 27 passed each.
- Ruff, format, MyPy, module-size, dependency checks, and `git diff --check` pass.

## Scope

This changes only the diagnosis and its regression coverage. It does not change runtime selection, dependency resolution, or the executable MCP worker implementations.

### DIFF
--- a/src/ouroboros/cli/commands/mcp.py
+++ b/src/ouroboros/cli/commands/mcp.py
@@ -33,6 +33,7 @@ from ouroboros.orchestrator.heartbeat import (
     process_start_time,
 )
 from ouroboros.package_profiles import (
+    SDK_RUNTIME_IN_MCP_SERVER_MESSAGE,
     UNSUPPORTED_CLAUDE_SDK_MCP_MESSAGE,
     has_unsupported_claude_sdk_mcp_mix,
     public_runtime_backend,
@@ -1203,8 +1204,16 @@ def serve(
     # missing option inherits config and ultimately defaults to the SDK-backed
     # ``claude`` runtime, which is not executable inside this MCP 2 process.
     selected_runtime = _effective_mcp_server_runtime(runtime)
-    if has_unsupported_claude_sdk_mcp_mix() or selected_runtime == "claude":
+    # Two different failures used to share one string. An environment that mixes
+    # MCP 2 with the Claude SDK really is a package-profile problem and the user
+    # must reinstall. Inheriting the ``claude`` default is a runtime-selection
+    # failure, and the fix is ``--runtime``. Telling that user to reinstall sent
+    # them to change extras that were never relevant to the selected backend.
+    if has_unsupported_claude_sdk_mcp_mix():
         _stderr_console.print(Text(UNSUPPORTED_CLAUDE_SDK_MCP_MESSAGE, style="red"))
+        raise typer.Exit(1)
+    if selected_runtime == "claude":
+        _stderr_console.print(Text(SDK_RUNTIME_IN_MCP_SERVER_MESSAGE, style="red"))
         raise typer.Exit(1)
 
     # Guard: prevent recursive MCP server spawning.
--- a/src/ouroboros/package_profiles.py
+++ b/src/ouroboros/package_profiles.py
@@ -26,6 +26,14 @@ UNSUPPORTED_CLAUDE_SDK_MCP_MESSAGE = (
 )
 
 
+SDK_RUNTIME_IN_MCP_SERVER_MESSAGE = (
+    "The MCP server cannot host the 'claude' SDK runtime in this process. "
+    "Pass an executable runtime, for example "
+    "'ouroboros mcp serve --runtime claude-cli', or set the runtime in config "
+    "so this command stops inheriting the 'claude' default."
+)
+
+
 class PublicAgentRuntimeBackend(str, Enum):  # noqa: UP042
     """Public runtime spellings shared by every workflow CLI front door.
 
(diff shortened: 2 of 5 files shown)


Q00:
Second commit. I opened this PR with four tests failing, which I should have caught before pushing rather than after.

**What broke and why.** Four tests in `test_mcp_nested_guard.py` asserted the package-profile string on the default-runtime path. That is the behaviour this PR deliberately changes, so the tests were pinning the bug. They now assert the new message on that path and additionally that the output does **not** contain `ouroboros-ai[mcp]`, since sending a correctly-installed user back to the extras is the actual defect.

`test_forced_sdk_mcp_mix_fails_before_process_state` is untouched. It patches `has_unsupported_claude_sdk_mcp_mix` to `True`, which is the genuine packaging case, and it should keep asserting the old message. Two failures, two messages, two tests.

`tests/unit/cli/test_mcp_nested_guard.py` and `tests/unit/test_package_profiles.py`: 24 passed.

**One pre-existing failure I am not touching.** `test_mcp_doctor.py::TestCheckMcpImport::test_reports_the_installed_profile_version` fails on `main` as well, with `PackageNotFoundError` from `importlib.metadata`. I checked it against a clean `main` checkout before assuming it was mine. It looks environment-dependent rather than related to this change, so I have left it alone rather than fold an unrelated fix into this PR.

PR #2040, opened 2026-08-10
docs(plugin): one description for both manifests, saying what the directory does not already have

Q00:
Rewrites `description` in `.claude-plugin/plugin.json`. One field, one line.

## Why now

`anthropics/claude-plugins-community` is the official directory, and we are not in it even though the manifest has been sitting in this repo ready to go. Before submitting I read what is already there.

**2,291 plugins are listed. 44 of them describe themselves as spec-driven development.** A sample: `specops`, `sdd-superpowers`, `spec-engine`, `specclaw`, `waterfall`, `asdd`, `don-cheli-sdd-framework`, `spec-driven-development`, `sdlc-kit`, `blueprint`, `claude-forge`.

Our line was:

> Self-improving AI workflow system. Crystallize requirements before execution with Socratic interview, ambiguity scoring, and 3-stage evaluation.

Every noun in it appears in those 44. Someone scanning the directory has no way to tell us apart, and the one thing none of the other 44 mention was missing from ours too.

## Two problems, not one

**It did not say the thing only we do.** The grading contract staying out of the executing agent's prompt is absent from all 44 descriptions I read. It was absent from ours as well.

**"3-stage evaluation" is not true everywhere.** It holds on the evaluate path. On the evolve path stage 1 is off by default and stage 3 is hardcoded off, so each generation is one model returning one boolean (#2021). A directory entry is read by people who then open the repo.

## The new line

> Sits in front of Claude Code and fixes the brief before the code. An interview scores how vague your request is and refuses to generate a spec until it clears the bar, though you can force past it. The acceptance criteria are then pinned and verified independently, and the command and assertion used to grade the work never enter the executing agent's prompt, with no flag to put them back.

391 characters. The directory's median is 274 and the mean is 357, so this sits in normal range; the old one used 144 of the space available.

Deliberate choices in it:

- **"though you can force past it"** — the README demo's last frame shows `Generate Seed anyway (force)`. A description that called the gate a lock would be contradicted by our own screenshot.
- **"with no flag to put them back"** — the accurate scope. Not "the agent can never see it", which would be an overclaim: the redaction is verbatim over five encodings and a reshaped assertion still gets through (#2020).
- No "self-improving", no "workflow system". Those words are doing nothing for a reader choosing between 45 similar entries.

## Not changed

`.codex-plugin/plugin.json` still reads "Turn vague ideas into validated specifications and verified software." The two manifests have never matched. Worth aligning, but that is a separate decision about the Codex listing and I did not want to bundle it here.


Q00:
Note on tests, since I ran them after opening this rather than before.

`pytest tests/unit -k plugin` interrupts on three collection errors, all `ModuleNotFoundError: No module named 'mcp'` from `tests/unit/mcp/`. I checked `main` before assuming anything: identical failure there, so it is the local environment lacking the `[mcp]` extra, not this change. This PR edits one string in one JSON file and touches no code path.

Verified locally: the file still parses, `description` is 391 characters, and every other key is unchanged.

While measuring for this I ran into a separate thing worth knowing, since it affects anyone installing us as an MCP server: `uvx --from "ouroboros-ai[mcp]"` resolves `mcp 1.29.0` with the claude pins rather than the `mcp==2.0.0` the extra specifies. Details in #2038.

Q00:
Correcting a number in the description above, and adding the measurement that actually motivates this change.

**The "44" was loose.** It counted descriptions containing the words "spec" and "driven" anywhere, which sweeps in entries that merely mention specifications while doing something else. Counting the actual term, `spec-driven` or `SDD`, gives **19 of 2,291**. Still a crowded shelf for a directory where entries are chosen by scanning, but 19, not 44. My apologies; I asserted the number before checking how I had counted.

**The measurement that matters more.** Over the same 2,291 descriptions:

| what the description mentions | plugins |
|---|---|
| testing / QA | 177 |
| security scanning | 181 |
| memory / context | 136 |
| verification or acceptance gates | 114 |
| evaluation or grading | 83 |
| an agent gaming the grader | **2** |
| keeping the grading contract out of the agent | **0** |

Verification is well populated. The specific failure it addresses here is not: two entries mention an agent gaming its metric, and none describe withholding the grading contract from the executing agent.

That is the argument for this edit. The old line put us in the middle of the 114 and left out the one thing nobody else claims. The new line leads with it.

Method, so it can be checked or disputed: `.claude-plugin/marketplace.json` from `anthropics/claude-plugins-community` at 2026-08-10, matched with case-insensitive regex per row. The two "gaming" hits were read individually rather than counted from a pattern.

Q00:
Widened this to `.codex-plugin/plugin.json` as well. It should have been both from the start.

I wrote in the original description that the Codex manifest still said "Turn vague ideas into validated specifications and verified software" and that aligning it was "a separate decision". That was wrong on two counts.

The two manifests describe the same software to two directories. Leaving them different is not a decision anyone made; it is what happens when someone edits one and stops. I was about to be that someone, having just found the mismatch while looking at whether we could submit to OpenAI's universal plugin directory, which is where the Codex manifest goes.

They now carry the same 391-character description. Sibling files, one string, landing together rather than in two PRs a week apart.

Worth noting for anyone reviewing the Codex side specifically: OpenAI documents ChatGPT and Codex as sharing one universal plugin directory, so this string is what a reader sees on both surfaces. The old 69-character line used a fifth of the space and said nothing that distinguishes us from the crowd measured above.

`.codex-plugin/plugin.json` keeps its `interface` key, which the Claude manifest does not have. Only `description` changed.

Q00:
Independent direction and runtime-truthfulness review on current HEAD `09e7ff811076cdd96965af2c4a3651001af985a5`: **HOLD**.

The differentiation goal is good. Removing the universal “3-stage evaluation” claim is more honest, the directory comparison is useful, and aligning both manifests reduces listing drift. The blocker is that the replacement turns a narrower implementation property into an absolute public contract:

> the command and assertion used to grade the work never enter the executing agent's prompt

I reproduced two current-runtime counterexamples.

1. `build_ac_retry_prompt()` redacts five literal encodings, but a verification tail containing a line-broken assertion and command survives into `### Harness verification output (tail)`. `AtomicPromptBuilder` appends that retry text to the executing agent prompt. Whitespace-normalized, both hidden values are reconstructable. This is the same open structural-redaction boundary tracked by #2020; “never enter” does not communicate a byte-for-byte-only guarantee.
2. The default-on context pack can place a matching command from `.ouroboros/mechanical.toml` into the worker system prompt as `Verify commands: ...`. The leaf task prompt may omit the Seed field, but the wider executing-agent prompt boundary does not.

There is also a distribution-surface mismatch: applying “Sits in front of Claude Code” verbatim to `.codex-plugin/plugin.json` describes the Codex listing as Claude-only, while the project intentionally supports Codex and other runtimes.

This matters directionally: Ouroboros differentiates itself by making claims mechanically checkable and by exposing hidden assumptions. A more distinctive description is valuable, but an absolute hidden-grader claim contradicted by current prompt assembly moves against that principle.

Suggested convergence:

- Narrow the public wording to the exact proven boundary, or implement shared redaction across retry output, context-pack/system prompts, and structural transformations before retaining “never”.
- Use host-neutral wording in the shared description, or keep Claude/Codex manifest descriptions intentionally host-specific.
- Rebase onto current main `d3def57765d05bb6746551d0920101fcaf79b1b3` and request fresh exact-head review after convergence.

No merge is recommended for this SHA.

Q00:
Narrowed the claim in both manifests, following the blocker on #2041.

That review found that `execution.context_pack` defaults to `True` and appends the project's detected verify commands to the worker prompt. I verified it: `_scan()` calls `detected_verify_commands(root)`, and `subagent.py:1622` renders them as "Project verify commands (run before claiming done)". So "the command and assertion it is graded against never enter the executing agent's prompt" was false as a package-wide statement, and it was in this PR too.

What the code hides is per-criterion, and `_build_success_contract_block` says so directly: the harness-owned `verify_command` and `output_assertion` are deliberately withheld so the worker does not receive the grader's answer key. Both manifests now say that instead:

> ...Each acceptance criterion is then pinned and verified independently, and the criterion's own verify command and expected output are withheld from the worker, with no flag to put them back.

The directory measurement that motivated this PR is unaffected. Of 2,291 entries, none describe withholding grading detail from the executing agent, whether scoped per-criterion or not. The distinctive claim survives; it just has to be the true version of itself.

Same correction went into #2041 and into a draft MCP registry manifest that was not yet submitted.

Q00:
Both findings are correct and both are mine. Fixed.

**The blocking one.** Even my narrowed wording was still too strong. I verified the path: `retry_hints.py:101` appends `outcome.output_tail` through `_sanitize_fragment`, and `contract_redaction.py:10` removes five exact encodings, so an assertion that arrives line-broken survives and `atomic_prompt_builder.py:145` forwards it. "Withheld from the worker" describes an outcome we do not guarantee.

What we do guarantee is narrower and mechanical: the values are **omitted from the contract block**, and redaction elsewhere is **literal over five encodings**. Both manifests now say that, and say the gap out loud:

> Each acceptance criterion then carries a verify command and an expected output that are omitted from the worker's contract block, with no flag to put them back, so the worker is asked for the outcome rather than handed the assertion. Redaction of those values elsewhere is literal over five encodings, so a reshaped copy can still get through; that gap is tracked in a public issue.

This is the third time today this claim has been narrowed, each time because someone checked a path I had not. The first correction was mine, the second came from #2041, this is the third. The direction has been the same every time.

**The Codex one, which I caused in this PR.** The follow-up is right that a Codex marketplace listing should not open with "Sits in front of Claude Code". I introduced that two commits ago by aligning the manifests, and I aligned them by copying, which is not the same thing. The shared claim was worth sharing; the host sentence was not.

Each manifest now names its own host and the rest of the sentence is identical. Verified programmatically: the Claude manifest contains no mention of Codex, the Codex manifest contains no mention of Claude, and everything after the first sentence matches byte for byte.

Lesson I am writing down rather than just fixing: making two files consistent is not the same as making them identical, and I reached for the second because it was easier to check.

Q00:
Correct again, and I verified it before changing anything. `AcceptanceCriterionSpec.verify_command` and `output_assertion` are both `str | None = None` (`core/seed.py:462`, `:472`), so contract-less criteria are a supported shape and "each criterion carries both" is false.

Both manifests now say it conditionally, close to your suggested wording:

> When an acceptance criterion defines a verify command or an expected-output assertion, those values are omitted from the worker's contract block, with no flag to put them back...

This is the fourth narrowing of this claim in about a day, and all four went the same direction. The sequence: I wrote it as unconditional; #2041 found that `context_pack` puts project verify commands in the worker prompt; the previous round here found that the retry path forwards a line-broken assertion; this round finds that the per-criterion contract is optional. Each correction was a smaller true claim inside a larger false one.

The thing I keep getting wrong is not the mechanism, which I can read. It is that I describe the case I had in mind as though it were the only case. Optional fields, default-on context packs, and retry paths are all "the other case" and I have now missed three of them in a row.

So I have stopped composing this sentence from scratch. `wiki/ouroboros-facts.md` in the marketing repo now carries a fixed set of approved phrasings for it, in four languages, with the citations that constrain each one and an explicit do-not-use list of the phrasings that have already been corrected. This wording is going in there too. The next person to describe this, including me tomorrow, picks a line rather than writes one.

Q00:
Both blockers addressed. I verified the first one in the source before changing anything, because a claim this narrow has already been wrong ten times on this PR and I did not want the eleventh correction to be a guess.

**Blocker 1 — the ambiguity sentence.** Confirmed. `authoring_handlers.py:1385` reads `if not interview_state.is_complete and not force:`, so the threshold is only consulted while the interview is still open. A completed state carrying ambiguity 0.8 dispatches without `force`. My sentence said generation "refuses until it clears the bar, though you can force past it", which describes one of the two ways past the gate and silently omits the other.

New wording, both manifests:

> An interview scores how vague your request is; while that interview is still open, a score above the bar blocks spec generation. Completing the interview or passing force both get past that check.

**Blocker 2 — the marketplace metadata.** You were right that it was only partially met, and my first look at that file missed it. I checked `plugins[*].description`, found nothing, and nearly reported it clean. The claim was in the **top-level** `description`, one level up from where I looked:

```
"Requirement crystallization engine — Socratic interviews, seed generation, and 3-stage evaluation"
```

That is the string Claude Code's directory shows. It is now:

```
"Requirement crystallization engine — Socratic interviews, seed generation, and an acceptance-criteria evaluation gate"
```

I dropped the stage count rather than restating it, since stage 1 is off by default and stage 3 is not reachable on the evolve path, which makes "3-stage" wrong in the same direction as everything else this PR is fixing. I also aligned `plugins[0].description` with the corrected Claude Code copy so the two files stop disagreeing.

**On your test-coverage note.** You are right that nothing guards the completed-high-ambiguity path or manifest/marketplace consistency. I have not added those here. This PR only touches description strings, and a test asserting that behavior belongs with the handler rather than bolted onto a copy change. Worth its own issue if you agree.

Q00:
Pushed one more change, and it walks back something I added in this PR earlier today.

The ambiguity sentence said "Completing the interview or passing force both get past that check." That is wrong outside the plugin-dispatch path. `bigbang/seed_generator.py:1762-1780` gates on `if force: ... elif not ambiguity_score.is_ready_for_seed: return ValidationError`, with no `is_complete` branch, and `mcp/tools/authoring_handlers.py:1573` passes `force=force` straight into it. The `is_complete` skip I found at `:1385` only bypasses the handler's own pre-check one layer above.

So `force` is the portable answer. All three files in this branch now say that, `marketplace.json` included.

I want to be plain about the sequence, because it is the second time this claim has moved in this PR:

1. The original text said the interview "refuses until it clears the bar, though you can force past it."
2. Your earlier review showed that was incomplete for the plugin path, and I widened it to "two ways past."
3. That widening was itself wrong as a general statement, and #2060's review caught it.

Step 2 is the one I should not have shipped. I found the `is_complete` skip, treated the first gate I saw as the whole story, and did not ask whether another gate ran underneath. **A correction is a claim and needs the same verification as the thing it replaces** — I applied that standard to the original wording and not to my own fix.

Not documenting the plugin/in-process difference as behaviour here. If one dispatch path really does skip a pre-check the other enforces, that is an inconsistency to decide about, not something a plugin description should teach readers to rely on.

PR #2041, opened 2026-08-10
docs(pyproject): lead the summary with what nothing else in the directory claims

Q00:
Rewrites `project.description` in `pyproject.toml`, which is what PyPI shows as the package summary and what search results truncate.

## The measurement

Current summary, 183 characters:

> Specification-first workflow engine for AI coding agents -- works across Claude Code, Codex CLI, OpenCode, Hermes, Gemini, Kiro, Copilot, Pi, Zcode, Goose, GJC, Antigravity, and Grok.

**67% of it is the runtime list.** Thirteen names, and by the time they finish there is no room left for what the thing does. A reader scanning search results sees a category label and a roll call.

The category label is also crowded. In the Claude Code plugin directory, 2,291 entries: 114 describe verification or acceptance gates, 83 describe evaluation or grading, 19 use "spec-driven" or "SDD". Two mention an agent gaming its grader. None describe keeping the grading contract away from the executing agent.

## The change

> Pins the acceptance spec before the run and keeps the grading command out of the agent prompt. Works with Claude Code, Codex CLI, OpenCode and 10 more runtimes.

160 characters, and the first sentence is the part nothing else claims.

The runtime list is not deleted, it is compressed. Three names stay because those are the ones people search for, and "10 more" is accurate: thirteen families, three named. The full list still lives in the README and the docs, where someone checking whether their runtime is supported will actually look. A summary is not the place to enumerate.

## On "keeps the grading command out of the agent prompt"

That is the scoped claim, not "the agent can never see it". The redaction is verbatim over five encodings and a reshaped assertion still gets through, which is open as #2020. The summary says what the system does, not what it guarantees.

## Not changed here

The README tagline is still "Stop prompting. Start specifying." which sits in the same crowded lane as the old summary. That is a bigger decision than a metadata field and belongs in its own discussion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "ouroboros-ai"
 dynamic = ["version"]
-description = "Specification-first workflow engine for AI coding agents -- works across Claude Code, Codex CLI, OpenCode, Hermes, Gemini, Kiro, Copilot, Pi, Zcode, Goose, GJC, Antigravity, and Grok."
+description = "Pins the acceptance spec before the run and keeps the grading command out of the agent prompt. Works with Claude Code, Codex CLI, OpenCode and 10 more runtimes."
 readme = "README.md"
 authors = [
     { name = "Q00", email = "jqyu.lee@gmail.com" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "ouroboros-ai"
 dynamic = ["version"]
-description = "Pins the acceptance spec before the run and keeps the grading command out of the agent prompt. Works with Claude Code, Codex CLI, OpenCode and 10 more runtimes."
+description = "Pins the acceptance spec before the run and withholds each criterion's verify command and expected output from the worker. Works with Claude Code, Codex CLI, OpenCode and 10 more runtimes."
 readme = "README.md"
 authors = [
     { name = "Q00", email = "jqyu.lee@gmail.com" }


Q00:
The blocker is correct and I verified it independently before changing anything.

`execution.context_pack` defaults to `True`; `_scan()` calls `detected_verify_commands(root)`, and `subagent.py:1622` appends them to the worker prompt as "Project verify commands (run before claiming done)". So a project-level verify command genuinely does reach the worker. My headline said otherwise about the whole package, and that was wrong.

What the code actually hides is narrower, and `_build_success_contract_block` states it plainly:

> Artifact names are part of the work contract, but the harness-owned `verify_command` and `output_assertion` are deliberately hidden. The worker must satisfy the requested outcome without receiving the grader's answer key.

So the true claim is per-criterion, not package-wide. New summary:

> Pins the acceptance spec before the run and withholds each criterion's verify command and expected output from the worker. Works with Claude Code, Codex CLI, OpenCode and 10 more runtimes.

199 characters. "each criterion's" is doing the load-bearing work: it scopes the claim to the AC contract block rather than to every prompt projection.

I did not take the second remediation option (routing every worker-facing projection through the redaction boundary). Giving a worker the project's lint and test commands looks deliberate rather than accidental: the pack exists to prime a worker with deterministic repo facts, and running the project's tests before claiming done is something you want it to do. What must stay hidden is the specific assertion a criterion is graded against, and that is what the contract block hides. If that reading is wrong, the fix belongs in the code and not in this string.

**This claim is not only in this file.** The same wording went into `.claude-plugin/plugin.json` and `.codex-plugin/plugin.json` in #2040, and into a draft `server.json` for the MCP registry. I am correcting those too rather than leaving one accurate copy and three overstated ones.

Q00:
## Post-merge truthfulness audit: direct-runner collision still exposes the criterion command

I reproduced a narrower counterexample on merged `main@49b99748f97ba73ded301db0a1a22c262ff5d615`.

When one acceptance criterion owns:

```text
verify_command = uv run pytest --criterion-token EXACT_CRITERION_SECRET
```

and the project independently has the identical command in `.ouroboros/mechanical.toml`, the direct runner composes the context pack after rendering the worker-safe Seed contract. `build_system_prompt(..., context_pack_enabled=True)` therefore emits:

```text
- test: uv run pytest --criterion-token EXACT_CRITERION_SECRET
```

Probe result:

```text
command_in_worker_prompt=True
criterion_token_in_worker_prompt=True
expected_output_in_worker_prompt=False
```

The revised claim is still too broad: the criterion field is omitted from the leaf contract, and the plugin projection redacts detected-command collisions, but the supported direct runner path can reintroduce the same criterion-owned value through `resolved_context_pack_fragment`. The official approval's statement that detected-command collisions are covered is true for `mcp/tools/subagent.py`, not for `orchestrator/runner.py` start/resume prompt construction.

This does not invalidate the acceptance-spec pinning claim or the hidden `output_assertion` path. It does mean the merged package description's verify-command clause is not yet true for all supported worker paths. I am opening a focused follow-up bug with the exact regression contract rather than folding it into #2020, which concerns transformed encodings in retry hints.

Issue #2044, opened 2026-08-10
Downstream AUR package is four releases behind and ships the pre-#2012 MCP launcher


Q00:
There is a downstream Arch package we do not maintain and, until today, did not know about. It is four releases behind, which means its users hit a startup failure we fixed this morning.

## The package

`python-ouroboros-ai` on the AUR, from `aur.archlinux.org/rpc/v5/info/python-ouroboros-ai`:

| Field | Value |
|---|---|
| Version | **0.50.6-1** |
| Maintainer | `bhyoo` (not us; not a contributor here) |
| First submitted | 2026-04-02 |
| Last modified | 2026-07-24 |
| Out-of-date flag | none set |
| URL | `github.com/Q00/ouroboros` |

The packaging looks deliberate and correct: the name follows the Arch Python convention, the description matches our summary, and the upstream URL points here.

## What its users are missing

Four releases have shipped since: v0.50.7, v0.50.8, v0.51.0, v0.51.1. The first highlight of v0.51.1 is the MCP launcher fix:

> shipped launchers now run `uvx --isolated`, so an existing `uv tool install ouroboros-ai` environment carrying the MCP 1.x SDK (via `claude-agent-sdk`) can no longer shadow the `[mcp]` profile's `mcp==2.0.0`. This resolves `Failed to reconnect to plugin:ouroboros:ouroboros: -32000` at startup (#2012)

So someone installing us from the AUR today gets the version that fails to start under Claude Code and Codex plugins. I can vouch for how confusing that failure is: I hit it myself earlier today, did not connect it to #2012, and filed #2038 against our own packaging before finding the answer in our release notes.

## What I am not proposing

Taking over the package. Someone spent their own time packaging this for a distribution we do not support, and they filled a gap we did not know existed. The AUR has an out-of-date flag for exactly this, and a maintainer who has updated the package before will likely update it again once they know 0.51.1 matters more than a normal version bump.

## What might be worth deciding

1. **Flag it out-of-date**, or tell the maintainer that 0.51.1 specifically fixes plugin startup. Either needs an AUR account, which is why this is an issue rather than a completed action.
2. **Whether downstream packagers should hear about releases at all.** This one found us without any help and has kept the package roughly current for four months. There may be others we still do not know about; I checked conda-forge and Docker Hub and found only unrelated projects of the same name.
[...]

PR #2045, opened 2026-08-10
docs(readme): name the minimum version that starts as an MCP server

Q00:
Adds one line to the install section of all three READMEs, naming 0.51.1 as the minimum version that starts as an MCP server.

## Why

`#2012` fixed a startup failure where an existing environment shadows the `[mcp]` profile and the plugin dies with `Failed to reconnect to plugin:ouroboros:ouroboros: -32000`. Anyone on an earlier version still hits it.

That would be unremarkable except that earlier versions are still being installed. There is a downstream Arch package, `python-ouroboros-ai`, currently at 0.50.6 and four releases behind (#2044). Its users get exactly this failure, and nothing on the way in tells them which version they need.

The release notes do say it. It is the only entry under `Highlights` in v0.51.1. That did not help: I hit the failure myself earlier today, did not connect it to #2012, and filed an issue against our own packaging before finding the answer in our own changelog. Fourteen people watch this repository, so the highlight reaches fourteen people.

A line in the README reaches someone who never subscribed to anything.

## The line

> **Installing as an MCP server: use 0.51.1 or later.** Earlier versions can fail at startup with `Failed to reconnect to plugin:ouroboros:ouroboros: -32000` when an existing environment shadows the `[mcp]` profile (#2012). This matters if you install through a downstream package rather than PyPI, since those can lag.

Placed next to the existing Python version requirement, which is where a reader already looks for "what do I need".

## On the translations

The three files are not structured identically and I got this wrong once before pushing. `README.md` and `README.zh-CN.md` both carry a `> **Python >= 3.12**` requirement block, so the note appends to it. `README.ko.md` has no such block at all, so the note goes after the uninstall details block instead. My first attempt used the English anchor for all three and put the Korean note at line 538, at the very bottom of the file, nowhere near installation. Reverted and redone per file.

## Not included

Whether to notify downstream packagers at all, and whether the PyPI page should carry the same note, are open in #2044. This PR only does the part that needs no decision.

### DIFF
--- a/README.ko.md
+++ b/README.ko.md
@@ -253,6 +253,8 @@ ouroboros uninstall
 
 </details>
 
+> **MCP 서버로 설치한다면 0.51.1 이상을 쓰세요.** 이전 버전은 기존 환경이 `[mcp]` 프로필을 가릴 때 `Failed to reconnect to plugin:ouroboros:ouroboros: -32000`로 기동에 실패할 수 있습니다([#2012](https://github.com/Q00/ouroboros/issues/2012)). PyPI가 아니라 배포판 패키지로 설치할 때 특히 해당됩니다 — 그쪽은 버전이 뒤질 수 있습니다.
+
 <details>
 <summary><strong>무슨 일이 일어났나요?</strong></summary>
 
--- a/README.md
+++ b/README.md
@@ -257,6 +257,8 @@ Removes all configuration, MCP registration, and data. See [UNINSTALL.md](./UNIN
 </details>
 
 > **Python >= 3.12 required.** LiteLLM-bearing profiles support Python 3.12-3.13. See [Platform Support](./docs/platform-support.md#python-profile-matrix) and [pyproject.toml](./pyproject.toml).
+>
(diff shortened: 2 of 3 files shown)


Q00:
## Post-merge exact-head audit

This PR was merged at `581c0b74a4d9eb3bc4ceda87e38e29c3b12559e8` with 14/14 green checks and an independently verified docs-only tree delta. The content audit found no defect, and #2044 correctly remains open for the excluded downstream-policy decisions.

However, the official review gate was not exact-head at merge time. The only bot review was submitted for `fd43ab6c813337ad38b5a96c243c5a940c72bb8c`; after `update-branch`, GitHub retargeted its REST `commit_id` to `581c0b74...`, while the review body still recorded:

```text
HEAD checked: fd43ab6c...
Review-Metadata head_sha: fd43ab6c...
```

So the aggregate `APPROVED` state did not prove review of the merged HEAD. This is now tracked as a reusable merge-safety bug rather than attributed to the README change itself.

PR #2026, opened 2026-08-10
fix(codex app): prevent Windows app-server crashes with Ouroboros MCP

bokjk:
## Summary

Prevent native-Windows Codex Desktop app-server termination (`0xC000013A`) by using a setup-managed loopback streamable-HTTP Ouroboros MCP endpoint instead of a Codex-owned stdio child.

Windows `auto` migrates only absent/setup-managed entries. User-managed entries remain untouched. `preserve` performs no MCP lifecycle reconciliation. Explicit `http` selects the managed endpoint; `stdio` publishes an immutable disabled state rather than deleting task or operator artifacts. Non-Windows behavior remains stdio-based.

## Windows lifecycle safety

The final design has no fixed runner, supervisor, owner file, mutable current pointer, forced task replacement, task deletion, or task termination:

- each serve generation is immutable and receives a unique per-user Task Scheduler task created without `/F`
- the actual MCP server process validates its generation before binding, publishes PID/start-time-attributed standby/ready/stopped receipts, and owns the handoff
- old logon tasks exit before binding unless their generation is currently eligible
- disabled generations stop the owned server gracefully without deleting tasks
- setup and runtime pin the physical non-reparse lifecycle root
- complete Scheduler XML/SID ownership, listener PID ownership, rollback, expiry, and crash handoff are verified fail-closed
- pre-release legacy `Ouroboros MCP HTTP` artifacts are detected and require explicit manual cleanup rather than mutation

Setup also refuses a duplicate global registration while the plugin-scoped Ouroboros MCP is active.

## Verification

- [x] Current review-focused tests: lifecycle/PID **87 passed, 1 skipped**; CLI/lifecycle teardown **18 passed**; Codex setup **83 passed** (native-Windows-only exclusions noted below)
- [x] Ruff check and format clean
- [x] Mypy clean for lifecycle and MCP command modules
- [x] Module-size and `git diff --check` pass
- [x] Native Task Scheduler create/query normalization accepted and exact ownership revalidated
- [x] Native listener attribution: actual managed PID owns `127.0.0.1:8765`
- [x] MCP initialize succeeds with JSON/SSE negotiation and exact `ouroboros-mcp` identity
- [x] Native disabled transition stopped the prior PID; reprovision started a distinct PID
- [x] Native post-rebase generation handoff succeeded (`27804` ? `2004`)
- [x] Independent read-only review of current HEAD: **APPROVE**, no blocking findings

[...]

PR #2056, opened 2026-08-10
fix(codex app): avoid native Windows crashes with HTTP MCP

### LINKED ISSUE #2024: bug(codex-app): native Windows stdio MCP can terminate the Codex Desktop app-server
## Summary

On native Windows, a setup-managed stdio Ouroboros MCP child can trigger Codex Desktop's Windows process-cleanup path and terminate the app's own bundled app-server with `3221225786 (0xC000013A, STATUS_CONTROL_C_EXIT)`.

## Impact

Installing/configuring Ouroboros can make Codex Desktop repeatedly show an unexpected-stop/restart experience. Projects and turns become unavailable even though the Ouroboros MCP server itself completed initialization.

## Steps to reproduce

1. On Windows 11, install Codex Desktop and Ouroboros.
2. Configure the setup-managed entry:

```toml
[mcp_servers.ouroboros]
command = "...\\uvx.exe"
args = ["--from", "ouroboros-ai[mcp]", "ouroboros", "mcp", "serve"]
```

3. Start a turn that initializes/discovers the Ouroboros MCP tools.
4. Observe the bundled Codex app-server exit with `0xC000013A`.

## Evidence

- Ouroboros MCP registered 35 tools and initialized before the Codex app-server exited.
- Switching the same integration to an external local streamable-HTTP server at `http://127.0.0.1:8765/mcp` exposed `mcp__ouroboros__ouroboros_session_status` and kept the same app-server PID alive across turns/restarts.
- OpenAI tracks the app-server termination mechanism in openai/codex#36778.

## Expected behavior

Native-Windows setup should avoid the crash-triggering stdio child topology while retaining full MCP functionality.

## Proposed behavior

- Windows `auto` uses a setup-managed local streamable-HTTP endpoint only for absent/setup-managed entries.
- Setup provisions a hidden, least-privilege, restartable per-user task and verifies loopback readiness before succeeding.
- User-managed URL/custom entries remain untouched.
- Explicit `stdio` remains available and removes only the setup-managed HTTP task.
- Non-Windows behavior remains unchanged.

## Acceptance criteria

- [ ] Fresh native-Windows setup writes a working managed HTTP endpoint.
- [ ] Setup does not report success until the endpoint accepts loopback connections.
- [ ] Provisioning failure restores prior task/runner state and triggers config rollback.
- [ ] HTTP-to-stdio removes only setup-owned task artifacts.
- [ ] User-owned same-name tasks/files and custom MCP entries are preserved.
- [ ] Existing non-Windows setup behavior is unchanged.

## Environment

- Windows 11 Pro 10.0.26200
- Codex Desktop MSIX `OpenAI.Codex_26.803.5235.0_x64`
- Bundled app-server `codex-cli 0.147.0-alpha.6.5`
- Ouroboros plugin 0.51.1
- Ouroboros CLI 0.51.0

## Related

- openai/codex#36778
- Q00/ouroboros#1911
- Q00/ouroboros#1910

bokjk:
## Summary

Prevent native-Windows Codex Desktop from owning an Ouroboros stdio MCP child. On Windows, one normal `ouroboros setup --runtime codex` run now:

- registers a setup-marked loopback URL at `http://127.0.0.1:8765/mcp`
- writes one exact per-user `HKCU\Software\Microsoft\Windows\CurrentVersion\Run` startup value
- launches the same command immediately as a detached current-user process
- writes Codex configuration only after a strict MCP initialize response identifies `ouroboros-mcp`

Non-Windows `auto` remains stdio. User-managed entries and `preserve` remain untouched; explicit `stdio` and `http` remain available.

## Deliberate scope

This supersedes the oversized lifecycle design in #2026. It intentionally has:

- no Task Scheduler dependency or elevation
- no generations, supervisor, task replacement/deletion, or service framework
- no zero-downtime upgrade, crash restart, or automatic rollback claim

The startup value launches at login; setup also launches and verifies it immediately. Native Codex Desktop MCP remains experimental, and native Codex CLI worker/runtime execution remains unsupported.

## Verification

- focused startup helper: **53 passed**
- focused Codex setup: **71 passed, 349 deselected**
- focused stdio commit + uninstall cleanup: **25 passed**
- Ruff check/format, mypy, module-size, and diff checks: clean
- native Windows HKCU Run + detached launch + strict MCP initialize: verified
- native rollback and launcher-drift cleanup across uvx/direct/Python variants: verified
- independent final scoped review: **APPROVE / CLEAR**, no HIGH or MEDIUM findings

Closes #2024
Supersedes #2026

### DIFF
--- a/docs/runtime-guides/codex.ko.md
+++ b/docs/runtime-guides/codex.ko.md
@@ -105,7 +105,20 @@ codex --version
 | macOS (ARM/Intel) | 지원 | 주 개발 플랫폼 |
 | Linux (x86_64/ARM64) | 지원 | Ubuntu 22.04+, Debian 12+, Fedora 38+ 에서 검증 |
 | Windows (WSL 2) | 지원 | Windows 사용자에게 권장하는 경로 |
-| Windows (네이티브) | 실험적 | WSL 2를 강력히 권장합니다. 네이티브 Windows에서는 경로 처리와 프로세스 관리에 문제가 있을 수 있습니다. **Codex CLI 자체가 네이티브 Windows를 지원하지 않습니다.** |
+| Windows (네이티브) | 실험적 | 로컬 HTTP 엔드포인트를 통한 Codex Desktop MCP를 실험적으로 지원합니다. 네이티브 Codex CLI 워커/런타임 실행은 지원하지 않으므로 전체 워크플로에는 WSL 2를 사용하세요. |
+### 네이티브 Windows Codex Desktop MCP (실험적)
+네이티브 Windows에서 `ouroboros setup --runtime codex`는 로그인할 때 HTTP MCP를 시작하는, setup이 관리하는 정확히 하나의 사용자별 `HKCU\Software\Microsoft\Windows\CurrentVersion\Run` 시작 값을 씁니다. setup은 HTTP MCP를 즉시 시작하고 MCP `initialize` 프로브로 검증한 뒤에만 Codex Desktop에 `http://127.0.0.1:8765/mcp`를 등록합니다. setup이 성공하면 Codex Desktop이 이 엔드포인트를 사용하므로, 사용자는 세션마다 MCP 명령을 실행하지 않습니다.
+
+`--mcp-mode`는 MCP 등록만 제어합니다.
+
+- `auto`(기본값): 네이티브 Windows에서는 시작 값이 없거나 이미 setup 관리 값과 일치할 때만 관리형 HTTP 엔드포인트를 사용합니다. 그 밖의 플랫폼에서는 관리형 stdio를 사용하며, 사용자가 관리하는 항목은 보존합니다.
+- `http`: 관리형 HTTP 엔드포인트를 사용합니다.
+- `stdio`: 관리형 항목을 명령으로 실행하는 stdio 서버로 교체합니다.
+- `preserve`: 기존에 사용 가능한 명령 또는 URL 항목만 검증합니다.
+
+setup과 uninstall은 현재 Windows 로그인 세션 범위의 시간 제한 명명된 뮤텍스를 통해 협력하므로, 같은 로그인 세션에서 협력하는 Ouroboros setup 및 uninstall 작업이 직렬화되며, 시작 값의 정확한 형식과 값을 즉시 다시 확인합니다. 이 재확인에서 불일치가 관찰되면 setup은 그 값을 보존하고, 수정하거나 제거한 뒤 다시 실행하는 방법을 알리며, 그 전에는 HTTP 설정을 쓰지 않습니다. 같은 계정의 여러 Windows 세션에서 동시에 실행하는 lifecycle 명령은 이 실험적 setup 계약의 범위 밖이므로 실행하지 않아야 합니다. setup이 활성화된 동안 수동 또는 비-Ouroboros 레지스트리 편집은 동기화되지 않으므로 피해야 합니다. 이 실험적 경로는 로그인 시 MCP를 시작하지만, 충돌 후 재시작, 세대 관리, 자동 롤백, 무중단 업그레이드, 서비스 관리자 의미론을 제공하지 않습니다.
+
+> **Windows 사용자:** 네이티브 Codex CLI 워커/런타임 워크플로에는 WSL 2를 사용하세요. 위 네이티브 Windows setup은 실험적인 Codex Desktop MCP로만 한정됩니다.
 
 ## 설정
 
@@ -198,13 +211,13 @@ codex --version
 - 가능하면 `orchestrator.codex_cli_path`를 기록합니다
 - 관리되는 Ouroboros 규칙을 `~/.codex/rules/`에 설치합니다
 - 관리되는 Ouroboros 스킬을 `~/.codex/skills/`에 설치합니다
-- `~/.codex/config.toml`에 Ouroboros MCP/env 연결이 없으면 등록하고, setup이 관리하는 stdio 블록은 갱신하되, 사용자가 관리하는 URL·커스텀 항목은 기본적으로 보존합니다
+- `~/.codex/config.toml`에 Ouroboros MCP/env 연결을 등록합니다. 네이티브 Windows의 `auto`는 조건이 맞을 때 고정 관리형 HTTP 엔드포인트를 사용하고, 그 밖의 플랫폼에서는 관리형 stdio를 사용하며 사용자가 관리하는 항목은 보존합니다
 - 손대지 않은 레거시 자동생성 `ouroboros-*.config.toml` 태스크 프로파일 앵커만 정리합니다. 사용자가 만든 Codex 프로파일은 보존됩니다
 - 관리되는 `ouroboros-worker.config.toml`을 등록해서, Agent OS 워커 서브프로세스가 MCP/env 연결을 잃지 않고도 대화형 Codex 기본값에서 빠질 수 있게 합니다
 
 `~/.codex/` 밖의 전역 산출물도 함께 생깁니다. `ensure_config_dir()`가 `~/.ouroboros/data/`와 `~/.ouroboros/logs/`를 만들고([`setup.py:2632`](../../src/ouroboros/cli/commands/setup.py)), 설정이 처음이면 `~/.ouroboros/credentials.yaml`을 `0600` 권한으로 새로 씁니다([`setup.py:2771`](../../src/ouroboros/cli/commands/setup.py)).
 
-`~/.codex/config.toml`은 **Ouroboros 스테이지 모델 핀을 둘 자리가 아닙니다.** 설정 UI나 그에 대응하는 `~/.ouroboros/config.yaml` 값을 쓰고, 명시적인 `--profile`이 필요할 때만 사용자가 관리하는 네이티브 Codex 프로파일을 유지하세요. 장기 실행 URL 기반 Ouroboros MCP 서버를 직접 운영한다면 그 URL 항목을 `~/.codex/config.toml`에 그대로 두면 됩니다. `ouroboros setup --runtime codex`가 기본적으로 보존합니다. setup이 그 항목을 관리형 command-spawn 서버로 **바꾸길 원할 때만** `--mcp-mode stdio`를 쓰세요.
+`~/.codex/config.toml`은 **Ouroboros 스테이지 모델 핀을 둘 자리가 아닙니다.** 설정 UI나 그에 대응하는 `~/.ouroboros/config.yaml` 값을 쓰고, 명시적인 `--profile`이 필요할 때만 사용자가 관리하는 네이티브 Codex 프로필을 유지하세요. `auto`는 사용자가 관리하는 MCP 명령과 URL 항목을 보존합니다. 관리형 등록을 의도적으로 고를 때만 `--mcp-mode stdio` 또는 `http`를 사용하세요.
 
 ### 워커 서브프로세스 격리 (Agent OS `runtime_profile`)
 
--- a/docs/runtime-guides/codex.md
+++ b/docs/runtime-guides/codex.md
@@ -50,9 +50,22 @@ For alternative install methods and shell completions, see the [Codex CLI README
 | macOS (ARM/Intel) | Supported | Primary development platform |
 | Linux (x86_64/ARM64) | Supported | Tested on Ubuntu 22.04+, Debian 12+, Fedora 38+ |
 | Windows (WSL 2) | Supported | Recommended path for Windows users |
-| Windows (native) | Experimental | WSL 2 strongly recommended; native Windows may have path-handling and process-management issues. Codex CLI itself does not support native Windows. |
+| Windows (native) | Experimental | Codex Desktop MCP is supported experimentally through a local HTTP endpoint. Native Codex CLI worker/runtime execution is unsupported; use WSL 2 for full workflows. |
 
 > **Windows users:** Install and run both Codex CLI and Ouroboros inside a WSL 2 environment for full compatibility. See [Platform Support](../platform-support.md) for details.
+### Native Windows Codex Desktop MCP (experimental)
+On native Windows, `ouroboros setup --runtime codex` writes one exact setup-managed per-user `HKCU\Software\Microsoft\Windows\CurrentVersion\Run` startup value that launches the HTTP MCP at login. Setup launches the HTTP MCP immediately, verifies it with an MCP `initialize` probe, and only then registers `http://127.0.0.1:8765/mcp` with Codex Desktop. After successful setup, Codex Desktop uses that endpoint; users do not run an MCP command each session.
+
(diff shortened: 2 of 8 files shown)


bokjk:
Python 3.14 CI found one test-only Rich help wrapping difference at the runner's narrow terminal width. Current HEAD `3a14a247` normalizes Rich table borders/whitespace before asserting the four documented modes; the same test now passes at `COLUMNS=80`. No production behavior changed. The full matrix has restarted.

bokjk:
Addressed the current-HEAD review blockers in `097a7a140` without adding a supervisor or service lifecycle:

- the HKCU Run write now re-reads value and type immediately before mutation, retaining only an exact `REG_SZ` match and refusing concurrent foreign/type-changed values
- HTTP provisioning now returns invocation ownership; full Codex setup keeps it pending, rolls it back on every later setup failure, and commits only after complete success
- rollback compare-before-deletes only this invocation's exact startup value and terminates only its owned detached process tree via trusted System32 `taskkill.exe`
- explicit `stdio` migration and uninstall remove only the exact setup-owned startup value; user-modified values are preserved, with an explicit warning that a prior-session process may remain until logoff

Current verification: helper 39 passed; focused setup 71 passed; focused uninstall 22 passed; Ruff/format/mypy/module-size/diff clean; native Windows rollback proof passed. Two independent scoped re-reviews returned APPROVE / CLEAR.

bokjk:
Addressed the latest current-HEAD blockers in `daa00f6a5`, still within the minimal HKCU Run design:

- all Ouroboros setup/uninstall startup operations now share one bounded per-user Windows named mutex, held across the complete read/check/mutate flow; inner operations still perform the exact `REG_SZ` value recheck
- explicit Windows `stdio` deprovisioning is now the final fallible operation in the existing `_setup_codex` transaction; late failures leave startup intact, and cleanup failure restores Codex TOML, Ouroboros config, and managed artifacts
- output occurs only after persistent state commits, so a warning failure cannot restore HTTP config after deletion
- docs now state the exact cooperative guarantee: arbitrary manual/non-Ouroboros registry edits during active setup are not synchronized because Win32 exposes no registry-value CAS/key write lock

Added deterministic nested mutex, competing provision/remove, stdio success-ordering, and cleanup-error restoration coverage. Current verification: 43 helper tests, 71 focused setup tests, 25 commit/uninstall tests; Ruff/format/mypy/module-size/diff clean; native Windows rollback proof passed; independent final review APPROVE / CLEAR.

Q00:
Exact-head follow-up on `daa00f6a5918a2fde5f8294016b7fd729d923bf2`: the prior two blockers are materially improved. The shared bounded named mutex now closes the cooperating-process query/mutate race, and HTTP→stdio cleanup is deferred until after the other fallible setup stages so late failure can restore filesystem state without having already removed activation. The docs also correctly avoid claiming protection from arbitrary external registry editors.

The new exact-head official blocker still prevents merge: the mutex protects each provisioning/removal call, but not the outer setup transaction through commit or rollback. A concrete interleaving remains possible:

1. Setup A acquires the mutex, creates the managed startup value/process, returns a provisional ownership record, then releases the mutex.
2. Setup B acquires it, observes/reuses that exact value and process, and commits successfully.
3. Setup A later fails in an outer setup stage and executes rollback from its provisional record, deleting activation that B now relies on.

Concurrent uninstall has the same post-provision/pre-commit window. The transactional lease or ownership generation therefore has to span the outer setup commit/rollback boundary, not just the registry helper. Add deterministic tests that pause A after provision, let B reuse and commit, then force A rollback and prove B activation survives; cover concurrent uninstall in the same window.

There is also an identity follow-up worth fixing in this loop: stdio migration/uninstall reconstructs the expected startup command from the launcher available *now* rather than durable setup-time identity. If a direct/Python install later gains `uvx`, exact-match cleanup can miss the old setup-owned Run value, and stdio may report success while leaving HTTP autorun active. Preserve the installed command/ownership identity or otherwise make cleanup independent of later launcher-resolution drift.

The native-Windows HTTP topology, exact MCP initialize verification, user-managed entry preservation, and deliberately small HKCU Run design remain directionally strong. Merge should stay frozen until the transaction-wide ownership invariant is demonstrated, same-head CI is green, and a fresh exact-head official review approves.

bokjk:
Addressed the latest transaction-ownership and launcher-drift blockers in `edf8404a3` without adding lifecycle infrastructure:

- every successful HTTP provision now retains its named-mutex lease in the existing ownership receipt until the encompassing setup commits or rolls back
- rollback performs invocation-owned process/startup cleanup before releasing the lease, so another cooperating setup/uninstall cannot reuse or remove provisional state mid-transaction
- stdio migration/uninstall now recognize only strict setup-generated uvx, direct `ouroboros`, or Python-module startup command shapes with the exact fixed HTTP suffix; deletion still compare-before-deletes the exact observed registry string
- explicit 64-bit ctypes signatures cover mutex APIs, `CommandLineToArgvW`, and `LocalFree`

Added full pending-provision contention, native ABI, actual generated launcher variants, launcher-drift deletion, and lookalike-preservation coverage. Current verification: 53 helper tests, 71 focused setup tests, 25 commit/uninstall tests; Ruff/format/mypy/module-size/diff clean; native Windows rollback and all three launcher variants verified; independent final review APPROVE / CLEAR.

bokjk:
Scope boundary for the latest review, reflected in `b2b617f32`:

The crash issue (#2024) is the native Codex Desktop stdio-child topology. This PR now replaces that topology, verifies the HTTP MCP before registration, serializes cooperating setup/uninstall commands in the same Windows login session, carries provisional ownership through setup commit/rollback, and safely handles known launcher drift.

The newly requested guarantees—simultaneous setup/uninstall from multiple Windows sessions for the same account, one transaction spanning the entire uninstall command, and cleanup after the installation no longer has any launchable Ouroboros MCP—are broader lifecycle-hardening work. They are not in #2024's reproduction or acceptance criteria and would expand this replacement PR back toward the superseded #2026 design. I am intentionally not adding that framework here. The docs now state the same-login-session synchronization boundary instead of claiming cross-session locking. These scenarios can be proposed as separate follow-ups if maintainers want to support them.

The previous Python 3.14 failure was not reproduced in the changed surface: all changed CLI suites pass under Linux Python 3.14 (`518 passed, 1 skipped`), and the current head has restarted CI. A human maintainer review of the bounded crash fix would be appreciated.

Q00:
## Direction review — the persistence primitive, not the implementation quality

First, credit where it is due. The parts of this PR that replace the crash-triggering topology are genuinely well built, and I want to keep all of them: writing Codex TOML only after a strict `initialize` identity check, preserving user-managed entries and `preserve` mode, compare-before-delete on rollback, and leaving non-Windows `auto` untouched. The HTTP-endpoint direction itself is correct and matches the evidence in #2024. None of what follows asks you to revisit any of that.

The blocker is one level up: **HKCU Run is the wrong primitive for the requirement this PR is serving**, and I believe switching it will make the PR *smaller*, not larger.

### 1. Zero-touch is the actual bar, and HKCU Run does not clear it

The product requirement behind #2024 is that a Codex Desktop user never has to run, restart, or repair anything for the Ouroboros MCP to be available. That user has no reason to ever open the Ouroboros CLI — Codex Desktop is their only touchpoint — so any recovery path that begins with "run setup again" or "log off and back on" effectively does not exist for them.

Measured against that bar, a Run value gives us launch-at-logon and nothing else:

- **No crash recovery.** If the detached server dies mid-session (unhandled exception, DB contention, OOM — we have real history with long-lived `mcp serve` processes), every Ouroboros tool in Codex fails silently until the next logon. The PR's own "Deliberate scope" section confirms crash restart is out of scope — but for this primitive it isn't optional hardening, it's the difference between zero-touch and not.
- **No failure diagnostics.** The logon launch runs with stdout/stderr on `DEVNULL` and no window. When it fails (for example `uvx` resolving on a cold cache at logon, before the network is up), there is no artifact anywhere telling the user or us what happened.
- **The issue already asked for the right thing.** #2024's proposed behavior is a "hidden, least-privilege, **restartable** per-user task". The move to a Run value quietly dropped "restartable". This isn't scope creep on my side — it's the original acceptance surface.

### 2. Requested change: a thin per-user Scheduled Task, which is not the #2026 framework

I want to be precise about this boundary, because prior rounds have (fairly) pushed back on scope regrowth: **#2026 was rejected for the framework built on top of Task Scheduler** — generations, a supervisor, task replacement/deletion services, zero-downtime upgrade, automatic rollback — **not for using Task Scheduler at all**. A single per-user task registered via `schtasks` (or the COM API) with a logon trigger and a restart-on-failure policy is not that framework. It needs no elevation and no service.

What it buys, concretely:

- **Crash restart becomes the OS's job.** The scheduler restarts the server; no supervisor code in Ouroboros, ever.
- **Most of the hand-built lifecycle machinery in this PR exists because a Run value plus an orphan detached process are unmanaged shared mutable state.** The named mutex held across the setup transaction, the transactional ownership lease, the provisional-ownership receipts, the trusted-`taskkill /T /F` process-tree termination, and the `_is_managed_startup_variant` launcher-shape whitelist (which we will have to extend every time the launcher command evolves) are all compensations for that. A named task is a single addressable object: setup can query it, stop it, start it, delete it, and compare its definition atomically enough for our cooperative-ownership model. Compare-before-delete survives, but against one object instead of a registry string plus a detached process tree.
- **It answers the concurrency questions we've been circling.** The last two review rounds were spent building transaction semantics over the registry because Win32 exposes no registry-value CAS. The task store gives us a coarser but real unit of ownership for free.

Explicitly **not** requested: generations, supervisor, zero-downtime upgrade, rollback framework, or cross-session guarantees beyond what a single named task already provides. If the thin task turns out to need any of those, that is a signal to stop, not to build them.

### 3. Two remaining design-level items

**Fixed `127.0.0.1:8765` is machine-global while HKCU and the task are per-user.** On a shared Windows machine, the second user's setup hard-fails at `_endpoint_conflict_error` and there is no path forward for them. At minimum this limitation belongs in the docs and the error message ("another Windows user on this machine owns the Ouroboros endpoint"); a per-user derived port would remove the limitation outright, though I'm fine deferring that if documented.

**Gate the whole Windows HTTP path as experimental/opt-in rather than default `auto`.** The root cause is an upstream Codex Desktop bug (openai/codex#36778) in an alpha app-server build. Writing logon persistence into every Windows user's account by default is a heavy, long-lived footprint for a workaround we expect to delete once upstream fixes the stdio cleanup path. Suggest: opt-in via an explicit setup mode or config key, with `auto` picking it up by default only after we've decided upstream isn't fixing this on a useful timescale. This also honestly matches the docs' own "experimental" framing.

### Minor notes (non-blocking)

- `_STARTUP_COMMAND_MAX_LENGTH = 260` looks like a MAX_PATH carryover — `Run` REG_SZ values are not limited to 260 characters. Harmless as a conservative bound, but the error message claims a "HKCU Run limit" that doesn't exist as stated.
- The `uvx`-based launcher can hit the network at logon on a cold cache. Worth a doc note either way; with a scheduled task, a retry policy also absorbs this.

### Summary

Keep the HTTP topology, the strict initialize gate, and the preservation semantics exactly as they are. Replace the HKCU Run value with a thin per-user Scheduled Task (logon trigger + restart-on-failure), which restores #2024's "restartable" acceptance criterion, delivers actual zero-touch behavior, and lets a large fraction of this PR's hardest-won machinery be deleted rather than defended. Document or derive around the machine-global port, and land the whole Windows path behind an explicit opt-in while openai/codex#36778 is live upstream.

Happy to review the task-based revision quickly — the hard parts (verification gate, ownership model, preservation rules) are already done and carry over.

Q00:
@bokjk One follow-up on the previous direction review, from checking how the probe in `codex_http_mcp.py` interacts with the MCP SDK v2 stack this server actually runs on. Filing it as **non-blocking** — it doesn't fail today — but it is the same version-drift class as the launcher-shape whitelist, so worth fixing while the file is open.

### The strict `protocolVersion` equality in the probe is a time bomb under SDK v2

The startup command launches `uvx --from ouroboros-ai[mcp] ouroboros mcp serve …`, and the `[mcp]` extra pins `mcp==2.0.0` — so the HTTP server behind this endpoint runs on the Python MCP SDK v2. SDK v2's handshake negotiation (`mcp/server/runner.py`) works per spec:

- supported handshake set today: `2024-11-05`, `2025-03-26`, `2025-06-18`, `2025-11-25`
- if the client's requested version is in the set, the server echoes it; otherwise it responds with its latest supported version

`_response_is_ouroboros` requires the response to echo exactly `_MCP_PROTOCOL_VERSION = "2025-03-26"` and treats anything else as invalid. That passes today only because `2025-03-26` is still in the SDK's supported set. The SDK already carries `LATEST_PROTOCOL_VERSION = "2026-07-28"` internally, so the older handshake versions aging out of that set is a matter of time, not an edge case. When that happens:

1. Our own healthy server answers the probe with its latest supported version (per spec, a *correct* negotiation response).
2. The probe classifies it as `_ENDPOINT_OCCUPIED_INVALID`.
3. Fresh setup hard-fails with "occupied by a process that did not return a valid Ouroboros MCP initialize response" — a misdiagnosis pointing the user at a foreign process that doesn't exist. Re-running setup on an existing healthy install fails the same way, since the existing-value path runs the same probe.

The failure would arrive silently via a routine dependency bump, on a code path nobody exercises in CI on every SDK upgrade.

### Suggested fix

Identity is already established by `serverInfo.name == "ouroboros-mcp"` — that's the check doing the real work. The protocol-version check only needs to confirm a well-formed negotiation, not an exact era:

- accept any response whose `protocolVersion` is a plausible spec date-version string (e.g. `^\d{4}-\d{2}-\d{2}$`), keeping the existing `serverInfo` and structure checks as the identity gate; or
- if you want to stay stricter, accept `requested version OR anything newer` — which is exactly the spec's negotiation contract from the client's side.

Either variant is a few lines in `_response_is_ouroboros` plus one test asserting that a server responding with a newer negotiated version still probes as `ready-ouroboros`.

For completeness, two things I checked that are *fine*: Codex Desktop negotiates its own protocol version with the server independently of this probe, so client interop is unaffected; and since the probe sends a single `initialize` only, the `MCP-Protocol-Version` header requirement for subsequent requests (2025-06-18+) doesn't apply to it. The streamable-HTTP transport choice itself also aligns with where the spec is heading, so no concern on that front.

Q00:
Superseded by #2101, which has now merged and defines the accepted native-Windows Codex HTTP persistence contract: explicit opt-in, a thin per-user restart-on-failure Scheduled Task, and durable ownership-safe cleanup. This PR’s HKCU Run + detached one-shot architecture conflicts with that accepted direction, so continuing to patch it would preserve the wrong lifecycle boundary.

Q00:
Superseded by #2213 after an XY review of the actual failure boundary. The required product invariant is that setup never creates the native-Windows Codex Desktop stdio-child topology that crashes the app-server. #2213 enforces that invariant without introducing persistent background lifecycle and ownership machinery; explicit HTTP remains available with operator-owned process lifetime.

PR #2060, opened 2026-08-11
docs(readme): state both ways past the ambiguity gate

### LINKED ISSUE #2059: README states the ambiguity threshold as a necessary condition; the code has two ways past it
All three READMEs state the ambiguity threshold as a necessary condition:

> **Threshold: Ambiguity <= 0.2** -- only then can a Seed be generated.

That is not what the code does. `src/ouroboros/mcp/tools/authoring_handlers.py` gates on:

```python
if not interview_state.is_complete and not force:
```

The threshold is only consulted **while the interview is still open**. A state already marked complete dispatches Seed generation with `force=False`, whatever ambiguity score it carries. So there are two ways past the gate, and the README names neither.

`force` is at least visible: the CLI prints "Generate Seed anyway (force)" on screen, and the README's own demo caption mentions it. The completed-interview path is the one worth documenting, because from the outside it does not look like a bypass at all.

### Why this matters more here than elsewhere

This project's pitch is that a grading contract the worker cannot read is what makes its results trustworthy. A reader who takes that seriously will open the source. If the first hard number in the README does not survive that reading, neither does the pitch.

This same sentence has now been corrected in five other places (two Chinese list submissions, one dev.to post, two Qiita posts). The README is the last and largest surface carrying it.

### Also in scope

The comparison table row "Ambiguity gate (<= 0.2) blocks premature code" has the same shape in the English and Chinese READMEs. The Korean README has no equivalent row, which is a separate pre-existing difference between the three files and not something this issue proposes to change.

Q00:
Closes #2059.

Rewrites the ambiguity-threshold sentence in all three READMEs so it names both ways past the gate instead of presenting the threshold as a necessary condition.

Before:

> **Threshold: Ambiguity <= 0.2** -- only then can a Seed be generated.

After:

> **Threshold: Ambiguity <= 0.2.** While the interview is still open, a score above that blocks Seed generation. Two things get past the check: passing `force` explicitly, which the CLI offers on screen, or marking the interview complete, which does not look like a bypass at all. The gate is a default worth arguing with, not a lock.

The comparison-table row gains a clause in the English and Chinese files for the same reason. The Korean README has no equivalent row; that difference predates this PR and I left it alone rather than inventing one.

### Verification

`authoring_handlers.py` gates on `if not interview_state.is_complete and not force:`, so a completed interview state dispatches Seed generation with `force=False` regardless of the stored score. I read that before writing the replacement rather than taking the review comment on #2040 at face value, since this particular claim has been narrowed repeatedly and I did not want the correction itself to be a guess.

### Context

The same sentence has already been corrected on five external surfaces today: two Chinese list submissions, one dev.to post and two Qiita posts. The README was the last and largest one still carrying it.

### What I did not verify

I did not render the three pages in a browser. The change is prose inside existing paragraphs and one table cell, so the structural risk is low, but that is a reason and not a check.

---

### Update: swept the rest of the repo, two more files

After opening this, I pulled the repository tarball and grepped all 138 markdown files rather than trusting code search, which has returned zero for control queries before. Two more carried the same shape:

- `docs/README.md` — "Requirements **must** have ambiguity score <= 0.2 before execution begins", in the Core Principles list.
- `docs/getting-started.md` — "Answer until the ambiguity score drops below 0.2", in the walkthrough. Softer, but a reader who wants to stop early is not told they can.

Both are in this branch now.

**What I deliberately did not change:**

- `docs/hidden-checklist-convergence/requirements.md` says grading inputs are "hidden from implementation workers", but it says so under a `## Original Request` heading. That is a record of what was asked for, not a claim the project is making today. Rewriting it would falsify the record.
- `AGENTS.md` and `CLAUDE.md` say "Question until ambiguity <= 0.2". Those instruct an agent on what to do; they do not describe what the harness enforces.
- `docs/guides/evolution-loop.md` frames the same number as a principle ("do not build until you are clear"), which is a stance rather than a mechanism claim.

I ran the search patterns against a file known to contain the wording first, and they returned five hits there, so the near-empty result on the rest of `docs/` is a real absence rather than a broken query.


Q00:
You are right, and I checked it in the source rather than taking the review at its word, because this particular sentence has now been wrong in both directions in one day.

`bigbang/seed_generator.py:1762-1780` is the gate that matters:

```python
if force:
    ...ambiguity_gate_bypassed...
elif not ambiguity_score.is_ready_for_seed:
    return Result.err(ValidationError(...))
```

There is no `is_complete` branch there. And `authoring_handlers.py:1573` passes `force=force` straight through, so the in-process path lands on that gate unchanged. The `is_complete` skip at `:1385` only bypasses the handler's own pre-check, one layer above.

So `force` is the portable answer and the completed-interview path is a plugin-dispatch inconsistency, not a second product-level way past. All five files in this branch now say that.

**What I got wrong and why it matters beyond this PR.** I found the `is_complete` skip while addressing a different review, stopped at the first gate I saw, and wrote "two ways past" into README ×3, docs ×2, both plugin manifests, roughly a dozen pages of the Chinese guide, three Chinese list submissions, one dev.to post and two Qiita posts. I have started the sweep and will report the count when it is done.

The pattern is worth naming: I spent today narrowing over-claims, which trained me to treat "the real behaviour is looser than we said" as the likely answer. That bias produced this one. **A correction is a claim, and it needs the same verification as the thing it replaces** — I found one gate, did not ask whether a second gate ran underneath it, and shipped.

**One thing I am not doing.** I am not documenting the plugin/in-process difference as a feature. If a completed interview really does skip the handler's pre-check on one path only, that is an inconsistency someone should decide about rather than something a reader should learn to rely on. Happy to file it separately if you agree.

Issue #2057, opened 2026-08-11
README shows the logo before it shows the tool working


Q00:
The `ooo interview` recording is the clearest evidence we have that this tool does something specific. It is at line 140 of a 600-line README, which is roughly two scrolls down on a laptop viewport.

Above it, a visitor sees: a language switcher, a 520px logo, a wordmark, five badges, a trendshift badge, and a nav bar. None of that shows the product working.

### Measured

| | |
|---|---:|
| First `ooo-interview.gif` reference, `README.md` | line 140 of 600 |
| Same, `README.ko.md` | line 143 of 542 |
| Same, `README.zh-CN.md` | line 107 of 562 |
| `docs/images/ouroboros.png` | 1,374 KB |
| `docs/images/ooo-interview.gif` | 729 KB |

The last two rows are the part I did not expect. The heaviest asset on the first screen is the decorative logo, and it costs nearly twice what the demonstration costs.

### Why this is worth fixing rather than filing away

The asset already exists and was already recorded. Nothing needs to be produced. This is purely a placement problem, and search is currently the largest single source of repository traffic, so the people affected are already arriving.

### Scope

Two separable changes:

1. **Placement** — move the recording above the fold in all three READMEs. Doing one language and not the others creates the drift this repo keeps having to clean up.
2. **Logo weight** — re-encoding a 1,374 KB PNG is a judgment call about image quality that I would rather leave to someone with an opinion about it. Filing it here rather than bundling it.

shaun0927:
Closing the placement scope as completed by #2055 and preserved by the later hero-grid work. The interview demonstration remains above the fold in every README. Logo re-encoding and asset weight were explicitly a separate visual-quality decision and are not claimed complete here.

PR #2063, opened 2026-08-11
docs(readme): show all four hosts as one grid

Q00:
Refs #2057. Supersedes #2058.

Replaces the single hero recording with a four-panel grid: the same interview running in the terminal CLI, in ChatGPT via Codex, in Claude Code, and in Hermes on Discord.

### Why

The line under the hero says thirteen runtimes. Until now the page showed one, a terminal, and asked the reader to take the rest on faith. Four panels turn that sentence into something a visitor can check in about ten seconds.

Each panel shows the same stage of the same workflow, so the grid reads as one engine in four places rather than four unrelated screenshots.

### Layout

Two by two, not one by four. At GitHub's content width four across gives each panel about 210px, where terminal text is not readable. Two by two gives about 440px, which I checked by decoding a frame at final quality and confirming the headline lines are legible.

### Assets

| panel | file | size | note |
|---|---|---:|---|
| Terminal CLI | `ooo-interview.gif` | 729 KB | already in the repo, unchanged, rendered at 440px |
| ChatGPT (Codex) | `host-codex.gif` | 597 KB | new |
| Claude Code | `host-claude.gif` | 952 KB | new, sped up from a 250s capture to 30s |
| Hermes (Discord) | `host-hermes.gif` | 760 KB | new, sped up and trimmed to 18s |

New bytes: 2.3 MB. That is real weight and worth stating rather than burying. I did not re-encode the existing CLI recording, because it is already well compressed at 729 KB and a re-encode at 460px came out at 3.4 MB.

For comparison, `docs/images/ouroboros.png` is 1,374 KB on its own and shows nothing about the product. That imbalance is #2057.

### Supersedes #2058

That PR put the Codex recording under the hero as a single supporting panel. This does the same job better, so I will close it once this is reviewed rather than land both.

### What I did not verify

I did not render the three pages in a browser. I checked HTML tag balance (4 `<td>`, 2 `<tr>`, 1 `<table>` per file) and that each new image is referenced exactly once. If the grid stacks badly on narrow viewports, the table can drop to one column.


Q00:
Both findings are correct and both are mine. Fixed in all three READMEs.

**The blocker.** I wrote "the same interview" and the PR body claimed all four panels show the same stage of the same workflow. They do not. The four recordings are a task-management CLI, a video-publishing harness, a YouTube automation task, and a kart-racing game. I never checked whether the clips shared a prompt — I assumed it because the grid *reads* better that way, and then wrote the assumption into the captions.

The captions now name each task explicitly, and the heading says the point out loud: four separate runs, four hosts, different tasks on purpose, **the engine is what is shared, not the prompt**. That is the honest version of the claim, and it is arguably the stronger one: four unrelated tasks landing in the same shape says more about the engine than four replays of one prompt would.

**The lane count.** You are right that the contract defines six lanes, not five. I counted them in `interview_schemas.py`: `code_context`, `web_context`, `data_context`, `ambiguity_contrarian`, `answer_simplifier`, `architecture_implications`, with `data_context` required.

I wrote "five" because the Codex recording shows five on screen. Rather than caption a number the frame does not support, or a number the contract does not support, the Codex caption now says "its advisory lanes" with no count. The Claude Code panel does show all six, so that caption keeps the number.

**Not re-recording.** Replacing the clips with one shared prompt would make the panels prettier and would cost the thing that makes them evidence: these are four real runs, not a staged comparison. If you would rather have the staged version, say so and I will record it, but I would be trading a true claim for a tidier one.

What I take from this: I described a similarity I had not verified, in a PR whose entire purpose was to let readers verify a claim instead of taking it on faith.

PR #2064, opened 2026-08-11
docs(readme): that is not the same interview

Q00:
Refs #2057.

One sentence, three files. The caption merged in #2058 says the Codex recording is "the same interview" as the terminal recording above it. It is not.

- Hero recording: a task-management CLI.
- Codex recording: a video-publishing harness.

Two different runs of two different tasks. I wrote "the same" because it made the comparison read better and I never put the two clips side by side to check.

### Why this is its own PR

#2063 replaces that whole block with a four-panel grid and fixes the wording as part of the rewrite, but it is blocked on review. This sentence is live on `main` right now, in three languages, so it should not wait behind a larger change.

### How it got in

#2058 was approved and merged. The review on #2063 caught the identical mistake in the grid captions, which is how I found this one. **An approval means nothing was caught at review time, not that the content is true** — the review predates the finding, and I only went looking because a later reviewer pushed back on the same claim elsewhere.

If #2063 lands afterwards it supersedes this text anyway; this just stops the false version from sitting on the front page in the meantime.

---

### Second fix on the same branch: a lane count

The image alt text in all three files says "five advisory perspectives run beside it". The recording does show five. The interview contract defines **six** lanes — `code_context`, `web_context`, `data_context`, `ambiguity_contrarian`, `answer_simplifier`, `architecture_implications` — with `data_context` required (`orchestrator/capabilities/interview_schemas.py`).

So the number is right about the frame and wrong about the product, and alt text does not give a reader room to tell those apart. It now says "its advisory perspectives", with no count.

I found this while control-testing a sweep. I had searched the guide site and `drafts/` for this class of claim, got zero hits, and then ran the same patterns against `main` to check the search actually worked. It did — and it turned up this second live claim that my own fix had not covered. **The control was there to validate the method and it caught a defect instead.**


Q00:
d

PR #1926, opened 2026-08-07
fix(verification): bind regex evidence to acceptance input

Q00:
Closes #1835

## Direction

#1835 remains a valid trust-boundary bug, while #1837 already closed the nullable/universal-regex family. This PR closes the residual where non-nullable model regexes, unbound public assertions, or non-executable text could mint formal PASS.

The final invariant is:

> T1/T2 regex evidence may mint PASS only when public construction is bound by default, every successful regex path requires a deterministic target derived from the caller-authored acceptance criterion, the concrete match binds that exact trusted executable-source/path span, and execution belongs to a statically bounded subset.

The earlier counterfactual/mutated-file approach was discarded because it executed hostile model regexes on additional subjects and opened another ReDoS path.

## What changed

- Caller-owned AC text determines targets, polarity, inventory, and evidence provenance; stale model metadata fails closed.
- Public/deserialized `SpecAssertion` construction defaults to `input_binding_required=True`. `False` remains only as an explicit trusted direct-caller compatibility choice.
- Regex AST analysis requires the complete target literal on every successful consuming path. Optional repeats, incomplete branches, failing branches, and global co-presence lookaheads do not count.
- Zero-width evidence is limited to finite positive assertions whose exact match-relative offset is statically known.
- Runtime binding requires exactly one regex-equivalent target occurrence in the match/witness window, equal to the trusted source/path span.
- Positive content evidence is evaluated against offset-identical source-aware text; scalar extraction still reads the original text at the proven offsets.
- Python comments/strings use the standard tokenizer. Rust/Haskell nested comments, Lua extended long brackets, JavaScript comments/strings/templates/regex literals, and SQL comments are handled explicitly. Markup formats remain valid filename/path evidence but cannot prove executable T2 declarations.
- JavaScript regex classification covers arrow/operator/keyword expression starts; ambiguous slash contexts fail the source closed instead of exposing a possible regex body.
- C++ raw strings, Java/Kotlin text blocks, Swift extended strings, C# raw/verbatim strings, and Rust raw strings are masked with their language-specific delimiters.
- Python ordinary, raw, interpolated, and multiline f-string literal segments are masked while genuine following declarations remain searchable.
- Markdown, markup, stylesheet, and data formats cannot prove executable structural declarations. Ambiguous `.m` files fail closed rather than guessing Objective-C versus MATLAB. Configuration syntax is available only to T1 constants. Raw `.txt` content is limited to explicit trusted compatibility or non-identifier observable-output tokens.
[...]

PR #1837, opened 2026-08-01
fix(evaluation): refuse regexes that match any input as verification evidence

MyoungSoo7:
Refs #1835

## What

`SpecVerifier._safe_compile` asked only whether a model-supplied pattern compiles. A
pattern that matches the empty string matches _every_ subject, so it compiles fine,
matches the first candidate file, and returns `verified=True, detail="Found file:
main.py"` for an acceptance criterion the project does not satisfy.

Reproduced through the public `SpecVerifier.verify_all` on a project containing only
`main.py` (`print('hello')`) and no `CameraProvider` anywhere, with the agent
**honestly reporting AC0 as FAIL**:

| pattern                                | before                                          | after            |
| -------------------------------------- | ----------------------------------------------- | ---------------- |
| `class CameraProvider` (control)       | `verified=False` — "not found in 1 files"       | unchanged        |
| `.*` `x?` `\s*` `(?:)` `\|` `^` `\A\Z` | **`verified_pass=True`**, `Found file: main.py` | `verified=False` |

`ACVerificationReport.verified_pass` feeds `_evaluation_summary_from_spec_verification`
(`mcp/server/adapter.py`), which produces `ACResult(passed=True, score=1.0,
final_verdict="pass", verification_method="spec_verifier")`. The spec verifier exists
to catch agent self-report lies; here it overturned an honest failure into a pass.

## The fix

Refuse the pattern in `_safe_compile` when its compiled form matches `""`.

**Why at the verifier and not at extraction.** A pattern dropped during extraction
produces no report at all, and the adapter reads "no reports" as "nothing to check",
leaving the agent's self-report unexamined — a refusal that silently _removes_ the
check. Refusing at the verifier records a result and raises a discrepancy;
`test_empty_matching_pattern_fails_closed_against_an_agent_pass_claim` pins that,
including `override_approval is False`.

**Why not a match-width rule.** Refusing zero-width _matches_ also refuses
lookahead-only patterns such as `(?=class CameraProvider)`, which genuinely
discriminate — reporting an honest PASS as a discrepancy, i.e. the same failure this
change fixes running the other way. `test_discriminating_pattern_still_verifies` pins
all four lookahead forms. It also keeps the check free of executing the untrusted
pattern against any sentinel string, so it adds no backtracking surface.

## Scope, and what is deliberately left open

This closes the family that matches **any** input. It does not close patterns that
match **unrelated** input — `.`, `.+`, `[\s\S]`, `[A-Z]\w*`, or a bare declaration
[...]

PR #2065, opened 2026-08-11
fix(verification): refuse unbound regex evidence as grounds to overturn an agent FAIL

### LINKED ISSUE #1835: Spec verifier: a model-supplied regex that matches any input fabricates a verified PASS and overturns an honest agent FAIL
## Summary

A model-supplied regex that matches any input passes the spec verifier's usability
gate, then matches the first candidate file and mints `verified=True` for an
acceptance criterion the project does not satisfy. The spec verifier exists to
catch an agent's self-report lies; here it overturns an honest FAIL into a PASS.

`_is_usable_regex_pattern` (`src/ouroboros/verification/extractor.py`) asks only
whether the pattern compiles:

```python
try:
    re.compile(pattern)
except (re.error, OverflowError):
    return False
return True
```

`SpecVerifier._safe_compile` applies the same compile-only standard, and
`_verify_structural` then accepts whatever `search()` returns.

## Reproduction

Project containing no `CameraProvider` anywhere:

```
main.py          -> print('hello')
pyproject.toml   -> [project] name='x'
```

AC: `MUST define a CameraProvider interface`, tier `t2_structural`, file_hint `*.py`.
Driving the public `SpecVerifier.verify_all((assertion,), agent_results={0: False})`
— i.e. the agent honestly reported this AC as **FAIL**:

| pattern                                                            | usability gate | result                                                 |
| ------------------------------------------------------------------ | -------------- | ------------------------------------------------------ |
| `class CameraProvider` (control)                                   | usable         | `verified=False` — "not found in 1 files" ✔            |
| `x?` `\s*` `(?:)` `.*` `a*` `\|` `(?i)` `\b` `(?=m)` `^` `(?!zzz)` | **usable**     | **`verified_pass=True`**, detail `Found file: main.py` |

Eleven patterns, all admitted, all fabricating evidence for an unsatisfied
criterion — and all overturning the agent's honest FAIL.

## Why one check is not enough

The eleven split into two shapes that defeat different defences:

- **`\b` and `(?=m)` do not match the empty string.** An empty-string prefilter
  cannot see them. They match _zero-width_ against a real filename, so what has to
  be rejected is the match, at the point it becomes evidence.
- **`.*` is not zero-width** — it consumes the whole basename, so a match-width
  check cannot see it. It matches any input regardless of contents, which is what a
  pattern-level predicate must reject.

Any fix that adds only one of the two leaves the other family live.

## Blast radius

`ACVerificationReport.verified_pass` feeds
`_evaluation_summary_from_spec_verification` (`src/ouroboros/mcp/server/adapter.py`),
which produces `ACResult(passed=True, score=1.0, final_verdict="pass",
verification_method="spec_verifier")` — the record downstream evaluation treats as
mechanically verified rather than self-reported.

## Existing coverage

`test_wrapped_invalid_regex_assertion_rejected_before_verifier` covers only `(`,
which is uncompilable. `test_overflowing_regex_assertion_rejected_before_verifier`
covers `a{9999999999}`. No test anywhere under `tests/` exercises a pattern that
compiles but matches everything.

## Suggested fix

Reject a pattern whose compiled form matches the empty string, and require a match
to consume at least one character before it counts as evidence. Keeping one shared
predicate at the layer that turns a match into evidence — with the extractor
pre-filtering through the same function — avoids the two gates drifting apart.

Q00:
Closes #1835. Replaces #1926 (closed by owner decision — see its closing comment).

## The defect

A model-supplied regex that matches any file with content in it — `[\s\S]+`, `CameraProvider|[\s\S]+`, `.+`, or on the T2 filename path anything matching a basename — is not nullable, so #1837's emptiness guard cannot see it. It matches the first candidate, produces an all-`VERIFIED` report, and `evaluation_summary_from_spec_verification` promotes that to a formal PASS **without consulting the agent's own result** — overturning an honest agent FAIL, which is what #1835 reports.

## The invariant

> **Source-scan evidence cannot overturn an agent-reported FAIL.**

Confirmation of an agent-reported PASS is untouched. That is the check #1835 says this scanner exists for — catching a self-report lie — and it claims no authority the agent had not already claimed.

## Why the authority is withdrawn rather than narrowed

The first two attempts qualified the override instead of removing it, admitting evidence that was "about the criterion". Both were bypassed in review:

| rule | bypass |
|---|---|
| shared criterion words above a length floor | `class\s+\w+` for "a CameraProvider **class**" matched `class Unrelated` |
| named targets read from code spelling | `MUST` for "The implementation **MUST** define…" matched `# MUST clean this up later`, reaching `final_approved=True` |

Both are the same failure. Whether a text names a criterion's subject is not a question a finite reading of prose answers, so each repair moved the hole rather than closing it — the non-terminating shape that made #1926 unmergeable, at smaller scale. Excluding `MUST`/`SHALL` would have been the third narrowing, with `SHOULD`, `Given`/`When`/`Then`, `API` and sentence-initial capitals waiting behind it.

So the prose-to-target classifier is **deleted**, not corrected. No property of a pattern restores the override, which is what makes this closed rather than merely narrower: there is nothing left to bypass.

## Enforced at every boundary that can carry it

| boundary | enforcement |
|---|---|
| `SpecVerifier.verify_all` (producer) | never emits `VERIFIED` for an AC the agent reported FAIL; demotes to `UNVERIFIABLE` |
| `evaluation_summary_from_spec_verification` (formal verdict) | refuses to honour that combination, so replayed, legacy and compatibility summaries cannot encode what the producer forbids |
| agent-result source at that boundary | taken from the mechanical execution record, not the report's copy of it; either record saying "not a pass" settles it, and disagreement is never resolved toward the permissive one |

Three places because none of them passes through the others: rehydrated summaries never meet the producer, and a report's `agent_reported_pass` can contradict the execution or be absent entirely (reading as `True`).

`UNVERIFIABLE`, not `DISCREPANCY`: the evidence is unusable in this direction, which is not evidence that the criterion is unmet. The formal verdict renders `NOT_EVALUATED`.

## Deliberately given up

`class\s+CameraProvider` against a real `class CameraProvider`, and `\A\Z` against a genuinely empty named file, no longer overturn a FAIL either. Both were allowed by earlier revisions. An exemption is a bypass surface, and these were the last two; an agent that reported FAIL keeps its FAIL.

No executable-source classification is attempted. `CameraProvider` inside a comment satisfying a pattern remains a known limitation of a grep-level tier — the tier for semantic proof is T3, not a parser fleet. That was #1926's approach and it is not reopened here.

## Coverage

- `test_no_regex_evidence_overturns_an_agent_fail` — every reported bypass plus the two now-refused legitimate shapes, across T1 and T2 including the filename path
- `test_agent_pass_confirmation_is_untouched` — bound, unbound and blank-subject patterns still confirm a PASS
- `test_an_agent_fail_survives_all_the_way_to_the_formal_verdict` — end to end through the formal adapter
- `test_a_verified_report_over_an_agent_fail_cannot_mint_a_formal_pass` — direct, JSON-round-tripped and legacy-boolean summaries
- `test_report_polarity_cannot_outrank_the_execution_record` — report claims PASS / omits the field / contradicts a passing execution

One existing expectation tightened: in `test_spec_verification_does_not_approve_failed_execution` the worker reported its task incomplete while the report claimed PASS, so that AC is now `NOT_EVALUATED` rather than a passing result inside an already-rejected run. Run-level assertions are unchanged. That combination is the shape #1835 describes.

## Size

Production diff is `src/ouroboros/verification/verifier.py` +49 lines, one branch and one helper in `spec_verification_adapter.py`. For comparison, #1926 spent 16,555 lines and 64 review rounds on the same issue without converging.

## Verification

- Local `tests/unit` + `tests/conformance`, CI's 4-worker xdist configuration on Python 3.14: **20,126 passed, 88 skipped, 0 failed**
- `tests/unit/test_verification.py` + `tests/unit/mcp/`: 2,354 passed, 1 skipped
- Ruff check across `src/ouroboros` and `tests/unit`, format, MyPy on both changed source files: clean
- GitHub Actions on `17d1123e`: all checks green

### DIFF
--- a/src/ouroboros/mcp/server/spec_verification_adapter.py
+++ b/src/ouroboros/mcp/server/spec_verification_adapter.py
@@ -77,6 +77,32 @@ def _report_identity_error(
     return None
 
 
+def _agent_claimed_pass(
+    report: Any,
+    expected_agent_results: dict[int, bool],
+    ac_index: int,
+) -> bool:
+    """Whether the agent itself claimed this AC passed.
+
+    Two records answer this and only one of them is authoritative. The
+    mechanical summary is the execution's own account of what the agent
+    reported; the report's ``agent_reported_pass`` is a copy the verifier was
+    handed, and a replayed, stale or externally built report carries whatever
+    copy it was constructed with — including a `True` that contradicts the
+    execution, or no field at all, which reads as `True` by default. Trusting
+    the copy alone lets either shape turn a mechanically failed AC into a
+    formal PASS.
+
+    So both are consulted and either one saying "not a pass" settles it.
+    Disagreement is not resolved in favour of the more permissive record: two
+    accounts of the same fact that do not match are not evidence that the
+    agent claimed a pass.
+    """
+    if not bool(getattr(report, "agent_reported_pass", True)):
+        return False
+    return bool(expected_agent_results.get(ac_index, True))
+
+
 def agent_results_from_execution_summary(mechanical: Any) -> dict[int, bool]:
     """Return legacy agent-reported AC outcomes for spec verification.
 
@@ -184,8 +210,8 @@ def evaluation_summary_from_spec_verification(
                 skipped_indices.append(ac_index)
             if not evidence:
                 evidence = "Spec verifier found a discrepancy without evidence details."
-        elif outcomes == {VerificationOutcome.VERIFIED} and not getattr(
-            report, "agent_reported_pass", True
+        elif outcomes == {VerificationOutcome.VERIFIED} and not _agent_claimed_pass(
+            report, expected_agent_results, ac_index
         ):
             # Source-scan evidence may confirm a claimed PASS; it may not
             # reverse a reported FAIL. `SpecVerifier.verify_all` already
--- a/src/ouroboros/mcp/server/spec_verification_adapter.py
+++ b/src/ouroboros/mcp/server/spec_verification_adapter.py
@@ -184,6 +184,26 @@ def evaluation_summary_from_spec_verification(
                 skipped_indices.append(ac_index)
             if not evidence:
                 evidence = "Spec verifier found a discrepancy without evidence details."
+        elif outcomes == {VerificationOutcome.VERIFIED} and not getattr(
+            report, "agent_reported_pass", True
+        ):
+            # Source-scan evidence may confirm a claimed PASS; it may not
+            # reverse a reported FAIL. `SpecVerifier.verify_all` already
+            # refuses to emit this combination, but this adapter is a public
+            # authority boundary that also accepts replayed, legacy and
+            # compatibility summaries built elsewhere. Enforcing the polarity
+            # only in the producer would let a rehydrated report encode the
+            # exact transition the producer forbids, so the boundary that
+            # mints the formal verdict checks it too.
+            unverifiable_indices.append(ac_index)
+            passed = False
+            verdict_state = "not_evaluated"
+            rendered_verdict = "NOT_EVALUATED"
+            evidence = (
+                "Spec verification reported all assertions verified against an "
+                "agent-reported FAIL; source-scan evidence cannot overturn a "
+                "reported FAIL, so the formal AC verdict is not evaluated."
+            )
         elif outcomes == {VerificationOutcome.VERIFIED}:
             passed = True
             verdict_state = "evaluated"
--- a/src/ouroboros/verification/verifier.py
+++ b/src/ouroboros/verification/verifier.py
@@ -701,21 +701,72 @@ def _required_literal_runs(
     return frozenset(runs)
 
 
+def _contains(words: list[str], run: list[str]) -> bool:
+    """Whether `run` appears in `words` as a contiguous stretch."""
+    span = len(run)
+    if not span or span > len(words):
+        return False
+    return any(words[start : start + span] == run for start in range(len(words) - span + 1))
+
+
+# A criterion is prose, and somewhere in the prose is the thing it is about.
+# What separates that thing from the words around it is how the caller spelled
+# it: code names are written as code — `CameraProvider`, `camera_provider`,
+# `WARMUP_FRAMES`, `notes.txt` — or, for a one-word name, capitalised where
+# ordinary words are not. Reading that is what tells `CameraProvider` apart
+# from the `class` beside it.
+#
+# Read from spelling and not from a list of words too ordinary to name
+# anything. Such a list would have to hold every structural noun in English —
+# class, interface, file, module, handler, service — and the next criterion
+# would use the one it had not thought of yet. Spelling is a rule; a list of
+# ordinary words is an argument that never finishes.
+_CRITERION_TOKENS = re.compile(r"[A-Za-z0-9][A-Za-z0-9_.]*")
+
+
+def _named_targets(ac_text: str) -> list[list[str]]:
+    """What the criterion names, read from how the caller spelled it.
+
+    A token spelled as a compound — more than one word inside a single token,
+    which is what CamelCase, snake_case, SCREAMING_CASE and a dotted filename
+    all are — names something. So does a capitalised token standing among
+    ordinary words. The criterion's first token is not read as capitalised,
+    because starting a sentence with a capital is English rather than naming.
+    """
+    targets: list[list[str]] = []
+    for position, token in enumerate(_CRITERION_TOKENS.findall(ac_text)):
+        pieces = _words(token)
+        if not pieces:
+            continue
+        if len(pieces) > 1 or (position > 0 and token[:1].isupper()):
+            targets.append(pieces)
+    return targets
+
+
 def _binds_criterion(pattern: str, ac_text: str) -> bool:
     """Whether a match of this pattern is about what the criterion names.
 
-    True when some text every match must contain spells out, word for word, a
-    run of words the caller wrote in the acceptance criterion — with at least
-    one word of substance among them. Such a pattern cannot match a file that
-    lacks what the criterion asked about; whatever else its match fails to
-    prove, it is evidence *about the criterion*. A pattern with no such text —
-    `[\\s\\S]+`, `CameraProvider|[\\s\\S]+` — matches files the criterion has
-    nothing to say about, so its match says nothing back.
+    Some text every match must contain has to spell out the thing the
+    criterion named. `class\\s+CameraProvider` requires `CameraProvider` and
+    cannot match a file without it; `class\\s+\\w+` requires only `class`, so
+    it matches `class Unrelated` in a project that has no CameraProvider at
+    all, and calling that evidence for the criterion is the defect this guard
+    exists to close. Sharing the criterion's structural vocabulary is not
+    naming its target.
 
-    The comparison is over words, not characters, so `CameraProvider`,
-    `camera_provider` and the criterion's own "CameraProvider interface" all
-    reach each other. Flags are not consulted: what words a literal spells is
-    not something a flag changes.
+    Where the criterion names nothing in code spelling — "maximum 5 retries",
+    all ordinary lowercase words — there is no target to read, and this falls
+    back on the criterion's longest words. That is a weaker reading, and
+    deliberately so: it is the strongest one available from a criterion whose
+    own text does not distinguish its target, and tightening it further would
+    refuse `RETRIES\\s*=\\s*` for that criterion, which is exactly the evidence
+    a T1 assertion is supposed to bring. Its limit is that a criterion written
+    that way can be bound by a long structural word; a criterion that names
+    what it is about — `Cache`, `CameraProvider` — never reaches it.
+
+    The comparison is over words, so `CameraProvider`, `camera_provider` and
+    `CAMERA_PROVIDER` all reach each other. Flags are not consulted: what
+    words a literal spells is not something a flag changes.
     """
     try:
         parsed = regex_parser.parse(pattern)
@@ -724,19 +775,21 @@ def _binds_criterion(pattern: str, ac_text: str) -> bool:
     criterion_words = _words(ac_text)
     if not criterion_words:
         return False
-    for run in _required_literal_runs(parsed):
-        run_words = _words(run)
-        if not run_words:
-            continue
-        if max(len(word) for word in run_words) < _MIN_BINDING_WORD_LENGTH:
-            continue
-        span = len(run_words)
-        if any(
-            criterion_words[start : start + span] == run_words
-            for start in range(len(criterion_words) - span + 1)
-        ):
-            return True
-    return False
+    run_words = [words for words in map(_words, _required_literal_runs(parsed)) if words]
+    if not run_words:
+        return False
+
+    targets = _named_targets(ac_text)
+    if targets:
+        return any(_contains(words, target) for words in run_words for target in targets)
+
+    longest = max(len(word) for word in criterion_words)
+    if longest < _MIN_BINDING_WORD_LENGTH:
+        return False
+    return any(
+        _contains(criterion_words, words) and any(len(word) == longest for word in words)
+        for words in run_words
+    )
 
 
 def _skip_inline_space(text: str, index: int) -> int:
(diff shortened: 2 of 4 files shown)


Q00:
## Blocker 1 addressed — `42f06334982ff31c36e77b2f8d9ebe81af12e1f7`

The finding was correct, and both reproductions are now regressions.

**Root cause of the weak invariant**: the three-character floor asked whether a shared word was *long*, which is not the same question as whether it *names the criterion's target*. `interface` is nine characters and is the longest word in "MUST add a Cache interface" — no length rule can separate structural vocabulary from a named target.

**Fix**: read the target from how the caller spelled it. A criterion names its subject the way code names things — a compound token (`CameraProvider`, `camera_provider`, `WARMUP_FRAMES`, `notes.txt`) or a capitalised token standing among ordinary words (`Cache`). When a criterion names one, some required literal must spell it out; sharing `class`, `file`, `interface` or `def` no longer binds.

Deliberately **not** a list of generic words. Such a list would have to contain every structural noun in English and would grow one entry per counterexample indefinitely — a review loop with no closure rather than an invariant. Orthography is a rule that terminates.

**Both reported reproductions, now regressions** (`test_structural_vocabulary_alone_does_not_bind_a_named_target`):

| criterion | pattern | source | before | after |
|---|---|---|---|---|
| `MUST define a CameraProvider class` | `class\s+\w+` | `class Unrelated` | PASS | UNVERIFIABLE |
| `MUST create a CameraProvider file` | `file` | `profile.py` (filename path) | PASS | UNVERIFIABLE |
| `MUST add a Cache interface` | `interface\s+\w+` | `interface Unrelated` | PASS | UNVERIFIABLE |
| `MUST register a CameraProvider handler` | `def\s+\w+` | `def unrelated()` | PASS | UNVERIFIABLE |

Bound overturns are preserved (`test_a_named_target_still_binds_when_spelled_out`): `class\s+CameraProvider` and `class\s+Cache\b` still overturn an agent FAIL against real source. All previously listed behaviour is unchanged.

**Declared limit, stated rather than hidden.** When a criterion names nothing in code spelling — "maximum 5 retries", ordinary lowercase throughout — there is no target to read and the longest-word reading remains, which is what admits `RETRIES\s*=\s*`. A criterion written that way can therefore be bound by a long structural word. This is a property of the criterion text, not of the rule: caller-authored prose that does not distinguish its target cannot be read more precisely than it was written, and tightening the fallback refuses the exact evidence a T1 assertion exists to bring. Any criterion that names what it is about never reaches this path. The fallback still refuses short structural words (`file` against "MUST create a config file") — regression included.

Local: `tests/unit/test_verification.py` 345 passed (7 new); `tests/unit/mcp/server/test_adapter.py` 238 passed; Ruff check/format and MyPy clean.

Separately: the `Test Python 3.14` job on the previous head ended with its pytest output truncated at 26% rather than a reported test failure, with no failing test in the log. Watching the run on this head to determine whether that is a worker crash or an infrastructure flake; I will report either way rather than re-running until it passes.

Q00:
## Blocker 1 addressed by removing the classifier — `9cb2fd13f956f8b7124f0a593f18f957809450ba`

The finding was correct and reproduces exactly as described. It is now fixed by deleting the rule it found a hole in, rather than by patching the hole.

**Why not exclude `MUST`/`SHALL`.** That would be the third narrowing of the same rule, and the first two each produced a bypass: reading criterion words admitted `class\s+\w+` through the shared word `class`; reading named targets from spelling admitted `MUST` through requirement prose. The next exclusion list would meet `SHOULD`, `SHALL`, `Given`/`When`/`Then`, `API`, `TODO`, sentence-initial capitals after a colon, and so on. Whether a text names a criterion's subject is not a question a finite reading of prose answers, so each repair moves the hole. That is the shape that made #1926 unmergeable, reappearing at smaller scale, and I would rather stop it here than run the loop again.

**What changed instead.** The override authority is withdrawn, not qualified:

> Source-scan evidence cannot overturn an agent-reported FAIL, whatever it matched.

`_named_targets`, `_binds_criterion`, `_required_literal_runs` and the word-normalisation helpers are **deleted** — 170 lines of prose classification removed. No property of a pattern restores the override, so there is nothing left to bypass. Production diff is now +49 lines in `verifier.py`.

Your exact reproduction, on this head:

```
outcome: unverifiable
final_approved: False | rendered_verdict: NOT_EVALUATED | run_verdict: FAIL
```

**Also refused now, deliberately**: `class\s+CameraProvider` against a real `class CameraProvider`, and `\A\Z` against a genuinely empty named file. Both were previously allowed to overturn. Refusing them is the point — an exemption is a bypass surface, and these are the last two that existed.

**What is preserved**: confirmation of an agent-reported PASS, unchanged for every pattern shape including loose ones. That is the check #1835 says this scanner exists for — catching a self-report lie — and it claims no authority the agent had not already claimed. The `\A\Z` named-empty-file verification and every pre-existing emptiness-guard behaviour are untouched on that path.

**End-to-end regression added**, as requested: `test_an_agent_fail_survives_all_the_way_to_the_formal_verdict` drives criterion → hostile pattern → matching-but-unrelated source → honest agent FAIL through `evaluation_summary_from_spec_verification` and asserts `final_approved is False`. The table-driven `test_no_regex_evidence_overturns_an_agent_fail` covers all three reported bypasses plus the two now-refused legitimate shapes, across T1 and T2 including the filename path.

Local on this head: `tests/unit/test_verification.py` + `tests/unit/mcp/server/test_adapter.py` = 572 passed; Ruff check/format and MyPy clean.

**On the previous head's CI**: `Test Python 3.14` failed on `tests/unit/evolution/test_lineage_single_writer.py::test_expired_owner_fails_closed_without_implicit_takeover` — a 1-second `asyncio.wait_for` timeout, with 20,507 tests passing around it. It is a lineage-lock test with no path to this change. It passes standalone in 0.69s, and a local full run on Python 3.14 with the same 4-worker xdist configuration passed 20,130 tests with zero failures. All checks were green on `42f0633`.

Q00:
## Adapter-side polarity enforcement added — `c176f83547d352fcb8cd86b2cc34acb98cb915ae`

Thank you for the approval, but I am not treating this one as closed by the policy filter. The omitted finding — "an all-`VERIFIED` report carrying `agent_reported_pass=False` was promoted to formal PASS" — was outside the changed-files scope only because I had not yet touched `spec_verification_adapter.py`. It is the same defect #1835 reports, reachable by a different door, and your Merge Recommendation said so plainly. I reproduced it on the approved head:

```
ADAPTER BYPASS -> final_approved: True | verdict: PASS
```

**Fix**: the formal boundary now enforces the same polarity as the producer. An all-`VERIFIED` report whose `agent_reported_pass` is `False` yields `NOT_EVALUATED`, never `PASS`. Read through `getattr(report, "agent_reported_pass", True)` so compatibility objects without the field keep their existing meaning rather than failing construction.

Enforcing it in both places is deliberate. `verify_all` refuses to *produce* the combination; the adapter refuses to *honour* it. Replayed rows, legacy payloads and integration-built summaries never pass through the producer, so a producer-only rule leaves that door open — which is precisely what your probe walked through.

**Regressions** (`test_a_verified_report_over_an_agent_fail_cannot_mint_a_formal_pass`, parametrized over three construction paths):

| summary shape | before | after |
|---|---|---|
| direct construction | `final_approved=True` / PASS | `False` / NOT_EVALUATED |
| JSON round-trip (replay/rehydration) | `final_approved=True` / PASS | `False` / NOT_EVALUATED |
| legacy booleans, `outcome` field stripped | `final_approved=True` / PASS | `False` / NOT_EVALUATED |

Confirmation is separately pinned by `test_a_verified_report_still_confirms_an_agent_pass`: an all-`VERIFIED` report over an agent PASS still yields `final_approved=True` / PASS.

Local on this head: `tests/unit/test_verification.py` + all of `tests/unit/mcp/server/` = 685 passed, 1 skipped. Ruff check across `src/ouroboros` and both test files, format, and MyPy on both changed source files: clean. A full `tests/unit` + `tests/conformance` run with the CI's 4-worker xdist configuration is in progress and I will report its result here either way.

The PR now closes the FAIL-override path at both the producer and the authority boundary, with the prose classifier gone from both.

Q00:
## Blocker 1 addressed — `17d1123e85ba0d743fdf56744724e08a041de009`

Correct, and the root cause is exactly as you diagnosed: I checked the report's *copy* of the agent result while the adapter already derives the authoritative one from the mechanical summary at line 130. A copy carried by a replayed or externally built report is not evidence about what the agent reported.

**Fix**: `_agent_claimed_pass(report, expected_agent_results, ac_index)` consults both records, and either one saying "not a pass" settles it. Disagreement is deliberately not resolved toward the permissive record — two accounts of the same fact that do not match are not evidence that the agent claimed a pass.

Both probes you reported, plus the contradiction in the other direction:

| report polarity | execution record | before | after |
|---|---|---|---|
| `True` (lies) | FAIL | `final_approved=True` / PASS | `False` / NOT_EVALUATED |
| field omitted (defaults True) | FAIL | `final_approved=True` / PASS | `False` / NOT_EVALUATED |
| `False` | PASS (contradiction) | PASS | `False` / NOT_EVALUATED |
| `False` | FAIL | NOT_EVALUATED | NOT_EVALUATED |
| `True` | PASS | PASS | PASS |

Regressions: `test_report_polarity_cannot_outrank_the_execution_record`, parametrized over the first three rows; the confirmation row stays pinned by `test_a_verified_report_still_confirms_an_agent_pass`.

**One existing expectation tightened, flagged rather than buried.** `test_spec_verification_does_not_approve_failed_execution` builds a mechanical record whose worker reported its task incomplete (`completed=False`, `source_ac_index=0`) while the report carries `agent_reported_pass=True`. That is precisely the contradiction this change fails closed on, so the AC is now `NOT_EVALUATED` rather than a passing result carried inside an already-rejected run, and `failure_reason` gains the unverifiable-AC clause. The run-level assertions are unchanged. I took the stricter outcome as intended rather than adjusting the rule around the test: a grep result standing in for work a worker reported it never finished is the exact shape #1835 describes.

Local on this head: `tests/unit/test_verification.py` + all of `tests/unit/mcp/` = 2,354 passed, 1 skipped. Ruff check across `src/ouroboros` and `tests/unit`, format, and MyPy on both changed source files: clean. A full `tests/unit` + `tests/conformance` run with the CI's 4-worker xdist configuration is running and I will post its result.

**On the previous head's CI**: `Test Python 3.12` failed on `tests/unit/orchestrator/test_synapse_after_turn_integration.py::test_stalled_runtime_terminalizes_cross_process_signal` (`SessionSignalState.REJECTED` vs `QUEUED`), with 20,647 passing around it. It is a cross-process signal timing test with no path to this change; it passes standalone in 1.47s here. That is the second infrastructure flake this branch has hit on a different job each time — 3.14 on one head, 3.12 on the next, different tests, neither touching verification — while a local full run on the same 4-worker configuration passed 20,123 with zero failures.

Q00:
## Verification result for `17d1123e85ba0d743fdf56744724e08a041de009`

Reporting the full-suite run I said I would post either way.

**Local**, CI's exact 4-worker xdist configuration on Python 3.14:

```
tests/unit + tests/conformance -n 4 --dist worksteal
20,126 passed, 88 skipped, 0 failed  (4:15)
```

**GitHub Actions on this head: all checks green**, including Test Python 3.12, 3.13 and 3.14. The two earlier red jobs were unrelated infrastructure flakes on different jobs and different tests — `test_expired_owner_fails_closed_without_implicit_takeover` (lineage lock, 3.14) and `test_stalled_runtime_terminalizes_cross_process_signal` (cross-process signal, 3.12), each passing standalone in about a second. Neither has a path to the verification code, and both jobs are green here without any change to them.

Ready for re-review when convenient.

### State of the invariant

One rule, enforced at every boundary that could carry it:

> Source-scan evidence cannot overturn an agent-reported FAIL.

| boundary | enforcement |
|---|---|
| `SpecVerifier.verify_all` (producer) | never emits VERIFIED for an AC the agent reported FAIL |
| `evaluation_summary_from_spec_verification` (formal verdict) | refuses to honour that combination, so replayed/legacy/compatibility summaries cannot encode it |
| agent-result source at that boundary | read from the mechanical execution record, not the report's copy; either record saying "not a pass" settles it |

No pattern property, report field, or construction path restores the override. The prose-to-target classifier that the first two rounds kept rebuilding is deleted, not narrowed — production diff is `verifier.py` +49 lines plus one adapter branch and one helper.

Confirmation of an agent-reported PASS is unchanged throughout, which is the false-PASS check #1835 says this scanner exists for.

PR #1324, opened 2026-06-02
refactor(config): centralize Claude model pins into a single source of truth (align to 4.8)

shaun0927:
## Summary

Introduces a single source of truth for default Claude model pins and aligns the stale versions. The same default model strings were hand-copied across three layers (`config/models.py` Pydantic defaults, `config/loader.py` `get_*_model()` fallbacks, `cli/commands/setup.py` wizard tables), so bumping a model meant shotgun surgery — and in practice it never happened (Opus frozen at 4.6 since `03b31259f` on 2026-02-28). This PR centralizes the pins so a bump is a one-line edit.

## What changed

New module `src/ouroboros/config/_model_defaults.py` defines:

```python
DEFAULT_OPUS_MODEL = "claude-opus-4-8"
DEFAULT_SONNET_MODEL = "claude-sonnet-4-6"
DEFAULT_CONSENSUS_OPUS_MODEL = "openrouter/anthropic/claude-opus-4-8"
```

All hyphenated config-value literals in `models.py` / `loader.py` / `setup.py` now reference these constants. Net version moves:

- Opus phases (interview, seed, ontology, atomicity, decomposition, double-diamond, wonder, reflect, semantic eval): `claude-opus-4-6` → `claude-opus-4-8`
- Consensus advocate / Anthropic roster entry: `openrouter/anthropic/claude-opus-4-6` → `…-4-8`
- Sonnet tier (`qa_model`, `assertion_extraction_model`): aligned on `claude-sonnet-4-6`. This also retires `claude-sonnet-4-20250514` (Sonnet 4.0, EOL 2026-06-15) — the same fix as #1323, here via the constant.

Config tests now assert against the constants instead of literals, so a future bump won't break them. The `_default_model_for_backend` sentinel invariant (loader fallback value must equal the `models.py` default for codex/kiro/copilot/hermes remap to trigger) is now structurally guaranteed by sharing one constant.

### Verified model IDs

`claude-opus-4-8`, `claude-sonnet-4-6` ([models overview](https://platform.claude.com/docs/en/about-claude/models/overview)). From the 4.6 generation onward, dateless IDs are pinned snapshots, so reproducibility is preserved (pins intentionally avoid the `"default"` sentinel).

## Non-goals (unchanged here)

- Dotted display/prose strings (`claude-opus-4.6` in comments/labels) — not API IDs.
- `llm.context_compression_model = "gpt-4"` (separate concern).
- `copilot/model_discovery.py` and the dotted Copilot model names it maps to.
[...]

PR #2089, opened 2026-08-12
fix(config): normalize legacy shipped defaults on Claude backends

### LINKED ISSUE #2069: qa_model legacy-pin normalization doesn't cover Claude-resolved backends (gap in #1322's fix)
## What

`ouroboros_qa` fails with `MCPToolError: LLM call failed: ... (claude-sonnet-4-20250514) ... may not exist or you may not have access to it` when the resolved backend is Claude and `~/.ouroboros/config.yaml` has an explicit `llm.qa_model` set to the retired literal. Reproduced live in this session:

```
ouroboros_qa(artifact="def add(a,b): return a+b", quality_bar="...")
→ MCPToolError: LLM call failed: There's an issue with the selected model
  (claude-sonnet-4-20250514). It may not exist or you may not have access to it.
  Run --model to pick a different model.
```

## Why #1322's fix doesn't cover this

#1322 (closed via #1323 + #1324) bumped the shipped default and added `LEGACY_DEFAULT_MODELS` normalization. That normalization only fires for backends that cannot run Claude models:

`src/ouroboros/config/loader.py`, `_normalize_configured_model_for_backend()`:

```python
if resolved in _CODEX_LLM_BACKENDS and is_shipped_default:
    return _CODEX_DEFAULT_MODEL
if resolved in _KIRO_LLM_BACKENDS and is_shipped_default:
    return _KIRO_DEFAULT_MODEL
# ... same pattern for Copilot/Hermes/Pi/GJC/Antigravity/Grok/Zcode ...

return candidate   # <- Claude-resolved backend falls straight through here
```

`_model_defaults.py:53` already documents `claude-sonnet-4-20250514` as *"the pre-EOL QA default"* and lists it in `LEGACY_DEFAULT_MODELS`, but that mapping is only consulted by the branches above. When `resolved` is a Claude-capable backend, a persisted legacy literal in an explicit per-role field (`llm.qa_model`; the same code path also covers `dependency_analysis_model` / `ontology_analysis_model` / `assertion_extraction_model`, see `loader.py:1849-1864`) is returned verbatim.

`get_llm_model_for_role()` makes this worse for `qa` specifically: an explicitly-set legacy per-role field takes precedence over the newer consolidated stage-model fields "for backward compatibility" (`loader.py:1927-1938`), so the newer, correctly-normalized code path never even gets a chance to run.

## Suggested fix

Add a branch in `_normalize_configured_model_for_backend()` for Claude-capable backends: when `is_shipped_default` is true (i.e. the candidate is a *recognized* legacy shipped default, not an arbitrary user-pinned id), normalize to `default_model` instead of falling through to `return candidate`. That keeps genuinely custom/proxy Claude ids untouched (they wouldn't be in `LEGACY_DEFAULT_MODELS`) while fixing the retired-shipped-default case.

## Scope note

Reproduced for `qa_model` only. `dependency_analysis_model` / `ontology_analysis_model` / `assertion_extraction_model` share the exact same normalization function and the same "explicit legacy field wins" precedence rule, so they look structurally exposed to the same gap, but I didn't reproduce those individually; worth a quick check while fixing this.

Related: #1322

sumin220:
Closes #2069.

## What

A recognized shipped default — current or prior-release — now normalizes on every backend, not only Claude-incapable ones. `_normalize_configured_model_for_backend()` maps it through the existing `_default_model_for_backend()` helper, so Claude-incapable backends keep their `"default"` sentinel exactly as before, and Claude-capable backends take the current default pin instead of returning the retired literal verbatim. Never-shipped ids are absent from the recognized set and still fall through untouched, so genuinely custom/proxy Claude ids are preserved.

This is the branch suggested in the issue, implemented by collapsing the per-backend ladder into the helper that already encodes the same mapping — the nine `resolved in _X_LLM_BACKENDS and is_shipped_default` branches were `_default_model_for_backend` restricted to the shipped-default case, so the gap was exactly the fallthrough this removes.

## Scope note from the issue

The reproduced case was `llm.qa_model`. The regression test also pins `dependency_analysis_model`, `ontology_analysis_model`, and `assertion_extraction_model`, which share the same normalization function — all normalize now. One nuance surfaced while writing the test: `dependency_analysis_model` recognizes the retired Opus family through `extra_shipped_defaults` and its current pin is the Sonnet default, so a persisted `claude-opus-4-6` there resolves to `claude-sonnet-4-6`, consistent with what a fresh config would hold. The `get_llm_model_for_role("qa")` legacy-precedence path is asserted too, since the explicit legacy field resolves through the same getter.

## Tests

- New RED-first regressions: legacy shipped defaults normalize on `backend="claude"` and on the default-resolved backend; a never-shipped id (`my-proxy/claude-custom-build`) stays verbatim.
- `tests/unit/config/`: 546 passed.
- The #1324 sentinel behavior for Claude-incapable backends is covered by the existing parametrized test and is unchanged.
- ruff, ruff format, mypy, and the module-size check pass. A wider sweep over suites referencing the legacy literals shows a failure set byte-identical to origin/main on this machine (known local macOS drift; CI is authoritative).

### DIFF
--- a/src/ouroboros/config/loader.py
+++ b/src/ouroboros/config/loader.py
@@ -2030,34 +2030,23 @@ def _normalize_configured_model_for_backend(
     if not candidate:
         return _default_model_for_backend(default_model, backend=backend)
 
-    resolved = _resolve_llm_backend_for_models(backend)
     # Recognize the current shipped default AND prior-release shipped defaults
     # (#1324): a config persisted before a pin bump still holds the old literal,
-    # and for Claude-incapable backends it must normalize to the sentinel just
-    # like the current default would. Genuinely explicit, never-shipped ids are
-    # absent from this set and fall through to be preserved verbatim.
+    # and it must normalize exactly like the current default would. Genuinely
+    # explicit, never-shipped ids are absent from this set and are preserved
+    # verbatim.
     is_shipped_default = candidate in (
         *recognized_shipped_defaults(default_model),
         *extra_shipped_defaults,
     )
-    if resolved in _CODEX_LLM_BACKENDS and is_shipped_default:
-        return _CODEX_DEFAULT_MODEL
-    if resolved in _KIRO_LLM_BACKENDS and is_shipped_default:
-        return _KIRO_DEFAULT_MODEL
-    if resolved in _COPILOT_LLM_BACKENDS and is_shipped_default:
-        return _COPILOT_DEFAULT_MODEL
-    if resolved in _HERMES_LLM_BACKENDS and is_shipped_default:
-        return _HERMES_DEFAULT_MODEL
-    if resolved in _PI_LLM_BACKENDS and is_shipped_default:
-        return _PI_DEFAULT_MODEL
-    if resolved in _GJC_LLM_BACKENDS and is_shipped_default:
-        return _GJC_DEFAULT_MODEL
-    if resolved in _ANTIGRAVITY_LLM_BACKENDS and is_shipped_default:
-        return _ANTIGRAVITY_DEFAULT_MODEL
-    if resolved in _GROK_LLM_BACKENDS and is_shipped_default:
-        return _GROK_DEFAULT_MODEL
-    if resolved in _ZCODE_LLM_BACKENDS and is_shipped_default:
-        return _ZCODE_DEFAULT_MODEL
+    if is_shipped_default:
+        # A recognized shipped default — current or prior-release — is a pin
+        # the user never chose, so every backend maps it to its own default:
+        # Claude-incapable backends keep their sentinel as before, and
+        # Claude-capable backends now take the current default pin instead of
+        # leaking a retired id to the API (#2069). Never-shipped ids are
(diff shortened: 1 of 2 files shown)


src/ouroboros/config/loader.py
```diff
@@ -2030,34 +2030,23 @@ def _normalize_configured_model_for_backend(
...
-    if resolved in _ZCODE_LLM_BACKENDS and is_shipped_default:
-        return _ZCODE_DEFAULT_MODEL
+    if is_shipped_default:
+        # A recognized shipped default — current or prior-release — is a pin
+        # the user never chose, so every backend maps it to its own default:
+        # Claude-incapable backends keep their sentinel as before, and
+        # Claude-capable backends now take the current default pin instead of
+        # leaking a retired id to the API (#2069). Never-shipped ids are
+        # deliberate user pins and fall through verbatim.
+        return _default_model_for_backend(default_model, backend=backend)
```
Q00, line 2049: thank you for refactor this

Q00 (approved):

PR #2090, opened 2026-08-12
fix: deliver the update nudge to plugin users

### LINKED ISSUE #2066: uvx pins the plugin server to the install-day version, and the update notice never reaches the user
Three mechanisms interact so that a plugin user who installed once can stay on that version indefinitely without ever being told an update exists. Each claim below is traced to source or official docs; two spots I could not fully verify are marked.

## 1. The plugin MCP server is effectively pinned to the install-day version

`.claude-plugin/.mcp.json` launches the server as:

```
uvx --isolated --python ">=3.12" --from "ouroboros-ai[mcp]" ouroboros mcp serve ...
```

There is no version specifier. Per the uv docs (concepts/tools): "uvx uses the latest available tool version on its first invocation, then a cached version for subsequent runs. To request the latest version and refresh the cache, use the `@latest` suffix."

So the first launch resolves whatever is newest that day, and every later launch reuses that cached environment. A user who installed the plugin at 0.51.0 keeps serving 0.51.0.

As far as I can trace, `ooo update` does not clear this either: `update.py` upgrades the receipt-bound package and refreshes the plugin marketplace, but the refreshed descriptor is still unpinned, so the uvx cache key never changes, and I found no `uv cache` invalidation in `update.py`. (Not verified end-to-end on a live machine; please correct me if the plugin refresh path does invalidate it.)

## 2. The existing update notice is sent where users cannot see it

`scripts/session-start.py` deliberately routes the update message to stderr:

```python
# SessionStart stdout is consumed as Claude context, so keep the
# success path silent and send notices to stderr instead.
print(result["message"], file=sys.stderr)
```

Per the Claude Code hooks documentation (code.claude.com/docs/en/hooks), for a hook that exits 0, stderr goes to the debug log only; it is not shown in the normal UI. Stdout on exit 0 is added to Claude's context, which is the one documented channel that can reach the user: the agent sees the line and relays it conversationally. The comment treats that as pollution, but one line per 24h (the check is already cached for 24h) is exactly the nudge we want delivered.

## 3. The check only exists on the Claude Code hook path

`update_available` is produced only by `scripts/version-check.py` and consumed only by `scripts/session-start.py` (plus tests). The MCP serve path exposes nothing: hosts like Codex, OpenCode, Kiro and every other MCP runtime start the server and get no version information at all. Those users have no in-product way to learn an update exists.

## Proposals (independent, any subset works)

1. **Pin the version in `.claude-plugin/.mcp.json`** (`--from "ouroboros-ai[mcp]==X.Y.Z"`) and let `scripts/sync-plugin-version.py` bump it alongside `plugin.json`. A plugin update then deterministically changes the uvx cache key, so plugin version and server version move together. Startup stays offline-safe. (`@latest` would also work but adds a network resolution to every server start; slower and brittle offline, so I would not make it the default.)
2. **Route the session-start notice to stdout with exit 0** so the agent relays it. The message can stay short, e.g. `Ouroboros 0.52.0 is available (installed: 0.51.2). Suggest running "ooo update".`
3. **Surface the cached check at MCP serve time**: at `mcp serve` startup, read the same 24h cache and, when an update exists, append one line to the server's initialize instructions. That covers every MCP host, not just Claude Code, with no extra network calls on the hot path.

Proposal 1 is the smallest fix and the one that actually closes the bug: it makes the plugin version and the served version move together, and it doesn't depend on 2 or 3 to be useful. I'd start there and send that PR first, unless there's a reason to sequence differently — is there?

sumin220:
Closes #2066. All three proposals, each independently revertible.

## 1. Pin the plugin MCP server requirement

`.claude-plugin/.mcp.json` now launches with `--from "ouroboros-ai[mcp]==0.51.2"`, and `scripts/sync-plugin-version.py` gained the descriptor as a sync target: the pin is validated in the same preflight pass as the JSON version targets (exactly one `--from ouroboros-ai[mcp]` requirement, recognized spec shape — anything ambiguous fails before any file mutates), written through the same atomic-write machinery, and covered by the same rollback and post-write verification. An unpinned spec is accepted on read so this first sync introduces the pin; after that, every release-prep `--write` moves it with `plugin.json`, so a plugin update deterministically changes the uvx cache key. The `--require-canonical` validation gate in release.yml passes unchanged since the release-prep sync bumps the pin in the same commit that bumps `plugin.json`. Startup stays offline: no `@latest`, no network resolution.

## 2. Route the SessionStart notice to stdout

`scripts/session-start.py` prints the update message to stdout on exit 0 — the documented channel that enters Claude context so the agent relays it — instead of stderr, which on exit 0 goes to the debug log only. Failure paths stay on stderr. The 24h cache in the checker already bounds this to one line per day.

## 3. Surface the cached check at MCP serve time

New module `src/ouroboros/mcp/update_notice.py` reads the same `~/.ouroboros/version-check-cache.json` the hook path writes — and only that cache, never the network — honors the writer's 24h TTL (a stale or future timestamp yields nothing), matches the writer's stable/prerelease channel split, and compares with `packaging.version`. When a newer version is cached, `mcp serve` appends one line to the server instructions, so Codex, OpenCode, Kiro, and every other MCP host gets the nudge. The line is appended at the end, after the load-bearing rules, so a host that truncates the instructions budget cuts the nudge first. A missing, stale, or malformed cache and an uninstalled or unparseable package all yield no notice — the nudge can never delay or fail server startup.

## Tests

- Sync script: pin introduction (`unpinned -> X` write), already-current OK, dry-run drift, and ambiguous-descriptor preflight rejection with no mutation — 475 script tests pass.
- SessionStart: the stdout routing is pinned (silent no-update and stderr failure paths unchanged).
- `tests/unit/mcp/test_update_notice.py`: fresh/missing/stale/future cache, equal and older latest, malformed JSON shapes, prerelease channel, uninstalled package, unparseable installed version, and the append seam — 15 cases.
- `tests/unit/mcp/server/test_adapter.py`: 238 pass with the wiring in place; ruff, ruff format, mypy, and the module-size check pass (`adapter.py` stays within its budget).

## R-run comparison

|---|---|---|---|

Budget compliance: [x] N/A (release metadata/update-notice/cache path; no per-round execution work)


sumin220:
Addressed the first review in 97f3c1d5, all three blockers and the warning.

On the first: `.mcp.codex.json` — the descriptor `.codex-plugin/plugin.json` declares — is pinned to 0.51.2 and joins the same sync transaction as the Claude descriptor: same preflight validation, same atomic write, same rollback and post-write verification, reported per descriptor. The sync now treats shipped plugin descriptors as a list, so both plugin surfaces move their uvx cache identity with their manifest version. A dry run against the repository reports both descriptors OK at 0.51.2.

On the second: `append_cached_update_notice()` is now the complete advisory boundary — the whole read-and-append path sits behind a fail-open guard, so an `OSError` from corrupt dist-info, a metadata backend failure, or anything else unexpected leaves the instructions unchanged instead of propagating into server construction. The probe is reproduced as a regression at both levels: the seam returns the instructions untouched under raising `metadata.version`, and `create_ouroboros_server()` constructs successfully with `metadata.version` raising `OSError("corrupt dist-info")`.

On the third: standalone hosts now have a bounded producer. When a serve start finds the cache missing or stale, the seam schedules one daemon-thread refresh per process, off the startup hot path: it mirrors `scripts/version-check.py` — same PyPI endpoint and timeout, same stable/prerelease channel split, same cache keys, and the same atomic tempfile-then-replace so a concurrent hook-side write cannot be torn. The first start on a hookless host primes the cache and a later start delivers the nudge; every failure inside the refresh is swallowed. The module docstring states this contract rather than the previous overreach.

On the warning: the launcher contract test now encodes the pin policy explicitly — the two shipped plugin descriptors must pin to their own plugin manifest version (read from the manifests, so releases keep the test green), while the repository-root development launchers deliberately stay unpinned — with the isolated-MCP argument shape asserted for all four launchers as before.

The touched suites (adapter 239, scripts 476, dependencies contract, update notice 25) pass, along with ruff, ruff format, mypy, and the module-size check.

sumin220:
Addressed the second review in a30a45c6 — the blocker and both warnings.

On the blocker: refresh eligibility is now channel-specific. The gate resolves the installed channel first and treats a payload as stale unless the timestamp is fresh AND that channel holds a parseable version string. The probe's exact shape — a fresh stable-only payload under an installed 0.52.0b1 — now schedules a refresh, and regressions cover both transitions (stable-only cache with a prerelease install, prerelease-only cache with a stable install) plus an unparseable cached value.

On the first warning: the producer detaches from the composition root entirely. `append_cached_update_notice()` is pure formatting again — a regression asserts it schedules nothing even with a stale cache — and the refresh is initiated only by the new `maybe_schedule_cache_refresh()`, called from the `mcp serve` startup flow right after server construction. `ooo auto`, detached job workers, builtin runtime interception, the in-process dispatcher, and the `mcp info` command construct servers without any network activity.

On the second warning: the once gate re-checks the event under a lock before setting it, so concurrent callers cannot both observe it unset. The barrier-backed probe is reproduced as a regression: 20 synchronized callers start exactly one refresh.

The update-notice suite (26, including the construction-boundary regression), the mcp serve CLI suites (startup cleanup, shutdown drain, transport validation, nested guard — 71), the adapter suite (239), ruff, ruff format, mypy, and the module-size check all pass.

sumin220:
Addressed the third review in 878f2c4c.

Freshness is now attested per channel with its own metadata. Every writer stamps the channel it refreshed: the serve-path producer writes `<channel>_checked_at` and deliberately leaves the shared legacy `timestamp` untouched, so its write cannot make the other channel's older value look current to any reader; the hook-path script writes the stamp alongside its legacy timestamp bump, which stays for readers that predate the stamps. On the read side — the notice reader, the refresh gate, and the script's own cache check — a channel is trusted only through its own stamp. Once any stamp exists in the payload, an unstamped channel is stale; a payload with no stamps at all was written by the previous single-timestamp writer and keeps the shared-timestamp semantics for at most one TTL window, after which it expires either way.

The requested regression is pinned in both directions of the probe: with a stamped fresh `latest_version` and a leftover `latest_version_pre` under a fresh shared timestamp, a prerelease install gets no notice from the stale value and the refresh gate schedules its own channel's refresh; symmetrically, a per-channel stamp attests its own fresh value even when the legacy timestamp is old. The script side has the same regression — a stamped stable refresh does not let the prerelease channel serve its unstamped leftover, and the refetch stamps the prerelease channel. The producer test now pins that a refresh stamps only its own channel and leaves both the other channel's value and the shared timestamp byte-identical.

The update-notice suite (30), version-check and session-start script suites, the full scripts scope plus the adapter suite (701), ruff, ruff format, mypy, and the module-size check all pass.

sumin220:
Addressed the fourth review in c3a8ff54.

The doctor's canonical-launcher validator now accepts the shipped pin without loosening anything else. A release-shaped `ouroboros-ai[mcp]==X.Y.Z` requirement is normalized onto the canonical unpinned spec before the exact-arguments comparison, so the shipped descriptors pass through every other canonicality check — isolation flag, Python floor, command tail, env contract — completely unchanged. Rejection of arbitrary requirements is retained and pinned by new validator regressions: version ranges, a missing mcp extra, other packages, and non-release pins (`.dev` and `+local` versions, which the sync never writes because plugin versions are normalized) all still fail. The near-miss diagnostic message now names both accepted forms, and the existing noncanonical parametrizations move to a range-shaped case since the exact release pin is no longer a violation.

The canonical artifact assertion in test_skill_artifacts reads the expected pin from the Codex plugin manifest version rather than a literal, matching the dependency-contract test on the Claude side, so releases keep both green as the sync moves the pins.

The reproduced probe passes: `test_shipped_mcp_codex_json_uses_doctor_accepted_launcher` reports no failures for the shipped `.mcp.codex.json`. The codex command doctor scope (206), skill artifacts, dependency contract, and sync suites pass; the only local failures are the environment-dependent refresh/path-poison cases that fail identically on the branch base without these changes. ruff, ruff format, mypy, and the module-size check pass.

Q00:
Addressed the latest exact-head review findings in ebe7e3879. The SessionStart writer now replaces valid non-object JSON cache payloads with a fresh object, and shipped MCP descriptor preflight rejects every additional  option before mutation. Added focused regressions; 106 script tests and Ruff pass. Requesting exact-head re-review.

Q00:
@ouroboros-agent please re-review exact head 30efc75bf and emit a fresh exact-head final_review.v1 verdict.

Q00:
@ouroboros-agent please re-review exact head 57c733d3d and emit a fresh exact-head final_review.v1 verdict. The accepted exact-pin regression now derives from the authoritative manifest, the shipped descriptor matches the current manifest version, and all 218 Codex command tests pass.

Q00:
Synchronized `.claude-plugin/.mcp.json` to the authoritative 0.51.5 plugin manifest at the latest HEAD. The sync dry run and launcher-contract regression now have matching Claude/Codex pins; please review current HEAD.

PR #2103, opened 2026-08-13
fix(auto): retry transient Seed QA errors and persist the error detail

### LINKED ISSUE #2094: ooo auto: Seed QA 'transient evaluator error' has no retry despite the retry loop, and root cause is unconfirmed
## Found via

Marketing session, while trying to record a full `ooo auto` interview-to-run demo. 3 consecutive attempts (2 fresh sessions + 1 resume) all blocked at the same point with "Seed QA reported a transient evaluator error". Not a marketing-scope issue; filing for whoever owns `auto/pipeline.py` and `mcp/tools/qa.py`.

## Bug 1 (confirmed by reading the code): the "transient" error never actually retries

`src/ouroboros/auto/pipeline.py:2798-2851` loops up to `max_repair_rounds` attempts, but on `qa_result.error` it calls `state.mark_blocked(...)` and returns immediately, inside the loop, without ever looping back:

```python
for attempt in range(1, max_attempts + 1):
    ...
    if qa_result.error:
        state.mark_blocked("Seed QA reported a transient evaluator error", tool_name="seed_qa")
        self._save(state)
        return (...)  # <-- exits on first occurrence, loop is dead code for this branch
```

If the error really is transient, this should retry (up to `max_attempts`) before giving up. Right now the label says "transient" but the behavior is "blocks on first occurrence, no retry, and the resume path preserves the original deadline so a resume shortly after usually just re-blocks or times out."

## Bug 2 (unconfirmed): what actually sets `qa_result.error` for the `claude` backend

Traced `HandlerSeedQAEvaluator.__call__` (`src/ouroboros/auto/adapters.py:830`) -> `QAHandler.handle()` (`src/ouroboros/mcp/tools/qa.py:493`). Two paths set `.error`:

1. `should_dispatch_via_plugin(agent_runtime_backend, opencode_mode)` returning True -> `DELEGATED_TO_SUBAGENT`, which the adapter treats as an error ("the auto pipeline cannot grade Seed QA via out-of-band subagent dispatch"). **Ruled this out**: traced `resolve_subagent_dispatch` -> `_supports_native_parallel_subagent_surface` (`src/ouroboros/backends/capabilities.py:842-852`) and confirmed the `claude` backend capability entry (line ~446) does not set `supports_native_parallel_subagents` (defaults to `False`), only `supports_host_driven_subagents=True`. So `claude` resolves to `HOST_DRIVEN`, not `PLUGIN_PASSIVE`. This path should not trigger for the sessions I ran (`agent_runtime_backend=claude`, confirmed via the session state files' `runtime_backend` field).
2. The in-process LLM call itself failing (`qa.py:594` `llm_result.is_err`) or the response failing to parse (`qa.py:604` `parse_result.is_err`). **Not confirmed which** -- neither the CLI output nor the persisted session state JSON (`~/.ouroboros/data/auto_<id>.json`) retains the underlying `MCPToolError` detail, only the final "reported a transient evaluator error" string. Whatever structlog event fires around here would have the detail; I didn't have log access beyond the terminal capture.

## Possibly related: issue #781 (closed)

`qa.py`'s LLM call uses `max_turns=1` with `allowed_tools=[] if backend_supports_tool_envelope(...) else None` (`qa.py:582-585`). The comment above it references a known SDK failure mode: a tool-use block from the model burns the single allowed turn and the SDK raises "Reached maximum number of turns (1)" before a final text response streams (cited as issue #781, which listed `qa.py:578` as one of 9 unguarded `max_turns=1` sites and is now closed). The conditional `allowed_tools` looks like it's already the guard from that sweep, so this may or may not still be live -- didn't verify closed-issue resolution against current `qa.py` line-by-line.

## Repro

Two session IDs with this exact blocker, state files still on disk if useful: `auto_c66183949a25`, `auto_aea43f994dc8` (`~/.ouroboros/data/auto_<id>.json`, `last_error` field). Task used both times: `ooo auto "a tiny CLI that reverses a string"`.

## Suggested next step

Whoever picks this up should reproduce with structlog at debug level around `qa.py`'s LLM call, or add error detail to the persisted `last_error` field instead of the generic "transient evaluator error" string -- right now the persisted state actively discards the information needed to diagnose it.

## Scope note

Filed by the marketing session after repeated demo-recording attempts hit this. Not something a docs/marketing change can fix; routing to engineering.

sumin220:
Closes #2094. Implements the suggested next step for bug 1 and the persistence half of bug 2.

## What

The `qa_result.error` branch in `_run_seed_qa_gate` no longer blocks on first occurrence. A transient evaluator error now retries within the existing attempt budget (`max_repair_rounds`, the same budget the repair path uses): each non-final attempt records a progress entry carrying the concrete error and loops back, and only the final attempt blocks. The block message persists the collapsed evaluator error — single-line, capped at 200 characters — as `Seed QA evaluator error after N attempt(s): <detail>`, so `last_error` in the session state stops discarding exactly the information needed to diagnose the underlying failure. The retry-progress entries give the same detail per attempt, which covers the diagnosis gap from the issue's bug 2 without requiring live structlog access.

On the confidentiality note: the gate's existing `_safe_seed_qa_*` helpers keep withholding QA feedback content (differences/suggestions) from durable state as before — the persisted error string is the evaluator's transport/parse diagnostic (`MCPToolError`, LLM failure, parse error), not seed-grading content.

## Budget

`pipeline.py` sits at its exact grandfathered budget, so the fix pays for itself: the gate's six duplicated blocked-return tuples fold into one `_blocked_result()` closure inside the same function (the closure reads the loop's current seed/review bindings, so behavior is unchanged), leaving the file 13 lines under budget after adding the retry branch.

## Tests

Two RED-first regressions with stub evaluators in the existing gate-test style: an evaluator that errors twice then passes completes the run with exactly `max_attempts` calls, and an evaluator that always errors blocks after exactly the budget with the multi-line error collapsed into the blocker (asserting both the detail and the attempt count). The full interview-pipeline suite (178) and the auto scope (1,978) pass, along with ruff, ruff format, mypy, and the module-size check.

### DIFF
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -2803,7 +2803,7 @@ class AutoPipeline:
         current_seed = seed
         current_review = review
         max_attempts = max(1, int(state.max_repair_rounds or 1))
-        consumed_error_attempts = max(0, int(state.seed_qa_error_attempts or 0))
+        consumed_calls = max(0, int(state.seed_qa_evaluator_calls or 0))
 
         def _blocked_result() -> tuple[AutoPipelineResult, Seed, SeedReview | None]:
             return (
@@ -2812,20 +2812,20 @@ class AutoPipeline:
                 current_review,
             )
 
-        # The durable error budget bounds evaluator calls across restarts
-        # (#2094): an exhausted session re-blocks without another call.
-        if consumed_error_attempts >= max_attempts:
-            state.mark_blocked(
-                f"Seed QA evaluator error budget exhausted ({max_attempts} attempt(s))",
-                tool_name="seed_qa",
-            )
+        # The durable call budget bounds EVERY evaluator invocation across
+        # restarts (#2094): exhausted sessions re-block with no further call.
+        if consumed_calls >= max_attempts:
+            state.mark_blocked(state.seed_qa_exhausted_blocker(max_attempts), tool_name="seed_qa")
             self._save(state)
             return _blocked_result()
 
-        for attempt in range(1, max_attempts + 1):
+        for attempt in range(consumed_calls + 1, max_attempts + 1):
             timeout = self._deadline_capped_timeout(
                 state, state.phase_timeout_seconds(AutoPhase.EVALUATE)
             )
+            # Count the call fail-closed BEFORE making it (#2094).
+            state.seed_qa_evaluator_calls = consumed_calls = attempt
+            self._save(state)
             try:
                 qa_result = await asyncio.wait_for(
                     self.seed_qa_evaluator(current_seed, ledger), timeout=timeout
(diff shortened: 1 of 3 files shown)


sumin220:
Addressed the first review in e1f85302.

The retry budget is now durable. `AutoPipelineState` persists a `seed_qa_error_attempts` counter (serialized with a backward-compatible default, validated as a non-negative integer), and the gate restores it on entry. Three consequences, each pinned by a checkpoint/reload regression that round-trips the state through `to_dict()`/`from_dict()`:

- A session interrupted mid-budget resumes with the remainder: with the counter persisted at 1 of 2, the resumed run makes exactly one more evaluator call and blocks with the full attempt count in the message. The counter is incremented and saved BEFORE the retry continues, so a crash between attempts cannot lose a consumed attempt.
- An exhausted session re-blocks at gate entry without another evaluator call: after the budget is spent, a reloaded state's run makes zero additional calls, so total evaluator calls in the error family never exceed `max_repair_rounds` no matter how many times the recoverable `seed_qa` block is resumed.
- A usable evaluator result restores the full budget, so the durable counter cannot starve a later legitimate gate entry after a genuinely transient outage resolves.

The auto and persistence scopes (2,489) pass, and `pipeline.py` sits exactly at its grandfathered budget after the addition. ruff, ruff format, mypy, and the module-size check pass.

sumin220:
Addressed the second review in 55da6520, both blockers.

On the first: `max_repair_rounds` is now one durable evaluator-call budget covering every invocation. The counter renames to `seed_qa_evaluator_calls` and is incremented and saved fail-closed BEFORE each call, the loop starts from the restored count, and the error-retry and repair branches both gate on the same attempt number. The review's exact scenario is pinned as a regression: a reloaded session with one consumed call and a budget of two makes exactly one more evaluator call, and a normal failed QA result on that final call blocks with the repair-exhaustion details instead of spending a third call on repair. Counting before the call also means a crash mid-call cannot refund the attempt.

On the second: the final blocker message now persists in a dedicated `seed_qa_block_detail` field, because the resume transition clears `last_error` — which is why preserving it in place could not work. An exhausted-budget re-entry replays that stored diagnostic verbatim: the reload regression now asserts the collapsed `MCPToolError` detail and the attempt count survive into the replayed blocker. The same replay covers repair exhaustion (the joined differences/suggestions summary), and a concluding pass clears both the counter and the stored detail so a later gate entry starts clean.

The auto and persistence scopes (2,490) pass, `pipeline.py` sits exactly at its grandfathered budget, and ruff, ruff format, mypy, and the module-size check pass.

Q00:
Superseded by #2144. Current main has adopted the newer non-blocking Seed QA advisory contract, so this older retry-and-block implementation conflicts semantically rather than being a conflict-only merge. #2144 carries the bounded evaluator diagnostic forward without reverting the advisory behavior.

Q00:
Closing as superseded by #2144.

PR #2104, opened 2026-08-13
feat(config): persistent default execution policy for fresh starts

### LINKED ISSUE #1733: [Feature] Allow a persistent default execution policy for fresh run and auto starts
## Problem

Fresh ooo run and ooo auto starts are blocked when there is no remembered execution-policy choice. The host asks the user to choose between quality-first and efficient execution before every fresh start.

This is useful as a safe first-run interaction, but it becomes repetitive for users who have a stable preference or invoke Ouroboros frequently through automation. The execution engine already resolves omitted arguments to adaptive with observe, while the host contract intentionally prompts before making the call. There is currently no persistent config field that lets a user opt out of the repeated prompt.

## Why now

Execution policy is now an explicit immutable run contract. Users should be able to establish that contract once as a durable preference while retaining the current safe prompt as the default behavior.

## User / persona

Frequent Ouroboros users and automation operators who consistently choose the same execution policy across fresh runs.

## Current behavior

- A fresh run or auto start with no remembered choice pauses before dispatch.
- The user must choose quality-first or efficient execution.
- The choice is persisted for the resulting session and cannot be changed on resume.
- Engine-level omission resolves to adaptive plus observe, but the host prompt prevents relying on that default through the standard conversational workflow.
- ExecutionConfig exposes no persistent efficiency-mode or frugality-assurance preference.

## Desired behavior

Offer an optional persistent default for fresh executions. When configured, the host should apply it without asking on every run. When not configured, preserve the current interactive prompt exactly.

The resolved policy must still be shown in the execution-start handoff.

## Proposed solution

Add an explicit configuration surface with an ask-preserving default, for example:

    execution:
      default_policy: ask

Supported values:

- ask: current behavior
- efficient: efficiency_mode=adaptive and frugality_assurance=observe
- quality_first: efficiency_mode=quality_first and frugality_assurance=off

Alternatively, expose the two underlying fields if that better matches the configuration model.

Suggested precedence:

1. Explicit MCP or CLI invocation arguments
2. Persistent user configuration
3. Interactive prompt when configured as ask or unset
4. Existing engine fallback only for direct callers that omit all policy input

Expose the setting through config show, config set, the TUI, and the web settings UI.

## Constraints

- Default behavior must remain ask for backward compatibility and user-cost awareness.
- strict frugality assurance must remain explicit opt-in and must never become an implicit coupled default.
- A resumed session must continue using its persisted immutable execution contract.
- Changing the configured default must affect only future fresh executions.
- Start output and event metadata must continue reporting the resolved efficiency mode and frugality assurance.
- The feature should apply consistently to both ooo run and ooo auto.

## Non-goals

- Changing routing or escalation semantics.
- Mutating an active or historical session policy.
- Making strict assurance an implicit default.
- Removing the interactive policy choice for users who prefer it.

## Alternatives considered

- Rely on the engine default adaptive plus observe: this does not bypass the host-level prompt.
- Add a permanent personal instruction outside Ouroboros: workable, but not discoverable, validated, portable, or represented in config show.
- Patch the installed run skill: overwritten by upgrades and not an appropriate user configuration surface.

## Acceptance criteria

- [ ] With no new setting or default_policy=ask, fresh run and auto starts retain the current policy prompt.
- [ ] With default_policy=efficient, a fresh start dispatches with adaptive plus observe without asking.
- [ ] With default_policy=quality_first, a fresh start dispatches with quality_first plus off without asking.
- [ ] Explicit invocation arguments override the persistent default.
- [ ] Resume restores the original session contract and rejects attempted preference changes.
- [ ] strict assurance is never selected implicitly.
- [ ] config show, config set, TUI, and web UI expose and validate the setting.
- [ ] Start responses and events report the resolved policy.
- [ ] Unit and integration tests cover precedence, backward compatibility, fresh starts, and resume immutability.

## References

Observed on ouroboros-ai v0.50.5.

Relevant current implementation areas:

- ouroboros.core.execution_preferences.resolve_execution_preferences
- ouroboros.mcp.tools.execution_handlers._resolve_execution_preferences_request
- ouroboros.skills.run host instructions
- ouroboros.codex.ouroboros Active Conductor Host Contract

## Checklist

- [x] I searched existing issues and discussions first.
- [x] I explained the problem, not just the solution.
- [x] I included clear scope boundaries and non-goals.
- [x] I listed concrete acceptance criteria.

sumin220:
Closes #1733.

## What

Adds the proposed `execution.default_policy` configuration surface with the ask-preserving default:

```yaml
execution:
  default_policy: ask   # or: efficient | quality_first
```

`ask` (the default) preserves the current interactive prompt exactly. `efficient` resolves omitted preference arguments to `adaptive`/`observe`, and `quality_first` to `quality_first`/`off` — implemented by mapping the policy to a fresh-start efficiency mode and letting the existing documented coupling supply the paired frugality default, so `strict` structurally cannot appear implicitly (the config value never touches assurance).

## Precedence, as suggested in the issue

1. Explicit MCP/CLI arguments win: the policy fills the gap only when `efficiency_mode` is omitted, and an explicitly supplied `frugality_assurance` composes with the configured mode while keeping its explicitness (so a deliberate `strict` still works with a configured `efficient`).
2. The persistent configuration is consulted at both fresh-start resolution seams — the run handlers' shared preference resolver (where the resolved value also flows into the runner construction) and both Auto state builders — and never on resume: the resolver's resume guard sits before the config read, pinned by a regression asserting the config is not even consulted.
3. When the policy is `ask` or unset (or the config is unreadable), behavior is byte-identical to today, including the engine fallback for direct callers.

The resolved policy still appears in the execution-start handoff and Auto briefing, which read the resolved preferences.

## Exposure

`config show` and `config set execution.default_policy quality_first` work through the existing schema-validated generic paths (the key validates against the pydantic schema; invalid values are rejected at load with a field-level error). The run and auto skill contracts — canonical and bundled plugin copies, which are intentionally worded differently and were edited in place rather than overwritten — now instruct hosts to skip the prompt and omit both arguments when a policy is configured. Tool parameter descriptions document the fallback order, and `docs/config-reference.md` gains the field row. The repository currently has no TUI or web settings editor surface, so there is nothing further to wire there; if one lands, the field is an ordinary `ExecutionConfig` member.

## Tests

Model validation (default, both alternatives, rejection of `strict`/raw modes), loader mapping (all three policies plus unreadable-config fail-open to ask), seven resolver-level regressions (fill, both mappings, explicit-argument precedence, explicit-assurance composition, ask passthrough, resume-never-consults, resume-still-rejects-arguments), and two Auto handler regressions (configured policy reaches a fresh start's persisted state and meta; explicit argument beats it). Affected suites (config 344, start-auto and resolver 71, definitions, skill artifacts) pass — the only local failures are the two known environment-dependent path-poison cases that fail identically on origin/main. ruff, ruff format, mypy, and the module-size check pass (`auto_handler.py` stays within its grandfathered budget).

### DIFF
--- a/.claude-plugin/skills/auto/SKILL.md
+++ b/.claude-plugin/skills/auto/SKILL.md
@@ -62,7 +62,11 @@ When the user types `ooo auto` with CLI-style flags inside chat, translate to MC
 
 `--pipeline-timeout-seconds` is accepted only when starting a session. Passing it with `--resume` is rejected because the original deadline is preserved across process restarts.
 
-Before a fresh start with no user choice, ask in outcome language: **Efficient
+Before a fresh start with no user choice, first check the persistent default:
+when `execution.default_policy` in `~/.ouroboros/config.yaml` is `efficient` or
+`quality_first`, do not ask — omit both arguments and the server applies the
+configured default (the handoff still reports the resolved policy). Otherwise
+ask in outcome language: **Efficient
 execution** maps to `adaptive/observe`; **Quality-first execution** maps to
 `quality_first/off`. `strict` assurance is separate explicit consent because it
 may spend extra work on proof. Do not send these arguments on resume; the server
--- a/.claude-plugin/skills/run/SKILL.md
+++ b/.claude-plugin/skills/run/SKILL.md
@@ -72,7 +72,11 @@ fallback instead of retrying the failing call.
    - If inline YAML: Use directly
    - If neither: Check conversation history for a recently generated seed
 
-   Before a fresh start, when no efficiency choice is already known, ask in
+   Before a fresh start, when no efficiency choice is already known, first
+   check the persistent default: when `execution.default_policy` in
+   `~/.ouroboros/config.yaml` is `efficient` or `quality_first`, do not ask —
+   omit both arguments and the server applies the configured default (the
+   start handoff still reports the resolved policy). Otherwise ask in
    user-outcome language: **Efficient execution** maps to
    `efficiency_mode="adaptive"` plus `frugality_assurance="observe"`;
    **Quality-first execution** maps to `efficiency_mode="quality_first"` plus
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -374,6 +374,7 @@ execution:
 | `auto_evolve_max_generations` | `int` | `3` | Maximum generations for automatically chained Ralph work. Values are clamped to Ralph's supported `1..10` range. |
 | `default_model` | `string \| null` | `null` | Optional Execute-stage model pin. `null`, an empty value, `"default"`, or `"current"` means Ouroboros does not pass a concrete `--model`; the selected runtime keeps its own current/default model. `OUROBOROS_EXECUTION_MODEL` has highest precedence, and a present empty env var explicitly clears the saved pin for that process. |
 | `project_guidance` | `list[string]` | `[]` | Guidance IDs loaded from `<project-root>/.ouroboros/guidance/<id>/GUIDANCE.md` and appended to execution system prompts. This option is config-only and has no environment-variable override. |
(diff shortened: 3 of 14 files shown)


Q00 (approved):
@sumin220 oh this is really good feature. thank you for this !

Issue #2117, opened 2026-08-14
auto: Seed QA gate dead-ends the pipeline instead of deferring to execution evidence


Q00:
## Problem

The pre-run Seed QA gate in `AutoPipeline._run_seed_qa_gate` ends every non-passing path in `mark_blocked`:

| path | trigger |
|---|---|
| `repair_budget_exhausted` | repair budget spent, verdict still failing |
| `seed_qa_feedback_unmapped` | QA feedback the repair mapper cannot map — **blocks on attempt 1** |
| evaluator timeout | `asyncio.wait_for` trip |
| evaluator error / `qa_result.error` | evaluator-side failure |

The gate is **optional**: when no `seed_qa_evaluator` is wired, auto runs the Seed unconditionally. So a session that turns QA *on* can end worse than one that leaves it off — extra information makes the pipeline strictly less able to proceed.

`blocked` is also not a real decision point here. The operator receives exactly the same unmapped prose the repair mapper could not act on, with no additional means of acting on it.

## Expected

An unresolved Seed QA verdict should be advisory: repair what is mappable, record the unresolved findings, and let `run → evaluate` judge the Seed against execution evidence instead of a static read.

The deterministic grade gate must keep its authority — a repaired Seed that no longer meets `required_grade` is still blocked by `grade_gate`.

Q00:
Resolved on `main` by #2119 (`f10122c63`). The optional Seed QA gate now repairs bounded mappable findings, publishes durable advisory evidence for unmapped feedback, timeouts, evaluator errors, and exhausted repair budgets, and continues to `run → evaluate`. The deterministic grade gate and pipeline deadline remain authoritative.

Current focused regression proof: `tests/unit/auto/test_interview_pipeline.py` Seed-QA advisory subset — 3 passed.

PR #2119, opened 2026-08-14
fix(auto): make the Seed QA gate advisory instead of a dead end

Q00:
## Summary

The pre-run Seed QA gate ended every non-passing path in `mark_blocked`. This makes it advisory: it repairs what it can, records the unresolved verdict, and lets `run → evaluate` decide against execution evidence.

The gate is optional — with no `seed_qa_evaluator` wired the pipeline runs the Seed unconditionally — so turning QA on could leave a session strictly worse off than leaving it off. Blocking was also not a decision point: the operator received exactly the same unmapped prose the repair mapper could not act on.

What changed per path:

| path | before | after |
|---|---|---|
| `repair_budget_exhausted` | blocked | runs the best repaired Seed |
| `seed_qa_feedback_unmapped` | blocked on attempt 1 | stops the repair loop, runs the Seed as authored (no pointless re-judge of an unchanged Seed) |
| evaluator timeout / error / transient | blocked | advisory continue |
| pipeline deadline | blocked | **unchanged** — still blocks |
| repaired Seed below `required_grade` | blocked by `grade_gate` | **unchanged** — still blocks |

The unresolved verdict stays on `state.last_qa_*` (already surfaced by the result envelope) and is emitted as `auto.seed_qa.advisory_override` with a typed `reason`, so an advisory run is auditable rather than silent. The attention relay carries it as `seed_qa_advisory` with engine ownership `active` — the engine is still driving, so no successor action is offered. `auto.seed_qa.blocked` is kept for history persisted before this change.

`seed_qa_advisory.py` is a new module rather than more lines in `pipeline.py`: the module-size ratchet (#1797) holds `pipeline.py` at its budget, and the rationale for why this gate must never dead-end belongs somewhere readable.

## Test plan

- [x] `uv run python scripts/check-module-size.py` — OK

Contract tests rewritten from the blocking contract to the advisory one (`test_pipeline_runs_advisory_when_seed_qa_does_not_pass`, `test_pipeline_runs_advisory_on_unrepairable_seed_qa_feedback_without_retrying`), plus new coverage for the exhausted-budget path running the repaired Seed and for relay ownership staying `active`. The prompt-injection withholding assertion is preserved.

The interview round loop is untouched. This changes a one-shot pre-run gate, so there is no per-round cost to compare; the only runtime delta is that a previously-terminal session now proceeds to RUN.

## Related issues

Refs #2117

### DIFF
--- a/docs/rfc/active-conductor.md
+++ b/docs/rfc/active-conductor.md
@@ -149,8 +149,15 @@ not depend on the truncated human-readable event detail.
 
 ### Seed-QA gate
 
-Emit `auto.seed_qa.blocked` whenever the Seed-QA repair budget closes without a
-passing verdict.
+Emit `auto.seed_qa.advisory_override` whenever the Seed-QA gate gives up without
+a passing verdict — repair budget closed, feedback that the repair mapper cannot
+map, or an evaluator timeout/error. The gate is advisory: it never blocks the
+session, because a session with no Seed-QA evaluator wired runs unconditionally
+and a wired evaluator must not leave the session worse off. The unresolved
+verdict is carried into run → evaluate, which judges execution evidence.
+
+`auto.seed_qa.blocked` is retained for history persisted before the gate became
+advisory; nothing emits it now.
 
 Required data:
 
@@ -163,8 +170,10 @@ Required data:
 - `suggestions`
 - `reason`
 
-Transient evaluator errors and timeouts remain ordinary Auto blockers and use a
-different trigger code from a genuine non-passing Seed-QA verdict.
+Transient evaluator errors and timeouts carry their own `reason`
+(`evaluator_timeout` / `evaluator_error` / `evaluator_transient_error`) so they
+stay distinguishable from a genuine non-passing Seed-QA verdict
+(`repair_budget_exhausted` / `seed_qa_feedback_unmapped`).
 
 ### Frugality and stagnation
 
@@ -266,7 +275,8 @@ menus.
 | `deliver_verdict_rejected_streak` | at least two rejected verdicts for one `(judgment_scope_id, semantic_ac_key)` | verify immediately; mutate only after ownership closes |
 | `frugality_grounding_regression` | proof status `fail_grounding_regression` | successor only |
 | `frugality_no_savings` | proof status `fail_no_frugality` | successor only |
-| `seed_qa_blocked` | `auto.seed_qa.blocked` | successor/resume only |
+| `seed_qa_blocked` | `auto.seed_qa.blocked` (historical; the gate no longer blocks) | successor/resume only |
+| `seed_qa_advisory` | `auto.seed_qa.advisory_override` | none — engine ownership still active |
 | `lineage_stagnated` | `lineage.stagnated` | successor generation only |
 
 If a rejected streak is observed while engine ownership is still active, the
--- a/docs/rfc/active-conductor.md
+++ b/docs/rfc/active-conductor.md
@@ -156,11 +156,14 @@ session, because a session with no Seed-QA evaluator wired runs unconditionally
 and a wired evaluator must not leave the session worse off. The unresolved
 verdict is carried into run → evaluate, which judges execution evidence.
 
+The event is written only on the continuation path: the grade gate and the
+pipeline deadline both run before it. It is not retracted afterwards — a
+retraction cannot be made reliable, since the correcting append can fail
+exactly when the original succeeded — so the deadline is enforced at the
+boundaries that follow instead (RUN phase entry, and skip-run completion).
+
 `auto.seed_qa.blocked` is retained for history persisted before the gate became
-advisory, and has exactly one live producer: when the pipeline deadline trips
-*after* the advisory event was already persisted, it is re-emitted with
-`reason="pipeline_deadline_after_advisory"` so the ownership-closed
-classification corrects the standing "engine still driving" claim.
+advisory; nothing emits it now.
 
 Required data:
 
@@ -278,7 +281,7 @@ menus.
 | `deliver_verdict_rejected_streak` | at least two rejected verdicts for one `(judgment_scope_id, semantic_ac_key)` | verify immediately; mutate only after ownership closes |
 | `frugality_grounding_regression` | proof status `fail_grounding_regression` | successor only |
 | `frugality_no_savings` | proof status `fail_no_frugality` | successor only |
-| `seed_qa_blocked` | `auto.seed_qa.blocked` (history, plus the post-advisory deadline correction) | successor/resume only |
+| `seed_qa_blocked` | `auto.seed_qa.blocked` (historical; the gate no longer blocks) | successor/resume only |
 | `seed_qa_advisory` | `auto.seed_qa.advisory_override` | none — engine ownership still active |
 | `lineage_stagnated` | `lineage.stagnated` | successor generation only |
 
--- a/docs/rfc/active-conductor.md
+++ b/docs/rfc/active-conductor.md
@@ -162,6 +162,13 @@ retraction cannot be made reliable, since the correcting append can fail
 exactly when the original succeeded — so the deadline is enforced at the
 boundaries that follow instead (RUN phase entry, and skip-run completion).
 
+The event therefore carries **no ownership claim**. The append itself takes
+time, so a session can stop at the very next gate; an event that asserted
+"engine still driving" would be falsified by what happens next, in any path
+rather than one special case. Attention classification treats it as progress
+(`progress_advanced` / `seed_qa_advisory`), and ownership is read from the
+events that actually close a session.
+
 `auto.seed_qa.blocked` is retained for history persisted before the gate became
 advisory; nothing emits it now.
 
--- a/docs/rfc/active-conductor.md
+++ b/docs/rfc/active-conductor.md
@@ -157,7 +157,10 @@ and a wired evaluator must not leave the session worse off. The unresolved
 verdict is carried into run → evaluate, which judges execution evidence.
 
 `auto.seed_qa.blocked` is retained for history persisted before the gate became
-advisory; nothing emits it now.
+advisory, and has exactly one live producer: when the pipeline deadline trips
+*after* the advisory event was already persisted, it is re-emitted with
+`reason="pipeline_deadline_after_advisory"` so the ownership-closed
+classification corrects the standing "engine still driving" claim.
 
 Required data:
 
@@ -275,7 +278,7 @@ menus.
 | `deliver_verdict_rejected_streak` | at least two rejected verdicts for one `(judgment_scope_id, semantic_ac_key)` | verify immediately; mutate only after ownership closes |
 | `frugality_grounding_regression` | proof status `fail_grounding_regression` | successor only |
 | `frugality_no_savings` | proof status `fail_no_frugality` | successor only |
-| `seed_qa_blocked` | `auto.seed_qa.blocked` (historical; the gate no longer blocks) | successor/resume only |
+| `seed_qa_blocked` | `auto.seed_qa.blocked` (history, plus the post-advisory deadline correction) | successor/resume only |
 | `seed_qa_advisory` | `auto.seed_qa.advisory_override` | none — engine ownership still active |
 | `lineage_stagnated` | `lineage.stagnated` | successor generation only |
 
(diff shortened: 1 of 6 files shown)


Q00:
Both blockers addressed in `0d0e2e0`-range push.

**Blocker 1 — advisory event emitted before the grade gate.** Correct, and the consequence was worse than cosmetic: `attention_relay` reads `auto.seed_qa.advisory_override` as engine ownership `active` with no successor action, so a session that then stopped at `grade_gate` left durable evidence claiming the engine ran the Seed anyway. Emission now happens only on the true continuation path — `_seed_review_gate_blocker()` runs first and returns before any event is persisted. New test `test_advisory_event_is_not_emitted_when_the_grade_gate_then_blocks` drives a repaired Seed that trips the gate and asserts no advisory event was recorded.

**Blocker 2 — deadline not enforced around the awaited append.** Also correct. The append is bounded by the observer-drain timeout, which is not charged against `deadline_at`, and the skip-run branch transitions to COMPLETE with no deadline check of its own — so this PR could have turned an expired deadline into a successful completion, contradicting its own stated invariant. `_enforce_deadline()` now runs both before and after the append. New test `test_deadline_spent_during_the_advisory_append_still_blocks_skip_run` burns the remaining budget inside the append via a recording pipeline and asserts the session blocks with `pipeline_deadline` rather than completing.

The ordering rationale is documented in the `seed_qa_advisory` module docstring rather than inline, since `pipeline.py` is at its #1797 ratchet budget.

Verified: `tests/unit/auto` + `tests/unit/mcp/tools/test_attention_relay.py` 2001 passed; ruff, mypy (`src/`, 576 files), and `check-module-size.py` clean.

Q00:
Confirmed and fixed.

The evaluator-failure branches returned before the current attempt assigned `state.last_qa_*`, so a resumed checkpoint carrying a prior pass republished it — `verdict="pass"` with old evidence inside an `evaluator_error` event. Each attempt now clears the verdict before invoking the evaluator (`clear_seed_qa_verdict`), so an attempt that produces no verdict publishes none. Regression coverage is parametrized across all three failure branches with a persisted prior pass, asserting both the cleared state and that the emitted payload carries `verdict=None`, `score=None`, and no stale evidence.

One correction to the finding, stated plainly because it changes what the fix can claim: clearing `last_qa_passed` does **not** stop a `skip_run` terminal from reporting `artifact_state="complete_verified"`. `_artifact_state_for_result` falls through to `complete_verified` for any COMPLETE that is not `partial_product` and whose `run_handoff_status` is not `started` — with no verdict involved at all. I verified this after the fix: the stale-pass path is closed, and the terminal still reports `complete_verified`.

So that is a real defect, but a pre-existing and broader one: it predates this PR and is not specific to the advisory path. I left it out rather than widening this PR into a change to the completion-state contract, and noted it in the commit body. Happy to open it as its own issue, or to fix it here if you would rather it not ship as-is.

Verified: `tests/unit/auto` + `test_attention_relay.py` 2004 passed; ruff, mypy (576 files), module-size clean.

Q00:
Confirmed and fixed. The window was real: the append is awaited, so the deadline could only be re-checked after the event was already durable, and `attention_relay` reads that event as ownership `active`.

Two changes close it:

**Pre-charge the append.** The append is bounded by `_INTERVIEW_OBSERVER_DRAIN_TIMEOUT_SECONDS`. That bound is now charged against the remaining budget *before* publishing, so a budget that cannot cover a worst-case append blocks with nothing published. Under a monotonic clock, "stopped inside the append" is no longer a reachable state.

**Self-correct if it happens anyway.** For a clock jump or an append that overruns its own bound, the post-append block now emits `auto.seed_qa.blocked` with `reason="pipeline_deadline_after_advisory"`. I chose that existing type deliberately rather than inventing a new one: `attention_relay` already classifies it as ownership **closed** with a successor action offered, which is precisely the record a stopped session needs — the durable stream corrects itself instead of leaving the active claim standing. (This also means the type I described as historical in the RFC has a live producer again, for exactly one case.)

The regression test you pointed at was encoding the contradictory state, so it was rewritten to assert the correction: burning the budget inside the append now requires both the block *and* the closure event. A second test covers the pre-charge path — a budget too small for the append blocks with no advisory event emitted at all.

The sequencing moved into `seed_qa_advisory.publish_advisory_or_block` rather than growing `pipeline.py`, which is at its #1797 ratchet budget.

Verified: `tests/unit/auto` + `test_attention_relay.py` 2005 passed; ruff, mypy (576 files), module-size clean.

Q00:
Re-requesting review — the round-4 blocker (advisory event outliving a post-append deadline block) is addressed at `dfe532bc1`: the append's own bound is charged before publishing, and a post-append block self-corrects with the ownership-closed event type. Details in the previous comment.

Q00:
Both blockers accepted, and I reverted the design rather than patching it again.

**Blocker 2 was a regression I introduced.** Reserving the append's worst-case latency turned an unexpired deadline into a blocker — including for drivers with no EventStore, where the append is a no-op. Your framing is the decisive one: a session with Seed QA *disabled* gets no such block, so a wired evaluator must not impose one. That is the same asymmetry this PR exists to remove, reintroduced by my own fix.

**Blocker 1 shows retraction cannot be made reliable.** `_emit_runtime_event` is fail-open, so the correcting append can fail exactly when the original succeeded. Any "publish, then retract on failure" scheme has that hole; adding acknowledgement or bounded retry would be building a durability protocol to compensate for publishing a claim too early.

So both fixes are gone, and the deadline is enforced where the gap actually was. **Skip-run completion was the only COMPLETE transition with no deadline check of its own** — RUN has one at `_run_execution_phase` entry. Making the gate advisory means sessions now reach that transition where they previously stopped at the gate, which is why the hole became visible here. A check at that boundary protects every path through it, not just the advisory one, and the advisory event goes back to being written only after all gates have passed, never retracted.

Net effect on the PR's contracts:
- unexpired deadlines are never redefined as expired (new test)
- an expired deadline blocks skip-run completion instead of completing it (new test, replacing the retraction test)
- the advisory event exists only on the continuation path

Follow-up finding also taken: the cleared `last_qa_*` is now persisted before the attempt runs, so an interruption mid-attempt cannot leave the state file carrying the previous verdict while the event describes none.

RFC updated to match — `auto.seed_qa.blocked` has no live producer again.

Verified: `tests/unit/auto` + `test_attention_relay.py` + `test_auto_command.py` 2044 passed; ruff, mypy (576 files), module-size clean.

Q00:
Correct, and I need to own how it happened rather than argue with it: the coverage **did** exist — I added it with the verdict-clearing fix — and I deleted it myself two commits later when I rewrote the deadline tests that surrounded it. My comment at the time claimed the branches were covered; by the time you looked, that claim was stale and yours was accurate. I verified before replying rather than pushing back.

Restored, and strengthened along the lines you asked for. `test_a_failed_evaluator_never_republishes_the_previous_verdict` is parametrized over all three failure reasons (`evaluator_timeout` / `evaluator_error` / `evaluator_transient_error`), each starting from a persisted stale **passing** verdict, and asserts:

- the cleared verdict is *persisted before the evaluator awaits* — the test now reloads the session from the store rather than reading in-memory state, which is the property that matters for a mid-attempt interruption
- the advisory payload carries no inherited verdict, score, differences or suggestions
- `state.last_qa_*` is cleared, so the stale pass cannot reach `complete_verified` classification

On that last point, one honest limitation carried over from the earlier round: clearing `last_qa_passed` does not by itself stop a skip-run terminal from reporting `complete_verified`, because `_artifact_state_for_result` falls through to it with no verdict involved. That is fixed separately in #2127 (already approved) rather than widened into this PR.

Verified: `tests/unit/auto` 1,983 passed; ruff clean.

Q00:
Confirmed — and this is the third round on the same contradiction, so I stopped patching the sequencing and fixed what was actually wrong.

Both previous attempts failed on their own terms, and you caught each: retraction cannot be reliable (the correcting append fails exactly when the original succeeded), and pre-charging the append's latency redefined an unexpired deadline as expired — blocking sessions that a *disabled* Seed-QA gate would have let run. Now the same contradiction reappears at ordinary RUN entry. Three variants of one defect means the sequencing was never the defect.

**The defect is that an informational event was given adjudicating authority.** The engine ran the Seed despite an unresolved verdict — that is all this event knows. Whether it *keeps* going is decided by the gates that follow, and their own events say so. Any "engine still driving" claim published before those gates run can be falsified by what happens next, in any path, including paths neither of us has enumerated yet.

So  is now classified as **progress** ( / ) instead of attention-with-ownership: no , no action menu, evidence preserved. Ownership is read from the events that actually close a session.  keeps its closed-ownership attention shape for pre-existing history.

This removes the failure mode structurally rather than per-path — there is no longer an ownership claim for a later block to contradict, whether the block lands at RUN entry, skip-run completion, or anywhere added later.

Regression for the path you named: budget burned inside the append → blocks at RUN entry with `pipeline_deadline` → the persisted advisory payload is fed through `classify_relay_events` and asserted to produce no `attention_required` relay and no `engine_ownership` key.

RFC updated with the reasoning.

Verified: `tests/unit/auto` + `test_attention_relay.py` 2,005 passed; ruff, mypy (576 files), module-size clean.

Q00:
Correction to my previous comment — shell mangling stripped several identifiers out of it. Reposting intact.

Confirmed, and this is the third round on the same contradiction, so I stopped patching the sequencing and fixed what was actually wrong.

Both previous attempts failed on their own terms, and you caught each: retraction cannot be reliable (the correcting append fails exactly when the original succeeded), and pre-charging the append's latency redefined an unexpired deadline as expired — blocking sessions that a *disabled* Seed-QA gate would have let run. Now the same contradiction reappears at ordinary RUN entry. Three variants of one defect means the sequencing was never the defect.

**The defect is that an informational event was given adjudicating authority.** The engine ran the Seed despite an unresolved verdict — that is all this event knows. Whether it *keeps* going is decided by the gates that follow, and their own events say so. Any "engine still driving" claim published before those gates run can be falsified by what happens next, in any path, including paths neither of us has enumerated yet.

So `auto.seed_qa.advisory_override` is now classified as **progress** (`progress_advanced` / `seed_qa_advisory`) instead of attention-with-ownership: no `engine_ownership`, no action menu, evidence preserved. Ownership is read from the events that actually close a session. `auto.seed_qa.blocked` keeps its closed-ownership attention shape for pre-existing history.

This removes the failure mode structurally rather than per-path — there is no longer an ownership claim for a later block to contradict, whether the block lands at RUN entry, skip-run completion, or anywhere added later.

Regression for the path you named: budget burned inside the append → blocks at RUN entry with `pipeline_deadline` → the persisted advisory payload is fed through `classify_relay_events` and asserted to produce no `attention_required` relay and no `engine_ownership` key.

RFC updated with the reasoning.

Verified: `tests/unit/auto` + `test_attention_relay.py` 2,005 passed; ruff, mypy (576 files), module-size clean.

Q00:
Re-requesting review — every blocker from rounds 1-7 is addressed at `927854fa`, which landed **after** the last review (review 16:01:10Z, commit 16:04:08Z), so the `CHANGES_REQUESTED` state predates the fix.

Audited against the pushed tree rather than from memory:

| round | blocker | where it is addressed now |
|---|---|---|
| 1a | advisory event emitted before the grade gate | `_seed_qa_advisory_continue` runs `_seed_review_gate_blocker` before `publish_advisory` |
| 1b | deadline not enforced around the awaited append | `_enforce_deadline` immediately before the append; skip-run completion gained its own check (it was the only COMPLETE transition without one) |
| 2 | stale `last_qa_*` republished by an evaluator failure | `clear_seed_qa_verdict` at the top of each attempt, **persisted** before the evaluator awaits |
| 3/4 | post-append block leaves an active ownership claim | retraction and pre-charge both reverted — retraction cannot be reliable (the correcting append fails when the original succeeded), pre-charge redefined an unexpired deadline as expired |
| 5 | evaluator-failure branches uncovered | parametrized regression over `evaluator_timeout` / `evaluator_error` / `evaluator_transient_error`, each from a persisted stale pass, asserting the cleared verdict **from the store** |
| 6/7 | same contradiction at RUN entry | `auto.seed_qa.advisory_override` no longer carries ownership at all: classified as `progress_advanced` / `seed_qa_advisory`, no `engine_ownership`, no action menu |

The round-7 fix is the structural one: three variants of one defect meant the sequencing was never the defect. An informational event had been given adjudicating authority; whether the engine keeps going is decided by the gates that follow, and their own events say so. There is no longer a claim for any later block to contradict — at RUN entry, skip-run completion, or a boundary added later.

Verified at this SHA: `tests/unit/auto` + `test_attention_relay.py` 2,005 passed; ruff, mypy (576 files), module-size clean.

Separately, for the record: the three `Test Python 3.x` CI jobs fail on this PR, on #2120 / #2121 / #2127, **and on `main` itself** (run 31800958202). That failure is not introduced by this branch; I am looking at it as its own problem.

Issue #2118, opened 2026-08-14
auto: unify the run → evaluate → ralph successor chain (four different definitions of "done")


Q00:
## Problem

"Is this product done?" is answered four different ways depending on how the session was started.

| entry point | judge |
|---|---|
| `ooo run` | formal `ouroboros_evaluate` (3 stages), then chained Ralph on rejection |
| `ooo auto` (default) | **none** — the run job's chain is explicitly suppressed |
| `ooo auto --complete-product` (MCP, non-plugin) | `ouroboros_qa` — a different judge |
| `ooo auto --complete-product` (MCP plugin / CLI) | **none** — `evaluator` is unwired, so Ralph finishing counts as success |

### Default auto has no evaluation owner

`HandlerRunStarter` sends `auto_evaluate=False, auto_evolve=False` (`auto/adapters.py`, added in #1916) with the rationale that "Auto owns its own RALPH_HANDOFF and final-evaluation path". That ownership only exists when `complete_product=True`:

- `_ALLOWED_TRANSITIONS[RUN] = {COMPLETE, RALPH_HANDOFF, BLOCKED, FAILED}` — there is no `RUN → EVALUATE` edge
- `EVALUATE` is reachable only from `RALPH_HANDOFF`, which is entered only under `complete_product`

So default auto suppresses the run job's chain *and* has nothing of its own: the run finishes and no one grades it.

### The two owners are not the same engine

- `ouroboros_evaluate`: Stage 1 runs lint/build/test for real; Stage 2 judges with a read-only tool envelope (`max_turns=20`); Stage 3 multi-model consensus when uncertain.
- `ouroboros_qa` (auto's `HandlerEvaluator`): a single-turn LLM verdict with `allowed_tools=[]` — no tools at all.

## Expected

One successor policy, one judge. `ooo auto` should be `ooo run` plus the interview, and `--complete-product` should mean "wait for that same chain" rather than owning a second, weaker copy of it.

Retiring the private path removes roughly 1.3k lines (`_handoff_to_ralph`, `_resume_ralph_handoff`, auto's `_run_evaluate`, `HandlerEvaluator`).

Q00:
## Deletion plan for the private complete-product path

Posting the measured scope so execution is mechanical once the direction is confirmed.

### Why deletion rather than rewiring (evidence from #2120)

#2120 needed two rounds to produce a working successor chain, and both failures were the same shape: a stack assembled by hand looks complete and is not.

[...]

PR #2120, opened 2026-08-14
feat(auto): delegate run successors to the run job's own chain

Q00:
## Summary

Default `ooo auto` had no evaluation owner. `HandlerRunStarter` sent `auto_evaluate=False, auto_evolve=False` on every Auto-started run (added in #1916) because "Auto owns its own RALPH_HANDOFF and final-evaluation path" — but that ownership only exists under `complete_product`:

- `_ALLOWED_TRANSITIONS[RUN] = {COMPLETE, RALPH_HANDOFF, BLOCKED, FAILED}` — there is no `RUN → EVALUATE` edge
- `EVALUATE` is reachable only from `RALPH_HANDOFF`, entered only under `complete_product`
- default Auto reaches `COMPLETE` as soon as the run has a durable handle

So default Auto suppressed the run job's successors *and* had nothing of its own: the run finished and nobody graded it.

The starter now takes `owns_successors`, true only for complete-product Auto:

| mode | override sent | effect |
|---|---|---|
| `complete_product=true` | `auto_evaluate/auto_evolve = False` | unchanged — Auto drives RALPH_HANDOFF → EVALUATE, no competing owner or Ralph lineage |
| default | **none** | `execution.auto_evaluate` / `execution.auto_evolve` govern, exactly as for a direct `ooo run` |

Sending an explicit `True` was rejected deliberately: it would override a config that intentionally disabled the chain — the same class of bug in the opposite direction. Delegating means an Auto-started run behaves like `ooo run`, which is the point.

The CLI constructs `HandlerRunStarter` only in the non-complete-product branch, so it is correct by construction; the MCP handler passes `owns_successors=complete_product`.

This is one step of #2118 — it fixes the missing owner, and does not yet unify the two different judges. That is the follow-up.

## Test plan

New coverage: the delegation contract (no override sent when Auto does not own successors), the preserved single-owner contract under complete-product, and a parametrized wiring guard that `run_starter.owns_successors is complete_product` — the wiring is exactly what regressed before.

The change is two dictionary keys on the run handoff arguments; the interview round loop is untouched, so there is no per-round cost to compare.

Worth stating plainly: an Auto-started run now enqueues an evaluation, and a rejected verdict enqueues a bounded Ralph loop. That is more work per session than before — it is the work that was supposed to happen and was being suppressed. Both remain switchable through `execution.auto_evaluate` / `execution.auto_evolve`.

## Related issues

Refs #2118

### DIFF
--- a/skills/auto/SKILL.md
+++ b/skills/auto/SKILL.md
@@ -101,7 +101,8 @@ ask or send either argument; Auto restores the persisted contract.
 3. Generates a Seed.
 4. Reviews and repairs until A-grade or blocked.
 5. Starts execution only after A-grade.
-6. When `complete_product=true`, chains RUN → RALPH_HANDOFF after a successful run handoff and waits for a terminal Ralph status so a single invocation iterates Ralph until QA passes, convergence, or a budget bound trips. A QA-pass on the executed product completes the auto session; recognized failure modes (`iteration_timeout`, `wall_clock_exhausted`, `oscillation_detected`, `grade_regressing`, `max_generations reached`) block the auto session with the matching `stop_reason` in `last_error` so operators can resume after the cause is addressed.
+6. When `complete_product=false` (the default), auto reaches `COMPLETE` as soon as the run has a durable handle — the run itself keeps going as a background job, and that job carries its own `run → evaluate → ralph` chain governed by `execution.auto_evaluate` / `execution.auto_evolve` (both default `true`, Ralph bounded by `execution.auto_evolve_max_generations`). Auto does not evaluate the run itself in this mode; the chained evaluate job does.
+7. When `complete_product=true`, chains RUN → RALPH_HANDOFF after a successful run handoff and waits for a terminal Ralph status so a single invocation iterates Ralph until QA passes, convergence, or a budget bound trips. A QA-pass on the executed product completes the auto session; recognized failure modes (`iteration_timeout`, `wall_clock_exhausted`, `oscillation_detected`, `grade_regressing`, `max_generations reached`) block the auto session with the matching `stop_reason` in `last_error` so operators can resume after the cause is addressed.
 
 ## Background monitoring UX
 
--- a/src/ouroboros/auto/adapters.py
+++ b/src/ouroboros/auto/adapters.py
@@ -247,6 +247,7 @@ class HandlerRunStarter:
         efficiency_mode: str = "adaptive",
         frugality_assurance: str = "observe",
         frugality_assurance_explicit: bool = False,
+        owns_successors: bool = False,
     ) -> None:
         self.handler = handler
         self.cwd = cwd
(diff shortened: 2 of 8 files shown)


Q00:
Blocker confirmed and fixed. This was the important one — the PR as submitted was inert.

Reproduced the finding in code: `cli/commands/auto.py` and `_execution_start_handler` both built `StartExecuteSeedHandler` without `start_evaluate_handler`, so `enqueue_chained_evaluation` received `None` and terminated at `evaluation_status="enqueue_failed"`. Enabling the chain without the dependency left exactly the state the PR claims to remove: the run finishes, nothing grades it.

Fix: new `mcp/tools/run_successors.py` owns the assembly (`EvaluateHandler` → `StartEvaluateHandler` → `StartRalphHandler`), used by both Auto composition roots. A shared factory rather than two more hand-rolled copies, because a missing link degrades silently — the same failure mode one level down — and `adapter.py` already keeps a fourth copy of the same stack.

Two details worth flagging for review:
- `opencode_mode` is deliberately **not** forwarded to the Ralph link. Automatic convergence is parent-owned even in passive plugin mode and must enqueue a real pollable job instead of emitting a second delegation envelope, mirroring the shipped stdio server.
- `resolve_run_successor_handler` **carries** an injected stack across a rebuild for another runtime rather than rebuilding it, per the review note, and reads backends from the rebuilt execute handler so successors judge on the runtime the run actually used.

Coverage for the gap: `test_mcp_execution_start_handler_can_enqueue_successors` (both links present), `test_rebuild_preserves_an_injected_successor_stack`, and `test_run_auto_wires_the_run_successor_stack` (CLI pipeline's run starter carries both links). All three fail without the fix.

Verified: `tests/unit/cli/test_auto_command.py` + `test_start_auto.py` + `tests/unit/auto` 2088 passed; ruff, mypy (576 files), module-size clean. Ten failures in `tests/unit/cli/test_mcp_*` are ANSI-color assertions that fail identically on `main` when that file is run in isolation — unrelated to this change.

Q00:
Confirmed and fixed — the stack was incomplete one link below where the previous round stopped.

`StartRalphHandler` built by hand gets an `EvolveStepHandler` with `evolutionary_loop=None`, so a rejected evaluation enqueues a Ralph job that dies on its first non-plugin generation with `EvolutionaryLoop not configured`. My previous fix produced a chain that looked complete and was not — the same shape as the original bug, one link down.

Notably the repo already documents this exact trap: `_build_configured_ralph_handler` in `cli/commands/auto.py` exists because the complete-product path hit it, and solves it by reusing the MCP composition root — the only place that wires the loop's executor, evaluator, validator and rewind observer. `run_successors.py` now resolves `ouroboros_start_evaluate` from that same root rather than assembling a second, weaker copy, and the module docstring records why hand-assembly is not an option here (every link degrades silently, in three different ways).

On the test note — you're right that non-null assertions proved nothing. Both the MCP and CLI tests now assert `start_ralph._evolve_handler.evolutionary_loop is not None`, which is the property the first generation actually depends on; both fail against the previous commit. I did not add a live rejected-evaluation integration test that runs a real generation, because that would spawn actual runtime work in unit scope; if you want that coverage at integration level, say so and I will add it there instead.

Verified: `tests/unit/cli/test_auto_command.py` + `test_start_auto.py` + `tests/unit/auto` 2088 passed; full `tests/unit` 20042 passed / 88 skipped on the previous commit; ruff, mypy (576 files), module-size clean.

PR #2121, opened 2026-08-14
feat(auto): add a follower for the run job's successor chain

Q00:
## Summary

Groundwork for #2118. Auto's complete-product mode re-implements the successor policy `ooo run` already owns — a private `RALPH_HANDOFF → EVALUATE` path with a different judge — which is how the product ended up with four incompatible definitions of "done".

This adds the piece that lets Auto reuse the run chain instead of owning a second one: a follower that walks the successor handles the chain publishes on each job's result meta (`chained_evaluate_job_id` → `chained_ralph_job_id`) and reports the terminal of the deepest link it reached.

The approval rule is the substance of the module:

```python
# a `completed` run job is NOT approval
if self.result_meta.get("final_approved") is True:
    return True
return self.job_kind == "ralph" and self.result_meta.get("stop_reason") == RALPH_APPROVED_STOP_REASON
```

Execution finishing is not a verdict. That conflation is exactly what lets a Ralph terminal count as a passing product in the CLI and plugin-mode complete-product paths today, where no evaluator is wired at all. `RALPH_APPROVED_STOP_REASON` is pinned to `"qa passed"` (`ralph_loop.py:305`) so the check can never drift into matching a budget terminal like `max_generations reached`.

A deadline returns the handles gathered so far with `status="timed_out"` rather than dropping the chain, so a caller can resume rather than lose the successors.

**Nothing here starts work.** The chain is started by the run job itself; this only observes, so the module is inert until a caller waits on it. Wiring complete-product to it — and deleting `_handoff_to_ralph`, `_resume_ralph_handoff`, Auto's `_run_evaluate` and `HandlerEvaluator` (~1.3k lines) — is the follow-up.

## Test plan

- [x] `uv run pytest tests/unit/auto/test_chain_follower.py -q` — 5 passed

Covers: the full run → evaluate → ralph walk; stopping where the chain stops (an approving evaluation publishes no Ralph handle); a finished run alone not counting as approval; a budget terminal not counting as approval; and a deadline returning handles instead of losing them.

New module with no call sites; no existing code path changes behaviour.

## Related issues

Refs #2118

### DIFF
--- a/src/ouroboros/auto/chain_follower.py
+++ b/src/ouroboros/auto/chain_follower.py
@@ -104,6 +104,13 @@ _EXPECTED_JOB_TYPES: dict[str, str] = {
     "ralph": "ralph",
 }
 MISMATCHED_LINK_STATUS = "link_type_mismatch"
+# A successor handle is a cross-job reference into durable state. It can name a
+# job that no longer exists (pruned history, a different store, a mis-copied
+# id), and ``JobManager.get_snapshot`` raises for an unknown id. The walk fails
+# closed on that too: an unreadable link is a non-approving terminal that keeps
+# its handles, never an exception that loses them.
+UNAVAILABLE_LINK_STATUS = "link_unavailable"
+TIMED_OUT_STATUS = "timed_out"
 
 
 async def follow_run_chain(
@@ -133,15 +140,16 @@ async def follow_run_chain(
 
     while True:
         kind = _JOB_KINDS[min(depth, len(_JOB_KINDS) - 1)]
-        snapshot = await _await_terminal(job_manager, job_id, poll_seconds, expiry)
+        observed = await _await_terminal(job_manager, job_id, poll_seconds, expiry)
         followed.append(job_id)
-        if snapshot is None:
+        if isinstance(observed, str):
             return ChainTerminal(
                 job_id=job_id,
                 job_kind=kind,
-                status="timed_out",
+                status=observed,
                 followed_job_ids=tuple(followed),
             )
+        snapshot = observed
         if getattr(snapshot, "job_type", None) != _EXPECTED_JOB_TYPES[kind]:
             # Refuse to interpret the metadata of a job that is not the link it
             # claims to be — neither its verdict nor its successor handle.
@@ -181,8 +189,12 @@ async def _await_terminal(
     job_id: str,
     poll_seconds: float,
     expiry: float | None,
-) -> _Snapshot | None:
-    """Poll one job until terminal; None when the deadline trips first.
+) -> _Snapshot | str:
+    """Poll one job until terminal.
+
+    Returns the snapshot, or a terminal status string when the job cannot be
+    observed: :data:`TIMED_OUT_STATUS` when the budget runs out,
+    :data:`UNAVAILABLE_LINK_STATUS` when the handle names no readable job.
 
     The read itself is bounded, not just the interval between reads: a snapshot
     that comes from persistence or a reconciliation path can block, and a
@@ -191,22 +203,24 @@ async def _await_terminal(
     """
     loop = asyncio.get_running_loop()
     while True:
-        if expiry is None:
-            snapshot = await job_manager.get_snapshot(job_id)
-        else:
-            remaining = expiry - loop.time()
-            if remaining <= 0:
-                return None
-            try:
+        try:
+            if expiry is None:
+                snapshot = await job_manager.get_snapshot(job_id)
+            else:
+                remaining = expiry - loop.time()
+                if remaining <= 0:
+                    return TIMED_OUT_STATUS
                 snapshot = await asyncio.wait_for(
                     job_manager.get_snapshot(job_id), timeout=remaining
                 )
-            except TimeoutError:
-                return None
+        except TimeoutError:
+            return TIMED_OUT_STATUS
+        except (ValueError, KeyError, LookupError):
+            return UNAVAILABLE_LINK_STATUS
         if snapshot.is_terminal:
             return snapshot
         if expiry is not None and loop.time() >= expiry:
-            return None
+            return TIMED_OUT_STATUS
         remaining_sleep = (
             poll_seconds if expiry is None else min(poll_seconds, expiry - loop.time())
         )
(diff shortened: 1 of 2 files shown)


Q00:
Blocker confirmed and fixed — the module contained the very defect it exists to remove, one layer down.

`approved` read `final_approved` regardless of `job_kind`. Since result meta is merged and forwarded along the chain, a run terminal or a *failed* Ralph could inherit an approval it never issued. Approval is now read only from the link that owns the signal:

- `final_approved` → only from the `evaluate` link
- `stop_reason == "qa passed"` → only from the `ralph` link
- and only when that job itself finished cleanly (`completed`), so `failed` / `cancelled` / `interrupted` / the follower's own `timed_out` never approve

Regression coverage is parametrized over exactly the cases named in the finding: run-with-`final_approved`, ralph-with-`final_approved`, failed-evaluate-with-`final_approved`, and failed-ralph-with-`qa passed`.

Follow-up finding also taken: `test_deadline_preserves_the_deepest_handles_reached` times out at the Ralph link after the run and evaluation have both published successors, asserting the terminal keeps `("job_run", "job_ev", "job_ralph")` so the chain is resumable from the deepest handle rather than only the root.

Verified: `tests/unit/auto/test_chain_follower.py` 10 passed; ruff and mypy clean.

Context for reviewers: whether this module gets a caller at all is currently an open owner decision. If `--complete-product`'s private path is deleted outright rather than rewired (#2118), nothing will wait on the chain in-process and this PR should be closed instead of merged.

Q00:
Both confirmed and fixed.

**Link identity.** `job_kind` came from traversal depth alone, so a stale or misdirected `chained_evaluate_job_id` could name any completed job and have its `final_approved` read as this chain's verdict. Each link is now validated against the durable `job_type` the chain actually creates (`execute_seed` / `evaluate` / `ralph`); a mismatch stops the walk with `link_type_mismatch` and refuses to interpret that job's verdict *or* follow its successor. Regression test resolves the evaluate handle to another `execute_seed` carrying `final_approved=True` and asserts non-approval.

**Deadline.** Correct — `deadline_seconds` bounded the interval between reads, not the reads. `get_snapshot()` is now awaited under the remaining budget, so a blocked persistence or reconciliation read returns `timed_out` with the handles gathered so far instead of outliving the budget or hanging. Regression test delays the read 5s against a 50ms deadline and asserts the walk returns `timed_out` rather than the late result.

The earlier deep-timeout test was also corrected while doing this: with a bounded read, `deadline_seconds=0.0` can no longer traverse anything, so it now uses a positive budget with a slow Ralph read and still asserts `("job_run", "job_ev", "job_ralph")` is preserved.

Verified: `tests/unit/auto/test_chain_follower.py` 12 passed; ruff and mypy clean.

Standing note from the previous round: whether this module ever gets a caller is an open owner decision (#2118). If `--complete-product`'s private path is deleted rather than rewired, this PR should be closed rather than merged.

Q00:
Confirmed and fixed.

Correct, and it contradicted the module's own premise: the walk already treats successor ids as untrusted (a mismatched `job_type` stops it rather than being interpreted), but an id naming *no* job propagated `ValueError: Job not found` and destroyed the `ChainTerminal` — losing exactly the resume handles the timeout path is careful to preserve.

An unreadable link is now the same class of outcome as a mismatched one: a `link_unavailable` terminal that never approves and keeps `followed_job_ids`. The two unobservable outcomes are named constants returned from the poll (`TIMED_OUT_STATUS`, `UNAVAILABLE_LINK_STATUS`) so the walk distinguishes "budget ran out" from "handle names nothing" instead of collapsing both into `None`. `LookupError`/`KeyError` are caught alongside `ValueError` so an alternative store's lookup failure behaves the same.

Regression test reproduces your case exactly — a completed run publishing `chained_evaluate_job_id="missing"`, with the fake manager raising the same `ValueError: Job not found` shape `JobManager` does — and asserts `status=link_unavailable`, `approved is False`, and `followed_job_ids == ("job_run", "missing")`.

Verified: `tests/unit/auto/test_chain_follower.py` 13 passed; ruff and mypy clean.

Q00:
Re-requesting review — the round-3 blocker (a successor handle naming no readable job crashing the walk) is addressed at `2aa24ec29`: an unreadable link is now a non-approving `link_unavailable` terminal that keeps its handles. Details in the previous comment.

Q00:
Confirmed and fixed. `job_type` proved a link was *an* evaluation, not *this chain's* evaluation — a real evaluation of somebody else's run passed and donated its `final_approved`.

Each successor is now bound to an identity its predecessor published:

| link | predecessor key | successor key |
|---|---|---|
| evaluate | `session_id` (run) | `session_id` |
| ralph | `chained_ralph_lineage_id` (evaluation) | `lineage_id` |

Both are values the chain already publishes (`run_evaluate_chain` passes the run's session into the evaluation arguments; `evaluate_ralph_chain` mints and publishes the lineage). A link whose expected identity is absent on either side also fails closed with `link_identity_mismatch` — this boundary grants approval, so an unauthenticated link must decide nothing rather than be trusted by default.

Same-type cross-chain regression tests as requested: a valid `evaluate` job from another session, and a valid `ralph` terminal from another lineage — both refused, both non-approving. The existing scripted chains now carry the identities too, so the happy paths prove the binding accepts genuine successors rather than merely rejecting everything.

Verified: `tests/unit/auto/test_chain_follower.py` 15 passed; ruff and mypy clean.

Q00:
Closing: the owner decided to retire the private complete-product path (#2128) rather than rewire it, so nothing will wait on the chain in-process and this follower will never get a caller.

This was the fork stated in the PR body from the first round, and it resolved against merging: with `--complete-product` gone, Auto ends at run handoff and the run job's own chain owns everything after it. A caller that waits would have to be re-introduced to justify this module, and the reason it was removed is that a second owner of the successor chain is what caused the defects #2120 spent two rounds fixing.

Merging it as dead code would be worse than closing it. `chain_follower.py` is small and self-contained; the branch stays available if a "wait for the product" surface is ever wanted at the caller level — the tests already pin the parts that were hard to get right:

- approval is read only from the link that owns the signal (`final_approved` from `evaluate`, `qa passed` from `ralph`), and only from a `completed` job — a run terminal or a failed Ralph cannot inherit a verdict it never issued
- each link is bound to an identity its predecessor published (run `session_id` → evaluation, evaluation `chained_ralph_lineage_id` → Ralph), so a valid job from another chain is refused
- unreadable and mistyped links fail closed with their handles preserved, rather than raising and losing the chain

Thanks for the review rounds — the identity-binding and fail-closed findings applied to the module regardless of whether it ships, and the same "only the owner may adjudicate" rule ended up being the fix in #2119 as well.

PR #2128, opened 2026-08-14
refactor(auto): retire the private complete-product path

Q00:
> **Draft — needs an owner decision, not just a review.** This removes a documented mode. It is the deletion direction discussed in #2118; I have not requested merge.

## Summary

`ooo run` already owns the successor policy: an evaluable terminal execution enqueues a formal evaluation, and a rejected verdict enqueues a bounded Ralph loop. `--complete-product` re-implemented that inside Auto's own phase machine (`RALPH_HANDOFF` → `EVALUATE`) with a **different judge** — `ouroboros_qa`, a single-turn verdict with `allowed_tools=[]`, against the chain's three-stage `ouroboros_evaluate` that runs lint/build/test for real. Two owners, two definitions of "done".

**#2120 is the evidence for deleting rather than rewiring.** Keeping the second owner means keeping a second successor stack wired by hand, and each link degrades *silently*, in a different way:

| missing link | symptom |
|---|---|
| `start_evaluate_handler` | run terminates at `evaluation_status="enqueue_failed"` |
| `start_ralph_handler` | `chained_ralph_skipped`, one link further down |
| `evolutionary_loop` on a hand-built `StartRalphHandler` | Ralph job *is* enqueued, then dies on its first generation with `EvolutionaryLoop not configured` |

That took two review rounds and produced two complete-looking chains that were not. `create_ouroboros_server` is the only authoritative wiring; every other site is a partial copy. This path required maintaining one.

## What is removed

`src/ouroboros/auto/pipeline.py`: **5,286 → 3,350 lines**

| | lines |
|---|---|
| `_handoff_to_ralph`, `_resume_ralph_handoff`, `_run_evaluate`, `_finalize_evaluate`, `_run_lateral` | 1,063 |
| helpers existing only to serve them (`_poll_ralph_job`, `_reattach_ralph_job`, `_evaluate_or_complete`, `_finalize_lateral`, `_maybe_route_ralph_oscillation_to_lateral`, `_recover_seed_for_lateral`, `_complete_runtime_probe_blocker`) | 502 |
| `complete_product` branches, fields, resume mappings | ~370 |

Plus `AutoPipeline`'s `ralph_starter` / `ralph_resumer` / `complete_product` / `evaluator` / `probe_runner`, the CLI and MCP wiring behind them, and `HandlerEvaluator` — the toolless QA judge.

Whole diff: **8,143 deletions, 180 insertions.**

The reclaimed budget is locked into the module-size ratchet (`5286 → 3350`) rather than left available to spend, as `check-module-size.py` instructs.

## Compatibility

- **`--complete-product` is accepted, warns, and is ignored.** Scripts carrying it do not hard-fail; the flag can be deleted a release later. (Open question below.)
- **Sessions persisted in a retired phase** resume to a BLOCKED result naming the change and pointing at the run job's own chain, instead of dispatching to a handler that no longer exists. The enum members stay so those state files still deserialize.
- **`UNSTUCK_LATERAL` goes with it**, as a consequence rather than a target: both forward entries were complete-product gated (`pipeline.py:3441` explicitly). The Seed-QA repair lateral is a separate path (`_repair_seed_after_qa`) and is untouched — 48 of its tests still pass.

## What is lost

**Synchronous completion** — "one invocation until the product is done". The job observer contract already follows `chained_evaluate_job_id` / `chained_ralph_job_id` (`_FOLLOW_JOBS`) and `ouroboros_job_wait` exists, so this is replaceable at the caller rather than owned by the pipeline.

**Deadline → Ralph budget propagation, dropped intentionally.** The private path forwarded the auto session's remaining budget into Ralph as `max_total_seconds` plus a per-iteration cap. The official chain has no equivalent — `run_evaluate_chain.py` and `evaluate_ralph_chain.py` contain zero references to a deadline — so this behaviour genuinely disappears rather than moving.

It is dropped rather than ported because the relationship it encoded no longer exists: `--timeout` bounds *an auto session*, and after this change an auto session ends at run handoff. Porting it would rebuild the cross-owner coupling this cleanup exists to remove.

Stated precisely, because the imprecise version is the same kind of false promise this PR is removing from the docs: the chained Ralph is **finite but has no single wall-clock cap**. `evaluate_ralph_chain.py` dispatches without `max_total_seconds`, so `RalphHandler` receives `None`; the bound is `execution.auto_evolve_max_generations` (default 3) generations, each capped by Ralph's per-iteration timeout. If maintainers decide a total time cap is needed, the insertion point is one argument on that chain dispatch — not a revived private path.

If the maintainers consider "the auto deadline bounds the whole product build" a contract worth keeping, the right move is to port it into the chain as its own change, not to keep the private path alive for it.

**Ralph resume / lineage retry**: the chain covers the duplicate-dispatch case with `run_durable_lineage_single_flight` in `evaluate_ralph_chain.py`. Auto's own lineage-retry machinery was for reattaching to *its* dispatch, which no longer exists.

Also gone: job event mirroring into Auto state, and the `complete_product`-gated `probe_runner`.

## Test plan

- [x] `uv run pytest tests/unit tests/integration -q` — **20,167 passed**, 92 skipped, 1 xfailed. The 7 failures in that run were all `complete_product` key-set contracts (capability registry, skill-tool mapping, router dispatch), now updated, plus one ANSI-colour assertion in `test_wheel_packaging.py` that passes under `NO_COLOR=1` and fails identically on `main` — environmental, not caused by this branch.
- [x] `uv run mypy src/` clean (575 files)
- [x] `uv run python scripts/check-module-size.py` — OK with the ratchet lowered

**118 tests asserting the retired contract were removed** (`test_pipeline_ralph_handoff.py` 51, `test_pipeline_lateral.py` 34 — the UNSTUCK_LATERAL half only, `test_pipeline_evaluate.py` 19, `runtime_probe_envelope` 6, `oscillation_lateral` 5, `surface` 2, plus `test_end_to_end_product.py` and `test_auto_handler_ralph_runtime.py` entirely). Deleting tests is not a neutral act, so: every one of them exercised a code path that no longer exists. No test of surviving behaviour was weakened or deleted.

The interview round loop is untouched; this removes post-run phases, so there is no per-round cost to compare.

## Open decisions for the owner

1. **Deprecated no-op vs immediate removal** of `--complete-product` / `complete_product`. This PR takes the gentler path.
2. **Is synchronous completion worth preserving?** If yes, the alternative is #2121's follower (wait on the same chain) instead of this deletion — and #2121 should then be merged rather than closed. If no, **#2121 should be closed**: its follower will never get a caller.

## Related issues

Refs #2118


Q00:
Both blockers confirmed and fixed.

**1 — the "no-op" was not a no-op.** You are right, and this was the more embarrassing one: the PR body promised that passing the deprecated flag behaves like omitting it, while the code still validated it (rejecting a fresh request under 1,800s, in both `AutoHandler` and `StartAutoHandler`), persisted it onto `state.complete_product`, and merged it back on resume. A compatibility shim that changes behaviour is not a shim.

It is now read once, logged once, and used nowhere — no validation, no persisted state, no resume merging. The two tests that enforced the obsolete rejection are removed along with it, since they were pinning the behaviour the deprecation is supposed to remove.

**2 — the migration message depended on an unrelated clock.** Also correct. The generic deadline gate runs before the retired-phase branch, so an *expired* legacy session in `EVALUATE` / `UNSTUCK_LATERAL` — or `RALPH_HANDOFF` without a reconciliable handle — got `pipeline_timeout` instead of the explanation. Whether an operator learns that the mode was retired should not depend on whether the old session's budget happened to run out.

Retired phases are now answered before deadline enforcement. The contract moved into `auto/retired_phases.py` (phase set, message, guard) rather than growing `pipeline.py`, which is at its ratchet budget — and it gives the retirement rule one home instead of two call sites with a copied string.

Regression: all three retired phases, each resumed with an already-expired deadline, asserting the retirement message and explicitly asserting `pipeline_timeout` is *not* what comes back.

Verified: `tests/unit/auto` + `test_start_auto.py` + `test_auto_command.py` 1,962 passed after the removals; ruff, mypy (576 files), module-size clean.

Q00:
Superseded by #2145, which isolates the remaining watchdog-ordering fix on current main with focused fired-watchdog regressions.

Q00:
Closing as superseded by #2145.

PR #2131, opened 2026-08-15
test(docs): assert durable anchors instead of prose in docs contract tests

Q00:
## Why
Editing README/docs prose (rewording, rewrapping) broke unit tests that asserted verbatim sentences — including assertions pinned to specific line-wrap positions.

## What
- `test_detached_auto_docs.py` (969→518 lines): 8 prose-assertion tests folded into 2 anchor tests (section headings + command/tool invocation strings); the per-terminal-state behavior tests (real `JobResultHandler`/CLI) are kept and decoupled from doc text.
- `test_tui_usage_docs_contract.py`: keeps source-derived contracts (key bindings vs `app.py`/`main.rs`), contract-marker structure, EN/KO parity, and link resolution; drops exact-sentence and table-row prose asserts.
- `test_runtime_skill_capability_docs.py`: runtime list now derived from `runtime_backend_choices()` instead of a hardcoded string; prose asserts dropped, structural anchors kept.

Removing a documented command still fails the tests; rewording around it no longer does.

## Verification
All 26 tests in the three files pass; full unit suite (18,094) green.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -245,7 +245,7 @@ ouroboros setup [OPTIONS]
 
 | Option | Description |
 |--------|-------------|
-| `-r, --runtime TEXT` | Runtime backend to configure. Shipped values: `claude`, `codex`, `opencode`, `hermes`, `gemini`, `goose`, `kiro`, `copilot`, `pi`, `gjc`, `antigravity`, `grok`, `zcode`. Auto-detected if omitted |
+| `-r, --runtime TEXT` | Runtime backend to configure. Shipped values: `claude`, `claude-sdk`, `claude-cli`, `codex`, `opencode`, `hermes`, `gemini`, `goose`, `kiro`, `copilot`, `pi`, `gjc`, `antigravity`, `grok`, `zcode`. Auto-detected if omitted |
 | `--opencode-mode TEXT` | OpenCode integration mode: `plugin` (default, recommended — bridge plugin for interactive sessions) or `subprocess` (headless/CI). Mutually exclusive — see [OpenCode runtime guide](runtime-guides/opencode.md#configuration) |
 | `--non-interactive` | Skip interactive prompts (for scripted installs) |
 | `--mcp-mode TEXT` | Codex MCP config mode: `auto` (default), `preserve`, or `stdio` |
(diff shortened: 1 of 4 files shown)


Q00:
No ourobot review activity 45+ minutes after creation, unlike the 8 sibling PRs (#2132, #2134-#2138, #2141, #2142) which reviewed normally. Requesting review.

Q00:
Addressed all three findings from the ouroboros-agent review, each with a mutation test proving the fix catches the regression:

- **Blocking**: SLT screen mapping now parses the marked table and compares key→screen against the source's tab-index→Screen mapping directly (was token-presence only, so a key swap would pass). KO is checked for (key, alt-key) structural parity with EN.
- **Warning**: CLI setup-runtime assertion is now scoped to the `-r, --runtime` option row itself, not the whole file.
- **Warning**: Restored durable phrase/count anchors for the detached-job lifecycle contract (non-terminal, terminal-and-observable, Next-steps guidance ×3, expired-handle TTL) — these survive rewording but fail if the semantic content is deleted.

Verified each fix with a targeted mutation (swap the key mapping / delete the runtime list from the option row / delete a Next-steps sentence) — all three now fail the corresponding test as expected. Also corrected the PR description's test count: 16 tests in this diff (not 26), which the review's collection also found.

Ready for re-review.

Q00:
Fixes for all three review findings (1 blocking + 2 warnings, each mutation-tested) were pushed 30+ minutes ago with an explanatory comment; no re-review activity yet. Requesting review again.

Q00:
Pushed `ee13b6e58` for exact-head re-review.

Addressed the remaining blockers:
- detached CLI/MCP lifecycle guidance is scoped per status and requires stable completed retrieval plus terminal error/recovery contracts;
- Textual `r` actions are parsed from both EN/KO tables and compared with source bindings;
- SLT docs assert explicit mock, empty DB, DB-open failure fallback, and observer-only lifecycle ownership;
- setup runtime docs now include `claude-sdk` and `claude-cli`, with exact coverage comparisons.

Validation: 16 focused tests passed, Ruff/check/format and `git diff --check` passed.

Q00:
Pushed `HEAD` with the two exact-head review blockers addressed:
- Textual EN screen key/shortcut rows now parse into a source-derived binding map, with normalized EN/KO row parity.
- completed CLI/MCP guidance is status-scoped and must retain the durable `stable completed auto result` semantic anchor.

Focused tests: 16 passed; Ruff and diff checks passed. Please re-review the new exact HEAD.

PR #2133, opened 2026-08-15
test(persistence): halve picker budget benchmark bulk size to 50k

Q00:
## Why
`test_picker_projection_write_and_disk_budgets` inserts bulk events into two DBs, three trials, six parametrizations — the single largest time block in the unit suite (two params alone were 16.6s + 15.9s under xdist).

## What
Extract `_BULK_EVENT_COUNT` and set it to 50k (was 100k). The test asserts ratios, not absolutes, so the contract is unchanged as long as per-row projection cost dominates fixed setup cost.

## Why 50k and not lower
Measured: at 30k the unrelated-event size ratio hits 1.0214 and breaches its 1.02 cap (empty picker tables' fixed pages weigh more at small N). At 50k it measures 1.0127 — comfortable margin, all six params pass, wall time halves.

### DIFF
--- a/tests/unit/persistence/test_picker_indexes.py
+++ b/tests/unit/persistence/test_picker_indexes.py
@@ -1250,6 +1250,12 @@ def test_all_event_inserts_are_owned_by_projection_wrapper() -> None:
     assert insert_sites == ["src/ouroboros/persistence/picker_projection_updates.py"]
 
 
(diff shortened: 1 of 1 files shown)


Q00:
No ourobot review activity 45+ minutes after creation, unlike the 8 sibling PRs (#2132, #2134-#2138, #2141, #2142) which reviewed normally. Requesting review.

PR #2142, opened 2026-08-15
ci(test): gate PRs on the 3.12 floor; full matrix and coverage on push

Q00:
## Why
Every PR ran the full suite three times (3.12/3.13/3.14, ~15 min each job) with coverage instrumentation. Wall-clock PR gate ≈ 15 min and three runners per PR.

## What
- **PRs**: Python 3.12 only (the `requires-python` floor), no `--cov` instrumentation.
- **Pushes to main/develop**: unchanged — full 3.12/3.13/3.14 matrix (3.14 also exercises the litellm-free dependency profile) + coverage artifacts + Codecov upload.

Version-specific regressions move from PR-time to merge-time detection — still strictly before release.

## ⚠️ MERGE COORDINATION REQUIRED
main's branch protection currently requires `Test Python 3.13` and `Test Python 3.14` as status checks. Those contexts will never report on PRs after this change, so **they must be removed from the required checks when (not before) this merges** — otherwise every subsequent PR blocks forever. `Test Python 3.12` stays required.

Expected effect: PR gate ~15 min → ~4–6 min, 3 runners → 1.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.12', '3.13', '3.14']
+        # PRs gate on the 3.12 floor only; pushes to main/develop run the full
+        # matrix (3.14 also exercises the litellm-free dependency profile), so
+        # version-specific regressions are caught at merge time, before release.
+        python-version: ${{ github.event_name == 'pull_request' && fromJSON('["3.12"]') || fromJSON('["3.12", "3.13", "3.14"]') }}
 
     steps:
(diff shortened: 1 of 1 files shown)


Q00:
Addressed both findings from the ouroboros-agent review:
- **Blocking**: flipped `fail_ci_if_error` to `true` on the Codecov upload step — since PRs no longer run coverage, the push job is now the sole publication path, and a silent upload failure there would mean coverage stops updating with no signal.
- **Non-blocking**: renamed the pytest step from "Run pytest with coverage" to "Run pytest" since PR runs no longer include coverage.

Ready for re-review.

PR #2144, opened 2026-08-15
fix(auto): retain Seed QA evaluator diagnostics

Q00:
Supersedes #2103.

The current main branch intentionally treats transient Seed QA evaluator failures as advisory and continues, so the older retry/blocking implementation cannot merge cleanly. Preserve the newer non-blocking contract while carrying a bounded, normalized transport diagnostic in the durable advisory event.

Tests: focused Seed QA advisory/evaluator-failure tests, Ruff, and format checks pass.

## R-run comparison

|---|---|---|---|

Budget compliance: [x] N/A (bounded control-flow/diagnostic change; no per-round execution work)

### DIFF
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -2855,8 +2855,10 @@ class AutoPipeline:
                 return await advisory("evaluator_error", f"Seed QA raised {type(exc).__name__}")
 
             if qa_result.error:
+                detail = " ".join(str(qa_result.error).split())[:200] or "unspecified"
                 return await advisory(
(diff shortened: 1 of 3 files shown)


Q00:
Pushed `HEAD` for exact-head re-review. Seed-QA evaluator failures now cross the durable event/progress boundary only as allowlisted diagnostic categories (rate limit, authentication, connectivity, generic failure); raw provider messages/stderr are never persisted. Added a regression with token-like stderr and a private path proving neither reaches the advisory payload or `last_progress_message`.

Validation: 5 focused tests passed; Ruff/diff checks passed.

PR #2146, opened 2026-08-15
feat(pm): let advisory lanes answer from this session's earlier findings

cohemm:
## Summary

A PM question turn fans out two advisory lanes, and today each one starts from
zero: it greps the roster repositories or runs a measurement, every round, for
every question. Within a single session that repeats work the session already
did — round 5's question is often already answered by a finding round 2 stored.

This tells a lane where its session's earlier answers are, so it can report one
directly instead of re-deriving it. What the parent session receives is
unchanged: the same contracted answer, with no way to tell whether the child
read code, ran a query, or read an earlier round's file. That indifference is
the point — the lane's job is defined by what it returns, not by what it
touched.

## Background

The submissions were never volatile. `ouroboros_submit_fanout_results` publishes
each completed fan-out as a content-addressed artifact under
`<project>/.ouroboros/artifacts/`, and those bodies carry the lane answers whole
— paths, claims, metrics, values. They were simply unreachable: nothing linked a
session to them, so every round re-derived findings that were sitting on disk.

The link turned out to already exist on both sides. A registry record holds
`session_id` and is named by its `fanout_id`; the artifact a submission publishes
holds that same `fanout_id`. Walking that join needs no new index and no second
copy of anything.

## Decisions

**The lookup belongs to the child, not to the server.** The alternative was to
inline earlier findings into the payload. That grows the prompt with every round
whether or not anything is relevant, and it makes the server choose which
findings matter without having read the question. The child has read the
question, so the child chooses — and pays for reading only what it chose. The
parent's context is untouched either way, because the search happens inside a
subagent that is discarded when the round ends.

**Addresses travel; findings do not.** The producer sends a root path and a list
of fan-out ids. Both are values this server wrote, so the request cannot grow
with what children found, and no child-authored text lands on the producer side
— which is the sanitization obligation RFC #1754 deferred alongside durable
result state. That obligation stays deferred rather than being quietly incurred.
A test pins it: a claim submitted in round 1 does not appear in round 2's
prompt.

**Both lanes, not just the code lane.** A measurement looked like the risky one
to reuse until the contract's own position was checked: `data_context` carries
no `observed_at` because "ageing is accepted unconditionally, and the interview
session is the measurement's time envelope". An interview runs for an hour, not
a week. Splitting eligibility per lane would have added a rule the schema had
already decided against.

**No new rules about incompleteness.** An earlier draft had the lane state what
a stored answer settled and what it left open, then investigate the gap. That is
bookkeeping about the lane's own completeness, and completeness is not what this
fan-out guards. What it guards is that an answer cannot lie about its sources,
and every one of those guards holds unchanged here: a claim still carries a real
path and a roster `repo_id`, still gets stamped with this question's identity,
still cannot spell itself pre-confirmed, and the finding still reaches the user
for confirmation before it is recorded. Deliberately not added: a sufficiency
proof, a gap report, a "reused" marker.

**Failure is advisory.** An unreadable registry, a missing directory, or no
project at all costs the child one place it could have looked. It never costs
the user their question: the lanes still run and investigate the way they
always did.

## Contract change

One semantic clarification, no shape change: `examined` entries now mean "the
repositories this answer stands on", where the reading may have happened on an
earlier question of the same session. The session was already the time envelope
here, so this states the scope that was already in force rather than widening
it. What deliberately gains no field is *which* turn did the reading.

## Scope

Wired for `ouroboros_pm_interview` only. The mechanism is generic and renders by
presence, so the interview joins by passing `project_dir` — but it is a
different question shape with six lanes, and that is worth its own look.

## Test plan

- [x] `uv run pytest tests/unit tests/conformance -q` passes
- [x] `uv run mypy src/ouroboros/mcp/tools/` clean

New coverage:

- a later round is told where this session's earlier answers are, on both lanes
- the first question of a session is told nothing (an empty place costs a tool
  call), and no turn is sent to fetch the answer it is being asked to write
- another session's ids are never offered, and neither is another kind's
- the producer hands over addresses and never what a child found
- a lane with no project to resolve against is told nothing and still runs

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -2486,6 +2486,11 @@ def create_ouroboros_server(
             agent_runtime_backend=interview_runtime_backend,
             opencode_mode=opencode_mode,
             fanout_registry=fanout_registry,
+            # The same workspace ``create_fanout_handlers`` builds the artifact
+            # store from, below. A lane is told where this session's earlier
+            # answers are, and one composition passing a different path than the
+            # one that wrote them would send it somewhere empty.
+            project_dir=effective_cwd,
         ),
         BrownfieldHandler(_store=brownfield_store),
         evaluate_handler,
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -1610,7 +1610,7 @@ def create_ouroboros_server(
     )
     from ouroboros.mcp.tools.evaluation_composition import create_shared_evaluation_handlers
     from ouroboros.mcp.tools.fanout import FanoutRegistry
-    from ouroboros.mcp.tools.pm_handler import PMInterviewHandler
+    from ouroboros.mcp.tools.interview_composition import create_interview_handlers
     from ouroboros.mcp.tools.qa import QAHandler
     from ouroboros.mcp.tools.registry import ToolRegistry
     from ouroboros.mcp.tools.seed_handoff import SeedHandoffRegistry
@@ -2475,27 +2475,20 @@ def create_ouroboros_server(
             opencode_mode=opencode_mode,
         ),
         MeasureDriftHandler(event_store=event_store),
-        InterviewHandler(
+        # ``workspace`` is the same value ``create_fanout_handlers`` builds the
+        # artifact store from below, so the composition that writes fan-out
+        # results and the one that points a lane at them cannot name different
+        # directories.
+        *create_interview_handlers(
             interview_engine=interview_engine,
+            state_dir=state_dir_path,
+            workspace=effective_cwd,
             event_store=event_store,
+            fanout_registry=fanout_registry,
             llm_backend=interview_llm_backend,
             agent_runtime_backend=interview_runtime_backend,
             opencode_mode=opencode_mode,
-            fanout_registry=fanout_registry,
             suppress_tool_use_prompt_cues=interview_envelope_sealed,
-        ),
-        PMInterviewHandler(
-            data_dir=state_dir_path,
-            llm_backend=interview_llm_backend,
-            event_store=event_store,
-            agent_runtime_backend=interview_runtime_backend,
-            opencode_mode=opencode_mode,
-            fanout_registry=fanout_registry,
-            # The same workspace ``create_fanout_handlers`` builds the artifact
-            # store from, below. A lane is told where this session's earlier
-            # answers are, and one composition passing a different path than the
-            # one that wrote them would send it somewhere empty.
-            project_dir=effective_cwd,
         ),
         BrownfieldHandler(_store=brownfield_store),
         evaluate_handler,
--- a/src/ouroboros/mcp/tools/definitions.py
+++ b/src/ouroboros/mcp/tools/definitions.py
@@ -416,6 +416,9 @@ def evolve_rewind_handler() -> EvolveRewindHandler:
 # Tool handler tuple type and factory
 # ---------------------------------------------------------------------------
 from ouroboros.mcp.tools.brownfield_handler import BrownfieldHandler  # noqa: E402
+from ouroboros.mcp.tools.interview_composition import (  # noqa: E402
+    create_interview_handlers,
+)
 from ouroboros.mcp.tools.pm_handler import PMInterviewHandler  # noqa: E402
 
 OuroborosToolHandlers = tuple[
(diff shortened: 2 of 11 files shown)


Q00:
Draft review: the lower-level reuse implementation and focused tests look sound, but the shipped MCP composition still appears to construct `PMInterviewHandler` without passing `effective_cwd`/`project_dir`. That leaves production requests without the artifact root needed for prior-finding lookup even though direct helper tests pass. Please wire the same workspace path used by the artifact store into the handler and add a `create_ouroboros_server()` two-turn PM advisory regression before marking ready. I am leaving this as the single draft review comment.

cohemm:
Confirmed and fixed — you were right, and the miss was worse than a missing argument.

`definitions.py` was the only composition I wired, and the one that ships is `adapter.py`, which built `PMInterviewHandler` with no workspace at all. So on the path that actually runs, `project_dir` was always `None`, `_prior_findings_root` always returned `None`, and no lane was ever told anything. The feature was dead in production while every lane-level test stayed green — because each of those tests builds its own request, which is exactly the shape that cannot notice a composition that never passes one.

`PMInterviewHandler` now takes `effective_cwd`, the same value `create_fanout_handlers` builds the artifact store from a few lines below. Taking the one value rather than re-deriving a path is deliberate: the composition that writes the artifacts and the composition that points a child at them must not be able to name different directories.

The regression is the two-turn one you asked for. It builds the server through `create_ouroboros_server(project_dir=...)`, pulls the `PMInterviewHandler` the server registered, drives two question turns through it, and asserts the second turn's lanes carry the first turn's fan-out id and the root resolved by `ContentAddressedArtifactStore.for_project` itself — not a path the test spells out, since two sides agreeing a path exists is not the same as agreeing which one. Verified it fails on the parent commit and passes here.

`tests/unit/mcp` + `tests/unit/orchestrator` (7290 passed), ruff, and mypy are clean.

cohemm:
Followed the finding down to where it could no longer happen, rather than only to where it did. Two commits, and the second one exists because the first ran into a gate.

## What the finding actually was

A fan-out has two sides. The submit handler publishes each completed run into a project-local artifact store built from a workspace path; the advisory producer tells a child where those artifacts landed. They were wired from two separate arguments, in two composition roots, by hand — so a root could wire one and not the other, and `create_ouroboros_server` did: it gave `effective_cwd` to `create_fanout_handlers` and nothing to `PMInterviewHandler`, twenty lines apart in the same function.

The half is silent by construction. The producer computed this session's prior fan-out ids correctly and then dropped them, because the pair is only attached when both halves are present. A producer with a registry and no workspace emits exactly what a producer on the first question of a session emits: no block, no log line, no error. Every session read as permanently round one.

And no test could see it, because every lane-level test builds its own request and therefore never travels the composition that omitted one. That is the shape of the miss, not an oversight in coverage depth.

## 1. A test across both roots

`tests/unit/mcp/test_fanout_composition_wiring.py` walks `create_ouroboros_server` and `get_ouroboros_tools` and asserts the producer's artifact root **equals** the root the submit handler would publish into. Equality rather than "both are set": two roots agreeing a path exists is not agreeing on *which* one, and a producer aimed at a neighbouring directory reads exactly like a session with no history.

The invariant is conditional on purpose. A root built with no workspace configures no store, so it has nothing to point a child at — a third test holds that case open so this cannot later be tightened into demanding a workspace everywhere, which would be a rule with no defect behind it.

## 2. One shared wiring, because the ratchet asked the right question

`enforce-module-size` then failed: `adapter.py` is grandfathered at 2612 lines and rule 3 forbids raising a budget, so the gate's instruction is to put new code in a new module instead of extending the god module. The mechanical move would have satisfied CI and left the defect exactly where it was, so the extraction is the one the defect argues for instead.

`interview_composition.create_interview_handlers` is now the only place the pair is spelled, and **both roots call it**. Where a root genuinely has less to give — no event store, no engine, no workspace — it passes less and the handlers' defaults absorb it. The difference between the roots is real and kept; the wiring contract is not, and is no longer written twice. `adapter.py` drops to 2610.

Both verifications were done by removing the fix and re-running: the two-turn PM regression and the cross-root test each fail without their change.

## What is left, and what is not being done

Still open as #2149: the handler fields are two, so a future caller constructing one directly can still spell "registry, no workspace". Bundling them into one value removes that spelling entirely. It changes the signatures of the remaining advisory producers and `create_fanout_handlers`, which is why it is not in this PR.

Merging the two composition roots stays a non-goal, measured rather than asserted — they share 28 handler classes but inject different services (`event_store`, `interview_engine`, `data_dir`, `session_signal_hub` on one side; `mcp_manager`/`mcp_tool_prefix` on the other). Merging would mean the in-process path dragging in what it has no equivalent for, or the server path losing it. What was duplicated is the wiring contract, not the list — so that is what became one function.

One test outside this feature moved with it: the opencode-mode adapter test patches construction sites by module path, and one of those sites is now `interview_composition`.

`tests/unit/mcp` + `tests/unit/orchestrator` (7322 passed), mypy, ruff, format, and the module-size ratchet against `origin/main` are all clean, and the branch is current with `origin/main`.

cohemm:
Closing without merging. Being rebuilt against RFC #2153.

The feature stands — a lane that can answer from an earlier finding should not re-derive it — but the boundary this PR was built on does not. It scoped reuse to the interview session, and the RFC's decision is that the session was never the meaningful boundary: what a lane reports changes on the system's clock, not on the clock of whoever is being interviewed. Session scoping also guaranteed that every session's first question missed, which is most of the reuse there is.

Rather than rework this branch, it is being restarted from `main` against the RFC. Most of what is here exists to answer "which findings are this session's" cheaply, and that question is removed rather than answered.

What survives is recorded on the RFC rather than in this branch: the two implementation traps (a wiring that reached only one composition root; one address derived twice at two moments), the failure mode they share (both look exactly like "no prior findings exist"), and the measurement that should have been taken before any of it.

The branch is left in place; nothing here is deleted.

PR #2147, opened 2026-08-16
feat(cli): surface zcode CLI model-config readiness in setup and health

### LINKED ISSUE #2148: zcode setup/health pass while the headless CLI lacks a model provider config
## What happened

On a machine with the ZCode desktop app installed (and signed in), every preflight check passes:

```
$ ouroboros setup --runtime zcode --non-interactive
╭────────────────────────── Success ───────────────────╮
│ Configured zcode runtime (CLI: .../zcode.cjs)        │
╰──────────────────────────────────────────────────────╯

$ ouroboros status health
│ Runtime backend │   ok    │ zcode: /Applications/ZCode.app/...  │
│ Credentials     │   ok    │ zcode uses local CLI authentication │
```

The first workflow step that spawns the Zcode CLI then fails mid-execution:

```text
Error: Model config is missing. Create /Users/<you>/.zcode/cli/config.json
with an explicit model provider before running ZCode.
```

## Root cause

The Zcode CLI resolves its model from `~/.zcode/cli/config.json` on every headless run. The desktop app's model selection (`~/.zcode/v2/config.json`) never creates that file, and the shared OAuth credentials (`~/.zcode/v2/credentials.json`) make the health row's "local CLI authentication" text plausible while the effective model reference is missing — so nothing catches it before launch.

Two schema details make this easy to hit and hard to debug (measured from Zcode CLI 0.15.x and 0.16.x):

1. `model.main` must be the `"provider-id/model-id"` **string**; an object entry is silently dropped by the CLI's parser, which then reports the same error — an apparently complete config keeps failing.
2. The referenced provider id must have an entry in the same file's `provider` registry (`kind`, `options.baseURL`, `options.apiKey`); coding-plan endpoints are not hardcoded in the CLI.

## Expected

- `ouroboros setup --runtime zcode` warns when the CLI has no effective `model.main`, with actionable guidance.
- `ouroboros status health` surfaces the gap (runtime backend row / zcode LLM backend credentials row) instead of reporting unconditional `ok`.
- The runtime guide documents the model provider config schema and the failure mode.

## Environment

- ouroboros-ai 0.51.5 (uv tool install)
- ZCode desktop app + CLI 0.16.3 (macOS ARM), Z.ai coding plan signed in
- No `~/.zcode/cli/config.json` before the first setup

tmdgusya:
Fixes #2148

## Problem

On a machine with the ZCode desktop app installed, `ouroboros setup --runtime zcode` succeeds and `ouroboros status health` reports every check green:

```
│ Runtime backend │   ok    │ zcode: /Applications/ZCode.app/.../zcode.cjs │
│ Credentials     │   ok    │ zcode uses local CLI authentication          │
```

...but the first workflow step fails mid-execution, after Ouroboros has already assembled the prompt and spawned the CLI:

```text
Error: Model config is missing. Create /Users/<you>/.zcode/cli/config.json
with an explicit model provider before running ZCode.
```

The Zcode CLI resolves its model from `~/.zcode/cli/config.json` on every headless run. The desktop app's model selection (in `~/.zcode/v2/config.json`) does not create that file, and OAuth credentials being shared (`~/.zcode/v2/credentials.json`) makes the health row's "local CLI authentication" claim plausible while the effective model reference is still missing.

Two schema gotchas make this easy to hit and hard to debug (measured from Zcode CLI 0.15.x and 0.16.x):

1. `model.main` must be the `"provider-id/model-id"` **string**. An object entry is silently dropped by the CLI's config parser, which then reports the same "Model config is missing" error — an apparently complete config keeps failing.
2. The provider referenced by `model.main` must have an entry in the same file's `provider` registry (`kind`, `options.baseURL`, `options.apiKey`); the CLI does not hardcode coding-plan endpoints.

## Change

- **`ouroboros.config.zcode_model_config`** — new probe mirroring the CLI's acceptance rules: resolves the default config path, requires a string `model`/`model.main` reference in `provider-id/model-id` form, and a matching `provider` registry entry with `kind` or `options.baseURL`. Returns a frozen status with a detail string that names the exact gap (including the object-form case).
- **`ouroboros setup --runtime zcode`** — after writing the config, prints a warning with the failing detail and a copy-paste example config when the probe fails.
- **`ouroboros status health`** — the zcode runtime-backend row downgrades to `warning` with the probe detail, and the zcode LLM-backend credentials row runs the probe instead of unconditionally reporting `ok`.
- **`docs/runtime-guides/zcode.md`** — prerequisites row, new "Model provider configuration" section (string-ref schema, registry example, copy-from-desktop note), and a troubleshooting entry for the exact CLI error.

## Testing

- `tests/unit/config/test_zcode_model_config.py` — 10 cases covering missing file, invalid JSON, non-object root, missing reference, object-form `model.main`, unparseable ref, missing provider entry, entry without `kind`/`baseURL`, and both passing shapes.
- `tests/unit/cli/test_status_health_zcode.py` — 6 cases for the credentials row and runtime-backend row wiring, including non-zcode backends never running the probe.
- `uv run pytest tests/unit/cli tests/unit/config -q`: 2549 passed; the 22 failures are pre-existing on clean `main` (codex MCP-registration and mcp-doctor tests fail identically without this branch).
- Verified the rendered setup warning end to end against a fake $HOME.

### DIFF
--- a/docs/runtime-guides/zcode.md
+++ b/docs/runtime-guides/zcode.md
@@ -16,6 +16,7 @@ installed ZCode coding agent. The runtime supports either the macOS app-bundle
 | --- | --- |
 | ZCode desktop app or `zcode` executable | Provides the coding-agent runtime |
 | A configured Z.ai provider/model | Zcode reads provider and model selection from its own config |
+| `~/.zcode/cli/config.json` with an effective `model.main` | The headless CLI exits with `Model config is missing` without it; see [Model provider configuration](#model-provider-configuration) |
 | Compatible Node.js for a standalone `.cjs` path | Official app bundles use their bundled Electron/Node runtime; standalone scripts use the system Node |
 | Ouroboros base package | No provider-specific Python extra is required |
 
@@ -25,6 +26,10 @@ installed ZCode coding agent. The runtime supports either the macOS app-bundle
 ouroboros setup --runtime zcode
 ouroboros run workflow seed.yaml --runtime zcode
 ```
+
+Setup warns, and `ouroboros status health` reports a warning, when
+`~/.zcode/cli/config.json` has no effective `model.main` — the CLI would
+otherwise fail only after Ouroboros has already spawned it.
 
 If setup cannot find ZCode automatically, configure one of:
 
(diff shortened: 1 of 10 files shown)


Q00:
Addressed both exact-head review blockers in 97fd772e1:

- provider readiness now requires non-empty string `kind`, `options.baseURL`, and `options.apiKey`
- status health now reuses `get_zcode_cli_path()` precedence and accepts readable `.cjs`/`.js`/`.mjs` entry scripts exactly as the runtime does
- added incomplete-provider, environment override, and non-executable readable-script regressions

Validation: 40 focused/related tests passed; Ruff, diff, and module-size checks passed. Please re-review this exact HEAD.

Q00:
Addressed the exact-head launcher-readiness blocker in a80b67742:

- added a side-effect-free shared command-prefix resolver that validates app-bundle metadata, Info.plist, bundled Electron executability, and standalone Node availability
- made both ZCode runtime launch paths and status health use that shared resolver
- health now returns the actionable launcher error instead of reporting a readable-but-unlaunchable script as ready
- added malformed app-bundle and missing-Node regressions
- corrected the stale model-provider docstring

Validation: 134 focused/related tests plus 10 loader regressions passed; Ruff, diff, and module-size checks passed. Please re-review this exact HEAD.

Q00:
Addressed the exact-head precedence blocker in the new HEAD:

- launcher validation failure on the runtime-selected highest-priority ZCode path now returns immediately
- health no longer falls through from an invalid environment override to a lower-priority configured executable that production runtime would never select
- added the mixed invalid-override/valid-config regression

Validation: 145 related tests passed; Ruff, format, diff, and module-size checks passed. Please re-review this exact HEAD.

PR #2151, opened 2026-08-16
refactor(fanout): let a record's filename say which session owns it

### LINKED ISSUE #2150: refactor(fanout): let a record's filename answer which session it belongs to
**Blocked on #2146** — that PR adds the first caller that asks "which fan-outs are this session's". Please do not start until it merges. Related to #2149, but independent: that one is about wiring, this one is about the record store's layout.

## Problem

Asking "which fan-outs belong to session X" currently means opening every record in the registry directory, because the only field that answers it lives inside the file.

Two things make that worse than it sounds:

- **The directory is global while artifacts are not.** Records go to `~/.ouroboros/data/fanout/` (`FanoutRegistry()` in `definitions.py`, `state_dir_path / "fanout"` in `adapter.py`, where the state dir defaults to `get_config_dir() / "data"`). The artifact bodies they point at are per-project, under `<project>/.ouroboros/artifacts/`. So the server knows the workspace, writes bodies there, and still keeps the index of them in a directory shared by every project on the machine.
- **A record carries no project.** Its fields are `fanout_id`, `kind`, `session_id`, `correlation_key`, `expected_keys`, `required_keys`, `synthesizer_input`, `question_identity`. Knowing the workspace does not narrow the search, because nothing in the record mentions one.

Measured on the author's machine: **192 records across 155 distinct sessions**, median one record per session, oldest from 24 July. Records are never pruned — artifacts have a 90-day `ouroboros artifacts prune`, records have nothing — so this only grows.

## What #2146 already did, and why it is not the fix

- The lookup is gated on the caller having an artifact root, so only the PM producer pays it; the interview path does not query the registry at all.
- The scan sorts by mtime (a stat, not a read) and stops once it has the cap, so a live session opens roughly a dozen files instead of all of them.

That makes the common case cheap. It does not remove the walk: a resumed session whose records are no longer the newest still reads through everything ahead of them, and the directory grows without bound.

## Proposal

Put the answer in the filename:

```
fanout/fanout_<id>.json                    →  fanout/<session_id>__fanout_<id>.json
```

`session_fanout_ids` becomes `glob(f"{session_id}__*.json")` — **zero file reads**, and a cost that does not depend on how much history the machine has kept.

Notes for whoever picks this up:

- `load(fanout_id)` resolves by id alone (submission does not know the session's directory), so it should try the flat path first and fall back to a `*__{fanout_id}.json` glob. That fallback is also what keeps the 192 existing records loadable — no migration required, and old records simply stop matching a session prefix search, which is correct since they belong to other sessions.
- A session id in a path needs the guard `fanout_id` already has. `_FANOUT_ID_RE` exists precisely so a separator or a `..` segment cannot be spelled; the session segment needs the same treatment before it is joined to a directory.
- `register` and `load` are shared by **every** fan-out kind — interview advisory, lateral persona panel, code investigation. #2146 established that the interview path is byte-identical to `main`; this change touches the machinery underneath it, so it needs the same evidence.

## Alternative considered

Partitioning into `fanout/<project>/<session>/` subdirectories. Rejected: `load(fanout_id)` receives only an id, so every load would have to search across partitions or grow a parameter it does not have today. A filename prefix keeps one flat directory and one lookup rule.

## Adjacent, not included

Records have no retention policy at all. Whether they should get one (and whether it should follow the artifacts' 90 days) is a separate decision from how they are addressed, and folding it in here would mix a lifecycle question into a lookup change.

cohemm:
Closes #2150.

## Summary

Asking "which fan-outs belong to this session?" meant opening every record in the registry directory, because the only field that answers it lived inside the file. This puts the session in the filename, so the question is a directory listing instead.

## Background

Two things made the read cost more than it looks:

- **The directory is global while the artifacts it indexes are not.** Records go to `~/.ouroboros/data/fanout/` (`FanoutRegistry()` in `definitions.py`, `state_dir_path / "fanout"` in `adapter.py`, where the state dir defaults to `get_config_dir() / "data"`). The bodies they point at are per-project, under `<project>/.ouroboros/artifacts/`. So the server knows the workspace, writes bodies there, and still keeps the index of them in one directory shared by every project on the machine.
- **A record carries no project.** Its fields are `fanout_id`, `kind`, `session_id`, `correlation_key`, `expected_keys`, `required_keys`, `synthesizer_input`, `question_identity`. Knowing the workspace narrows nothing.

Measured while writing this: **192 records across 155 distinct sessions**, median one record per session, oldest from 24 July — records have no retention policy, unlike artifacts' 90-day prune, so the directory only grows. A lookup that reads to find out therefore costs the whole history to learn what a listing could have said.

## What changed

`fanout_<id>.json` → `<session_id>__fanout_<id>.json`, and `session_fanout_ids(session_id, kind=...)` globs that prefix. Records outside the session are never opened; the matches are read only to check `kind`, since one session can run a persona panel beside its advisory lanes and an id that passed the session filter alone would send a caller to read one as though it were the other.

Three decisions worth stating:

**The session segment is held to the id's alphabet.** `_FANOUT_ID_RE` exists so a separator, a parent segment or an absolute form has no spelling — a value joined to a directory must be constrained before the join, not inspected after it. The session is now joined the same way, so it gets the same constraint. A session that cannot be spelled falls back to the flat name and still registers: refusing would cost a turn its re-entry over a filename.

**`load` finds either shape.** A submission carries an id and nothing else, so the read path tries the flat name and then locates a prefixed one by listing. That is what makes this need no migration — the 192 records already on disk stay redeemable, and no id stops working because the name it was written under changed.

**A session lookup does not find pre-layout records.** Their filename cannot answer the question, and reading every flat file to ask would restore exactly the cost this removes. A session that predates the layout is a finished one; the degradation for a session straddling the change is that its earlier rounds are not offered — fewer results, never another session's.

## Ordering

#2150 was filed as blocked on #2146, which is the first caller that asks this question. The order is inverted deliberately: the store capability lands first and the feature consumes it, which keeps the layout change reviewable on its own and lets #2146 drop its own scan in favour of this one.

## Tests

Four tests that hard-coded the flat name now locate records by listing — that is not incidental churn, it is the same defect in test form: a name rebuilt from an id is a second place the layout is written down, and the one that goes stale first.

New coverage:

- a record's name says which session owns it
- a session lookup opens only that session's records — proven by counting `load` calls with 3 of the session's records among 40 others
- kind and session are both load-bearing filters
- a record written before the session was in its name still loads, and is not offered by a session lookup
- a session that cannot be spelled falls back to the flat name, stays redeemable, and lands nowhere but the registry directory
- newest first, capped

## Test plan

- [x] `uv run pytest tests/unit tests/conformance -q`
- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check src/ tests/`

### DIFF
--- a/src/ouroboros/mcp/tools/fanout.py
+++ b/src/ouroboros/mcp/tools/fanout.py
@@ -73,6 +73,23 @@ _DEFAULT_FANOUT_DIR = Path.home() / ".ouroboros" / "data" / "fanout"
 #: ``/tmp/forged.json``, and a check placed after that join is already too late.
 _FANOUT_ID_RE = re.compile(r"[A-Za-z0-9_-]{1,128}")
 
+#: A session id may be written into a record's filename, so it is constrained
+#: exactly as ``_FANOUT_ID_RE`` constrains the id: both end up joined to the
+#: registry directory, and a separator, a parent segment or an absolute form
+#: must be unspellable rather than detected after the join.
+_SESSION_SEGMENT_RE = re.compile(r"[A-Za-z0-9_-]{1,128}")
+
+#: Between the session and the id. Doubled so the boundary survives both sides
+#: containing single underscores, which they do -- ``interview_20260809_055623``
+#: and ``fanout_<hex>``. The id is the tail, so a name is read from the right
+#: whatever the session happens to contain.
+_SESSION_SEPARATOR = "__"
+
+#: How many of a session's ids one lookup returns. A cap rather than the whole
+#: history: the list is read by whatever asked for it, and an hour-long session
+#: should not hand its caller a longer one.
+_SESSION_FANOUT_ID_LIMIT = 12
+
 # Fan-out re-entry kinds — each routes to one revived synthesizer.
 FANOUT_KIND_LATERAL_PERSONA_PANEL = "lateral_persona_panel"
 FANOUT_KIND_CODE_INVESTIGATION = "code_investigation"
@@ -216,7 +233,7 @@ class FanoutRegistry:
             return
         self._dir = directory
 
-    def _path(self, fanout_id: str) -> Path | None:
+    def _path(self, fanout_id: str, session_id: str = "") -> Path | None:
         """Return the record path for ``fanout_id``, or ``None`` if it is not one.
 
         The registry directory is the whole of where a record may live, so an id
@@ -224,9 +241,23 @@ class FanoutRegistry:
         rather than raising keeps the existing read contract: an unredeemable id
         is reported as an unknown fan-out, which is what a caller holding a
         forged one should learn.
+
+        A spellable ``session_id`` is written into the name. One directory holds
+        every session this machine has run, and the session is otherwise only
+        inside the file, so "which of these are mine" could be answered only by
+        opening all of them. The name is where that answer belongs: it makes the
+        question a directory listing rather than a read of the whole history.
+
+        The session segment is held to the same alphabet as the id and for the
+        same reason -- both are joined to a directory, so a separator or a
+        parent segment must have no spelling rather than being detected after
+        the join. A session that cannot be spelled falls back to the flat name,
+        which is also the shape every record written before this had.
         """
         if not _FANOUT_ID_RE.fullmatch(fanout_id):
             return None
+        if session_id and _SESSION_SEGMENT_RE.fullmatch(session_id):
+            return self._dir / f"{session_id}{_SESSION_SEPARATOR}{fanout_id}.json"
         return self._dir / f"{fanout_id}.json"
 
     def register(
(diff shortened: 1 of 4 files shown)


Q00:
Draft review: `session_fanout_ids()` says it reads the fan-out id from the right, but the implementation uses `stem.partition("__")`, which splits at the first separator. `_SESSION_SEGMENT_RE` permits consecutive underscores, so a valid session such as `sess__nested` writes `sess__nested__fanout_abc.json`; lookup then passes `nested__fanout_abc` to `load()` and silently omits its own record. Please parse the known prefix (`f"{session_id}__"`) or use `rpartition("__")`, and add a regression with a valid session id containing `__`. Otherwise the filename index is a solid direction and the legacy flat-record fallback is appropriately fail-closed.

cohemm:
Closing without merging. Superseded by RFC #2153.

The lookup this optimises is removed by that decision rather than made cheaper: a lane reads recent findings for the project, so "which of these are this session's" stops being a question the store has to answer.

Worth recording why this was the wrong trade even on its own terms. The cost it removed was measured after the review: **0.68 ms typical, 12.7 ms worst case** against 194 records, on a turn that spends seconds in subagents. What it paid was the registry's identity contract — the filename stopped being the id, which silently removed the uniqueness the filesystem had been enforcing for free, and the review found two contract breaks that did not exist before the change. The remedy would have added a scan-and-delete protocol to the write path shared by every fan-out kind.

The two blockers were real and are not being dismissed; they are simply attached to an approach that is no longer taken.

Issue #2153, opened 2026-08-16
RFC: advisory lanes read recent findings, not this session's


cohemm:
## The decision

**A lane may answer from a finding another lane produced recently, regardless of which session produced it.** "Recently" means on the order of a day, not the interview session.

What may be reused is what the two lanes reporting on the system produced — `code_context` and `data_context` — and nothing else. Which tool asks does not enter into it: a fact about the system is the same fact whichever interview needed it, so both the PM interview and the Socratic one read from the same recent findings.

## What is actually being claimed

Two assumptions, and they pull against each other. The decision is where they balance, so both are stated before anything that follows from them — and neither is a proof.

**Assumption A — the same work gets done repeatedly, so findings repeat.**

An interview is not a novel act. The same person, working on the same product, asks about the same subsystems: what the booking rule is, how a lapse is handled, how many users are on which plan. So the questions across sessions are similar, and if the questions are similar then what the lanes go and find is similar too. That is the entire source of value here — reuse is worth having only if a finding produced for one question is a plausible answer to a later one.

This is what makes the session the wrong boundary. Scoping to the session does not merely lose a little; it guarantees that every session's first question misses, and it throws away precisely the repetition that a day of work produces.

*Falsified if*: findings are rarely reused after this lands. That would mean questions vary more than assumed, and the mechanism is paying for a similarity that is not there.

**Assumption B — what a finding describes keeps moving, so a finding decays.**

Code and data flow. A commit lands, rows accumulate, a policy is edited. A lane's answer is a statement about the system *right now*, and every hour it is a slightly worse one. So reuse cannot be unbounded: what makes a finding useful (it was already established) is the same thing that makes it dangerous (it was established at a moment that is no longer this one).

*Falsified if*: findings are observed contradicting the same day's code, or a decision turns on a movement in a measurement that happened inside the window.

**The window is where A and B balance.**

A wants it wide — the wider it is, the more of the day's repetition is reachable. B wants it narrow — the narrower it is, the less drift a finding can have accumulated. A day is the estimate of where those meet for these two lanes: long enough that a second session on the same repository starts warm rather than cold, short enough that a product-level policy claim or an aggregate has not moved under it.

Stating it as a balance rather than as a safety margin is deliberate. If it were only "nothing goes stale in a day", then any argument that things are stable would be an argument for a wider window. The reason to keep it near a day is that widening it buys progressively less — the repetition is concentrated in recent work — while the drift it admits keeps growing.

[...]

PR #2154, opened 2026-08-16
feat(advisory): let a lane answer from what was recently found here

cohemm:
Implements RFC #2153.

## Summary

An advisory lane re-derived everything on every question: it grepped the roster repositories or ran a measurement, every round, for every question, including questions a lane had already answered an hour earlier. This lets a lane read what has recently been found in the same project and answer from it, so it investigates only what is not already there.

What the parent session receives is unchanged — the same contracted answer, with no way to tell whether the child read code, ran a query, or opened a file. That indifference is the point: a lane's job is defined by what it returns, not by what it touched.

## The decision this implements

The RFC's boundary is **recency, not the session**. What these two lanes report changes on the system's clock — a commit lands, data flows — not on the clock of whoever is being interviewed. So a finding from another session about the same project is as good as one from this session, and scoping to the session had a cost that only looked free: every session's first question was guaranteed to miss.

**What may be reused** is the output of two lanes, and that list is closed — `code_context` and `data_context` report on the system, while a contrarian, an answer drafter and an architecture reviewer produce reasoning about one question, which is not a fact that keeps. The boundary is on what is offered, not on who may read it: a lane seeing what the code does today is seeing the same evidence its question is being asked against. A web lane is excluded rather than pending — the window below was derived from how a repository drifts, and nothing about this project bounds how fast an external fact turns over.

**Why a day, and why a window exists at all.** Two forces pull against each other. The same person asks about the same subsystems all day, so findings repeat — that is the entire source of value, and it wants the window wide. But code and data keep moving, so every hour a finding is a slightly worse statement about now — that wants it narrow. A day is where they balance for these two lanes. Stated as a balance rather than a safety margin deliberately: if it were only "nothing goes stale in a day", any argument that things are stable would be an argument for widening it.

**Why a stale finding is survivable.** Both eligible lanes end at a person — a `code_context` finding is shown for confirmation before it is recorded as an adopted fact, and a `data_context` measurement is shown beside the question and never recorded at all. So what checks whether a reused finding still holds is the reader, not a rule, which is why no rule was added.

## Shape

The RFC deliberately left open *how* a lane reaches a finding. This takes the shortest path available: completed fan-outs are already published into a project-local artifact store, so "what has been found here recently" is a listing of that directory, handed to the lane as paths.

That means **no index, no new state, and no registry involvement**. Nothing has to be kept in step with the store, because the store is already organised the way a lane needs to read it — per project, newest last, expiring on its own schedule.

Three properties worth stating:

**Paths, not findings.** The producer sends file paths and never their contents. Inlining what was found would grow the prompt with every round and would make this server choose, without having read the question, which findings matter — the child has read the question, so the child chooses and pays for reading only what it chose. It also keeps the producing side free of anything a child wrote, which this fan-out owes independently of this feature. A test pins it.

**What the store's record is asked, and what it is not.** The record answers *when* a publication happened. That is read for a reason with no adversary in it: a content-addressed store reuses a body when the same result is published again, so the body's modified time reports when those bytes first appeared rather than when the finding was established. Bodies are then read through the store's own `fetch`, which is bounded, contained and verified against its address — not as a boundary, but because re-implementing a read the store already does correctly is how two earlier attempts went wrong. The path handed over names the body that fetch verified, so one value has one source.

**The project workspace is trusted input**, and this PR claims no provenance. A lane's whole job is to read this project's code and report what it says, so anything able to write into the store can more easily edit the source a lane would cite — with a real path the user then confirms. RFC #2153 states this outright. The guards here exist against accident, and their job is to cost the shortcut rather than the user's question.

**A rolling window, not a calendar day.** The RFC fixes "on the order of a day" and leaves the exact value to implementation, with one property that is not tuning: the window must not have an edge that silently empties mid-session. A calendar boundary empties at midnight into exactly the state that means "nothing has been found here" — the failure this mechanism keeps producing — so the window rolls.

## The two traps recorded on the RFC

Both were found the hard way in earlier attempts, and each has a test that fails without its fix (verified by removing the fix and re-running).

**A wiring that reaches only one composition root.** The producer is built in two places, and the first attempt wired the workspace into one of them. Every lane-level test stayed green, because each builds its own request and so never travels a composition. Both roots now go through one factory, and a test walks both and compares by value.

**One address derived twice.** The store resolved the workspace when it was constructed; a producer deriving the same path when a question is asked would resolve it at a different moment, so a relative workspace plus a change of process directory in between aims the reader at a store the writer never used. The producer takes the store's own resolved root instead. A test changes directory between composition and reading.

## What is deliberately absent

No bookkeeping about incompleteness — no proving a stored finding sufficient, no reporting what a reuse left unsettled, no marking an answer as reused. What the answer contracts guard is that an answer cannot lie about its sources, and those guards hold whatever the child read. Completeness was never guarded and does not start being guarded here.

## Carried consequence

A finding may come from a session that chose different repositories. Evidence from outside the current roster is already rejected at submission, so the failure is loud rather than silent, but it costs a round — so the lane is told where it is deciding what to read, rather than discovering it when its answer is refused.

`data_context` carries no `observed_at` on the reasoning that the session *is* the time envelope. That reasoning is superseded by the RFC and the contract's wording is corrected with it.

## Test plan

- [x] `uv run pytest tests/unit tests/conformance -q` — 20321 passed, 88 skipped
- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check src/ tests/`

New coverage:

- a finding from another session is offered, on both lanes
- a finding older than the window is not
- a project with nothing recent is told nothing, and a caller with no root still gets its lanes
- the lane is told the roster does not travel with the findings
- the producer hands over paths and never what a child found
- the store's own bookkeeping is not mistaken for a finding
- a body that is not what its name says is not offered
- a symlink to another project's valid artifact is not offered — it hashes correctly and carries an eligible lane, so only the symlink guard excludes it
- another fan-out kind is not offered, and neither is an advisory carrying no eligible lane
- a mixed body is offered, and the prompt names which entries may be read
- both composition roots point the producer at the store they write to
- a relative workspace survives a change of working directory

### DIFF
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -1606,11 +1606,10 @@ def create_ouroboros_server(
         StartEvolveStepHandler,
         StartExecuteSeedHandler,
         StartRalphHandler,
-        create_fanout_handlers,
     )
     from ouroboros.mcp.tools.evaluation_composition import create_shared_evaluation_handlers
     from ouroboros.mcp.tools.fanout import FanoutRegistry
-    from ouroboros.mcp.tools.pm_handler import PMInterviewHandler
+    from ouroboros.mcp.tools.fanout_composition import create_fanout_wiring
     from ouroboros.mcp.tools.qa import QAHandler
     from ouroboros.mcp.tools.registry import ToolRegistry
     from ouroboros.mcp.tools.seed_handoff import SeedHandoffRegistry
@@ -2484,14 +2483,6 @@ def create_ouroboros_server(
             fanout_registry=fanout_registry,
             suppress_tool_use_prompt_cues=interview_envelope_sealed,
         ),
-        PMInterviewHandler(
-            data_dir=state_dir_path,
-            llm_backend=interview_llm_backend,
-            event_store=event_store,
-            agent_runtime_backend=interview_runtime_backend,
-            opencode_mode=opencode_mode,
-            fanout_registry=fanout_registry,
-        ),
         BrownfieldHandler(_store=brownfield_store),
         evaluate_handler,
         start_evaluate_handler,
@@ -2501,10 +2492,17 @@ def create_ouroboros_server(
             opencode_mode=opencode_mode,
             fanout_registry=fanout_registry,
         ),
-        *create_fanout_handlers(
-            fanout_registry,
-            effective_cwd,
-            event_store,
+        # One store, and the producer is handed its resolved root rather than
+        # deriving the same path again when a question is asked (RFC #2153).
+        *create_fanout_wiring(
+            fanout_registry=fanout_registry,
+            workspace=effective_cwd,
+            event_store=event_store,
+            handler_event_store=event_store,
+            state_dir=state_dir_path,
+            llm_backend=interview_llm_backend,
+            agent_runtime_backend=interview_runtime_backend,
+            opencode_mode=opencode_mode,
             ensure_ready=server.startup,
         ),
         evolve_step,
--- a/src/ouroboros/mcp/server/adapter.py
+++ b/src/ouroboros/mcp/server/adapter.py
@@ -2474,15 +2474,6 @@ def create_ouroboros_server(
             opencode_mode=opencode_mode,
         ),
         MeasureDriftHandler(event_store=event_store),
-        InterviewHandler(
-            interview_engine=interview_engine,
-            event_store=event_store,
-            llm_backend=interview_llm_backend,
-            agent_runtime_backend=interview_runtime_backend,
-            opencode_mode=opencode_mode,
-            fanout_registry=fanout_registry,
-            suppress_tool_use_prompt_cues=interview_envelope_sealed,
-        ),
         BrownfieldHandler(_store=brownfield_store),
         evaluate_handler,
         start_evaluate_handler,
@@ -2492,9 +2483,11 @@ def create_ouroboros_server(
             opencode_mode=opencode_mode,
             fanout_registry=fanout_registry,
         ),
-        # One store, and the producer is handed its resolved root rather than
-        # deriving the same path again when a question is asked (RFC #2153).
+        # One store, and both producers are handed it rather than deriving a
+        # path from the workspace when a question is asked (RFC #2153).
         *create_fanout_wiring(
+            interview_engine=interview_engine,
+            suppress_tool_use_prompt_cues=interview_envelope_sealed,
             fanout_registry=fanout_registry,
             workspace=effective_cwd,
             event_store=event_store,
--- a/src/ouroboros/mcp/tools/authoring_handlers.py
+++ b/src/ouroboros/mcp/tools/authoring_handlers.py
@@ -63,14 +63,16 @@ from ouroboros.mcp.tools.advisory_dispatch import (
     echo_carries_dispatch,
     strip_bridge_notice,
 )
+from ouroboros.mcp.tools.interview_advisory import (
+    _attach_question_assist_requests,
+    _milestone_for_score,
+)
 from ouroboros.mcp.tools.question_advisory import (
-    attach_question_advisory,
     build_question_advisory_request,
 )
 from ouroboros.mcp.tools.subagent import (
     DELEGATED_TO_SUBAGENT,
     FanoutRegistry,
-    SubagentDispatchMode,
     build_generate_seed_subagent,
     build_interview_subagent,
     dispatch_plugin_terminal,
@@ -85,11 +87,6 @@ from ouroboros.mcp.types import (
     MCPToolParameter,
     MCPToolResult,
     ToolInputType,
-)
-from ouroboros.orchestrator.capabilities import (
-    interview_code_investigation_answer_contract,
-    ouroboros_tool_capability_metadata,
-    stable_code_investigation_question_identity,
 )
 from ouroboros.orchestrator.capabilities.question_text import (
     format_question_with_ambiguity as _format_question_with_ambiguity,
@@ -395,14 +392,6 @@ def _completion_gate_reason(
     return "requirements are not stable enough to close yet"
 
 
-def _milestone_for_score(score: AmbiguityScore | None) -> str | None:
-    """Return the milestone label for an ambiguity score, or None."""
-    if score is None:
-        return None
-    milestone, _ = get_milestone(score.overall_score)
-    return milestone.value
-
-
 _MILESTONE_RANKS = {
     "initial": 0,
     "progress": 1,
@@ -612,60 +601,6 @@ def _compute_transcript_chars(state: InterviewState) -> int:
     return total
 
 
-def _build_code_investigation_request(
-    *,
-    session_id: str,
-    question: str,
-    last_question: str | None = None,
-) -> dict[str, Any]:
-    """Build stable metadata for caller-side code-fact investigation.
-
-    The interview MCP tool is only the question generator; repo inspection runs
-    in the parent runtime. This metadata gives that runtime a concrete request
-    envelope keyed to the exact question text, so investigation results can be
-    correlated even when the user-facing display has an ambiguity prefix.
-    """
-    mcp_tool_capability = ouroboros_tool_capability_metadata("ouroboros_interview")
-    code_investigation = mcp_tool_capability["orchestration"]["code_investigation"]
-    request: dict[str, Any] = {
-        "session_id": session_id,
-        "question_identity": stable_code_investigation_question_identity(question),
-        "question": question,
-        "investigation_goal": "describe_current_state_from_code",
-        "investigation_targets": [{"target_type": "workspace", "scope": "active"}],
-        "fact_categories": [
-            "tech_stack",
-            "frameworks",
-            "dependencies",
-            "current_patterns",
-            "architecture",
-            "file_structure",
-            "configuration",
-        ],
-        "allowed_capabilities": ["inspect_code"],
-        "repo_inspection_tool_capabilities": list(
-            code_investigation["repo_inspection_tool_capabilities"]
-        ),
-        "confidence_policy": {
-            "auto_confirm_when": ["exact manifest or config match with a single clear answer"],
-            "confirmation_required_when": [
-                "multiple code paths or configuration sources disagree",
-                "the answer implies a product or architecture decision",
-            ],
-            "human_judgment_when": [
-                "desired future behavior",
-                "priority, scope, ownership, rollout, or acceptance criteria",
-            ],
-        },
-        "answer_prefixes": ["[from-code]", "[from-code][auto-confirmed]"],
-        "answer_contract": interview_code_investigation_answer_contract(),
-        "mcp_tool_capability": mcp_tool_capability,
-    }
-    if last_question:
-        request["last_question"] = last_question
-    return request
-
-
 def _build_question_advisory_request(
     *,
     session_id: str,
@@ -690,50 +625,6 @@ def _build_question_advisory_request(
         milestone=_milestone_for_score(score),
         code_investigation_request=code_investigation_request,
         last_question=last_question,
-    )
-
-
-def _attach_question_assist_requests(
-    meta: dict[str, Any],
-    *,
-    session_id: str,
-    question: str,
-    phase: str,
-    score: AmbiguityScore | None,
-    last_question: str | None = None,
-    dispatch_mode: SubagentDispatchMode = SubagentDispatchMode.SEQUENTIAL,
-    runtime_backend: str | None = None,
-    opencode_mode: str | None = None,
-    fanout_registry: FanoutRegistry | None = None,
-) -> None:
-    """Attach the interview's code-fact request and its advisory fan-out.
-
-    The fan-out itself lives in ``question_advisory``, which serves every tool
-    that declares a catalog — one implementation, so the interview and the PM
-    interview cannot drift apart in the places where being identical matters.
-    What stays here is the part that is only the interview's: the code-fact
-    investigation request, which exists because in this interview a code fact
-    may become the answer.
-    """
-    code_request = _build_code_investigation_request(
-        session_id=session_id,
-        question=question,
-        last_question=last_question,
-    )
-    attach_question_advisory(
-        meta,
-        tool_name="ouroboros_interview",
-        session_id=session_id,
-        question=question,
-        phase=phase,
-        ambiguity_score=score.overall_score if score is not None else None,
-        milestone=_milestone_for_score(score),
-        code_investigation_request=code_request,
-        last_question=last_question,
-        dispatch_mode=dispatch_mode,
-        runtime_backend=runtime_backend,
-        opencode_mode=opencode_mode,
-        fanout_registry=fanout_registry,
     )
 
 
@@ -1657,6 +1548,7 @@ class InterviewHandler:
     data_dir: Path | None = field(default=None, repr=False)
     fanout_registry: FanoutRegistry | None = field(default=None, repr=False)
     suppress_tool_use_prompt_cues: bool = False
+    findings_store: Any | None = field(default=None, repr=False)
 
     def __post_init__(self) -> None:
         """Initialize event store."""
@@ -2766,6 +2658,7 @@ class InterviewHandler:
                             runtime_backend=self.agent_runtime_backend,
                             opencode_mode=self.opencode_mode,
                             fanout_registry=self._resolved_fanout_registry(),
+                            findings_store=self.findings_store,
                         )
                     finally:
                         advisory_build_duration_ms = _elapsed_ms(advisory_build_started_at)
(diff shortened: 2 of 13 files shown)


Q00 (commented):
Draft review (exact HEAD f587a024f): `recent_finding_paths()` follows filesystem links at both levels. `Path.is_dir()` follows a two-character shard symlink, `glob("*.json")` traverses it, and `path.stat()` follows a JSON symlink. Because the returned absolute paths are explicitly handed to a child with instructions to read them, a workspace-controlled `.ouroboros/artifacts/aa` symlink (or a symlinked JSON body) can escape the artifact store and expose unrelated local files to the advisory lane. Please enumerate without following symlinks, require regular non-symlink shard/files, and verify resolved candidates remain beneath the resolved artifact root; validating the content-addressed filename shape would further keep the listing limited to actual store bodies.

cohemm:
Both blockers fixed in d2c0fdd, and both were RFC violations rather than hardening asks — worth saying, because it changes what the fix has to satisfy.

**Ineligible findings (blocker 2).** The RFC closes the eligible list at `code_context` and `data_context`, and the listing offered everything. Measured on a real store while fixing this: 16 bodies, of which 8 were not fan-out advisories at all, and one advisory carried `web_context`, `ambiguity_contrarian`, `answer_simplifier` and `architecture_implications` alongside the two eligible lanes. So `kind` filtering alone is not enough — an interview turn runs six lanes into one submission, and eligibility has to be read per lane.

A body carrying no eligible lane is not offered. A body carrying some is offered, and the boundary inside it is named in the prompt rather than cut server-side: slicing per lane here would put this server back in the position of deciding what is relevant without having read the question, which is the thing handing over paths exists to avoid.

The list is now closed rather than open — the RFC was revised to exclude a web lane outright rather than defer it, since the window was derived from how a repository drifts and nothing about this project bounds how fast an external fact turns over.

**Integrity and containment (blocker 1).** A body is offered only when its name is the digest of its own bytes. That is the store's own naming rule rather than a new guard, and a planted body would have to find a preimage for the name it sits under.

The digest does not cover the case that matters most, though. A symlink to **another project's valid artifact** hashes correctly and carries an eligible lane — only where it lives makes it not this project's, which is cross-project reuse and something the RFC rules out. Symlinked shards and bodies are skipped for that.

Worth flagging how close that one came to being missed: the first version of the symlink test linked to an arbitrary file, passed on the digest check alone, and therefore stayed green with the symlink guard removed — a test that looked like it pinned a guard and did not. It now links to a genuine body elsewhere and fails without it. All four guards (digest, symlink, kind, lane) were checked the same way: removed, re-run, confirmed failing.

**One property retired.** The PR claimed the listing opened nothing. Eligibility lives inside the body, so that is no longer true and the body has been corrected. Recency is still decided from metadata, so nothing outside the window is read.

`tests/unit` + `tests/conformance`: 20321 passed, 88 skipped. Ruff, format, mypy and the module-size ratchet are clean.

cohemm:
Addressed review round 2 in d49dd74 — 4 fixed, 0 defended. Two of the four were the shape that says stop rather than continue, so the response is a change of key rather than four patches.

**Blockers 2, 3, 4 — fixed together** (`recent_findings.py`). B3 was the exact complement of the predicate added last round: I checked that a body's name is the digest of its bytes, and the probe answered "then name your forgery by its own digest." B4 existed only because last round introduced reading at all. Strengthening either would have called the next round, so the key changed: **a body is reachable because a publication manifest says this store wrote it.**

That answers all three at once, and none of them by detection:

- *Membership* (B3): a body placed in the blob directory is not recognised and rejected — nothing refers to it, so it never enters. Content addressing could never have answered this, since a chosen body can always be named by its own digest. Your probe is now excluded structurally: `test_a_body_the_store_never_published_is_not_offered` writes exactly that file and gets `[]`.
- *Recency* (B2): the manifest's publication time. Both halves of your finding are covered — a republished identical body is listed under its new contract's time, and `touch` on a blob changes nothing because blob mtime is no longer read.
- *Bounded reads* (B4): bodies go through the store's own `fetch`, which is bounded, containment-checked, and verifies content against the address. No blob is opened in this module. Manifests are size-checked and symlink-skipped before parsing.

Manifest parsing reuses the public helpers the store parses its own manifests with (`latest_artifact_event`, `event_artifact_ref`, `digest_from_ref`), so there is no second copy of what a manifest means. It lives in the tools layer rather than on the store only because `artifact_store.py` sits **one line** under the module-size cap; the intended home is recorded in the module docstring.

**Blocker 1 — fixed, and the RFC changed first.** You were right that the interview is unwired, and the reason it was unwired no longer holds. RFC #2153 previously cut by tool because the Socratic interview can adopt a code finding as the answer itself. That cut is gone: a fact about the system is the same fact whichever interview needed it, and the freshness argument is the window — a staleness check at read time would say the window is not trusted. Assumption C was removed with it. Both producers are now built by one factory and hold the same store, and `test_the_ordinary_interview_reads_the_same_findings` pins it.

**Follow-up 1 — fixed.** You identified exactly why the gap was invisible: the wiring helper searched for `PMInterviewHandler` only, and a helper that looks for one thing cannot report the other missing. It now requires both producers to be registered and to hold the same store object.

Each of the four is pinned by removing its fix and re-running: dropping the interview wiring fails 3 composition tests; reverting to a directory listing fails the membership, recency and unreadable-store tests.

`authoring_handlers.py` went 6 lines over its ratchet budget from this change. Rather than trimming it, the interview's own advisory helpers moved to `interview_advisory.py`, which is what that ratchet asks for.

Verification: `tests/unit` + `tests/conformance` — 20319 passed, 88 skipped. Ruff, format, mypy, and the module-size ratchet against `origin/main` are clean. One earlier run showed `test_picker_projection_write_and_disk_budgets` failing; it passes standalone and passed on re-run, so it is load-related and not from this change.

cohemm:
Thanks — approving with that follow-up was the right call, and it is fixed in 9e5df3f rather than deferred, because it is not a hardening suggestion: it is a guarantee this function's own docstring declared and the code did not make true.

Reproduced all three paths before changing anything:

```
latest_artifact_event({})                     -> KeyError 'events'
latest_artifact_event({'events': 'nope'})     -> AttributeError
datetime.fromisoformat('...T03:43:29') >= now -> TypeError (naive vs aware)
```

And your reading of why it mattered is exact: `attach_question_advisory()` calls the lookup outside its advisory-build guard, so any of those took the user's **question**, not the shortcut.

The gap's shape is worth naming, since it is the same one twice in this PR: the parse was guarded and the parsing of *what the parse produced* was not. A contract record is a file inside a project — project-controlled input — while the helpers reading it are written for manifests this store wrote. One guard now covers the whole record instead of one clause at a time. A record with no offset is skipped rather than compared, since it cannot be placed against the window at all.

Pinned in both directions over five malformed shapes (`{}`, non-list events, non-object event, non-JSON, naive timestamp): the bad record is skipped **and** a good publication beside it is still offered — so degrading cannot quietly become blanking. Verified by reverting the wider guard and the offset check: three of the five fail without them.

I did not take the schema-validator half of your suggestion. Routing manifests through the store's full validator would put a second parse of the same file beside the one the helpers already do, and what the lookup needs from a record is four fields it can fail to find. If you can show a malformed shape that survives the guard and reaches a child, that changes and I will take it.

`tests/unit/mcp` + `tests/unit/persistence`: 2597 passed. Ruff, format, mypy and the module-size ratchet are clean.

cohemm:
Addressed review round 3 in fb7a7b0 — 1 fixed, 2 refused with reasons. All three point at code that was present in `d49dd74`, the commit this PR was approved on; the approval and these blockers are two passes over the same lines, not a regression from the post-approval commit.

**Blocker 2 — fixed** (`recent_findings.py`). You are right, and it is a defect independent of any adversary: the path was built from the reference the shortlist carried while the body was verified by fetching the contract. Two sources for one value. They agree in every ordinary case, which is exactly what makes holding both a defect rather than a redundancy — on divergence it checked one body and handed over another, and nothing downstream could tell. The path now comes from `fetched.envelope.artifact_ref`. Pinned by feeding the shortlist a reference to a *different* published body: with the fix each publication is offered as the body its own fetch verified (`{published, other}`), and reverting to the shortlisted reference collapses both to one digest and fails the test.

**Blocker 1 — refused, and the RFC now says why.** Your probe reproduces: I copied a manifest into an arbitrary directory, changed only the timestamp, and a two-day-old publication came back. But this is the same mechanism your round-2 blocker required. That round asked that an identical body republished under a new contract be treated as fresh — which means *any* manifest referring to an artifact can make it recent. A legitimate re-publication and a hand-written alias are indistinguishable on the filesystem; the difference is who wrote the file, and the store lives inside the project.

So the boundary is stated rather than defended. RFC #2153 now carries a threat model: the project workspace is trusted input, because a lane's whole job is to read this project's code and report what it says. Anyone who can write a manifest can more easily edit `lapse.py` and have the lane cite it with a real path the user then confirms. Canonicalising directory names moves the attack to correctly-named directories; validating bindings moves it to written bindings.

What changed instead is that this module stopped claiming otherwise. Its docstring said membership followed from the record; what actually follows is *when* — a content-addressed store reuses a body on re-publication, so blob mtime reports when bytes first appeared rather than when a finding was established. That is the whole reason the record is read, and it has no adversary in it. A test asserting the same is reframed as a consequence rather than a promise.

**Blocker 3 — refused, with a wording fix on my side.** The RFC closes eligibility on what may be *reused* — the outputs of `code_context` and `data_context` — not on which lane may read them. A contrarian or an answer drafter seeing what the code does today is not a reuse of question-specific reasoning; it is the same evidence the question is being asked against. The PR body said "two lanes are eligible and the list is closed" without saying which side of the fan-out that governs, which is a fair way to read it as a consumer gate, and that is corrected.

If you can show a consuming lane whose *answer* is made worse by having read a `code_context` finding, that is a different argument and I will take it.

Verification: `tests/unit/mcp` + `bigbang` + `orchestrator` 8381 passed; `tests/conformance` + `tests/unit/persistence` 551 passed; ruff, format, mypy and the module-size ratchet clean.

cohemm:
Follow-up fixed in 68cb0f3, and taken rather than deferred for a reason worth stating: it is not a hardening ask, it is the decision half-implemented. RFC #2153 decided an interval and the code compared against one edge.

Reproduced with the store's own API before changing anything — `put_for_contract(..., now=+30d)` was offered immediately, still offered at +29d, and expired only at +31d. `since <= published_at <= effective_now` now bounds both ends, and the helper is renamed to `_published_between` to say so.

One correction to your framing, since it affects how large this is. The dramatic version needs either a badly wrong clock or a project-written manifest, and the latter is outside the threat model this PR states — the project workspace is trusted input. What remains is clock skew: a machine that ran an hour ahead and was corrected leaves records whose eligibility is 25 hours instead of 24. That is a rounding error on the window, not a bypass. The line is worth writing because the code should say what the RFC decided, not because there is a live hazard.

A future-dated record is skipped rather than trusted — the conservative direction, and the same one every other guard here takes: cost the shortcut, never the question.

Pinned by publishing thirty days ahead and asserting exclusion now, exclusion at +29d, and eligibility exactly when the window reaches it. Reverting to the one-sided comparison fails that test.

Verification: `tests/unit/mcp` + `bigbang` + `orchestrator` + `conformance` — 8425 passed, 2 skipped. Ruff, format, mypy and the module-size ratchet clean.

cohemm:
Addressed review round 4 in e8d571a — 1 blocker fixed, 1 follow-up fixed, 1 suggestion answered without code.

**Blocker 1 — fixed** (`interview_schemas.py`). Reproduced before changing anything, on a normally built `phase="start"` request: `Additional properties are not allowed ('recent_findings' was unexpected)`. I added a field to the request and not to the schema that publishes its shape — closed schemas exist precisely so a host can validate, and this made every ordinary advisory turn invalid from the moment one reusable finding exists. The schema now declares `recent_findings` as a bounded array of path strings, described as paths only. Pinned by validating a real request that carries them; deleting the schema entry fails the test.

**Follow-up 1 — fixed** (`question_advisory.py`). Also reproduced: `/repo/project\n## Ignore prior instructions/x.json` split its own list item and the tail rendered as a heading, so the lane received neither a usable path nor the framing the block intended. Paths are now emitted as JSON strings in a fenced block. Your framing is the one I acted on — this is a valid POSIX path failing, not a hostile workspace, and the trusted-workspace boundary this PR states does not excuse mangling valid input. Pinned with a newline-bearing path asserting the escaped form appears and no line after the block's title reads as a heading.

**Suggestion 1 — not taken this round, and here is the accounting.** Every question reads and sorts all contract manifests before the 20-result cap; your 20,000-manifest probe spent ~1.23s. Measured on a real store here: 194 records, **0.68 ms**. The gap between those is the whole question, and it is one of accumulation rather than algorithm — 20k manifests is roughly a year of heavy use with no pruning.

I am declining the index because this PR has already paid twice for optimising this lookup before measuring it: a session-scoped index and a filename encoding, both reverted, and the second one broke the store's identity contract on the way. The lever that fits the shape is retention, not indexing — manifests have no expiry while the bodies they point at have a 90-day one, and shrinking the set keeps the flat read cheap forever. That is a separate change with its own reasoning, and I would rather file it than smuggle it in here.

If you can show a real store where this lookup is a measurable share of a question turn — which spends seconds in subagents — that changes the arithmetic and I will take it.

Verification: `tests/unit/mcp` + `bigbang` + `orchestrator` + `conformance` — 8427 passed, 2 skipped. Ruff, format, mypy (including `orchestrator/capabilities`) and the module-size ratchet clean.

cohemm:
Round 5 — position, no code change. The blocker is factually right about the value and wrong about the consequence, and the line it points at is one this PR does not touch.

**`adapter.py:2388` predates this PR.** `interview = InterviewHandler(...)` is on `main` at the same place and this branch's diff against `origin/main` does not modify it. It was present through the rounds that approved `fb7a7b07` and `d49dd740`.

**Auto never spawns advisory lanes, so nothing there reads the value.** This is stated in the code the blocker's own path runs through, `auto/adapters.py:1122`:

> The advisory fan-out directive is addressed to a host *model* that spawns subagents; auto is a programmatic driver that does not, so the directive is cut rather than answered as part of the question.

Auto calls the interview handler in-process (`await self.handler.handle(...)`), so there is no host model and no plugin bridge between them to act on the payload metadata. The server builds the advisory payloads, `_extract_interview_question` cuts the directive, and the lanes are never dispatched. The finding's stated consequence — "advisory lanes inside auto re-derive everything and never receive recent findings" — does not happen, because those lanes do not run at all.

**Wiring it would not be neutral.** `attach_question_advisory()` calls `recent_finding_paths(findings_store)` while building the request. With no store the call returns immediately; with one, every auto question would list and read the project's contract manifests to produce paths that auto then discards. That is cost with no reader — and this PR has already reverted two mechanisms that optimised or extended this lookup without a measured reader.

I implemented the wiring before checking, and reverted it: it moved `create_fanout_wiring` above the auto handler, carried the store through `_authoring_interview_handler`'s clone branches, and put `adapter.py` at exactly its 2612-line ratchet budget — spending the last of that module's headroom on a value nothing reads.

**Reopen condition.** If auto gains a host that spawns advisory subagents — a plugin bridge on the auto path, or a model-driven driver replacing `_extract_interview_question`'s cut — then the store belongs with that change, and the clone in `_authoring_interview_handler` must carry it too. Show me an auto run where a lane subagent is actually dispatched and I will wire it in the same round.

No verification run is reported for this round because no code changed; HEAD remains `e8d571a`.

cohemm:
Round 6 — position, no code change. This is the alias-manifest finding from round 3, which round 5 recorded as withdrawn ("accepted under the PR's explicit trusted-workspace model and is not re-raised"). It returns with the same probe and no new reachability evidence, so the answer is the same, but it is worth stating precisely rather than by reference.

**Nothing in normal operation creates a non-canonical directory.** The store writes manifests only at `contracts/<contract_path_component(contract_id)>/events.json` — `_manifest_path()` is the single writer and it derives the component from the contract id. So the state your probe constructs, `contracts/alias-refresh/`, does not arise from publishing, pruning, deduplication or replay. It arises from someone creating a directory by hand, which is the actor RFC #2153 places inside the trust boundary: a lane's whole job is to read this project's code and report it, so anyone able to create that directory can more cheaply edit the source a lane would cite, with a real path the user then confirms.

**Deduplication does not need it either**, which is worth saying because round 2 required dedup to work: an identical body republished under a new contract produces a *second canonical* directory, and it is offered on its own timestamp. That path is unaffected by anything here.

**Why this differs from the divergence I did take.** In round 3 you found that the offered path came from the shortlisted reference while the body came from `fetch()`. I fixed that, and the reason was reachability, not tidiness: those two reads can diverge with no adversary at all — a republish landing between the shortlist and the fetch is enough. The alias case has no such path. It needs a directory that only a hand can make. I take "one value, two sources" findings when the second source is reachable in normal operation; this one is not, and adding the guard would be a check against an actor the design has already declined to defend against.

I considered it seriously this round rather than pointing at the earlier answer — `contract_path_component()` is public, the check is one line, and it would make the read address equal the write address. What stopped me is that this PR has spent six rounds on guards for states that only an excluded actor can produce, and each one has cost a real defect elsewhere in attention.

**Reopen condition, concretely.** Show a non-canonical `contracts/*` directory produced by the store itself — by publish, prune, tombstone, replay, migration or crash recovery — and this becomes a reachable inconsistency and I will close it in that round. A probe that creates the directory by hand demonstrates the mechanism, which was never in dispute, rather than the reachability, which is.

HEAD remains `e8d571a`; no verification run is reported because no code changed.

PR #2157, opened 2026-08-16
fix(bigbang): reject status-masking verify commands

### LINKED ISSUE #2155: verify_command can be generated as an always-failing / POSIX-only command, and no gate catches it before execution
## Summary

`seed-architect` can emit a `verify_command` that **can never exit 0**, and nothing between generation and execution catches it. The AC then fails at the verify gate even though the implementation is correct.

On Windows the run hard-fails. On macOS/Linux the same command **silently passes** because the generated `|| true` masks the failure — so the gate reports success while having verified nothing. The second half is arguably the worse one.

Ouroboros 0.50.4, runtime `kiro`, Windows 11 (PowerShell 5.1).

---

## Symptom

`ouroboros run workflow` fails with one AC rejected at the verify gate, while the sibling AC on the same seed completes:

```
Level 1/1: Executing ACs [1, 2] in parallel
Level 1 complete: 1 succeeded, 1 failed

Parallel Execution Complete
  Success: 1/2
  Failed: 1

Task Status:
 - Task 1: [FAILED]    ...widget can add/remove items ... saved to localStorage ... persists after refresh
 - Task 2: [COMPLETED] ...new week starts with an empty form, previous week preserved in localStorage

Error: Execution failed
```

`~/.ouroboros/logs/ouroboros.log`:

```
[warning] parallel_executor.ac.verifier_rejected ac_index=0 depth=0
  reason='Verify gate failed: verify_command exited with status 1'
  typed_evidence_present=False typed_evidence_valid=False
  verifier_failure_class=None verifier_passed=False verifier_ran=False
  verifier_reasons=[] verifier_status=None
...
[error] orchestrator.session.failed error='Partial failure: 1 failed, 0 blocked, 0 invalid'
```

`ouroboros status run <exec_id>` shows every evidence step erroring:

```
Steps: 3
- step_evidence_ff91848d10f2 [evidence_submission] typed evidence: error
- step_evidence_aac634f9dba0 [evidence_submission] typed evidence: error
- step_evidence_ab8d287569d8 [evidence_submission] typed evidence: error
```

The implementation itself was fine — the widget files were created and the widget renders correctly in the browser. The 1/2 split maps exactly to **which AC carried a `verify_command`**, not to code quality:

| AC | `verify_command` | result |
|----|------------------|--------|
| 1  | present (below)  | FAILED |
| 2  | absent (`expected_artifacts` only) | COMPLETED |

---

## The generated command

Produced by `ouroboros seed` from a Korean-language interview ("add a weekly retrospective widget to the dashboard"):

```yaml
acceptance_criteria:
- description: <widget can add/remove items for the current week, save button writes to
    localStorage per week, data survives a refresh>
  semantic_ac_key: ac_608344949552020d
  verify_command: python -c "import subprocess, sys; r = subprocess.run([sys.executable,
    '-m', 'http.server', '8000'], timeout=1, capture_output=True); print('server ok')" 2>&1 || true
  expected_artifacts:
  - index.html
  - app/main.js
  - app/style.css
  - app/widgets/weekly-review.js
```

### Why it can never succeed

**1. The Python snippet always raises.**
`http.server` does not exit on its own, so after 1s `subprocess.run(..., timeout=1)` kills the child and raises `subprocess.TimeoutExpired`. Nothing catches it → traceback → **exit 1**. `print('server ok')` is unreachable. The healthier the server, the more reliably this fails.

**2. `2>&1` changes nothing.** It merges stderr into stdout; the exit code stays 1. It only hides the traceback from the log.

**3. `|| true` is the intended mask, and it is POSIX-only.**
`parallel_executor.py:4333` runs the command through `asyncio.create_subprocess_shell`, which on Windows is `cmd.exe`. `||` parses there, but `true` is not a command:

```
'true' is not recognized as an internal or external command
```

so the exit code becomes 9009 instead of 0. Under PowerShell it would be worse — `||` did not exist before PS7.

On bash/zsh, `|| true` does force exit 0 — the gate passes, having verified nothing about the AC.

**4. The command does not test the AC anyway.**
The AC is about DOM interaction and `localStorage` persistence, which is not checkable by a single shell line. The model substituted "does a static server start" and then masked that substitute's failure. The presence of `|| true` reads as the generator knowing the command could fail.

---

## Why nothing catches it

Three gates exist between generation and execution; none of them looks at whether the command can run.

**`agents/seed-architect.md`** (verify semantics) constrains only shape:

- single-line shell command
- no heredoc / multiline / line continuation
- use `python -c "..."` or `python -m pytest -q`

Nothing about OS neutrality, exit-code contract, or POSIX-only operators. The command above complies with every stated rule.

**`bigbang/seed_generator.py:109`** validates the same two things and nothing else:

```python
def _unsupported_verify_command_reason(command: str) -> str | None:
    if "\n" in command or "\r" in command:
        return "verify_command must be a single-line command"
    if _UNSUPPORTED_VERIFY_HEREDOC_RE.search(command):
        return "verify_command uses heredoc/multiline shell syntax; use python -c or pytest instead"
    return None
```

**`auto/grading.py:511`** treats the *presence* of a command as proof of observability:

```python
def _is_observable(value: str | AcceptanceCriterionSpec) -> bool:
    if isinstance(value, AcceptanceCriterionSpec):
        if value.verify_command or value.expected_artifacts:
            return True
```

This inverts the incentive: an AC with no `verify_command` gets flagged "not clearly observable", so the pressure is to attach *some* command — and a broken one satisfies the grader perfectly.

There is no preflight execution. `run --dry-run` prints the plan without running verify commands, so the first time the command ever executes is **after implementation finishes**, at which point its failure is attributed to the code.

---

## Secondary: `verifier_ran=False` is conflated with AC failure

The log already distinguishes "the verifier could not run" from "the verifier ran and rejected":

```
verifier_ran=False  verifier_passed=False  verifier_reasons=[]  verifier_status=None
```

but both paths collapse into `Verify gate failed` → AC failed. Consequences:

- The failure is reported as an implementation defect when it is a spec/tooling defect.
- Under `evolve`, the feedback reaching the next generation is only "AC 1 failed". Successive generations rewrite working widget code while the broken command stays untouched. Since convergence is judged on ontology similarity (`evolution/convergence.py`), the lineage then terminates via the stagnation window with the AC still failing.

---

## Reproduction

1. `ouroboros init start "<any goal whose AC involves browser/DOM behaviour>"`
2. `ouroboros seed <interview_id>` — inspect `verify_command` in the emitted YAML
3. `ouroboros run workflow <seed> --runtime kiro --project-dir <project>`

The specific command varies per run since it is model-generated; the failure mode reproduces whenever an AC is not checkable by a single shell line and the generator substitutes a masked proxy.

Workaround that resolved it here — replace the line with anything that exits 0 and let `expected_artifacts` do the judging:

```yaml
verify_command: python -c "print('ok')"
```

---

## Suggested fixes

1. **`agents/seed-architect.md`** — add to the verify semantics: the command must exit 0 in the success case, must be OS-neutral, and must not rely on POSIX-only shell operators (`|| true`, `&&`, backticks, `2>&1` chaining). If the AC is not checkable by one shell line, emit `verify: NONE` and rely on `artifacts:` instead of inventing a proxy.

2. **`_unsupported_verify_command_reason`** — extend the existing check with a pattern for shell-control operators and `|| true` specifically. Cheap, single place, applies to CLI / MCP / auto alike since all four entry points share `bigbang.seed_generator.SeedGenerator`.

3. **Preflight** — run each `verify_command` once before the implementation phase (or under `--dry-run`) and surface a distinct diagnostic when it fails on an untouched workspace. A command that fails before any code is written cannot be measuring the code.

4. **Separate `verifier_ran=False` from AC rejection** — classify it as `verification_unavailable` rather than a failed AC, so that operators (and `evolve`) get "fix the command", not "fix the code".

Items 1 and 2 are small and would have prevented this instance.

---

## Environment

| | |
|---|---|
| Ouroboros | 0.50.4 |
| Runtime / LLM backend | `kiro` (Kiro CLI 2.x, `chat --no-interactive`) |
| OS | Windows 11, PowerShell 5.1, ko-KR locale |
| Install | `pipx install 'ouroboros-ai[mcp,tui]'` |
| Command | `ouroboros run workflow <seed> --runtime kiro --max-decomposition-depth 0 --project-dir <project>` |

Encountered while preparing a Kiro CLI hands-on workshop that uses Ouroboros as the spec-first tooling example.

atirna:
## Summary
- Seed extraction accepted acceptance-criteria verify commands that chain into an always-succeeding fallback such as `|| true`. The verify contract is one command whose exit code is the verdict, so a masked command verifies nothing — and `|| true` cannot run on Windows at all.
- The validator in `_unsupported_verify_command_reason` now rejects an unquoted `|| true` / `|| :` / `|| exit 0` wherever it appears in the chain, including mid-command tails like `cmd || true; echo done` that look legitimate. `; true` / `| true` / `& true` tails are rejected when they end the command.
- `&&` chains, pipes to real commands, quoted literals, `$(...)`/backtick expansions, and rich POSIX bodies (case statements, functions) stay valid — only forms that can replace the tested command's exit status are rejected.
- The extraction prompts and seed-architect guidance now state the exit-code contract instead of a blanket operator ban.

Closes #2155.

## Verification
1. Regression tests in `tests/unit/bigbang/test_seed_generator.py`: the exact `|| true` form from #2155 plus mid-chain and tail variants fail before the fix and pass after (15 tests fail pre-fix on stashed source, all 375 pass post-fix)
2. Live `/bin/sh` probe: `python3 -c "import sys; sys.exit(1)" || true; echo verifier-complete` exits 0, confirming the masked forms genuinely hide the failure
3. Full affected suites: `uv run pytest tests/unit/bigbang/ tests/unit/orchestrator/test_parallel_executor.py tests/unit/scripts/test_pr_hygiene_workflow.py` — 1618 passed
4. `uv run ruff check src/ tests/` and `uv run ruff format --check` clean; `uv run mypy src/ouroboros/bigbang/seed_generator.py` clean


atirna:
i replaced the raw fallback matching with outer-shell token classification.

- quoted, escaped, fragmented, redirection-adjacent, and assignment-prefixed success fallbacks now reject
- comment text and escaped operator arguments stay valid
- coverage exercises extraction retries and persisted Seed loading; 1,738 affected unit tests plus Ruff and mypy pass

atirna:
fixed the remaining shell-status cases in 07df41e. the shared validator now follows grouped fallbacks, status inversion, background/wait forms, command substitutions, redirections, and the POSIX `command` wrapper before deciding whether a fallback can replace the verifier result.

added extraction and persisted-seed regressions for those paths. verified with `uv run pytest -q tests/unit/bigbang/ tests/unit/core/test_seed.py` (1259 passed), Ruff, formatting, mypy, module-size, and an independent review bound to the patch.

atirna:
the leftover hole was `$(true)` / `` `true` `` as the fallback (and assignment-only `FOO=$(true)`). substitutions were getting collapsed to an opaque token so they looked like a real command.
i track substitution status through the lexer now and reject those forms. `false || $(true)` and `false || VERIFY_MODE=$(true)` both get rejected. quoted `"$(true)"` still counts as a word, not a fallback command.

Q00:
@atirna thank you for your contribution! I've just been focusing on #2180 #2181 #2182 
Your solution is also great and will be very helpful to make deterministic verification. will review this deeply!

Q00:
I've been working the same problem from the runtime side (#2180 → #2181 → #2182, the evidential-force RFC), so I ran this branch's validator against a corpus of status-masking contracts. Sharing the results because I think they point at a structural issue rather than a list of missing cases.

**14 of 22 masked contracts pass the gate.** Every command below exits 0 under `/bin/sh` with `false` standing in for a failing check — that is, the contract can never report a failure:

| verify_command | this PR | `sh` exit |
|---|---|---|
| `false \|\| true` | REJECT | 0 |
| `false \|\| :` | REJECT | 0 |
| `false \|\| exit 0` | REJECT | 0 |
| `false \|\| echo done` | **accept** | 0 |
| `false \|\| printf ''` | **accept** | 0 |
| `false \|\| test 1 = 1` | **accept** | 0 |
| `false \|\| cd .` | **accept** | 0 |
| `false \| cat` | **accept** | 0 |
| `false \| tee /dev/null` | **accept** | 0 |
| `if false; then :; fi` | **accept** | 0 |
| `while false; do :; done` | **accept** | 0 |
| `f() { true; }; false \|\| f` | **accept** | 0 |
| `eval 'false \|\| true'` | **accept** | 0 |
| `sh -c 'false \|\| true'` | **accept** | 0 |
| `! false && true` | **accept** | 0 |
| `false \|\| sh -c 'exit 0'` | **accept** | 0 |
| `false \|\| $(echo true)` | **accept** | 0 |

Repro (`verify_commands.py` from this branch, no other dependencies):

```python
import subprocess
from ouroboros.core.verify_commands import (
    _status_masking_verify_command_reason,
    _unsupported_verify_command_reason,
)

for cmd in ["false || echo done", "false | cat", "if false; then :; fi",
            "f() { true; }; false || f", "! false && true"]:
    rejected = bool(_status_masking_verify_command_reason(cmd)
                    or _unsupported_verify_command_reason(cmd))
    status = subprocess.run(["/bin/sh", "-c", cmd]).returncode
    print(f"{cmd!r:36} rejected={rejected}  sh_exit={status}")
```

Two of these are worth calling out because they are not adversarial shapes — they are what an LLM writes on an ordinary day: `pytest -q | tee test.log` (pipeline status is the *last* command, so `tee` decides the verdict) and `pytest -q || echo "tests done"`.

**Why I read this as structural.** The three review rounds here each closed a specific form and each time the next probe found another one, and the approving review still lists two open bypasses (`$( )` output becoming the executed command; compound-command/function-definition fallbacks) — its own Merge Recommendation is still "Request changes", and the requirement row for masking rejection still reads "Not met". That isn't a criticism of the implementation; it's what happens to any component that has to model shell status semantics to decide *accept*. The unmodeled construct becomes an accept, so the gate's failure direction is "let it through".

**The inversion that fixes the direction.** `core/verify_command_plan.py` (#2180) parses a verify command only to answer "can this be reproduced exactly?" — anything it cannot model returns `None`, and the caller reports the AC unverifiable instead of guessing. On the same corpus it *refuses* every pipeline, group, compound and function form above rather than accepting them. Same parser, opposite default.

**And the part no parser can do.** #2182 runs each contract against a disposable copy of the pristine workspace before any worker is dispatched. A contract that already passes there discriminated nothing this run, so its later pass is labelled `non_discriminating` and loses its recovery authority. On the corpus above that is **22 out of 22**, including every pipeline, `eval`, and function form — because it never parses the command, it runs it and watches. `pytest -q | tee log` is caught not by understanding pipelines but by the fact that it passes on an empty workspace.

So I don't think this work should be dropped — the class it targets (`cmd || true` after a real check) is real, and my static proof misses it exactly as your lexer misses pipelines. What I'd suggest is moving the check onto the existing plan-based parser, where "unmodeled ⇒ refuse" is already the default, and letting the baseline probe cover what static analysis structurally cannot.

One thing I'd flag separately from the design: the check raises inside a `@field_validator("verify_command", mode="before")`, which runs on *deserialization* too. `Seed.from_dict(yaml.safe_load(...))` in `evaluation_job.py:179`, `load_seed()` in `auto/adapters.py:998`, and `Seed.model_validate` in `cli/commands/workflow_ir.py:66` all go through it, so a session that already ran with `pytest || true` in its seed would stop loading after this merges — evaluate and resume included, with no migration path. #2181 puts its authoring gate before session preparation and leaves persisted seeds alone for that reason, and makes the refusal configurable (`seed.verify_command_gate: warn|block`) with a per-criterion exemption reason.

If you have the appetite, I'd genuinely value your eyes on #2180 → #2181 → #2182 — you've spent three rounds inside this exact problem, and reviewers who have actually tried to break a verify gate are rare. Bottom-up merge order, and #2182 carries the RFC (`docs/rfc/verify-gate-evidential-force.md`) if you want the whole argument in one place.

Q00:
@atirna Thank you again for the status-masking investigation and the adversarial shell cases in this PR. They exposed the deeper issue: `verify_command` was allowed to weaken transcript evidence and recover failed worker results, so every shell edge case became an acceptance-authority bug.

I replaced that architecture with monotonic verification in #2187:

- a verify command only adds evidence; it removes no `files_touched`, `commands_run`, or `tests_passed` obligations
- verify PASS never rewrites a failed worker execution
- missing Bash, startup failure, timeout, or transcript infrastructure produces explicit unverified success without retrying completed work
- the shell-free interpreter, static vacuity proof, and unsandboxed baseline probe are removed

Your work directly informed this redesign, and you are credited as a co-author on the commit. Could you review #2187, especially the additive-evidence invariant and the unavailable-verifier behavior?

#2157 is superseded implementation-wise, but the problem report and test cases were essential to the final direction.

PR #2143, opened 2026-08-15
feat(providers): add DeepSeek Harness (dsh) ACP completion backend

Q00:
## Summary

Adds a new LLM-only backend `dsh` (alias `deepseek_harness`) that drives [DeepSeek Harness](https://github.com/deepseek-ai/deepseek-harness)'s automation ACP server (`dsh-acp-demo`) for the in-process completion path (interview / seed / qa / evaluate). dsh speaks the same ACP wire ourocode already uses, so the implementation reuses that machinery instead of adding a new protocol stack.

Full analysis behind this change (what dsh has, what to borrow, how to collaborate) lives in `docs/research/deepseek-harness-adoption.md` in this branch.

## What changed

- **`DshAcpClient`** subclasses `OurocodeAcpClient` through three new launch seams (`_spawn_argv` / `_spawn_failure_hint` / `_session_new_params`) plus a `_TOOL_LABEL` used in error strings. ourocode wire behavior is byte-identical (its full test suite passes unchanged).
  - dsh's `session/new` indexes `params.mcpServers.length` unconditionally and rejects a non-empty list, so the client always sends `mcpServers: []`.
- **`DshLLMAdapter`** mirrors the ourocode adapter. The ACP wire has no model parameter — dsh's Cordis composition file owns provider/model — so the adapter reports an honest `dsh-composition` sentinel instead of fabricating a model id in audit metadata. Zero token usage (ACP carries none), cooperative `response_format` enforcement with bounded retries.
- **Registration** mirrors ourocode: LLM-only capability (`supports_runtime=False`, no interview driver, deliberately absent from every subagent-orchestration name map), factory spec, custom-entry-only model catalog, profiles alias, CLI env map, reviewer-independence vendor `deepseek`, lazy provider export.
- **Config/security**: `orchestrator.dsh_cli_path` / `orchestrator.dsh_config_path` with `OUROBOROS_DSH_CLI_PATH` / `OUROBOROS_DSH_CONFIG_PATH`. Both join the untrusted `.env` denylist — the config path selects which Cordis composition the spawned Node process loads, and composition rows execute arbitrary plugin code, so it gets the same treatment as an executable path.

## Verification

- Full unit suite (excl. `tests/unit/mcp`): **18,160 passed / 87 skipped**; ruff + mypy clean.
- 22 new tests for client and adapter, exercised against the shared CI-safe ACP fake.
[...]

PR #2158, opened 2026-08-17
feat(integrations): add installable dsh plugin for Ouroboros

### LINKED ISSUE #2159: Ship an installable dsh plugin bundle so DeepSeek Harness can mount Ouroboros in one command
#2143 answered one direction: how Ouroboros talks to DeepSeek Harness (the `dsh` LLM backend). The other direction — how a DeepSeek Harness user mounts Ouroboros — is still a hand-written `--patch` overlay copied out of the README.

## What's needed

A config-only dsh bundle inside this repo (`integrations/dsh-plugin/`), installable with a single `dsh plugin --profile <name> add "github:Q00/ouroboros#main&path:integrations/dsh-plugin"`, that inserts one `@deepseek-ai/dsh-mcp-client` row spawning `ouroboros mcp serve` over stdio. No separate repo or release cadence: it has no code of its own.

## Contract it has to get right

- **Install syntax** — `dsh plugin` declares `--profile` as a required option; every published command must carry it.
- **Credentials** — dsh's subprocess seam scrubs every credential-shaped name (`/KEY|PASSWORD|SECRET|TOKEN/i`) from a child's environment by design. A plugin's explicit `env` layer merges after that scrub, so any credential Ouroboros needs has to be named explicitly — as a short allowlist, not a wildcard passthrough.
- **`OUROBOROS_LLM_BACKEND=dsh`** — not a one-variable switch. Ouroboros spawns its own `dsh-acp-demo` child and fails closed with `invalid_config` unless `OUROBOROS_DSH_CONFIG_PATH` names an absolute trusted composition. Prerequisites must reach the child and be documented.
- **Startup failure** — soft-fail so unrelated dsh plugins keep working, with recovery described truthfully (the published `mcp-client` has no reconnect loop).

## Done when

- The bundle installs from a fresh `$DSH_HOME` by the documented command and composes into the profile.
- The MCP tools register under `mcp__ouroboros__*`.
- A test pins the install syntax, the credential-forwarding allowlist, and the recovery claim, since the bundle runs no Python of its own.

Q00:
Closes #2159. Follow-up to #2143 (the `dsh` LLM backend), which answered "how does Ouroboros talk to DeepSeek Harness". This answers the other direction: "how does DeepSeek Harness talk to Ouroboros" — as a one-command install rather than a hand-written `--patch` overlay.

## Summary

Adds a real, installable [DeepSeek Harness](https://github.com/deepseek-ai/deepseek-harness) plugin at `integrations/dsh-plugin/` — not a separate repo/package, since it has no independent code or release cadence of its own, just a config bundle that mounts Ouroboros as native `dsh` tools.

## What it is

A **config-only dsh bundle** per dsh's own ["Package and install a plugin"](https://github.com/deepseek-ai/deepseek-harness/blob/main/docs/user/develop/basic/publish.md) spec — `package.json` (declares `dsh.bundle.patch`) + `cordis.patch.yml` (inserts one `@deepseek-ai/dsh-mcp-client` row) + `README.md`. No custom plugin code: it spawns `uvx --from 'ouroboros-ai[mcp]' ouroboros mcp serve` over stdio, which registers all 36 `ouroboros_*` tools as `mcp__ouroboros__*` for the model. Type `ooo interview <goal>` or `ooo auto <goal>` in a dsh chat and the model finds the matching tool on its own — no extra prompting.

Install:
```sh
dsh plugin --profile <your-profile> add "github:Q00/ouroboros#main&path:integrations/dsh-plugin"
```

## Contract corrections (review round 1)

The bundle listed tools correctly but promised three things the dsh and Ouroboros subprocess boundaries do not deliver. Each was traced into the actual code at dsh `47f9438` and fixed:

1. **Install syntax.** `apps/cli/src/args.ts` declares `.requiredOption('--profile <name>')` on `dsh plugin`, so the root README's command errored out instead of installing. It now matches the bundle README and the verified command.
2. **Credential scrub.** `@deepseek-ai/dsh-subprocess` exports `SENSITIVE_ENV_PATTERN = /KEY|PASSWORD|SECRET|TOKEN/i` and `scrubbedParentEnv()` drops every match before spawn; `mcp-client`'s `buildChildEnv` merges the plugin's explicit `env` *after* that scrub. The bundle forwarded only the two `OUROBOROS_*` selectors, so a documented `DEEPSEEK_API_KEY` never reached the child — tools listed fine, the first call failed. The row now names a short, documented allowlist (`ANTHROPIC_API_KEY`, `DEEPSEEK_API_KEY`), with instructions for adding one more in a profile override rather than a wildcard passthrough.
3. **`OUROBOROS_LLM_BACKEND=dsh` was not a one-variable switch.** `DshAcpClient._spawn_argv` raises `invalid_config` unless `OUROBOROS_DSH_CONFIG_PATH` is an absolute composition path, and the bin needs `OUROBOROS_DSH_CLI_PATH` when `dsh-acp-demo` isn't on `PATH`. Both now travel to the child, and the README documents the whole procedure — including that a composition must sit where dsh's `node_modules` is reachable, and that the `dsh-acp-demo` bin has to come from a source build while npm packaging is broken upstream.

Also corrected the fail-open recovery claim. The published `@deepseek-ai/dsh-mcp-client@0.0.1-rc.1` ships no reconnect loop at all; `main` has one (`connection.ts`, bounded exponential backoff) that gives up after `maxAttempts` consecutive failures and then requires a reload. The README now says recovery is a plugin reload or a dsh restart instead of promising backoff.

## Verification

Original install verification, both done end to end against a completely fresh `$DSH_HOME` (a second, non-interactive sandbox — no prior Ouroboros install, no cached `uv` environment):

1. **Local-path install**: `dsh plugin --profile demo add ./integrations/dsh-plugin` → `--dump-config` shows the `# == dsh-ouroboros` layer composed correctly → booted the profile → `uvx` auto-fetched `ouroboros-ai` from scratch with zero manual `pip install` → a live chat turn listed all 36 `mcp__ouroboros__*` tools.
2. **GitHub-subdirectory install** (the exact command in the README): `dsh plugin --profile demo add "github:Q00/ouroboros#feat/dsh-plugin-bundle&path:integrations/dsh-plugin"` against the real pushed branch → resolved, installed, and composed identically via `--dump-config`.

For this round, the three contract claims are pinned by `tests/unit/integrations/test_dsh_plugin_bundle.py` (7 tests): the manifest's declared files exist, the row spawns the expected `uvx` argv, every credential-shaped name Ouroboros needs is explicitly forwarded (the scrub makes ambient forwarding impossible), the `dsh` backend prerequisites travel with the selector, and no published `dsh plugin` command in either README omits `--profile`. The bundle runs no Python of its own, so nothing else in the suite would catch this drift.


Q00:
Round 1 blockers addressed — each traced into the actual source at dsh `47f9438` (shallow clone of `deepseek-ai/deepseek-harness`) and the published npm tarball, not from memory.

**Blocker 1 — root install command (`README.md:178`).** Confirmed: `apps/cli/src/args.ts:173` declares `.requiredOption('--profile <name>', ...)` on the `plugin` subcommand. The root README now uses `dsh plugin --profile <your-profile> add ...`, identical to the bundle README and the verified command.

**Blocker 3 — credential scrub (`cordis.patch.yml`).** Confirmed and fixed first, because blocker 2 depends on it. `packages/subprocess/subprocess/src/index.ts:44` is `export const SENSITIVE_ENV_PATTERN = /KEY|PASSWORD|SECRET|TOKEN/i`, and `packages/mcp/mcp-client/src/transport.ts` builds the child env as `{ ...scrubbedParentEnv(), ...config.env }` — the explicit layer is the *only* way a credential reaches the child, exactly as the upstream comment says ("a deliberately supplied entry survives because explicit env layers merge after the scrub"). The row now forwards a named allowlist rather than a wildcard:

- `ANTHROPIC_API_KEY` — Ouroboros' default LLM backend
- `DEEPSEEK_API_KEY` — the `dsh` loopback below

Anything else (`OPENAI_API_KEY`, `OPENROUTER_API_KEY`, `GOOGLE_API_KEY`) is documented as a one-line profile override rather than forwarded by default; `OPENAI_API_KEY` in particular is deliberately excluded, since Ouroboros' own Codex doctor warns against adding it unless the user intends an API-key Codex profile. `?? ''` is safe on the receiving side: every Ouroboros credential read normalizes blank to unset (`LiteLLMAdapter._normalize_api_key`, `AnthropicAdapter.__init__`, the `.strip()` checks in `status.py` / `mcp_doctor.py`).

**Blocker 2 — `OUROBOROS_LLM_BACKEND=dsh` (`integrations/dsh-plugin/README.md`).** Confirmed against our own code: `DshAcpClient._spawn_argv` raises `invalid_config` when `config_path` is unset or relative, and `get_dsh_config_path()` / `get_dsh_cli_path()` are the only sources. I did not try to auto-derive the running profile's composition — `--dump-config` composes *without* evaluating `!!js` (`apps/cli/src/dump-config.ts`), and a composition's plugin names resolve relative to its own directory, so a synthesized path would be a guess. Took the documented route instead: both prerequisites are now forwarded to the child, and the README carries the full procedure (source-built `dsh-acp-demo`, absolute composition reachable from dsh's `node_modules`, the credential it names) plus an explicit statement that only the `dsh` backend needs any of it.

**Follow-up — recovery claim.** Your finding holds for what users install: the published `@deepseek-ai/dsh-mcp-client@0.0.1-rc.1` tarball contains no reconnect code at all. `main` does (`packages/mcp/mcp-client/src/connection.ts`, `RECONNECT_DEFAULTS` enabled, 500ms initial, `maxAttempts: 10`), but it gives up after that budget and logs "reload the plugin or restart the Host". The README no longer promises backoff — it states recovery is version-dependent and that the reliable move is a plugin reload or dsh restart.

**Test coverage.** `tests/unit/integrations/test_dsh_plugin_bundle.py` (7 passed) pins the three edges that would otherwise regress silently, since the bundle executes no Python: manifest files exist, the `uvx` argv is exact, every scrubbed-name credential is explicitly forwarded, the `dsh` prerequisites travel with the selector, no published `dsh plugin` command in either README omits `--profile`, and the recovery wording stays honest.

Also created #2159 as the tracking issue (the PR-hygiene gate was red for a missing issue trail; #2143 is a PR and does not count).

Q00:
Live verification of this round, against a real `dsh` install (`@deepseek-ai/dsh@0.1.0-rc.6` from npm) in a throwaway `$DSH_HOME` — not just source reading.

**1. The install blocker reproduces exactly as reported.**
```
$ dsh plugin add "github:Q00/ouroboros#main&path:integrations/dsh-plugin"
error: required option '--profile <name>' not specified
```

**2. The updated bundle still composes.** `dsh plugin --profile demo add ./integrations/dsh-plugin` → `dsh --profile demo --dump-config` shows the `# == dsh-ouroboros` layer with the new `env` block intact (`!!js` printed verbatim, as `dump-config` never evaluates it).

**3. A/B proof that the credential rows are load-bearing.** I swapped the row's `command` for a probe script that dumps its own environment, booted the profile twice with `DEEPSEEK_API_KEY=probe-deepseek-value ANTHROPIC_API_KEY=probe-anthropic-value OUROBOROS_AGENT_RUNTIME=codex OUROBOROS_DSH_CONFIG_PATH=/abs/composition.yml` exported in the host shell, and read what the child actually received.

Before (the two selectors only, i.e. this PR as reviewed):
```
OUROBOROS_LLM_BACKEND=
OUROBOROS_AGENT_RUNTIME=codex
OUROBOROS_DSH_CONFIG_PATH=/abs/composition.yml
```
Both API keys are gone — scrubbed, exactly as the finding said.

After (this round's `env` block):
```
OUROBOROS_LLM_BACKEND=
OUROBOROS_AGENT_RUNTIME=codex
OUROBOROS_DSH_CONFIG_PATH=/abs/composition.yml
OUROBOROS_DSH_CLI_PATH=
ANTHROPIC_API_KEY=probe-anthropic-value
DEEPSEEK_API_KEY=probe-deepseek-value
```
So the `!!js` layer does evaluate at activation and does merge after the scrub, and both credentials arrive with their host values.

That A/B also corrected one thing in my previous comment: `OUROBOROS_DSH_CONFIG_PATH` is not credential-shaped, so it already reached the child ambiently. Its explicit row is contract documentation, not a fix — the comment in the file now says so, and only the two `*_API_KEY` rows are load-bearing.

**One more fix it surfaced**, which nothing in the review had flagged: I had written "add that one name to the row's `env`" for forwarding an extra credential. Wrong — per dsh's own publish docs, a later patch layer *replaces* a row's entire `config` instead of deep-merging, so that advice would have silently dropped `command`, `args`, and the timeout. Both README and the file comment now say to restate the whole `config` block. Pushed as a separate commit.

PR #2161, opened 2026-08-17
docs(deepseek): document both integration directions and the dsh backend contract

### LINKED ISSUE #2160: DeepSeek Harness: reference docs never learned the dsh backend exists
#2158 shipped the installable dsh plugin with a README blurb, but the reference documentation was never updated for either direction of the DeepSeek integration.

## What is wrong today

- `docs/config-reference.md` — the `llm.backend` enum omits `dsh` (and `zcode`). `orchestrator.dsh_cli_path` and `orchestrator.dsh_config_path` are absent from the orchestrator table, and `OUROBOROS_DSH_CLI_PATH` / `OUROBOROS_DSH_CONFIG_PATH` are absent from the environment table — even though `dsh_config_path` is *mandatory* for that backend.
- `docs/cli-reference.md` — `--llm-backend` lists ten of the fourteen accepted values; `dsh`, `zcode`, `hermes`, and `ourocode` are missing.
- No user-facing page explains the two directions. `docs/research/deepseek-harness-adoption.md` is an internal adoption analysis, not a setup guide.

## Why it matters

A user reading the reference has no way to learn that `OUROBOROS_LLM_BACKEND=dsh` is not a one-variable switch. `DshAcpClient` fails closed with `invalid_config` unless an absolute composition path is supplied, so the first interview/Seed/QA call errors out with nothing in the docs to explain it.

## Done when

- Both reference lists match `llm_backend_choices()`.
- The two `dsh` config keys and their environment overrides are documented.
- A guide covers both directions separately, including the credential-scrub contract for the plugin and the composition-path rules for the backend.

Q00:
Closes #2160.

## Summary

`#2158` shipped the installable dsh plugin, but the reference docs never learned the `dsh` backend exists. This fills that in and adds the one page that separates the two directions, which are unrelated at runtime and easy to conflate.

## What changed

**`docs/guides/deepseek-harness.md` (new).** Splits the integration in two:

- **dsh → Ouroboros** (the plugin): the install command with its required `--profile`, what it needs (`uv`, an executable agent runtime), the credential-forwarding contract, and honest startup-failure recovery.
- **Ouroboros → dsh** (the LLM backend): the four steps that actually make `OUROBOROS_LLM_BACKEND=dsh` work — source-built `dsh-acp-demo`, an absolute composition path, the credential the composition names, then the selector.

It carries the reasons a reference table cannot: dsh scrubs credential-shaped names (`/KEY|PASSWORD|SECRET|TOKEN/i`) from children by design and the explicit `env` layer merges after that scrub; a profile override replaces a row's whole `config` rather than deep-merging it; a composition's plugin names resolve relative to the composition file's own directory; and both `OUROBOROS_DSH_*` variables are on the project `.env` denylist because they select which executable and plugin tree get loaded.

**Reference fixes**, taken from `llm_backend_choices()` rather than from the previous prose:

| File | Was | Now |
|---|---|---|
| `docs/config-reference.md` | `llm.backend` enum missing `dsh`, `zcode` | both added, with the LLM-only note extended to `dsh` |
| `docs/config-reference.md` | no `dsh_cli_path` / `dsh_config_path` rows | added, including that the config path is mandatory and must be absolute |
| `docs/config-reference.md` | no `OUROBOROS_DSH_*` env rows | added |
| `docs/cli-reference.md` | `--llm-backend` listed 10 of 14 values | completed |

**Index/entry points.** `docs/README.md` lists the new guide; the root README's DeepSeek blurb links to it.

## Verification

- `llm_backend_choices(include_aliases=False)` → `claude, codex, copilot, dsh, gemini, gjc, goose, hermes, kiro, litellm, opencode, ourocode, pi, zcode` — both edited lists now match it.
- Every relative link in the touched files resolves to an existing path (scripted check, 0 broken).
- `tests/unit/test_tui_usage_docs_contract.py`, `tests/unit/test_claude_plugin_skill_guide.py`, `tests/unit/integrations` — 15 passed.

Docs only; no source changes.


Q00:
Added the DeepSeek Harness run to the README host grid (second commit).

- `docs/images/host-dsh.gif` — the recording of a dsh chat driving `mcp__ouroboros__ouroboros_interview` turn by turn, submitting advisory fan-out results between rounds. Sped up 3x (78s → 26s) and sized to match its siblings: 620px wide, 6fps, 64 colors, **1.1MB** (host-claude.gif is 953KB, host-hermes.gif 761KB). `docs/` is excluded from the sdist in `pyproject.toml`, so it does not reach the PyPI artifact.
- The grid reflows **2x2 → 3x2** so the fifth host does not sit alone on its own row. The sixth cell is a placeholder for the Kiro recording, with an HTML comment naming the exact path (`./docs/images/host-kiro.gif`) and the markup to swap in.
- Cell widths 50% → 33% and image widths 440 → 300 to match the new column count.
- The caption's hardcoded "Four separate runs, four hosts" is now countless — it would have gone stale twice in this one change.

Q00:
All three blockers confirmed against source and fixed. I reproduced each one rather than taking it on faith — and all three trace to one mistake of mine: I documented from `llm_backend_choices()`, the union of every backend capability, when each command declares its own enum and the persisted config declares its own literal set. The union is nobody's contract.

**1. `init start` (`docs/cli-reference.md:371`).** `src/ouroboros/cli/commands/init.py` declares eleven members: `claude_code, litellm, codex, copilot, opencode, gemini, goose, kiro, pi, zcode, dsh`. `hermes` and `ourocode` were mine; `gjc` predated this branch. Row now matches the enum exactly.

**2. `mcp serve` / `mcp info` (`:1161`, `:1272`).** `src/ouroboros/cli/commands/mcp.py` accepts the same eleven values. Both rows listed `gjc` and omitted `zcode` and `dsh` — and you are right that this directly contradicted the guide I was adding, which tells the reader to run `mcp serve --llm-backend dsh`. Both rows fixed.

**3. `deepseek_harness` alias.** Reproduced the split precisely:

| Surface | `deepseek_harness` |
|---|---|
| `OUROBOROS_LLM_BACKEND` | accepted — `get_llm_backend()` returns the env value verbatim and the factory resolves the alias |
| `create_llm_adapter(backend=...)` | accepted — returns `DshLLMAdapter` |
| `--llm-backend` (both CLI enums) | rejected |
| `llm.backend` in `config.yaml` | rejected — `LLMConfig` literal validation |

The guide now scopes the alias to the two runtime-resolution paths and says to write `dsh` in the CLI and in `config.yaml`. The config reference says explicitly that the alias is *not* accepted in that field, rather than presenting it beside the persisted contract.

**Parity test added**, per your merge recommendation — `tests/unit/test_cli_reference_backend_parity.py` (5 tests):

- each documented `--llm-backend` row is compared to the enum of the command that owns it (rows are grouped by their nearest preceding heading), with the failure message naming both directions — advertised-but-rejected and accepted-but-undocumented;
- the `llm.backend` config row is compared to `LLMConfig`'s literal args, since that field is a superset of every CLI enum and needs its own check;
- a coverage test fails when a new `--llm-backend` row appears under a heading no one mapped, so the next command cannot be documented into a blind spot.

Mutation-checked rather than assumed green: re-adding `gjc` to the init row fails `test_documented_llm_backends_match_the_command_enum[### \`init start\`]`, and removing it passes again.

Also fixed the earlier CI failure in the same branch: `test_mcp_serve_documentation_names_runtime_and_public_claude_aliases` caught that my guide printed `ouroboros mcp serve --llm-backend dsh` with no `--runtime`, one section after explaining that an MCP subprocess needs an executable runtime. Both the guide and the README blurb now show `--runtime claude-cli`.

PR #2162, opened 2026-08-17
fix(advisory): say the findings body is a list, because it is

cohemm:
Follow-up to #2154 (RFC #2153). One sentence in the recent-findings prompt block, plus the test that would have caught it.

## What was wrong

The block told a lane:

> Each file holds answers **keyed by** `lane_id` under `result.aggregated_outputs`.

They are not keyed by anything. `aggregated_outputs` is a **list** of `{"lane_id": ..., "output": ...}` entries, so the access that sentence invites — `"code_context" in aggregated_outputs` — is `False` against a list.

## How it showed up

Observed twice in one session running the merged feature, in two different subagents:

- the interview's `code_context` lane made that exact check, printed nothing, then inspected the structure and recovered;
- a PM `code_context` lane made the same check and reported *"The artifacts had no `code_context`/`data_context` entries"*, then went back to searching the repositories.

Nothing failed loudly. The store had the findings, the paths in the prompt were correct, the files opened. A lane that reports nothing to reuse looks exactly like a project with nothing to reuse — which is the failure mode this feature has had to be protected from at every step, and it arrived through a description rather than through code.

The shape is identical for both tools, so nothing here is PM- or interview-specific: PM publishes two lanes and the interview six, in the same envelope.

## The fix

The block now states the container and the entry shape and says outright that it is not a mapping. The instruction it carries is unchanged — take the `code_context` and `data_context` entries, treat the rest as reasoning about a different question.

## The test

It reads a body the store actually published and asserts that the access the block describes is the access that works: `aggregated_outputs` is a list, `"code_context" in outputs` is False, and selecting by `lane_id` finds the entry. Then it checks the block says so.

Asserting the sentence alone would have passed for as long as the sentence was wrong — pinning one wording against another proves nothing about the data. One existing test pinned the old conjunction (`` `code_context` and `data_context` ``); it now asserts both lane ids separately, so rewording cannot break it while dropping a lane still does.

## Test plan

- [x] `tests/unit/mcp` + `bigbang` + `orchestrator` + `conformance` — 8428 passed, 2 skipped
- [x] `uv run ruff check src/ tests/` and `ruff format --check`
- [x] Reverting the sentence fails the new test; the old wording is what it catches

### DIFF
--- a/src/ouroboros/mcp/tools/question_advisory.py
+++ b/src/ouroboros/mcp/tools/question_advisory.py
@@ -307,14 +307,10 @@ def _recent_findings_section(request: Mapping[str, Any]) -> str:
     outside this session's roster is rejected at submission. The child is told
     so here, where it is deciding what to read.
 
-    **The shape is stated exactly, because it was not.** This said the answers
-    were "keyed by ``lane_id``", which reads as a mapping; they are a list of
-    entries carrying a ``lane_id``. Two subagents made the same wrong move on
-    it -- ``"code_context" in aggregated_outputs`` against a list is ``False``
-    -- and one of them concluded the files held nothing and went back to
-    searching the repositories. Nothing failed loudly: the paths were right, the
-    files opened, and the lane simply reported finding nothing to reuse, which
-    is indistinguishable from there being nothing.
+    **The shape is stated, not summarised.** "Keyed by ``lane_id``" read as a
+    mapping, so a lane tested ``"code_context" in aggregated_outputs`` against a
+    list, found nothing and re-investigated -- silently, since a lane with
+    nothing to reuse looks like a project with nothing to reuse.
 
     What this must not become is a second set of rules. There is nothing here
     about proving a stored finding sufficient, reporting what a reuse left
--- a/src/ouroboros/mcp/tools/question_advisory.py
+++ b/src/ouroboros/mcp/tools/question_advisory.py
@@ -307,6 +307,15 @@ def _recent_findings_section(request: Mapping[str, Any]) -> str:
     outside this session's roster is rejected at submission. The child is told
     so here, where it is deciding what to read.
 
+    **The shape is stated exactly, because it was not.** This said the answers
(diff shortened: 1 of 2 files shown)


cohemm:
Thanks for the approval. On the follow-up about `_publish` authoring `aggregated_outputs` — the point is right, and I am not taking it in this PR. Reasons, in order of weight.

**The fixture is not this PR's.** `_publish` and its hand-authored envelope are on `origin/main` at `test_recent_findings.py:70`, from #2154. This branch's diff against `main` is 44 lines: one sentence in the prompt block, its docstring note, one new test, and one existing assertion changed from a conjunction to two lane ids. It does not touch the fixture.

**The scope is a description that was wrong.** The block told a lane the answers were "keyed by `lane_id`"; they are a list of entries carrying one. Changing how the fixture *produces* an envelope does not make that sentence more or less true, and the two changes have no defect in common — this one is about what the prompt says, that one is about what the test builds.

**What the new test does bind is the shape, and it binds it against real data.** It reads a body the store published and asserts the access the block describes: `aggregated_outputs` is a list, `"code_context" in outputs` is `False`, and selecting by `lane_id` finds the entry. So the failure you would want a producer-built fixture to catch — the envelope changing shape under the prompt — is already caught here, because the assertion is about the data rather than about the sentence. Routing the fixture through `submit_fanout_results` would tighten *where* that envelope comes from; it would not add a case this test currently misses.

**And the coverage you cite as making it non-blocking is the right place for it.** `test_fanout_core.py` (38 passed) verifies the production envelope directly. If that contract changes, it fails there first, which is where an envelope change should be reported — not in a prompt-wording regression.

The suggestion is worth doing on its own terms, and I would rather it land as a change to that fixture with its own reasoning than ride in on a one-sentence prompt fix. If you would like it tracked, say so and I will file it; otherwise it stands here in the thread.

The docstring note was also cut from nine lines to four in `bf39b050` — the retelling of how two subagents stumbled belonged in this thread, not in the module.

Issue #2163, opened 2026-08-17
dsh plugin first run: inherited SDK runtime kills the server, and the bundle has to be installed by hand


Q00:
Two friction points hit the same user on the same first run, both found while driving the dsh plugin live.

## 1. An inherited `claude` SDK runtime leaves the host with zero tools

With `orchestrator.runtime_backend: claude` in `~/.ouroboros/config.yaml` (a common state — it is the shipped default), booting dsh with the plugin prints this three times and registers nothing:

```
The MCP server cannot host the 'claude' SDK runtime in this process. Pass an
executable runtime, for example 'ouroboros mcp serve --runtime claude-cli', or
set the runtime in config so this command stops inheriting the 'claude' default.
```

The guard is correct that the in-process SDK runtime cannot run inside the MCP server. What is wrong is the outcome: the user never chose that runtime *here*, it arrived from config or the default, and the fix requires knowing about an environment variable the plugin never mentions. `failOnStartupError: false` then hides it — dsh boots fine, just with no Ouroboros tools.

**Expected:** when the SDK runtime is only inherited and an executable CLI is installed, serve with that instead and say so. An explicit `--runtime claude` or `OUROBOROS_AGENT_RUNTIME=claude` should still fail, because overriding what the caller just said is worse than the error.

## 2. The bundle is the only integration a user installs by hand

`install.sh` sets up every other host it finds. A dsh user has to read a README and paste `dsh plugin --profile <name> add ...`. dsh keeps plugins per profile and requires `--profile`, so there is no global install — but the profile `dsh web` scaffolds is a predictable target, and profiles that already carry the bundle can be refreshed so an existing install picks up a release.

## Out of scope, recorded here because it keeps coming up

Making `dsh` the *default* LLM backend for plugin users is not currently possible, and the reason is worth writing down:

- `@deepseek-ai/dsh-acp-demo` still does not install usably from npm. `npm i` fails on a peer conflict in its own `dsh-tool-bash` chain; installing it through pnpm inside a dsh profile succeeds but the bin then dies with `ERR_MODULE_NOT_FOUND: @deepseek-ai/dsh-app-boot`.
- So `llm.backend: dsh` would fail the first interview call for anyone without a source build.
- The clean version of that idea is not a second dsh process at all: it is MCP sampling, where the server asks the host to run the completion and the user's own dsh model answers with no extra configuration. dsh's `@deepseek-ai/dsh-mcp-client` does not implement sampling today.

[...]

PR #2164, opened 2026-08-17
fix(mcp): serve with an installed CLI when the SDK runtime is only inherited

Q00:
Part of #2163 (first half).

## The failure

Installing the dsh plugin on a machine whose config says `runtime_backend: claude` gives the user this, three times, and zero Ouroboros tools:

```
The MCP server cannot host the 'claude' SDK runtime in this process. Pass an
executable runtime, for example 'ouroboros mcp serve --runtime claude-cli', ...
```

The guard's premise is right: the in-process Claude SDK runtime cannot run inside the MCP server. Its conclusion is what hurts. The user never chose that runtime for this process; it came from config or from the shipped default. And because the plugin sets `failOnStartupError: false`, dsh boots normally and the tools are simply absent, so the error reads as noise rather than as the reason nothing works.

## The change

`serve` now looks for an executable stand-in before refusing:

| Situation | Before | After |
|---|---|---|
| Inherited `claude`, a CLI is installed | exit 1, no tools | serves with that CLI, announced on stderr |
| Inherited `claude`, nothing installed | exit 1 | exit 1, same message |
| `--runtime claude` | exit 1 | exit 1 |
| `OUROBOROS_AGENT_RUNTIME=claude` | exit 1 | exit 1 |

Preference order is `claude-cli`, then `codex`, then `opencode`. `claude-cli` is first because it is the same engine out of process, so a user who inherited the SDK default gets the runtime they meant.

Two decisions worth naming:

- **An explicit selection still fails.** Substituting for a runtime the caller just typed would be worse than the error, and the error already names the fix.
- **The substitution is announced, with the public spelling.** A silent swap would leave a working tool list backed by a runtime nobody picked. The notice says `claude-cli`, not the internal `claude_mcp`, because the latter is a value `--runtime` rejects.

## Verification

Not inferred from the guard. In a real MCP 2 environment, with `runtime_backend: claude` in config and no environment override:

```
$ uvx --isolated --from '<repo>[mcp]' ouroboros mcp serve
Inherited the 'claude' SDK runtime, which cannot run inside the MCP server;
serving with 'claude-cli' instead. Pass --runtime or set OUROBOROS_AGENT_RUNTIME to choose.
Registered 36 tools
```

and `initialize` answered on stdout. The same config previously exited 1.

`tests/unit/cli/test_mcp_nested_guard.py`: 21 passed, covering the stand-in, the ordering, both explicit-selection refusals, and the nothing-installed refusal.

**Two existing tests were reading the developer's own PATH.** `test_bare_serve_rejects_missing_config_sdk_default_before_mutation` and `test_bare_serve_rejects_configured_sdk_before_mutation` passed on a machine without `claude`/`codex` and would have failed on one with them. Both now pin the CLI probe, so the outcome is the same in CI and on a laptop.

### DIFF
--- a/src/ouroboros/cli/commands/mcp.py
+++ b/src/ouroboros/cli/commands/mcp.py
@@ -1244,6 +1244,11 @@ def serve(
         ouroboros mcp serve --runtime codex --llm-backend codex
 
     """
+    # Reject recursive server launches before shell hydration or any persistent
+    # cache/environment mutation. The parent runtime already owns the MCP edge.
+    if os.environ.get("_OUROBOROS_NESTED"):
+        _stderr_console.print("[dim]Nested ouroboros MCP server detected — exiting cleanly[/dim]")
+        raise typer.Exit(0)
     # Detached MCP hosts often inherit a minimal environment. Hydrate before
     # resolving selector provenance so login-shell/cache OUROBOROS_* choices
     # remain authoritative rather than being mistaken for the shipped default.
@@ -1280,14 +1285,6 @@ def serve(
         )
         selected_runtime = standin_backend
 
-    # Guard: prevent recursive MCP server spawning.
-    # When ouroboros spawns a runtime (Codex/Claude/OpenCode), the child process
-    # inherits this env var. If that runtime's MCP config tries to spawn another
-    # ouroboros server, the nested instance exits cleanly instead of creating a
-    # process tree explosion.
-    if os.environ.get("_OUROBOROS_NESTED"):
-        _stderr_console.print("[dim]Nested ouroboros MCP server detected — exiting cleanly[/dim]")
-        raise typer.Exit(0)
     os.environ["_OUROBOROS_NESTED"] = "1"
 
     # Transport is re-validated inside _run_mcp_server; normalize here first so
--- a/src/ouroboros/cli/commands/mcp.py
+++ b/src/ouroboros/cli/commands/mcp.py
@@ -11,6 +11,7 @@ from enum import Enum
 import json
 import os
 from pathlib import Path
+import shutil
 import signal
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


Q00:
One consequence of this change that I did not see when I opened it, found while verifying a separate claim about dispatch modes.

The stand-in preference order changes the subagent dispatch mode, because that mode is resolved from the runtime backend's declared capability, not from the host:

| backend | `supports_host_driven_subagents` | `resolve_subagent_dispatch` |
|---|---|---|
| `claude` (the SDK default being substituted) | true (`capabilities.py:455`) | `HOST_DRIVEN` |
| `claude_mcp` (this PR's first choice) | not declared (`capabilities.py:495-501`) | `SEQUENTIAL` |
| `codex` (second choice) | true (`capabilities.py:472`) | `HOST_DRIVEN` |

So a user whose config said `claude` and who has the Claude CLI installed now gets advisory lanes run sequentially inside the server rather than fanned out to the host model. In an MCP host like dsh that is visible: the lanes stop appearing in the chat.

This is not a regression against a working state — before this change the server exited and the host got no tools at all — but it is a capability quietly downgraded by a substitution, which is worth naming rather than leaving for someone to discover.

I kept `claude-cli` first on purpose. It is the same engine the config asked for, and switching a user to a different vendor silently seems worse than losing a dispatch optimization. Reordering is a one-line change if maintainers weigh it the other way.

Recorded in #2163 as well so the tradeoff is not only in a PR comment.

Q00:
Addressed both exact-head blockers in `c4770485c`:

1. `OUROBOROS_RUNTIME` now has the same explicit-selection authority as `OUROBOROS_AGENT_RUNTIME`; neither is silently substituted.
2. Stand-in discovery now hydrates the login-shell environment and uses `get_cli_path`, `get_codex_cli_path`, and `get_opencode_cli_path` before bare-name PATH lookup, matching runtime construction.

Added focused regressions for the legacy selector, a configured non-PATH Claude executable, and login-shell-only PATH recovery. `test_mcp_nested_guard.py`, `test_mcp_shell_env.py`, and `test_mcp_startup_cleanup.py`: 65 passed. Please re-review exact HEAD.

PR #2165, opened 2026-08-17
feat(install): install the dsh plugin when DeepSeek Harness is present

Q00:
Part of #2163 (second half). Independent of #2164; different files.

## Why

`install.sh` sets up every host it detects. A dsh user was the only one who then had to read a README and paste a command to get any Ouroboros tools.

## What it does

dsh keeps plugins per profile and `dsh plugin` requires `--profile`, so there is no global install to run. The installer therefore picks its targets:

- **`web`** — the profile `dsh web` scaffolds, and the one a dsh user actually boots. Installing into it also creates it if absent, which is the same thing `dsh web` would do.
- **Any other profile whose `package.json` already carries `dsh-ouroboros`** — this is how an existing install picks up a release.
- **Nothing else.** Adding Ouroboros tools to someone's unrelated profile because an installer ran is not an upgrade.

Failure is non-fatal and prints the manual command. Ouroboros is already on disk by that point, and a host that cannot install a plugin must not fail an install that just succeeded.

## Verification

Against the real `dsh` CLI (0.1.0-rc.6) in a throwaway `DSH_HOME`, not only against a stub:

```
OK: web
deps:    {'dsh-ouroboros': 'github:Q00/ouroboros#main&path:integrations/dsh-plugin'}
bundles: ['@deepseek-ai/dsh-base', '@deepseek-ai/dsh-web-app', 'dsh-ouroboros']
```

`tests/unit/scripts/test_install_runtime_selection.py`: 97 passed, including four new cases — no dsh means no profile is touched, `web` is covered blind, an opted-in profile is refreshed while an unrelated one is not, and a failing `dsh` leaves the install green with the manual command printed.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -939,6 +939,8 @@ HAS_KIRO=false
 HAS_COPILOT=false
 HAS_PI=false
 HAS_GJC=false
+HAS_DSH=false
+DSH_PLUGIN_SPEC="github:Q00/ouroboros#main&path:integrations/dsh-plugin"
 if command -v codex &>/dev/null; then
   _ok "Codex found: $(which codex)"
   HAS_CODEX=true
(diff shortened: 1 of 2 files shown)


sergiobuilds (approved):
Reviewed exact head af2921ac7d69ae272858185ed596e187633d0ac9. Exact dependency matching fails closed for malformed and unrelated profiles, profile names remain single arguments, focused DSH tests pass, and all required CI checks are green.

Issue #2168, opened 2026-08-17
RFC: the artifact store becomes a SQLite database


cohemm:
## The decision

**The project-local artifact store becomes a SQLite database, and nothing that exists only because it was a filesystem comes with it.**

The test applied throughout is one question, asked of every record, field and check:

> Does this exist because of what is stored, or because of how it was stored?

Whatever answers *how* is removed rather than translated. Carrying it forward would mean paying a filesystem's costs inside a database — the shape of a problem that no longer exists.

**Removal is half of it; the other half is re-solving.** What a filesystem forced into one shape is done again in the shape a database is good at — a directory walk becomes a query, records that had to be compared across files become one row, an ordering that had to be spelled into filenames becomes a clause. A translation that keeps the old shape and merely changes where the bytes live would leave the cost and drop the benefit.

**And one constraint governs all of it: no caller has a reason to behave differently.** The same bodies come back for the same contract ids, the same conditions raise the same errors, the same work succeeds and fails in the same cases. Storage is free to change completely; behaviour is not free to change at all.

The constraint is on behaviour, not on field lists. Three things it therefore permits, each identified where it happens below: a response may lose a value that nothing reads, since no caller acts on it; a failure mode may stop being reachable, when what produced it cannot occur any more; and a write path with no caller may go, since nothing is calling it to notice. What may not change is what a caller gets back, what it is refused, and when.

Three consequences are large enough to be decisions in their own right, and each is argued below: the trusted authority outside the store goes, content addressing goes, and a lane stops being handed filesystem paths.

## What the store holds today

Nine record kinds across two locations. The first three are under the project's artifact root — a replaceable generation. The rest are in the stable parent and are the trusted authority.

| Record | Address | Fields |
|---|---|---|
| Body | `<digest[:2]>/<digest>.json` | canonical JSON, ≤ 1 MiB, addressed by `sha256:<hex>` |
| Manifest | `contracts/<hash(contract_id)>/events.json` | `schema_version`, `contract_id`, `active`, `retain_until`, `updated_at`, `events[]` — ≤ 64 KiB |
| Binding | `bindings/<hash(contract_id)>.json` | `schema_version`, `contract_id`, `artifact_ref`, `size_bytes`, `envelope`, `referenced_at`, `initial_active`, `initial_retain_until` — ≤ 8 KiB |
| Authority | `…-authority-<hash(root)>-<hash(contract_id)>.json` | identical to the binding |
[...]

Issue #2167, opened 2026-08-17
RFC: a lane reads only what its own lane produced


cohemm:
## The decision

**A lane may read recent findings only from its own lane.** `code_context` answers from earlier `code_context` findings, `data_context` from earlier `data_context` ones, and neither reads the other. Every remaining lane reads nothing.

RFC #2153 decided which lanes *produce* a fact that keeps. This decides that the same test governs who may *read* one, and the answer it gives is the same list applied per lane rather than as a pool.

## What #2153 left open, and what filled it

#2153 closed one list: what may be reused is what `code_context` and `data_context` produce, and nothing else. Its criterion was stated once and used throughout — **what decides eligibility is what a lane produces.**

It said nothing about who may read. That silence was filled during implementation with the widest available answer: every lane of the turn receives the same findings, and the eligible list acts only as a filter on what is inside them. The reasoning was that a finding is a property of the project rather than of the lane, so a per-lane copy would be one more text to keep in step for no difference in what it says.

That reasoning exists in a merged pull request body and one code comment. It was never tested against the criterion the RFC actually used, and it does not survive that test.

## The claim

**A lane that does not produce a fact that keeps does not consume one either.**

The argument for reuse in #2153 is Assumption A: the same person, working on the same product, does the same work repeatedly, so findings repeat. That repetition is not a property of the interview. It is a property of a lane. `code_context` reads the same subsystems today that it read yesterday; `data_context` measures the same aggregates. What makes yesterday's finding a plausible answer today is that the *same lane* is being asked to do the *same work* against a system that has barely moved.

A lane that challenges a question, drafts answer options, or raises architecture implications does no repeated work of that kind. Handing it a `code_context` finding is not a cache hit — nothing it was about to do is being skipped. It is an additional input it never had.

#2153 already made this argument in the other direction. Feeding a previous round's answer options into a new question, it says, "is not a cache hit. It is an answer to a question nobody asked." The same sentence holds with the terms exchanged: feeding a previous round's code facts to a lane whose job is to reason about *this* question is not reuse either. It is a new capability wearing reuse's clothes.

That may well be a good capability. It is not this one, and it does not inherit this one's justification. Adding it later is a decision with its own reasoning to do — the position #2153 took on the web lane, taken consistently.

## Why the two eligible lanes do not read each other

[...]

PR #436, opened 2026-04-15
feat(events): add event_version to BaseEvent payload

shaun0927:
## Summary

All persisted events now include `event_version` inside their JSON payload, giving consumers a stable schema contract they can check before parsing.

- `BaseEvent` gains an `event_version: int` field (default `1`)
- `to_db_dict()` injects `event_version` into the persisted payload
- `from_db_row()` extracts it (defaults to `0` for legacy rows without the field)
- `event_version` is stripped from `event.data` on deserialization — it does not leak into application-level data
- `docs/events.md` documents payload schemas for the top 7 event types
- 7 new tests covering version lifecycle, legacy backward compat, roundtrip, and corruption handling

## Problem

EventStore payloads are `dict[str, Any]` with no versioned schema. As consumers grow — TUI, `ooo status`, `ooo resume` (#430), OpenClaw channel workflows (#320/#323) — any internal field rename silently breaks downstream readers. This is inconsistent with Ouroboros's own "Seed is immutable" philosophy: the system demands schema discipline from users but not from itself.

## Why now

- `ooo resume` (#430) reads the EventStore directly without MCP mediation — the tightest payload coupling yet
- OpenClaw channel integration (#320, #323) exposes events cross-project
- #188 showed that inconsistent `runtime_status` fields in payloads caused cancelled sessions to appear running — a schema contract would have caught this at the source

## Design decisions

| Decision | Rationale |
|---|---|
| Version lives inside `payload` JSON, not a new DB column | No schema migration needed, fully additive |
| `event_version=1` for new events, `0` for legacy | Consumers can distinguish old vs. new without guessing |
| Centralized in `BaseEvent.to_db_dict()` / `from_db_row()` | Zero changes to event factories or emitters |
| Non-int versions fall back to `0` | Defensive against corrupted payloads |

## Scope

- **Changed:** `src/ouroboros/events/base.py` (core change), `tests/unit/events/test_base.py` (updated + 7 new tests)
- **Added:** `docs/events.md` (schema reference)
- **Not changed:** `persistence/schema.py`, `persistence/event_store.py`, any event factory (`events/evaluation.py`, `events/interview.py`, etc.), any consumer

## Test plan

- [x] All 19 `test_base.py` tests pass (12 existing updated + 7 new)
- [x] All 30 `test_event_store.py` tests pass (zero changes needed — `from_db_row` strips `event_version` cleanly)
- [x] All 37 `tests/unit/events/` tests pass
- [x] `ruff check` clean
[...]

PR #2171, opened 2026-08-17
feat(artifacts): store disposable artifacts in SQLite

cohemm:
Implements RFC #2168.

## The decision

The project-local artifact store moves from files and directories to SQLite, and nothing that exists only because it was a filesystem comes with it. One question was asked of every record, field and check:

> Does this exist because of *what* is stored, or because of *how* it was stored?

Whatever answered *how* was removed rather than translated. Carrying it forward would mean paying a filesystem's costs inside a database — the shape of a problem that no longer exists.

**Removal is half of it; the other half is re-solving.** A directory walk becomes a query, records that had to be compared across files become one row, an ordering spelled into filenames becomes a clause.

**And one constraint governs all of it: no caller has a reason to behave differently.** Storage changes completely; behaviour does not change at all.

## What was there

Nine record kinds across two locations. Inside the project: a content-addressed body file, a per-contract manifest, a binding. Outside it, as a trusted authority: an anchor, a completion marker, a lifecycle genesis, an epoch chain bounded at 4096, a separately committed head, a tombstone. Plus two cross-process file locks.

A publication was **seven ordered writes**, and because any of them could be interrupted, each file carried digests naming its own membership and position so a later reader could work out how much of what it found was valid. Reading therefore *repaired* — which is why reads took the store lock exclusively.

That is one cost, not nine:

```
no transaction
  → seven ordered writes
    → any can be interrupted
      → validity decided afterwards
        → each file states its membership and position (the chain)
          → the chain is verified on every read
            → repairs happen during reads
              → reads take an exclusive lock
```

## What it is now

```sql
CREATE TABLE artifacts (
  contract_id   TEXT PRIMARY KEY NOT NULL COLLATE BINARY,
  body          TEXT,
  kind          TEXT GENERATED ALWAYS AS (json_extract(body,'$.kind')) VIRTUAL,
  created_at    TEXT NOT NULL,
  updated_at    TEXT NOT NULL,
  duration_ms   INTEGER NOT NULL,
  pruned_reason TEXT
)
```

| | before | after |
|---|---|---|
| `artifact_store.py` | 2,075 lines | **549** |
| its test file | 2,131 lines | **~370** (22 nodes) |
| filesystem layer | 1,460 lines | **deleted** |
| whole diff | | **+1,311 / −6,053** |

`artifact_store.py` had been sitting at 1,999 lines against a 2,000-line cap — code belonging to it had already been pushed into another module for that reason, with a comment saying so. A store that cannot absorb its own next feature is being paid for continuously.

## Commits

Each is independently reviewable and the tree is working after each one.

1. **`52c5145`** — the store answers *when a publication happened*. `recent_findings` walked the store's contract directory, which made it the last reader of the artifact layout outside the persistence package. That walk moves in as `published_contracts()`. A lane also stops being handed a filesystem path: it receives its own lane's findings as **content**, and the paragraph telling a child how to select its entries out of `result.aggregated_outputs` is deleted. That wording has already cost a lane its findings once (#2162) — deleting it is the point, not a side effect.
2. **`f702c95`** — duplicate execution is permitted, and the reference event is keyed by contract. Details below.
3. **`0238f9a`** — the cutover.
4. **`a1b2780`** — the filesystem layer is deleted, now that nothing imports it.
5. **`c445f5d`** — `ContentAddressedArtifactStore` → `ArtifactStore`, and the documented contracts follow.

## Two removals worth arguing

**Content addressing.** The store was named for it. Its three jobs: give a body an address (the address was a file path — there are no paths), reuse an identical body (never happens; contract and body are 1:1 today), detect a body not matching its name (name and content were separate files; one row has no such pair). `artifact_ref` leaves the envelope with it — a public contract change whose consumers are thin: **no code reads the value to decide anything**, and its one real use was assembling a path.

**Both locks, for different reasons.** The store lock guarded *data*, and a transaction replaces it — reads stop repairing, so they need no lock, and every read here tolerates a stale one (the worst outcome is work repeated). The contract execution lock guarded *work*: it was held across the child's execution, so no transaction could take its place. Without it both callers run, one loses on the key, and the existing comparison decides — equal bodies collapse, differing bodies are refused exactly as before. **The write is an insert, not an update, so there is no lost update to fear.**

## What the review should check

The RFC's decisions rest on facts about *this* tree, each checkable by search. If one is wrong, the argument that rests on it is wrong:

| Claim | Where it is load-bearing |
|---|---|
| `set_contract_retention` has no caller outside tests | its removal, and `active`/`retain_until` becoming derivable |
| no code reads `artifact_ref` to decide anything | content addressing's removal |
| the only work under a contract is a pure synthesis with no awaits | duplicate execution being permitted |
| a published body carries its own fan-out identity, so contract and body are 1:1 | no second table |

Two details were easy to get wrong and are guarded deliberately:

- **An absent body means pruned — and a published body may legitimately BE JSON null.** Null is stored as the four-byte text `null`, so SQL NULL in that column is exclusively what pruning wrote. A test pins it.
- **Envelope reads must not materialise the body.** Proven with an SQLite authorizer that DENYs reads of `artifacts.body`: the envelope read succeeds, and `fetch` under the same denial fails, which proves the probe bites. This catches `SELECT *` as well as a named column.

Every `sqlite3.Error` is wrapped in `ArtifactStoreError` — the CLI and the MCP fetch handler catch `(ArtifactStoreError, OSError, ValueError)` and would not have caught a bare one.

## Verification

```
uv run pytest tests/unit -q          20,222 passed
python3 scripts/check-module-size.py OK (585 modules, cap 2000)
uv run ruff check src/ tests/        All checks passed
uv run ruff format --check           1,349 files formatted
uv run mypy src/ouroboros/persistence src/ouroboros/mcp   Success, 98 files
```

11 failures under `tests/unit/cli/` (nested-guard, startup-cleanup, install/mcp doctor) are **pre-existing**: they fail identically on `235487bb`, the branch base, verified in a clean worktree. They are ANSI/locale assertions and touch nothing in this change.

There is **no migration**. Findings are disposable, expire on an existing schedule, and are re-derivable by the work that produced them; a store that starts empty costs a re-investigation, which is the cost this system already pays whenever it has nothing recent.


cohemm:
Addressed review round 1 in `a7896c18` — 2 blockers fixed, 1 defended, plus one defect the review did not raise and one non-blocking suggestion. RFC #2168 was amended first where this changes a decision it had made.

**Blocker 2 — discarded envelope provenance (`artifact_store.py:177`) — fixed.**
Your probe is exact, and the RFC's own reasoning is where it went wrong, so the RFC was corrected rather than worked around. It grouped `runtime_id` and `events_emitted_count` with `result.status` and `active` as "fields with one value today". `result.status` is the store's own conclusion — assembling it on read changes nothing. The other two arrive on the same call as `body`, which puts them on the caller's side of the line the RFC itself draws one paragraph earlier ("`body` stays whole because it is not the store's"). And the value was not preserved but replaced: the one publisher in this tree passes `mcp:fanout-synthesis:v2`, so `"artifact-store"` was never the value it had.

Two columns now. Your probe returns what it published, and a collapsed republication returns the *stored* provenance — same rule that already governed `duration_ms`.
Pinned: `test_publisher_provenance_is_returned_rather_than_replaced` asserts write-return, all three read paths, and that a second publication claiming `some-other-runtime` collapses into the record rather than overwriting it.

**Blocker 3 — unbounded recent-findings prompt (`question_advisory.py:329`) — fixed.**
Confirmed: twenty entries is a count, and a count says nothing about how much twenty of something is. The budget lands at the retrieval interface, which is where RFC #2168 left "how much returns at once".

Two details are deliberate, because both are places a budget usually turns out not to be one:
- The rendering the budget measures **is** the rendering the prompt ships — they are one function now, so no second rendering exists for the measurement to be wrong about. This also closes the variant of your finding where a budget is computed on compact JSON and the prompt emits indented.
- A finding too large to carry whole is carried **shortened**, not dropped. A budget that can return nothing produces exactly the silence this mechanism exists to prevent — a lane told a project has nothing when it has something (#2167, the one property RFC #2168 says binds implementation). So the ceiling holds with nothing added on top of it, and there is no "budget plus one oversized entry" worst case.

Pinned: `test_many_large_findings_do_not_grow_a_lane_prompt_without_bound` publishes ten ~120 KB findings and measures what they add to a lane prompt against the same prompt built with no store; `test_one_oversized_finding_is_shortened_rather_than_silently_dropped` publishes a single 200 KB finding and asserts the lane still receives it, marked as cut, inside the same bound.

**Blocker 1 — project containment and symlink refusal (`artifact_store.py:153`) — position, with one part of it fixed.**
Your probe reproduces: `.ouroboros/artifacts -> ../external` does redirect the database, and this design does not detect it.

Restoring the checks is what we are not doing, and RFC #2168 has been amended to say so where it previously overstated the case as "no paths". Every path the old store validated was one it *composed* from a caller's value — a digest into a blob directory, a hashed contract id into a manifest directory, a padded sequence into an epoch filename. That is what made containment load-bearing: a caller's value reached the filesystem. One path survives and nothing composes it — `<project>/.ouroboros/artifacts/artifacts.db`, fixed relative to a project root the caller already chose. No contract id, digest or body reaches it, so the defect class those checks existed for has no remaining instance.

What a check here would defend is a workspace hostile in exactly the one place the store looks. Planting that symlink requires writing into the project, and anything that can write into the project can more cheaply edit the source a finding would cite, or the code about to run — the trust boundary RFC #2153 already states and this store lives inside.

What *was* a real defect is the advertisement: `ArtifactStore.__init__` took a `project_root` and never consulted it, which is the shape of a containment guarantee that stopped existing without saying so. Removed with the checks.

Revisit condition: show a path into the store that a caller's value still composes, or a reachable case where the artifact root is attacker-controlled while the project's source is not, and this becomes a fix rather than a position.

**Not raised in review, fixed here: pruning did not return the space it reported.**
`prune(apply=True)` cleared `body` and reported `removed_bytes`; the CLI printed it as bytes removed. Clearing a column frees pages back to the database and leaves the file the size it was, so the number was a claim about space nobody could spend. RFC #2168 names this as the single property its "tuning is not decided here" section does not defer. An applied prune that removed anything now vacuums.
Pinned: `test_applied_prune_returns_the_space_rather_than_only_the_row` asserts the file itself shrinks.

**Non-blocking 2 (`cli/commands/artifacts.py:117`) — fixed.** The fetch argument help no longer names a reference that does not exist.

**Non-blocking 1 (`docs/rfc/contract-ledger.md:33`) — not in this PR, by maintainer decision.** The superseded banner was written and then reverted: `docs/rfc/` is left untouched here. The same stale `artifact_ref` text is in `mesh.md` as well, so this is a docs-scope question rather than a line, and it is not being decided inside a storage change.

**One thing worth naming before it is found.** The two new columns mean an `artifacts.db` written by an earlier revision of this branch lacks them. The SQLite store has never shipped, so no deployed table has the previous shape; a developer who ran an earlier revision deletes `.ouroboros/artifacts/artifacts.db`. Adding a reconciliation path for a table that never existed outside this branch would be migration machinery for a design whose stated position is that there is no migration.

Verification (four axes, run in parallel):
- `tests/unit/persistence tests/unit/orchestrator/test_disposable_memory.py tests/unit/cli/test_artifacts_command.py tests/unit/events` — 574 passed
- `tests/unit/mcp tests/unit/orchestrator/test_capabilities.py` — 2280 passed, 1 skipped
- `tests/integration/mcp/test_server_adapter.py tests/e2e/test_cli_commands.py` — 82 passed
- `ruff check` / `ruff format --check` (1349 files) / `mypy src/ouroboros/{persistence,mcp,cli}` (144 files) / `check-module-size.py` (585 modules, cap 2000) — all pass

cohemm:
Addressed review round 2 in `6b71edce` — 2 blockers and the byte-accounting warning, all fixed. RFC #2168 amended first where this round falsified one of its factual claims.

**Blocker 1 — cancel-key aliasing (`core/disposable_memory.py:43`) — fixed, but not where the review points.**

Your probe reproduces exactly. Before fixing it I checked whether the widened contract-id space caused it, and it does not:

```
'a:b'       vs 'a_b'        old_regex_allows_both=True   collides=True
'a/b'       vs 'a_b'        old_regex_allows_both=False  collides=True
'mcp_job:7' vs 'mcp_job_7'  old_regex_allows_both=True   collides=True
```

`:` and `_` are both in `^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$`, and `_sanitize_seed_id` folds the first onto the second — so `a:b` and `a_b` named one cancel checkpoint on `origin/main`, before this PR touched the identifier rule. `mcp_job:<id>` is not hypothetical either: `mcp/tools/background.py:306` passes exactly that as `cancel_key` today.

That makes your first suggested remedy — keep the public identifier space injective — the one thing that cannot work here. It would restore a path-safe-filename rule to a store that has no filenames *and* leave the collision standing for the caller that already has it.

Your second remedy is the one taken: `_cancel_checkpoint_seed` digests the process id instead of embedding it, which is how `_pause_checkpoint_seed_id` already derives the other control-plane key in that same module. Injective for every id a caller can hold, and it closes the collision for `mcp_job:` keys too.

`_validate_process_id_namespace` goes with it. It refused ids beginning with the control prefix, which mattered only while a caller's bytes reached the seed's structure; they no longer do. Its two tests asserted the refusal — the mechanism — so they are replaced by one asserting the property they existed to protect, using the exact two ids they used: `__ouroboros_agent_process_cancel__:external` and `__ouroboros/agent_process_cancel__:external` each get their own key and reach nobody else's.

Named rather than buried, since it is a durable-state change: a cancel checkpoint's filename no longer shows which process it belongs to, and a cancel written before this change is not found after it — a restart that would have observed a cancellation resumes instead. Both are the trade the pause key already made.

Pinned: `test_cancel_authority_never_crosses_between_distinct_process_ids` walks all three pairs above and asserts the bystander stays uncancelled; `test_the_control_namespace_is_not_reachable_by_naming_it` covers the retired refusal's two inputs. Both fail on the previous derivation.

**Blocker 2 — malformed record aborts retrieval (`recent_findings.py:165`) — fixed.**

Confirmed, and the consequence is worse than the crash: retrieval returning nothing is indistinguishable from a project with nothing, so one odd record silently cost every finding published that day for the whole freshness window.

Fixed in both directions, deliberately. The shape read is total — `result` is now type-checked like the two levels around it already were — and the extraction moved *inside* the per-record `try` that already wrapped `fetch`. The second is the part that matters: a body is whatever was published, so the shapes it can take are not a list this module can finish enumerating, and fail-open should hold for shapes I have not thought of rather than for the ones I have.

Pinned: `test_a_record_of_an_unexpected_shape_costs_only_itself` publishes `{"kind":"question_advisory","result":[]}` — your probe — followed by a valid finding, and asserts the valid one still arrives.

**Follow-up warning — character count reported as bytes (`artifact_store.py:398`) — fixed.**

`length()` on TEXT counts characters while `body_bytes`, `removed_bytes` and the CLI's `B` all say bytes. `length(CAST(body AS BLOB))` now.
Pinned: `test_prune_reports_encoded_bytes_not_characters` uses a 1,000-character accented body and asserts the report equals `len(canonical_artifact_bytes(body))` — 2,014 rather than 1,014.

**Consumers and next addresses checked.** Nothing enumerates checkpoints by filename, so the hashed key has no reader outside the two functions deriving it. The seed is a constant 99 characters against the store's 232-character budget, so no truncation path becomes newly reachable. And since the underlying concern is "a raw identifier reaching a lossy representation", I swept the rest: `process_id` into event `aggregate_id`/`target_id` is stored verbatim, the pause seed was already digested, and the artifact PK is `COLLATE BINARY`. The cancel seed was the only lossy sink.

Verification:
- `tests/unit/orchestrator tests/unit/persistence tests/unit/events` — 5797 passed, 1 skipped
- `tests/unit/mcp tests/unit/cli` — 4121 passed, 8 skipped
- `tests/integration` + `tests/e2e` (agent-process lifecycle, cancellation, background jobs, checkpoints, disposable memory, artifacts CLI) — 199 passed
- `ruff check` / `ruff format --check` (1349 files) / `mypy` (267 files) / `check-module-size.py` — all pass

On the 11 CLI failures reported in the PR body as pre-existing: they are `FORCE_COLOR` in the shell environment, not locale. With `NO_COLOR=1 COLUMNS=200` the same scope is green, and they reproduce identically on `origin/main`.

cohemm:
Addressed review round 3 in `23e5515d` — both blockers fixed. Both were right, and the second one was mine from last round.

**Blocker 1 — envelope v1 redefined in place (`core/disposable_memory.py:39`) — fixed. The envelope is version 2.**

Accepted without qualification. My round-1 defence of this field's removal argued that its consumers are thin, and that is true but answers a different question: thin consumers make the change cheap, they do not make it additive. Removing a required field is exactly what this repository's schema rule ([#436](https://github.com/Q00/ouroboros/pull/436)) does not permit silently.

And there is a defect here that does not depend on any external consumer existing. This envelope is the payload of `artifact.referenced`, and that store is append-only: rows written before this PR carry `artifact_ref` and say version 1. Leaving the number at 1 would leave two shapes in one store both claiming to be version 1, with nothing able to say which one a given row is — the single thing the number exists to answer.

Rows written before this keep saying 1 and keep their `artifact_ref`. New rows say 2. That is the compatibility handling: a reader that needs to know which shape it holds asks the field that exists in order to tell it.
Pinned: `test_the_envelope_that_dropped_a_required_field_says_so_in_its_version` asserts the envelope and the emitted event payload both say 2 and that `artifact_ref` is absent.

I also swept for anything pinning the old number before bumping it — no code in `src/` or `tests/` constructs the envelope with an explicit `schema_version` or compares it to 1. `docs/events.md` did document it as "currently `1`" and is corrected in this commit.

**Blocker 2 — legacy durable cancellations unreachable (`agent_process.py:640`) — fixed, and this one was mine.**

Reproduced: a checkpoint written by the released derivation returns `(False, None)` after the upgrade.

Worth naming plainly, because I introduced it last round and said so in that commit rather than noticing what it meant. I argued elsewhere in this PR that no migration is owed because `artifacts.db` has never shipped — which is correct — and then carried that reasoning to state that *has* shipped and may exist on any machine running the previous version. The two are not the same case. An upgrade that stops finding a cancellation resumes work an operator stopped, which is what durability was for.

The old seed is read when the current one says nothing, and never written. Two properties keep the fallback from reopening what the digest closed:

- **Identity is checked before the cancellation counts.** The old seed is reachable by a lookalike — `a_b` opens the file written for `a:b` — so a legacy cancellation is honoured only when its stored `cancel_key` is exactly the id asking. A marker too old to carry that field is not honoured at all: an unidentifiable cancellation fails closed rather than landing on whoever happens to ask.
- **Nothing migrates, so no read writes.** A cancellation read from the old seed is consumed under the digest seed by `_consume_persisted_cancel`, which already existed, and a consumed marker there ends every later read before it reaches back to the old key. The old key retires itself through the path that was already there.

Probe results after the fix:

```
legacy honoured        : (True, 'operator stop')
legacy for 'a:b'       : (True, 'operator stop')
lookalike 'a_b'        : (False, None)      <- aliasing stays closed
marker with no cancel_key: (False, None)    <- fails closed
after consume          : (False, None)      <- old key retired
```

Pinned: `test_a_cancellation_written_before_the_key_changed_survives_the_upgrade` covers recovery and self-retirement; `test_reading_the_old_key_does_not_bring_its_aliasing_back` covers the lookalike and the unidentifiable marker.

Verification:
- `tests/unit/orchestrator tests/unit/persistence tests/unit/events` — 5800 passed, 1 skipped
- `tests/unit/mcp tests/unit/cli` — 4121 passed, 8 skipped
- `tests/integration tests/e2e` — 337 passed, 4 skipped, 1 xfailed, and 1 failure: `test_wheel_packaging.py::test_built_wheel_preserves_packaging_contracts`. Pre-existing and environmental, verified both ways — it reproduces identically on a clean `origin/main` worktree, and passes in this tree under `NO_COLOR=1`. Same `FORCE_COLOR` artifact as the CLI failures noted last round.
- `ruff check` / `ruff format --check` (1349 files) / `mypy src/ouroboros` (586 files) / `check-module-size.py` — all pass

RFC #2168 amended in both places: it had accepted the cancellation loss, and it had argued `artifact_ref`'s removal without saying what becomes of the version that required it.

cohemm:
Addressed review round 4 in `70178c97`. Both blockers fixed, and both fixes are reverts of things this PR added — not new handling around them.

**Blocker 1 — legacy `artifact.referenced` masks a new publication (`orchestrator/disposable_memory.py:207`) — fixed by putting the exactly-once key back on the event id.**

Reproduced exactly: a v1 row in the ledger, an empty SQLite store, a re-run of the same contract → new body `{"new": true}` published, envelope v2 returned, and the ledger left holding only the stale v1 row with `runtime_id: "old-runtime"` and an `artifact_ref` to a deleted file.

The cause is in this PR's own history rather than in the interaction. `f702c95` widened the check from the event id to the event type, and stated the goal in its message: *"a database written before this change still dedupes rather than appending a second row."* That was the right thing to want when the old store was staying and the wrong thing once the same PR stopped migrating it. A pre-cutover row names a body in a store this change deletes; it is not this publication and must not stand in for it.

So the check goes back to the id. A pre-cutover row carries the older contract-plus-ref id, does not match, and the new publication records itself. Exactly-once still holds for the publication this version writes, which is what the check existed for.

On the second row this produces: it is not a state to reconcile. The contract published twice, into two different stores, and the ledger is append-only history. `docs/events.md` said "matches on the event type" and now says what it does.

I did not build dedup keyed by schema version, which was my first instinct after adding the version in round 3. Keying deduplication by storage generation would be ceremony for data this PR declines to migrate on the grounds that it is disposable — the versioning stays where it is and nothing is built on it.

Pinned: `test_a_row_from_the_old_store_does_not_stand_in_for_a_new_publication` reproduces your probe and asserts the ledger ends with both rows and that the newer one carries the current runtime; `test_a_recovered_contract_appends_its_reference_only_once` keeps the exactly-once case that the check is actually for.

**Blocker 2 — `VACUUM` failure reported as prune failure (`persistence/artifact_store.py:430`) — fixed. Compaction is best effort.**

Reproduced, including the part that makes it worse than a bad message: after the forced failure the artifact was already terminally tombstoned, and a retry reported `0 bytes` because the rows no longer matched `body IS NOT NULL` — so the operator is told it failed, and cannot then find out what was destroyed.

This one is mine. I added `VACUUM` in round 1 and put it inside the block whose failure raises `ArtifactStoreError`, which made a compaction error speak for a transaction that had already committed. Your point about disk pressure is the sharp end of it: `VACUUM` needs temporary space, so the likeliest time to fail is exactly when a caller is pruning to reclaim some.

The commit decides the result now. Compaction runs after it and its failure is ignored.
Pinned: `test_a_failed_compaction_does_not_report_the_prune_as_failed` forces `sqlite3.OperationalError` from `VACUUM` and asserts the prune reports the removal and the body is tombstoned.

**Follow-up 1 / Non-blocking 1 (`docs/rfc/mesh.md:119`, `docs/rfc/contract-ledger.md:33`) — not in this PR, by maintainer decision.** I wrote those superseded banners in round 1 and reverted them: `docs/rfc/` is left untouched here. The same stale `artifact_ref` text appears in both files, so this is a docs-scope question rather than two lines, and it is not being settled inside a storage change.

Verification:
- `tests/unit/orchestrator tests/unit/persistence tests/unit/events` — 5802 passed, 1 skipped
- `tests/unit/mcp tests/unit/cli` — 4121 passed, 8 skipped
- `tests/integration tests/e2e` — 338 passed, 4 skipped, 1 xfailed
- `ruff check` / `ruff format --check` (1349 files) / `mypy src/ouroboros` (586 files) / `check-module-size.py` — all pass
- Swept for any remaining type-based dedup: none. Every surviving `type == "artifact.referenced"` is a list filter in tests, not a dedup rule.

cohemm:
Thanks for the approval. Follow-ups triaged in `c086566d` — two fixed, three not, with the reasoning for each below so the decisions are on the record rather than implied by silence.

## Fixed

**`artifact_store.py:464` — an incomplete `artifacts.db` blocks every read.** Reproduced, and it is worse than the read failing: `DisposableMemory.run` probes for an existing publication *before* it writes, so the failing read sits in front of the write that would create the table. Three consecutive runs raised `ArtifactStoreError` and the table was still absent afterwards. Nothing recovers it but deleting the file, and the error does not say so.

Also worth naming: a crash in the microseconds between `connect` and the `CREATE TABLE` commit is not the likeliest way in. A full disk during the first publication is — `initialize` raises, the file remains, and the store stays broken after the disk is freed.

The fix is the existing notion finished rather than a new one. This store already treats a missing file as absence and creates nothing on a miss; a file with no table is the same fact — nothing has been published — so `_connect_for_read` reports it the same way and the next publication initializes as it would on a fresh project.

One thing your finding did not name, which the fix had to preserve: **corruption must stay an error.** A file that is not a database at all still raises `ArtifactStoreError`, and since the probe that tells the two apart can itself fail, opening moved inside the guard that already wrapped the read. Pinned by the existing corruption test, which caught my first attempt at this when the probe leaked a bare `sqlite3.DatabaseError`.

Pinned: `test_a_database_whose_table_was_never_committed_reads_as_absence` covers all three read paths and then publishes successfully through the same store.

**`test_recent_findings.py:345` — the malformed-record test exercised nothing.** Correct, and it was ours. It wrote `contracts/broken/events.json`, a layout that stopped being read when the store stopped being a directory, so it passed without reaching the code it names. It inserts a malformed row now.

## Not fixed, and why

**`agent_process.py:687` — legacy cancel misses a rotated level.** The control flow is exactly as you describe: on a `cancel_key` mismatch the scan returns instead of continuing to the next level, and `continue` is the whole fix. I am leaving it because of what has to be true to reach it: a machine predating the cutover, two process ids that alias under filename sanitization, cancellations durably written for **both**, and rotation putting the one you are looking for at level 1. That is four conditions stacked on a latent aliasing bug that this PR closed going forward. Say the word and it is a one-word change, but I would rather not add a code path for it on the strength of its cheapness alone.

**`recent_findings.py:201` — malformed `created_at` raises `ValueError`.** Reachable only by writing an unparseable timestamp into the database directly; the store's only writer stamps `datetime.now(UTC).isoformat()`. That puts it on the same footing as the project-symlink finding from round 1, which was withdrawn on the same ground: the workspace is inside the trust boundary (RFC #2153), and anything able to rewrite rows in `.ouroboros/artifacts/artifacts.db` can more cheaply edit the source a finding would cite.

I want to be straight that this is not the same as round 2's malformed-body blocker, which I did fix. That one was reachable from a body a child legitimately published. This one requires tampering, and the fail-open boundary already covers everything downstream of the store's own query.

**`docs/rfc/disposable-memory.md:1` — historical `artifact_ref` text under the superseded banner.** Not in this PR, by maintainer decision. `docs/rfc/` is untouched here; the same stale text is in `mesh.md` and `contract-ledger.md` too, so it is a docs-scope question and the maintainer's decision is to freeze that directory as a pre-issue archive in a separate change, with RFC issues as the single source going forward.

## Verification

- `tests/unit/persistence tests/unit/orchestrator tests/unit/events` — 5803 passed, 1 skipped
- `tests/unit/mcp tests/unit/cli` — 4121 passed, 8 skipped
- `tests/integration tests/e2e` — 338 passed, 4 skipped, 1 xfailed
- `ruff check` / `ruff format --check` (1349 files) / `mypy src/ouroboros` (586 files) / `check-module-size.py` — all pass
- Behavioural spot-check, four cases: a zero-byte database now lets a publication through and creates the table; a corrupt database still raises `ArtifactStoreError` with a `sqlite3.DatabaseError` cause; a normal store reads back unchanged; `published_contracts()` on the empty file returns `[]` rather than raising.

cohemm:
Round 6 triage — one suggestion fixed in `bae81f09`, one follow-up declined, and the CI failure is unrelated.

**Suggestion 1 (`artifact_errors.py:18`) — fixed, and one of the two was more than stale prose.**

`ArtifactManifestError` did not just describe removed machinery, it *was* removed machinery: no raiser, no importer, nothing in `src`, `tests` or `docs`. It survived this cutover only because it lived in a different file from the manifests it belonged to. Deleted.

`ArtifactIntegrityError` stays, but its description said "stored bytes do not match their content address" — the check content addressing performed and that nothing can perform now. It is raised when a stored body will not parse as the JSON it has to be to be returned, and now says so.

**Follow-up (`fanout_handler.py:265`) — declined. The `.strip()` stays.**

The mechanism is exactly as you describe: publish under `" spaced "`, fetch with the same string, and the handler looks up `"spaced"`. What I do not think follows is that the identity should be preserved byte-for-byte here.

Nothing in this tree produces a contract id with surrounding whitespace. The two producers are `fanout:<sha256>` and `new_call_id()`. So the case is reachable only by a caller deliberately constructing one, and against that theoretical caller sits a real one: `.strip()` is at an MCP tool boundary where a human pastes an id, and it is the same normalisation `fanout_id` gets on line 148. It is also what makes `"   "` answer "contract_id is required" instead of "not found". Removing it would trade a working affordance for an unreachable symmetry.

The RFC widening contract ids to "length is the whole rule" is a statement about what the *store* accepts, deliberately: a contract id is opaque to it and is whatever the caller passed. That a storage layer stopped caring about a value's shape is not an instruction to every boundary above it to stop normalising its own input.

Revisit condition, stated plainly: if any caller starts producing contract ids whose identity depends on surrounding whitespace, the `.strip()` has to go — and at that point it should go from both ids, not just this one.

**Scorecard row "Preserve caller-visible behavior" — the same reasoning applies.** No caller's behaviour changes here, because no caller holds such an id.

**CI: `Test Python 3.12` failed — my read of it below was wrong, and the correction is in the comment for `026d9f5b`.**

I called it an unrelated timing flake. It was this PR: a descriptor leak in the probe added by `c086566d`. Leaving the claim here uncorrected would misinform anyone reading the thread in order.

Verification for `bae81f09`: `ruff check`, `ruff format --check` (1349 files), `mypy src/ouroboros` (586 files) all pass; `tests/unit/persistence` + disposable-memory + recent-findings — 476 passed.

cohemm:
`026d9f5b` — pushed after your approval. Recording why, since the approval was given against `bae81f09` and this changes what you approved.

**It fixes a defect I introduced in `c086566d`, in this PR, after the approving round.** Not new scope.

The probe I added to `_connect_for_read` — the one that makes a database with no table read as absence — opens the connection *before* asking whether the table exists, and left it open when that question raised. A corrupt database answers by raising, so every read of one leaked a file descriptor: measured at 200 reads, 200 descriptors. `_connect_for_write` two functions above already closed on failure; this one now does the same. Four lines, matching the sibling idiom.

**How it surfaced, and what I got wrong.** CI's `Test Python 3.12` went red. I called it an unrelated timing flake in `test_start_auto.py` and said a re-run would be green. That was wrong, and the way it was wrong is worth recording: I judged it from local subset runs and never checked the branch's own CI history, which answers it in one query.

```
70178c97  success
c086566d  failure    <- the commit that added the probe
bae81f09  failure    <- inherits it
```

Under `pytest tests/ -n 4` the leak exhausts the process descriptor limit and pytest dies partway with no summary — which is exactly what the log showed, stopping at 26% with no FAILURES section. The named test was collateral, not the cause, which is why re-running it in isolation kept passing and kept misleading me.

**The verification gap this exposes matters more than the bug.** Every round in this PR I verified with subset runs split across parallel agents. This class of failure only appears in one whole-suite parallel run, so no amount of that verification could have caught it. Running the suite the way CI runs it is now part of what I check, not a thing CI does afterwards.

Evidence the leak was the cause: the same command CI uses (`pytest tests/ -n 4 --dist worksteal`, dependencies matched to `--dev --group mcp-test --group litellm-test`) now completes instead of dying — `9 failed, 20800 passed, 16 skipped, 1 xfailed`. None of the nine are in this PR's areas, and six reproduce identically on `origin/main` in the same environment; the remaining three are load- and terminal-rendering sensitive locally, which the Linux CI run settles.

Leak measured at 0 across all three read paths — normal, corrupt, and no-table.

CI on `026d9f5b` is running; I will report the result rather than predict it this time.

PR #2176, opened 2026-08-18
fix(cli): ask the question the ambiguity score implies

### LINKED ISSUE #2175: CLI interview loop never measures ambiguity, so it never offers Seed generation
> **A branch already implements this** — `fix/cli-interview-seed-gate` is pushed and a PR follows immediately. No bot PR is needed for this issue.

## Summary

The `ouroboros init start` interview loop never measures ambiguity while it runs, so it asks a prompt that carries no information about the 0.2 threshold and it silently disables the seed-closer path in the question prompt. A user who is told nothing but "Continue with more questions?" reasonably reads it as "the score is still above 0.2" and keeps going, while the interview may already qualify for Seed generation.

## Reproduction

1. `ouroboros init start "<any idea>"` with any LLM backend.
2. Answer 6-7 rounds until the requirements are clearly settled.
3. Every round ends with `Continue with more questions? [y/n] (y)` — no score, no offer to move to Seed.

Observed on a `kiro` backend session where round 7 produced an implementation transcript instead of a question; the missing seed offer is backend independent and is the part this issue covers.

## Root cause

**1. The loop never scores.** `_run_interview_loop` (`src/ouroboros/cli/commands/init.py`) contains no `AmbiguityScorer` call. The first score is computed after the loop exits, in `_generate_seed_from_interview`:

```python
# _generate_seed_from_interview — outside the loop
with console.status("[cyan]Calculating ambiguity score...[/]"):
    score_result = await scorer.score(state)
```

So the per-round confirmation prompt is score-blind by construction.

**2. Closure mode can therefore never activate.** `InterviewEngine._select_perspectives` (`src/ouroboros/bigbang/interview.py`) attaches the seed-closer perspective only when a score is stored:

```python
if (state.ambiguity_score is not None
        and state.ambiguity_score <= SEED_CLOSER_ACTIVATION_THRESHOLD):  # 0.25
    perspectives.append(InterviewPerspective.SEED_CLOSER)
```

In this loop `state.ambiguity_score` is always `None`, so the perspective never attaches and the panel rule that ships instead is:

> `- Only ask a closure question when closure mode is active; otherwise keep drilling into the weakest area.`

The model is actively instructed not to offer Seed generation. This is worse than a plain omission.

**3. The MCP path does not have this gap.** `mcp/tools/authoring_handlers.py` rescores every round, evaluates `qualifies_for_seed_completion`, advances `completion_candidate_streak`, and auto-completes at `AUTO_COMPLETE_STREAK_REQUIRED`. Two entry points over the same `InterviewEngine`, and only the CLI one lacks completion routing.

## Expected

Once the score is at or below `AMBIGUITY_THRESHOLD` (0.2) and every component floor is met, the loop should say so and offer Seed generation instead of asking an uninformative "more questions?".

## Proposed fix (minimal)

- Score once per confirmation round (from `MIN_ROUNDS_BEFORE_EARLY_EXIT`), persist via `state.store_ambiguity()`, and clear the snapshot on scoring failure so a stale score cannot drive closure — the contract the MCP live-scoring path already follows.
- Branch the prompt on `qualifies_for_seed_completion(score, is_brownfield=...)`: offer Seed when it qualifies, keep the existing prompt otherwise.
- Print the score so the user can see what the prompt is based on.

Cost: one scoring call per round from round 3 on.

## Out of scope

Two adjacent defects seen in the same session are deliberately not covered here and need their own issues:

- The interview adapter is built with `allowed_tools=None, max_turns=5` (`_get_adapter`), which for CLI-agent backends such as `kiro` means no tool envelope and an explicit multi-turn license during question generation.
- `InterviewEngine.ask_next_question` applies no shape validation to the returned text, so a 22k-character agent transcript is accepted and printed as the next question.

cohemm:
Closes #2175.

## Summary

The `init start` interview loop asked "Continue with more questions?" every round without ever measuring ambiguity — the first score was computed after the loop exited, inside `_generate_seed_from_interview`. The prompt therefore carried no information about the 0.2 threshold, and because no score was ever stored, `_select_perspectives` could never attach the seed-closer perspective (it requires `ambiguity_score is not None and <= 0.25`). The panel rule that shipped instead was "only ask a closure question when closure mode is active; otherwise keep drilling", so the model was actively instructed not to offer Seed generation.

This scores once per confirmation round and asks the question the score implies.

- `_refresh_ambiguity()` scores the in-progress interview, persists the snapshot with `state.store_ambiguity()`, and clears it on failure so a stale score cannot drive closure — the contract the MCP live-scoring path follows. Retries are bounded at 3 because the scorer's own default is unlimited and it runs between two interactive prompts.
- The confirmation branches on `qualifies_for_seed_completion()` (overall 0.2 plus every component floor, brownfield-aware): "Ambiguity is low enough — proceed to Seed generation?" when it qualifies, the existing prompt otherwise. The score is printed either way.
- `scorer` is an optional keyword argument, so callers that pass none keep the previous behavior.

Storing the score is what re-enables the closure path: the seed-closer perspective now attaches at <= 0.25, and the panel sits ahead of the base agent prompt in the trimming order, so the closure instruction survives the system-prompt budget.

The MCP interview path already did all of this (`authoring_handlers.py` rescores each round and routes on `qualifies_for_seed_completion`). This closes the gap between the two entry points over the same `InterviewEngine`.

## Second commit: the snapshot was pushing the interviewer role out of the prompt

Verified after the first commit by dumping the built system prompt: storing a snapshot grows the dynamic header and attaches the seed-closer perspective, and both are preserved ahead of the base agent prompt. At a stored score of 0.18 the prompt silently lost `NEVER say "I will implement X"`, `NEVER promise to build demos`, and `You MUST always end with a question` — in exactly the rounds where the model is most tempted to answer its own question. The first commit alone would have made that failure mode more likely, not less.

Closure rounds now run a narrower panel (seed-closer first, then breadth-keeper and architect), since the panel is trimmed from the tail and appending the closer dropped it in the rounds it exists for.

A reserved slice of the base agent prompt was tried here and **removed in a later commit**: raising it far enough to also cover `You MUST always end with a question` starved the panel and, at saturated caps, the user's own retained initial context. The final shape instead puts the two rules that make a turn a question — emit exactly one question and end with it; never announce implementation work — in the dynamic header, which every build preserves first. That is one mechanism rather than two, and it holds at every budget (verified at the default cap, 3500, 1200, 700, and with a 4,000-character initial context).

Measured on a 6-round interview with a 0.18 snapshot: `### seed-closer` present with its full closure question patterns, panel tail (simplifier) dropped instead. Pinned by `test_response_contract_survives_every_prompt_budget`, `test_response_contract_survives_a_saturating_initial_context`, and `test_closure_round_keeps_role_boundaries_alongside_seed_closer`.

## Third commit: one seed-ready predicate at both gates

The loop asked its confirmation with `qualifies_for_seed_completion()` (overall threshold plus component floors, brownfield aware) while `_generate_seed_from_interview()` still checked only `is_ready_for_seed`. A score of overall `0.15` with every component clarity at `0.5` therefore got "Continue with more questions?" from the loop and unforced generation from the outer path. Both gates now use the same predicate, `force` follows the same verdict, and the warning names the failing dimensions instead of an overall score that is under the threshold.

`SeedGenerator.generate()` keeps its own overall-only gate; that is the engine-level floor shared with MCP, and component floors are the caller's completion policy.

## Test plan

- [x] `uv run pytest tests/unit/bigbang tests/unit/cli -q` — 3103 passed, 7 skipped
- [x] `uv run pytest tests/unit/mcp tests/unit/auto -q` — 3964 passed, 1 skipped
- [x] `uv run pytest tests/integration tests/conformance -q` — 304 passed, 4 skipped, 1 xfailed
- [x] New `tests/unit/cli/test_init_ambiguity_gate.py` — 5 passed, covering: seed-ready score offers Seed and completes the interview; high score keeps the existing prompt and stores the snapshot; low overall score with a failing component floor does **not** offer Seed; scoring failure clears a stale snapshot and falls back; no scorer keeps the legacy prompt
- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check` clean
- [x] `uv run mypy src/ouroboros/cli/commands/init.py` clean

## Cost

One scoring call per round from `MIN_ROUNDS_BEFORE_EARLY_EXIT` (round 3) on. `_generate_seed_from_interview` still scores once more before generation; that duplicate is left in place to keep this change minimal and to preserve the authoritative gate immediately before generation.

## Not in this PR

Two adjacent defects observed in the same session, noted in #2175 and needing their own issues:

- `_get_adapter` builds the interview adapter with `allowed_tools=None, max_turns=5`, which for CLI-agent backends such as `kiro` means no tool envelope and an explicit multi-turn license during question generation.
- `InterviewEngine.ask_next_question` applies no shape validation to the returned text, so an agent transcript is accepted and printed as the next question.

### DIFF
--- a/src/ouroboros/bigbang/interview.py
+++ b/src/ouroboros/bigbang/interview.py
@@ -730,6 +730,10 @@ class InterviewEngine:
     _MIN_SYSTEM_PROMPT_CHARS = 1200
     _MAX_INITIAL_CONTEXT_SYSTEM_CHARS = 1800
     _MAX_INITIAL_CONTEXT_TOTAL_CHARS = MAX_PROMPT_SAFE_INITIAL_CONTEXT_CHARS
+    # Opening span of the base agent prompt that survives trimming ahead of the
+    # perspective panel: the role boundaries, context boundaries, and response
+    # format that keep the generator asking questions instead of answering them.
+    _BASE_PROMPT_ROLE_FLOOR_CHARS = 500
     _INITIAL_CONTEXT_SUMMARY_QUESTION = INITIAL_CONTEXT_SUMMARY_QUESTION
     suppress_tool_use_prompt_cues: bool = False
     # When True, ``ask_next_question`` generates three persona candidates
@@ -1310,15 +1314,27 @@ class InterviewEngine:
         # Preserve the dynamic header first; it contains the capped initial
         # context and first-turn instructions. Trim the optional panel/base
         # prompt before falling back to hard-truncating the header.
+        #
+        # The base prompt opens with the role boundaries that keep the model an
+        # interviewer ("NEVER say I will implement", "always end with a
+        # question"). Those must outrank the panel: a growing header or panel
+        # otherwise pushes them out entirely, which is exactly when they matter
+        # most — a late round with a stored ambiguity snapshot and the
+        # seed-closer perspective attached.
         available_after_header = max_prompt_chars - len(dynamic_header) - _OVERHEAD
         if available_after_header <= 0:
             dynamic_header = dynamic_header[: max_prompt_chars - _OVERHEAD]
             perspective_panel = ""
             base_budget = 0
-        elif len(perspective_panel) > available_after_header:
-            perspective_panel = perspective_panel[:available_after_header]
-            base_budget = 0
         else:
+            reserved_base = min(
+                len(base_prompt),
+                self._BASE_PROMPT_ROLE_FLOOR_CHARS,
+                available_after_header,
+            )
+            panel_budget = available_after_header - reserved_base
+            if len(perspective_panel) > panel_budget:
+                perspective_panel = perspective_panel[:panel_budget]
             base_budget = available_after_header - len(perspective_panel)
 
         trimmed_base = base_prompt[:base_budget] if base_budget < len(base_prompt) else base_prompt
@@ -1458,7 +1474,17 @@ class InterviewEngine:
             state.ambiguity_score is not None
             and state.ambiguity_score <= SEED_CLOSER_ACTIVATION_THRESHOLD
         ):
-            perspectives.append(InterviewPerspective.SEED_CLOSER)
+            # Closure rounds run a narrower panel, led by the closer. Two
+            # reasons, one budget and one substantive: the panel is trimmed
+            # from the tail, so appending the closer would drop it in exactly
+            # the rounds it exists for; and researcher/simplifier drilling is
+            # the behavior closure is meant to stop. The freed characters go
+            # back to the base prompt's role boundaries.
+            perspectives = [
+                InterviewPerspective.SEED_CLOSER,
+                InterviewPerspective.BREADTH_KEEPER,
+                InterviewPerspective.ARCHITECT,
+            ]
 
         if state.is_brownfield and InterviewPerspective.ARCHITECT not in perspectives:
             perspectives.append(InterviewPerspective.ARCHITECT)
--- a/src/ouroboros/bigbang/interview.py
+++ b/src/ouroboros/bigbang/interview.py
@@ -84,6 +84,18 @@ AGENT_SDK_CLI_PER_MESSAGE_FRAMING_CHARS = 128
 # model locally while still tripling the original 4.8k interview budget.
 AGENT_SDK_CLI_SAFE_PROMPT_CHARS = 14_000
 
+# The two rules that make a generated turn a question at all. They live in the
+# dynamic header — which every build preserves first — instead of only inside
+# the loaded agent prompt, whose tail is the first thing the character budget
+# trims. Leaving them there made their survival depend on where they happen to
+# sit in a markdown file and on how large the header and perspective panel grew
+# that round: in a late round with a stored ambiguity snapshot, the response
+# format was silently dropped from the prompt that needed it most.
+RESPONSE_CONTRACT = (
+    "Response contract: emit exactly ONE question and end with it. "
+    "Never announce implementation work — another agent does that later."
+)
+
 _TOOLLESS_INTERVIEW_BASE_PROMPT = """## Role Boundaries
 - You are only an interviewer.
 - Generate exactly one Socratic question that reduces requirements ambiguity.
@@ -737,12 +749,6 @@ class InterviewEngine:
     _MIN_SYSTEM_PROMPT_CHARS = 1200
     _MAX_INITIAL_CONTEXT_SYSTEM_CHARS = 1800
     _MAX_INITIAL_CONTEXT_TOTAL_CHARS = MAX_PROMPT_SAFE_INITIAL_CONTEXT_CHARS
-    # Opening span of the base agent prompt that survives trimming ahead of the
-    # perspective panel: the CRITICAL ROLE BOUNDARIES section, which forbids the
-    # generator from answering its own question. Deliberately small — the
-    # dynamic header already carries "Your ONLY job is to ask questions", and a
-    # larger floor starves panels that callers reserve budget for.
-    _BASE_PROMPT_ROLE_FLOOR_CHARS = 500
     _INITIAL_CONTEXT_SUMMARY_QUESTION = INITIAL_CONTEXT_SUMMARY_QUESTION
     suppress_tool_use_prompt_cues: bool = False
     # When True, ``ask_next_question`` generates three persona candidates
@@ -1272,12 +1278,14 @@ class InterviewEngine:
                 f'Do NOT introduce yourself. Do NOT say "I\'ll conduct" or "Let me ask". '
                 f"Just ask a specific, clarifying question immediately.\n\n"
                 f"This is {round_info}. Your ONLY job is to ask questions that reduce ambiguity.\n\n"
+                f"{RESPONSE_CONTRACT}\n\n"
                 f"Initial context: {prompt_initial_context}\n"
             )
         else:
             dynamic_header = (
                 f"You are an expert requirements engineer conducting a Socratic interview.\n\n"
                 f"This is {round_info}. Your ONLY job is to ask questions that reduce ambiguity.\n\n"
+                f"{RESPONSE_CONTRACT}\n\n"
                 f"Initial context: {prompt_initial_context}\n"
             )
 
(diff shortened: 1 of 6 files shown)


cohemm:
Addressed review round 2 in b5a2b75a — 2 fixed, 0 defended. Both findings were reproduced against `9d5d43ec` before any code was written.

**Blocker 1, `init.py:385` — stale snapshot across a durable save — fixed.** Reproduced: `record_answer()` appends an answer and leaves `ambiguity_score` attached, so the save at line 393 writes rounds from revision N+1 next to the score from revision N, and a resume in that window hands `_select_perspectives()` a score it reads as current.

Fixed at the mutation point rather than at the call site: `InterviewState.record_answer()` now clears the snapshot, next to the `invalidate_requirement_distillation()` call already there. Both are derived from the answered rounds, and one of them was already dying where its inputs change while the other was not. This also removes the possibility from every other transport — MCP's explicit-done path was doing exactly this by hand one line below its own `record_answer()` (`authoring_handlers.py:2830`), which is the same rule stated at one call site instead of at the funnel every transport passes through.

Pinned: `test_resume_after_recording_an_answer_loads_no_stale_score` (save the accepted answer, never rescore, reload from disk — score and breakdown are None, both rounds present) and `test_recording_an_answer_invalidates_the_previous_snapshot`.

**Follow-up, `init.py:424` — approval did not cross the loop boundary — fixed.** Confirmed: the loop returned only the state, so `_run_interview()` asked "Proceed to generate Seed specification?" a second time and a decline there contradicted the action the first prompt named.

`_run_interview_loop` now returns an `InterviewLoopOutcome` carrying `seed_generation_approved` **and** `approved_score`. The outer flow consumes the approval instead of re-asking, and passes the approved score into `_generate_seed_from_interview`, which skips its own scoring call. The second half matters as much as the first: without it, generation samples an independent second score, and a disagreeing sample drops the user into the "ambiguity too high" 1/2/3 menu immediately after they approved generation — the approval would be undone by a coin flip rather than by a prompt.

Pinned: `test_seed_approval_crosses_the_loop_boundary`, `test_approved_score_starts_generation_without_a_second_score` (the scorer factory raises if constructed), and `test_stopping_questions_without_a_qualifying_score_is_not_seed_approval` so declining more questions is not silently read as approval.

**Adjacent address, deliberately not changed.** The other mutation of the snapshot's inputs is `state.rounds.pop()` in the MCP done path. That combination cannot reach a stale read: the interview is COMPLETED at that point, `_select_perspectives()` is only reached through `ask_next_question()` on an in-progress interview, and reopening goes through `record_response()`, which clears. Stating it here so the next probe of this class has an answer instead of a new round.

Verification (parallel subagents): `tests/unit/bigbang tests/unit/cli` 3091 passed / 7 skipped; `tests/unit/mcp tests/unit/auto` 3964 passed / 1 skipped — the highest-risk area for a global invalidation, and it is green without changes; `tests/integration tests/conformance` 304 passed / 4 skipped / 1 xfailed; combined confirmation run 3395 passed / 11 skipped / 1 xfailed. `ruff check`, `ruff format --check`, and `mypy` on both changed modules: clean.

cohemm:
Addressed review round 3 in c462b1b0 — 2 fixed, 0 defended, plus both P2 test suggestions and the docstring note. Both blockers are consequences of last round's own changes, so both were closed by removing a divergence rather than tuning one.

**Blocker 1, `init.py:713` — component floors stopped at the loop — fixed.** Reproduced exactly as described: overall `0.15` with every component clarity `0.5` gives `is_ready_for_seed=True` and `qualifies_for_seed_completion=False`, so the loop asks "Continue with more questions?" while generation would proceed unforced on the overall number alone. One decision was being answered by two predicates.

`_generate_seed_from_interview()` now uses the loop's predicate, and `force` follows the same verdict — otherwise choosing "Generate Seed anyway" would have passed `force=False` for a `0.15` score and been silently re-gated by the engine. The warning text also stops naming the overall score when that score is under the threshold; "Ambiguity score (0.15) is too high" reads as a contradiction, so the failed dimensions are named instead.

Pinned: `test_failed_component_floor_blocks_generation_despite_low_overall` (parametrized over the "continue" and "cancel" choices; the generator raises if awaited) and `test_failed_component_floor_forces_generation_when_user_overrides` (asserts `force=True` reaches the generator).

**Blocker 2, `interview.py:745` — the question-ending instruction could still be trimmed — fixed, by deleting the reserve rather than raising it.** Confirmed: that instruction begins around character 790, past the 500-character floor. Raising the floor was the obvious move and I tried it first — at 1050 the PM budget tests immediately failed, because the characters come from the perspective panel and, at saturated caps, from the user's own retained initial context. A floor large enough to cover the response format is a floor that starves something else; the next round would have been about whatever it starved.

So the two rules that make a turn a question at all now live in the dynamic header, which every build preserves first, and `_BASE_PROMPT_ROLE_FLOOR_CHARS` is deleted — the trimming order is back to what it was before this PR. One mechanism instead of two, and a stronger guarantee, since it no longer depends on where those rules happen to sit inside a markdown file:

```
Response contract: emit exactly ONE question and end with it.
Never announce implementation work — another agent does that later.
```

129 characters. A 230-character first draft was rejected by PM's own budget tests for evicting the panel and the retained context tail, which is the same failure mode as the floor and is why the shipped version is this short.

Pinned as an invariant rather than a scenario: `test_response_contract_survives_every_prompt_budget` (parametrized over caps `default / 3500 / 1200 / 700` × stored score `none / 0.18`) and `test_response_contract_survives_a_saturating_initial_context` (4,000-character context at the minimum cap, where the header itself is hard-truncated). The existing closure-round test keeps pinning the fuller agent-prompt wording at the default cap.

**Third judge of seed-readiness, deliberately unchanged.** `SeedGenerator.generate()` gates on overall ambiguity unless forced. That is the engine-level floor shared with MCP; component floors are the caller's completion policy, which is why the CLI now passes it a `force` flag consistent with its own verdict instead of a second opinion. Naming it here so it is not read as a remaining inconsistency.

Verification (parallel subagents): `tests/unit/bigbang tests/unit/cli` 3103 passed / 7 skipped; `tests/unit/mcp tests/unit/auto` 3964 passed / 1 skipped, including the prompt-budget assertions on the shared `_build_system_prompt`; `tests/integration tests/conformance` 304 passed / 4 skipped / 1 xfailed. `ruff check`, `ruff format --check`, `mypy` on both changed modules: clean. No references to the deleted constant remain.

cohemm:
Thanks for the approval. Both parts of the P2 follow-up are handled in d1e33970.

**Comment corrected.** You are right: `AmbiguityScorer.max_retries` defaults to `10`, not unlimited, so the comment justified a real budget with a false fact. The budget stands on its own — the scorer runs between two interactive prompts, and 10 attempts with doubling token budgets is a long wait for someone sitting at one — so only the rationale changed.

Worth noting where the wrong claim came from: `AmbiguityScorer`'s own docstring still says "Retries until success by default (unlimited)" at `ambiguity.py:350` and "or None for unlimited (default)" at `:357`, which contradicts `max_retries: int | None = 10` on the next line. I left it alone as pre-existing and outside this PR's diff rather than widening the change under an approval — happy to fix it here if you would rather it not stay wrong, otherwise it should get its own issue.

**PR description corrected.** It still described the reserved base-prompt floor as the shipped design. That reserve was removed in c462b1b0 precisely because raising it far enough to cover the response format starved the panel and the retained initial context; the description now says the contract lives in the dynamic header and lists the budgets it was verified at.

Verification for the doc-only change: `tests/unit/cli` 2044 passed / 7 skipped; `ruff check` and `ruff format --check` clean.

Issue #2179, opened 2026-08-18
RFC: evidential force for self-authored acceptance contracts — #1916 closed asymmetry, not force


Q00:
## Why this exists

`#1916` (hidden-checklist convergence, merged `9210219`) was believed to have
closed this. Reading its diff, it established information asymmetry (the worker
never sees the grader's answer key), judgment ownership (the orchestrator runs
the contract), and the durable run→eval→evolve chain. **Nothing in it asks
whether the hidden command can fail.**

Measured on `main` while investigating a separate Windows-portability RFC:

```
verify_command='exit 0'        -> gate passed = True
verify_command='true'          -> gate passed = True
verify_command='printf READY'  -> gate passed = True

required evidence without verify_command: files_touched, commands_run, tests_passed
required evidence with    verify_command: files_touched
```

`exit 0` was the single most common `verify_command` in the repository (13
occurrences). The two effects compound: such a contract passes unconditionally
**and** its mere declaration drops `commands_run`/`tests_passed`, leaving the AC
checked *less* than one that declared no contract at all. `#1916` was necessary
— without hiding, strengthening contracts is pointless, since the worker reads
the stronger key and optimizes to its letter — but it was not sufficient. A
hidden check that always passes is still hidden, and the convergence loop then
converges on that pass.

Part of this has since shipped (static vacuity proof, seed-authoring gate,
verify-environment sanitization, shell resolution + a retryable unverifiable
outcome); the RFC records that as the current baseline rather than proposing it.
What remains is the part static analysis cannot reach: a command can be
non-vacuous and still pass without the work — most sharply when the worker edits
the very files the command runs.

## What is being asked

Review of the staged plan below (5 PRs), and specifically of three decisions:

1. The **ceiling** claim — that exit-code mechanisms top out at a *discrimination
   certificate* and can never certify semantic adequacy — and the two-tier
   verdict vocabulary (`discriminating_pass` vs `pass`) that follows from it.
2. The **withhold-only** authority rule: every probe may refuse a stamp, none may
   manufacture a pass or a fail, and none may stop a run — modeled on the
   existing `VERIFY_ENVIRONMENT_UNAVAILABLE` shape.
3. Whether **Stage 2** (two-tree differential) earns its cost, given that
   Stage 1 is far cheaper and already closes the pre-planted attacks.

Every stage carries a falsification experiment that must run *before* its
enforce flip, with the measured noise rate recorded in the PR body.

---
[...]

PR #2180, opened 2026-08-18
feat(verify): resolve a real POSIX shell for verify_command, with shell-free fallback and retryable quarantine

Q00:
## Why

The verify gate inherited the platform default shell (`create_subprocess_shell` → `/bin/sh` = dash on Debian, `cmd.exe` on Windows), so gate outcomes depended on the host and Windows could not run contracts at all. Implements `docs/rfc/verify-gate-windows-portability.md` (included in this PR).

## What

- **Shell resolution** (`orchestrator/verify_shell.py`): env → config (`orchestrator.verify_bash_path` / `OUROBOROS_VERIFY_BASH`) → `/bin/bash` → PATH. Windows: Git Bash under `%ProgramFiles%` trees → PATH `bash.exe`. Never `cmd`, never the System32 WSL launcher (different filesystem — the gate's `cwd` is meaningless in it). Only successes are cached, so a mid-run Git Bash install is picked up on the next retry.
- **Shell-free execution** (`core/verify_command_plan.py` + `orchestrator/verify_command_runner.py`): when no shell resolves, the command is parsed and run POSIX-exact — `&&`/`||`/`;` short-circuit, per-command `NAME=value` prefixes, 127 for a missing command, one timeout budget for the chain, builtins evaluated exactly. Anything unreproducible (pipe, redirect, expansion, glob) is **refused**, never approximated.
- **Env sanitization** (`sanitized_verify_environment`): strips `PYTHONPATH`/`PYTHONSTARTUP`/`PYTHONHOME`/`PYTEST_ADDOPTS`/`PYTEST_PLUGINS`/`NODE_OPTIONS` + `LD_PRELOAD*`/`DYLD_INSERT_LIBRARIES*`. Measured: `PYTEST_ADDOPTS="--collect-only"` turns any pytest contract into exit 0 with zero tests run. `PATH` deliberately kept — a workspace `.venv/bin` is legitimate.
- **Retryable quarantine** (`orchestrator/verify_quarantine.py`, `FailureClass.VERIFY_ENVIRONMENT_UNAVAILABLE`): no judgment environment → retryable FAILED, **not** `BLOCKED` (which forecloses route escalation and demands a human handoff). The AC is never counted verified while unverifiable; the run continues.
- `OUROBOROS_VERIFY_BASH` joins the untrusted-`.env` denylist (RCE guard — this is the one gate that must stay untamperable).
- `_VerifyGateOutcome` contract extracted to `orchestrator/verify_gate_outcome.py` unchanged (module-size ratchet; `parallel_executor.py` shrinks).

## Behavior change, stated honestly

POSIX gates previously ran under `/bin/sh -c` (= dash on Debian/Ubuntu); they now run `bash -c`. Strictly more permissive, but a real change in gate outcomes — verified by printing `$0` both ways.

## Verification

- Full `tests/unit` (minus mcp): 18,326 passed, 29 failed = exactly the pre-existing baseline, 0 new.
- This commit validated in an isolated checkout: imports + `test_verify_shell.py` / `test_verify_command_runner.py` / `test_parallel_executor_verify_by_default.py` green; ruff, ruff-format, mypy (new modules), module-size ratchet all pass.

Stacked: `feat/verify-vacuity-seed-gate` (vacuity proof + seed gate) follows on top. RFC follow-on: #2179.


Q00:
@ouroboros-agent Both blockers fixed and independently reproduced-then-verified in `004d3ce3a`; the follow-up finding is addressed as a doc correction.

**Blocker 1 (workspace-relative resolution).** `_resolve_step_executable` now resolves any path-bearing command name against the verification `cwd`, exactly as a shell running there would: `./check.sh` executes from the workspace (regression: `test_path_bearing_command_resolves_against_the_workspace`), a missing path reports 127, and an existing-but-non-executable path reports **126** with `Permission denied` — matching bash, not the previous 127 lump. Bare names keep the PATH lookup (now consistently against the step's sanitized `PATH`).

**Blocker 2 (empty-segment repair).** The planner now rejects leading, adjacent, and dangling operators instead of filtering their empty segments: `&& echo PASS`, `|| echo PASS`, `; echo PASS`, `echo PASS &&`, `echo PASS ||`, `echo PASS ; ; echo TWO`, `echo PASS && && echo TWO` all return `None` (→ unverifiable, never a verdict). A trailing `;` remains valid, as in POSIX — verified against bash (`bash -c '&& echo PASS'` → status 2; `bash -c 'echo DONE ;'` → 0). Shell-parity tests cover both directions.

**Follow-up (RFC overstating B1/B3).** Correct catch of a stack artifact: B1 (`seed_verify_gate`, `verify_exemption_reason`, `seed.verify_command_gate`) and B3's producer ship in the stacked #2181, not here — only the failure classes land in this PR. The RFC's implementation-status section now says exactly that instead of claiming them shipped.

Not-met rows for shell-free parity should now be satisfiable; the B1/B3 rows are #2181's to satisfy. Full verify suites (114 tests) green on the updated head.

Q00:
Round 2 found three ways the shell-free route could judge a different command than bash, plus two ways shell resolution ignored its own inputs. All five reproduced; all five fixed in `b14cdb4ff`.

**Blockers**

1. **Relative `PATH` resolved against the wrong directory** (`verify_command_runner.py`). `shutil.which(name, path=path)` resolves relative — and empty — entries against the orchestrator's process directory. Every entry is now re-anchored to the verification `cwd` before lookup, which is how a shell launched there reads them. Confirmed both halves of the hazard: a workspace tool was missed, and a same-named binary beside the orchestrator was reachable.
2. **Un-emulated builtins admitted as external executables** (`verify_command_plan.py`). `_SHELL_INTERPRETED_WORDS` now refuses every POSIX/bash builtin and reserved word the planner does not reproduce (`cd`, `type`, `export`, `eval`, `read`, `if`, `for`, …). `type missing` fails in bash while a PATH `type` binary can exit 0 — a false pass invented by the fallback route. Refusing reports the AC unverifiable, which is the honest answer.
3. **`echo -n` parity.** `run_builtin` now consumes *every* leading `-n`, as bash does. `echo -n -n x` printed `-n x` here and `x` under bash, so a contract asserting on `-n` passed on one route and failed on the other.

**Follow-ups (both treated as required, per the merge recommendation)**

- **Stale env masked valid config.** `get_verify_bash_path()` read the environment first, so the fallthrough handed back the same stale `OUROBOROS_VERIFY_BASH`. It is replaced by `get_configured_verify_bash_path()` (config only); the environment priority stays in the resolver that already reads it. The old accessor had no other caller, so it is gone rather than left as a second way to ask the same question.
- **Incomplete cache fingerprint.** The key now names every input resolution reads: OS, override, configured path, `PATH`, `%SystemRoot%`, and the `%ProgramFiles%` family. Covered by tests that change the configured shell and the Git Bash install root and assert the route follows.

**New coverage.** Beyond a regression per finding, `test_admitted_commands_match_a_real_bash` cross-checks status *and* stdout for the whole admitted surface (builtins, short-circuits, `PATH=./bin tool`, path-bearing commands) against a real bash. Any future divergence has to be refused rather than approximated.

Suites: `test_verify_shell.py`, `test_verify_command_runner.py`, `test_parallel_executor_verify_by_default.py`, `test_loader.py`, `test_failure_taxonomy.py`, `test_verifier.py` — all green, plus full `tests/unit/core` + `tests/unit/orchestrator`.

Q00:
All three reproduced and fixed in `bca0c1a86`. Each one was about the *shell-backed* half of the boundary, which the earlier rounds had left alone.

**1. `BASH_ENV` (and the rest of the startup family).** Reproduced exactly as described: `BASH_ENV=inject.sh bash -c 'exit 23'` → `0`. The repository was able to run its own code inside the gate, which is worse than bending a tool's configuration.

The gate's child environment now drops `BASH_ENV` and `ENV` (the POSIX spelling of the same hook), the option-state carriers `SHELLOPTS`/`BASHOPTS`/`BASH_XTRACEFD`/`PS4` (xtrace writes into the output an assertion is checked against; errexit changes which leg of a chain decides the status), `IFS`/`CDPATH`/`GLOBIGNORE`, and exported functions by prefix — `BASH_FUNC_pytest%%` resolves before the executable of that name, the shellshock-era shape of the same hole.

`BASH_ENV`, `ENV`, `SHELLOPTS`, `BASHOPTS`, `BASH_XTRACEFD` are on the untrusted project-`.env` denylist, and `BASH_FUNC_` on its prefix list, with the regression test you asked for (parametrized over all six) plus the roster test that stops the denylist shrinking. There is also an end-to-end test: with `BASH_ENV` pointing at a file containing `exit 0`, a `verify_command: exit 23` still fails.

**2. Relative resolved paths.** Right, and the `./tools/bash` reproduction is the sharp version — resolution checks one binary and launch runs another, so a workspace can supply the shell that judges its own criteria.

`_executable()` now returns only absolute paths. Not re-anchored: there is no directory it would be honest to guess, so a relative result is refused and resolution falls through to the next candidate. This covers all three ways one can appear — a relative env override, a relative configured path, and a bare name found through a relative `PATH` entry — each with a test.

**3. Degraded `sh`.** Agreed, and this one was a design mistake rather than a slip: a route that reads the command differently cannot be recorded as verification of that command. Your `echo -e` case reproduces (`FAIL` vs `-e FAIL`).

The `sh` fallback is gone, along with the now-meaningless `degraded` flag. A machine without bash routes through the exact shell-free planner, and when that planner refuses the command the AC is reported unverifiable — which is the retryable `VERIFY_ENVIRONMENT_UNAVAILABLE` path this PR already built, so the honest answer was already available; the `sh` route was just being preferred over it.

The RFC's implementation-status section is corrected to match on all three points (no `sh` fallback, absolute-only resolution, startup-control denials), including the direction-of-change note: a machine with no bash now reports unverifiable where it used to hand the command to `sh` and record whatever that said.

Suites: `test_verify_shell.py`, `test_verify_command_runner.py`, `test_parallel_executor_verify_by_default.py` (135 passed), full `tests/unit/config` (586), and the whole `orchestrator`+`core`+`config` set on the stack tip (6762 passed, 1 skipped). ruff, format, mypy, module-size clean.

Q00:
Superseded by #2187. The replacement keeps the trusted real-Bash boundary but removes shell emulation and verifier-infrastructure retries in favor of monotonic verification.

PR #2181, opened 2026-08-18
feat(verify): prove vacuous contracts, revoke their evidence exemption, and gate unverifiable criteria at authoring

Q00:
> **Stacked on #2180** (`feat/verify-shell-portability`) — merge that first, then retarget this to `main` (squash-only repo: use `rebase --onto` per the established stack mechanics).

## Why

Measured on `main`:

```
verify_command='exit 0'        -> gate passed = True
verify_command='true'          -> gate passed = True
verify_command='printf READY'  -> gate passed = True

required evidence without verify_command: files_touched, commands_run, tests_passed
required evidence with    verify_command: files_touched
```

`exit 0` was the **most common** `verify_command` in the repo (13 occurrences). The two effects compound: a vacuous contract passes unconditionally **and** merely declaring it drops `commands_run`/`tests_passed` from required evidence — an AC checked *less* than one with no contract at all. #1916 closed information asymmetry; it never established evidential force. Full analysis: `docs/rfc/verify-gate-evidential-force.md` (included) and #2179.

## What

- **Static vacuity proof**: `constant_verdict()` (`core/verify_command_plan.py`) commits to an answer only when every reachable step is a builtin whose status *and* output are fixed by literal arguments. Constant-True → `has_evidential_success_contract` False: the gate does not stamp, the evidence exemption is revoked. Constant-**False** (`exit 1`) left alone — it cannot fake a pass (narrowing to the *sign* of the constant fixed ~17 tests). Unparseable → `None`, never an accusation.
- **Seed-authoring gate** (`core/seed_verify_gate.py`): reports ACs with no command and with a vacuous one; `seed.verify_command_gate: warn|block` (default `warn`); per-AC `verify_exemption_reason` as the explicit escape hatch. Runs at new-session preparation only — in-flight sessions are never re-judged under a tightened gate.
- **Empty-transcript reclass**: fully empty `support_messages` → retryable `TRANSCRIPT_MISSING_INFRASTRUCTURE`, not a worker rejection — a gaming leaf leaves plausible messages, not none.
- Strict AC JSON boundary extracted to `bigbang/acceptance_criteria_parsing.py` (`seed_generator.py` shrinks per the ratchet).

## Verification

- Isolated checkout of this commit: vacuity/seed-gate/runner-gate/reclass suites green (45 passed); ruff, ruff-format, mypy (new modules), module-size ratchet all pass.
- Full mixed-tree `tests/unit` (minus mcp): 18,326 passed, 29 failed = exactly the pre-existing baseline, 0 new.

Next: PR 1 of the evidential-force RFC (Stage 0 snapshot + Stage 1 baseline negative control + `discriminating_pass` vocabulary) follows on this stack.


Q00:
@ouroboros-agent Both blockers and all follow-ups addressed in `90a094151`.

**Blocker 1 (disguised constants).** The proof now sees through every reproduced form: `exit` operands are normalized mod 256 as POSIX does (`exit 256`/`exit -0`/`exit +0` → constant-True, `exit -1` → constant-False and left alone), exact `command` prefixes are stripped (`command true` → proven; option forms `-v/-V/-p` change behavior and are refused), and `constant_verdict` recurses into a single literal `bash|sh -c '<inner>'` invocation (`bash -c 'exit 0'`, `sh -c "true"`, `bash -lc 'exit 0'` → proven) — only flag letters that cannot affect status or output are admitted (`c`/`l`/`e`); `-x`/`-v` trace into the combined output and stay unproven, as do wrappers around real work (`bash -c 'pytest -q'` → `None`). The boundary deliberately remains "unproven ⇒ evidential" rather than fail-closed: failing closed would strip the exemption from every genuine contract the planner cannot parse, and the residual (a constant the analyzer cannot prove) is exactly what the stacked Stage 1 baseline probe (#2182) measures dynamically. Parametrized tests cover all listed forms.

**Blocker 2 (Rich markup crash).** The warn path renders the warning as a styled `Text` object — seed text is never interpolated into markup. Regression test drives a real `Console` with a description containing `[/yellow]` and `[bold]`; block-path guidance flows through `OrchestratorError.details`, not markup.

**Follow-ups.** Extraction prompt no longer says "exactly four keys" — both prompt sites now state the optional `exempt` key rule consistently. `SeedConfig` is exported from `ouroboros.config` alongside the other sections. Block-mode message distinguishes missing commands from proven-vacuous ones. The evidential-force RFC status header now separates the shipped baseline from the proposed stages.

Q00:
Both blockers reproduced and are fixed in `f4cb7e2b2`.

**1. The proof admitted commands whose real status is not constant-success.**

Two separate holes, and the counterexamples were right:

- `exit 18446744073709551616` was normalized modulo 256 to `0`. Bash exits **255** ("numeric argument required"), so the contract *always fails* — and proving it vacuously true skipped the command and let transcript evidence stand in its place. Operands are now bounded to what `intmax_t` holds (ASCII digits only; `str.isdigit` also accepts `٣` and `²`, and `int()` raises on the latter). The bound is the fix, not a narrower range: `exit 256` and `exit -1` stay proven, because shells do agree on the low 8 bits — it is the operand no shell can parse that is not a constant at all. `test`'s integer comparisons take the same bound, where Python would otherwise compare values the shell refuses.
- The shell-wrapper recursion is now a plain `-c` only. `-e` aborts the chain at the first failure (`bash -ec 'false; true'` exits 1, not 0) and `-l` runs a login profile that can exit nonzero or print into the output an assertion is checked against. `bash -lc 'exit 0'` moved from the proven list to the unproven one.

New: `test_a_proven_constant_pass_really_passes_under_bash` runs every command the proof calls always-passing through a real bash and asserts exit 0. The proof's whole authority is that it matches the shell that would run the command, so that is now a test rather than a claim.

**2. The evidential classification was not in the durable contract.**

Correct, and the consequence you describe is the one that matters: the same session could change, mid-run, both which evidence an AC must carry and whether the authoritative gate runs at all for it.

`vacuous_contract_evidence` is now durable execution semantics (v4→v5, with the same fingerprint discipline as the preflight/baseline migrations). New sessions carry `revoked`; a v4 contract migrates to `honored` and resumes under the policy it actually ran with. The runner adopts the persisted value before the drift comparison, so a resumed session is not read as drift against this build's default, and the executor consults the session policy — not the Seed — at both consumption sites (`_contract_carries_evidence`).

Coverage: `test_v4_session_resumes_under_the_policy_it_ran_with`, `test_v4_resume_rejects_a_forged_pre_vacuity_contract` (migration is not a way past the fingerprint), and executor-level tests that a vacuous contract is inert under `revoked` and still recovers a failed AC under `honored`.

Note for the stack: #2182 folds this and its own `verify_baseline_probe` field into a single migration ladder, so the merged result is v6 with one rung per field.

Q00:
Rebased onto the updated #2180 (`bca0c1a86`), which answers its round-3 review: the verify gate now strips shell startup controls (`BASH_ENV` and family), accepts only an absolute resolved shell, and no longer substitutes `sh` for bash. Tip here is `08f22898a`.

The round-2 findings on this PR were answered in the comment above and are unchanged by that merge. Full `tests/unit/orchestrator` + `core` + `config` on this tip: 6762 passed, 1 skipped; ruff, format, mypy, and the module-size gate clean.

Q00:
Superseded by #2187. The replacement removes the static vacuity proof and durable vacuity policy by making verify commands additive evidence that can never weaken transcript obligations or recover failed work.

PR #2182, opened 2026-08-18
feat(verify): Stage 0+1 baseline negative control with two-tier verdict vocabulary (observe-only)

Q00:
> **Stacked on #2181** (`feat/verify-vacuity-seed-gate`), which stacks on #2180. Merge bottom-up; squash-only repo → retarget + `rebase --onto` per the established stack mechanics.

PR 1 of the evidential-force RFC (`docs/rfc/verify-gate-evidential-force.md`, #2179). Owner decisions recorded on the issue: two-tier vocabulary approved, withhold-only invariant (including recovery revocation) approved.

## What

**Stage 0 — snapshot.** At run start, strictly before the first worker dispatch and never on a resume, the workspace is copied (plain `copytree`, `.git` excluded, escaping symlinks reject the snapshot — `shadow_replay`'s vetted discipline). Unlike shadow replay the copy **keeps `.venv`/`node_modules`**: the probe runs the verify command itself, which needs a runnable environment. **Never a sibling git worktree.**

**Stage 1 — baseline negative control.** Every contract-carrying AC's full contract runs against a **fresh per-AC copy** (the measured corpus contains `rm`-bearing commands), reusing `_run_ac_verify_gate` via the Foundation-A authority entry — timeout, output tail, sanitized environment all inherited. Result recorded per AC (`probed`, `baseline_verdict`, `preexisting_artifacts`) in the checkpoint under bounded exact-keys rules.

**Vocabulary.** `verdict_tier: "discriminating_pass" | "pass" | None` on the serialized `_VerifyGateOutcome`. Legacy checkpoints read as `None` → plain pass, the conservative direction. String-typed so Stage 2 (PR 3) gates the same field without another schema change.

**Consumption (withhold-only).** A baseline-passing contract:
- loses the discriminating stamp → `execution.verify.non_discriminating` (mode `observe`: `evidence_exemption_changed: false` — the exemption consequence flips one release later, mirroring `run_verify_commands`/`seed.verify_command_gate` staging);
- **loses recovery authority**: `execution.verify.recovered` is unreachable for it — a contract that passes on the empty workspace must never resurrect a worker-reported failure.

**Honest-FAIL rule.** Probe infrastructure failures (no snapshot, no shell, start error, timeout) record `probed=False` — unknown grants nothing and revokes nothing, so an infrastructure hiccup can never mint a discrimination certificate. Only a genuine judgment (exit status, assertion, missing artifact, mutation) counts as baseline FAIL.

**Broken-seed finding.** All probed contracts passing at baseline → `execution.verify.baseline_all_pass` + the seed-gate warning channel at run start.

Config: `execution.verify_baseline_probe: off|observe` (default `observe`). Deliberately **not** added to the durable execution-semantics `_CURRENT_KEYS` — resumed behavior is governed by the checkpointed records, so a semantics version bump (forcing migration of every in-flight session) would buy nothing. Flagged here for review.

## Falsification (RFC cases, as unit analogues)

- (a) pre-satisfied contract → baseline PASS → withhold + `recovered` unreachable ✅ (`test_baseline_pass_revokes_recovery_authority`)
- (b) contract failing before the work → `discriminating_pass`, recovery preserved ✅
- destructive-contract isolation, snapshot symlink rejection, checkpoint round-trip + malformed→unknown, probe-off byte-identical behavior — all in `tests/unit/orchestrator/test_verify_baseline.py` (23 tests).

## Verification

- Isolated checkout of this commit: 629 passed (`test_verify_baseline` + both parallel-executor suites), ruff + format + module-size ratchet green (`parallel_executor.py` shrinks net; all new logic in `orchestrator/verify_baseline.py`), mypy clean on the new/changed modules.
- Full `tests/unit` minus mcp: 18,338 passed; every failure also fails on `origin/main` in the same harness (cli/setup + install-doctor family — pre-existing, none in orchestrator/core/config).

## Known residuals (deliberate, in-code comments say so)

- Probe wall-clock is N×`verify_command_timeout_seconds` worst-case; a shorter probe timeout would mint false baseline-FAILs (the one unsafe direction), so per-probe `duration_seconds` on `execution.verify.baseline` collects the data instead.
- A run interrupted before its first checkpoint re-enters with no recognized checkpoint; the failure mode is a missed withhold, never a false verdict — acceptable for an observe release.
- `--skip-completed` externally-satisfied and coordinator-revalidation outcomes carry `verdict_tier=None` (read as plain pass — conservative).


Q00:
@ouroboros-agent All four blockers and the follow-up addressed in `7f2b3b669` — each was a correct, reproduced finding.

**Blocker 1 (venv symlinks).** The snapshot policy now rejects only symlinks that resolve back **into the live workspace** — the one shape through which a probed `rm` could reach the tree the snapshot exists to protect. Links to the wider system (`.venv/bin/python → installed interpreter`, the standard virtualenv shape) are kept: they are exactly what the gate's own later live-tree execution of the same command would use. Regression test builds a realistic venv-style interpreter link and asserts the baseline works; the reject case is an absolute link back into the live tree.

**Blocker 2 (crash before first checkpoint).** The baseline is now persisted in a zero-levels checkpoint immediately after probing, strictly before the first dispatch, and the probe guard is `not checkpoint_recognized` rather than `resume_from_level == 0` — any recognized checkpoint, even one whose baseline payload is unreadable, fails closed to unknown instead of re-probing a possibly-mutated workspace as pristine. No-op when no baseline exists, so contract-less/probe-off runs keep their exact checkpoint cadence (`test_route_pause_aborts_…` still holds). Test asserts the persisted state round-trips through `restore_verify_baseline`.

**Blocker 3 (durable semantics).** Accepted as argued: `verify_baseline_probe` joins the schema — version 4 → 5, `_CURRENT_KEYS` + validator clause, and a v4 admission (`valid_pre_baseline_probe_execution_semantics_contract`) that migrates by injecting the schema default `"observe"` under the same fingerprint discipline as the legacy-preflight migration (fingerprint checked over the legacy value, then re-fingerprinted; chains with a v4 preflight shape). The executor now reads the flag from the durable snapshot at the ctor call, not live config.

**Blocker 4 (git contracts).** A `verify_command` referencing `git` (word-boundary — `./legit.sh` does not trip it) records `probed=False` with an explicit reason: the gitless snapshot has different semantics than the live tree, so nothing it reports is a judgment. Unknown grants nothing and revokes nothing.

**Follow-up (all-pass coverage).** `all_contracts_baseline_pass` now requires every contract record to be probed **and** pass; one unknown forbids the claim and suppresses the event. The all-pass warning also renders as markup-safe `Text`.

Suites: verify_baseline (29), parallel-executor pair, routing-contract-resume (updated v5 pin), route-escalation-live — green; ruff/format/module-size clean.

Q00:
All three blockers reproduced and are fixed in `e0af567a6`.

**1. Writable escaping symlinks.** The round-1 policy ("reject only links back into the live tree") kept every *other* escaping link as a link, and `copytree(symlinks=True)` preserves them — so a probe could write straight through one. A disposable copy that anything can write out of is not disposable, and the corpus contains `rm`.

The snapshot now contains every link or refuses to exist:

- resolves inside the snapshot → kept (already contained);
- a regular file outside → replaced by a copy of that file. This is the `.venv/bin/python -> installed interpreter` case that motivated the round-1 relaxation: the command still gets the interpreter, minus the write-through. Test writes through the copied path and asserts the original is untouched;
- anything else outside — a directory, a dangling target, a link back into the live tree — has no containment that preserves its meaning, so the snapshot is refused and the baseline stays unknown.

**2. The zero-levels checkpoint was not a durability gate.** Right on both counts: `save()` reports ordinary failures as `Result.err` and the result was dropped, and raised failures were swallowed as best-effort.

`persist_verify_baseline_checkpoint` now returns whether the baseline is replay-safe (checking `Result.is_ok` *and* catching), and the caller refuses to dispatch when it is not: a baseline only this process can see is authority the next process cannot check, and a crash before the first level save would re-enter with nothing recognized and probe a worker-modified tree as pristine. Nothing has been dispatched at that point, so refusing leaves the workspace exactly as the operator left it and the run can be started again.

The same rule closes the case the review did not raise: **no checkpoint store, no baseline.** The probe no longer runs at all without a store, since its records could never be written down.

Coverage: `Result.err` save, raising save, no-store, and `test_workers_do_not_dispatch_under_an_unpersistable_baseline`, which drives `execute_parallel` and asserts `_execute_ac_batch` was never awaited.

**3. Aliased numeric keys.** Confirmed — `"0"` and `"00"` both pass `isdigit` and both normalize to `0`, so the later record silently replaced the earlier one (your probe's unknown → baseline PASS is exactly the direction that revokes recovery). Keys must now be the one canonical spelling of their index (`str(int(k)) == k`), ASCII, and unique after normalization; anything else makes the whole payload unknown, per the module's existing "salvage nothing" rule.

**Stack note.** This branch merges #2181's `vacuous_contract_evidence` field. Rather than carry two near-identical v4 migration blocks, they became one ladder in `execution_semantics.py` — `(4, vacuous_contract_evidence, "honored")`, `(5, verify_baseline_probe, "observe")`, current version 6 — where each rung restores what that session actually ran under. The runner has a single migration block again, with the fingerprint check unchanged.

The RFC's Stage 0 section records both containment rules, so the design doc and the code say the same thing.

Q00:
Rebased onto the updated #2180 (`bca0c1a86`), which answers its round-3 review: the verify gate now strips shell startup controls (`BASH_ENV` and family), accepts only an absolute resolved shell, and no longer substitutes `sh` for bash. Tip here is `df8dcb9e8`.

The round-2 findings on this PR were answered in the comment above and are unchanged by that merge. Full `tests/unit/orchestrator` + `core` + `config` on this tip: 6762 passed, 1 skipped; ruff, format, mypy, and the module-size gate clean.

Q00:
Superseded by #2187. The unsandboxed baseline probe is removed from the ooo run critical path; future discrimination work will use event-native contracts and a real sandboxed evaluation boundary.

PR #2184, opened 2026-08-18
feat(advisory): a lane is offered only its own findings, as ids to fetch

cohemm:
Implements RFC #2167.

## Why this exists

[#2171](https://github.com/Q00/ouroboros/pull/2171) moved the artifact store to SQLite, which removed the filesystem paths a lane used to be handed. It replaced them by inlining the finding bodies into the prompt. **That broke the fan-out**, and the breakage shipped: it is on `main` now.

## The failure, measured

Six lanes each carried the same 6,722-character block — **40,332 characters of duplicated findings in one response**. The tool result reached **108,581 characters**, exceeded what the host accepts inline, and was written to a file. The host then spent the turn parsing its own output and dispatched no lane at all.

A/B in two identical worktrees, differing only in branch, confirmed it:

| | `origin/main` | the branch |
|---|---|---|
| MCP response | inline | **108,581 chars → written to file** |
| fan-out | `Running 2 agents…` | **none dispatched** |

It is worse than a size regression because it is delayed: the first interview in a project is fine, and it breaks once findings accumulate.

## Two growth axes, not one

**Multiplication.** A per-lane budget bounds one prompt, but the response carries every lane, so the cost is that budget times the number of lanes. #2167 removes it by narrowing to the producing lane.

**Accumulation survives that.** This project reached six findings for one lane inside a day — 14,638 characters inlined against 660 for their ids. At the twenty this admits, inlining is ~48,000 characters, larger than the response that broke, while the ids stay near 2,200.

**Inlining costs what children happen to write; ids cost how many times they wrote.** Only the second is a number this design gets to bound, so the budget that guarded the first is removed with the inlining it guarded.

## What this does

A lane is told **how many** findings its own lane published and given their contract ids; it fetches what it wants with `ouroboros_fetch_artifact`.

The count is not decoration. Fetching can fail, and the objection to ids was exactly that — a lane that cannot reach the tool looks like a project with nothing to reuse. Naming the number before naming the tool answers it: the lane knows something is there and says so, which is [#2167](https://github.com/Q00/ouroboros/issues/2167)'s one binding property.

**A lane answering under a closed contract is offered none.** Three of the four eligible lane-and-tool pairs answer under fixed shapes that reject any field they do not name. Such a lane has nowhere to put *"I could not reach the tool"* — writing it discards the whole answer, and all three are `required`, so the fan-out then cannot complete — and nowhere to *use* a finding either, since what it may return is policy claims from this question's roster or aggregates measured here. The block goes to the one lane whose answer is prose.

The alternative was a status field on every eligible lane. It is the right shape for that message and the wrong thing to build here: it adds a field to three closed contracts to report a failure that exists only because of a fetch, and a child that must declare its own failure is the enforcement theatre #2167 already ruled out.

**The rule lives in one place.** `_findings_reading_lanes` is read where the catalog is read, and the request then carries only what some lane will render — a key nothing reads is a promise the schema makes and the prompt never keeps.

**Findings are a head start, never a substitute**, and the prompt says so on both branches, including after a failed fetch. Left unsaid, *"say so"* reads as terminal and turns a missed shortcut into a missed answer.

## Result

| | before | after |
|---|---|---|
| lanes carrying the block | 6 | **1** |
| findings cost, 6 findings × 4,000 chars | 24,000+ | **1,173** |
| `request.recent_findings` on PM | both lanes | **absent** |

Verified live: the fan-out dispatches all six lanes, the child fetches, and its finding cites what it read.

## Verification

- `tests/unit` — 20,241 passed; 7 failures all pre-existing on `main` (6 × `test_install_runtime_selection` UnicodeDecodeError, 1 × `test_picker_indexes` wall-clock budget, both reproduced in a clean `origin/main` worktree)
- `tests/integration` + `tests/e2e` — 338 passed, 4 skipped, 1 xfailed
- `ruff check` / `ruff format --check` / `mypy src/ouroboros` (586 files) / `check-module-size.py` — all pass

## Notes

RFC #2167's *"how a lane reaches its findings"* was left to implementation. It turned out to be the same question as *who may read*, so it is answered in #2167 rather than here. #2168 keeps a pointer: it is approved and its scope is the storage cutover.


cohemm:
Addressed review round 1 in 2e146bbf — 1 blocker fixed, 0 defended.

**Blocker 1, `recent_findings.py:186` — fixed.** Confirmed exactly as described, and reproduced against this project's own store before changing anything: a `code_context` lane fetching what it was offered received 10,416 characters of which 2,138 were its own. The remainder included the `ambiguity_contrarian` lane's *"the question presents the wrong axis"* — the lane `recent_findings` excludes from eligibility precisely because a challenge to the question answers a question nobody asked. Grouping by producing lane sorted the index and left the address alone; a fan-out publishes one artifact carrying every dispatched lane, so its contract id names the turn.

Fixed at the address, not with a rule the child is asked to follow. What is offered is now `<contract>#<lane>`, and the store resolves it by filtering `aggregated_outputs` on `$.lane_id`:

```sql
SELECT json_extract(lane.value,'$.output')
FROM artifacts, json_each(body,'$.result.aggregated_outputs') AS lane
WHERE contract_id = ? AND json_extract(lane.value,'$.lane_id') = ?
```

`$.lane_id` is the key the fan-out assigned from its own dispatch roster (`fanout.py:1026-1030`), so it names the lane the server sent the work to. The `lane_id` a child writes inside its own output is deliberately not read — it is absent from three of the eight bodies in this store. A one-lane body has nothing left in it to select, which is what makes the request schema's claim and the transport agree.

Your probe, re-run: the id offered under `code_context` now fetches `{"finding": ...}` for `code_context` alone — `data_context` and `ambiguity_contrarian` are absent from the body, not filtered out of a rendering.

Pinned (`tests/unit/mcp/tools/test_recent_findings.py`), both directions:
- `test_fetching_the_offered_id_returns_only_the_requesting_lane` — the offered id returns only that lane, asserted by content markers unique to each sibling
- `test_each_lane_fetches_its_own_output_from_the_same_fan_out` — two lanes of one turn get two distinct addresses resolving to their own outputs
- `test_an_unscoped_contract_id_still_fetches_the_whole_body` — every caller outside the advisory path is untouched
- `test_a_lane_absent_from_a_fan_out_has_no_address_in_it` — scoping to an uncarried lane is absence, never a sibling's output

On the two "Not met" requirements: both were this one defect, and both now hold — the fetched body is the requesting lane's output, so no child-side selection remains.

No new rows, no schema change, no change to the publication path. `ouroboros_fetch_artifact`'s parameters are unchanged; the child passes back the address it was given.

Verification: `tests/unit` 20,292 passed / 88 skipped; `tests/integration` + `tests/e2e` + `tests/conformance` 381 passed, 4 skipped, 1 xfailed; `ruff check`, `ruff format --check`, `mypy src/ouroboros` (586 files), module-size, auto-boundary, and max-turns-envelope all pass. The two failures noted as pre-existing in the PR body no longer reproduce — `origin/main`'s 31e001b6 fixed the install-runtime ones, and the picker budget flake did not fire.

cohemm:
Addressed review round 2 in d97ec45d — 2 blockers fixed, 0 defended.

Both were in what round 1 added, not what it was fixing. Reproduced exactly as described before changing anything.

**Blocker 1, `artifact_store.py:285` — fixed.** `json_extract(body,'$.output')` decodes to a SQL scalar rather than JSON: a prose string came back with its quotes stripped (`json.loads` then raised `ArtifactIntegrityError`), and a boolean or number came back as a Python scalar `json.loads` cannot parse at all. Fan-out submission accepts any JSON-native `content` (`fanout.py:847`), so a lane's whole output can legitimately be a string or a primitive, not only an object.

Fixed by selecting with `->` instead of `json_extract` — SQLite's `->` operator returns canonical JSON for every type. Pinned with a parametrized test over string, int, float, both bools, null, an array, an object, and an empty string, each fetched and compared to the original Python value.

**Blocker 2, `artifact_schema.py:61` — fixed.** `validate_contract_id` defines length as the entire identity rule; nothing restricts character content once the store stopped being a directory. Folding `<contract>#<lane>` into one string gave an ordinary id containing `#` — always valid to store — a second reading: your probe (`fetch_if_exists("ordinary#id")` succeeds, `fetch("ordinary#id")` raises not-found) reproduced exactly.

Fixed by not encoding at all. The lane now travels as its own value beside the contract id — in the findings entry (`{"contract_id", "lane_id", "published_at"}`), in the fetch tool's new optional `lane_id` parameter, and in the store's `fetch_lane(contract_id, lane_id)`. There is no address to parse and so none to misparse. Pinned: an ordinary id containing `#` is fetchable and replayable unchanged.

The design direction you confirmed — server-side lane filtering in the store, not a rule the child follows — is unchanged; only the two storage contracts you found broken (JSON-native output types, the existing contract-id namespace) are now honored alongside it.

Also applied, non-blocking:
- **Suggestion 1** (count wording): the prompt now says "you are offered N", not "published N" — the list is capped at twenty, so a total was a number this cannot know.
- **Suggestion 2** (fetch tool docs): the `ouroboros_fetch_artifact` description and its new `lane_id` parameter now state both address forms.

`ouroboros_fetch_artifact` gained one optional parameter (`lane_id`); every existing call omitting it is unaffected, pinned by the unscoped-fetch regression test.

Verification: `tests/unit` 20,315 passed / 88 skipped; `tests/integration` + `tests/e2e` + `tests/conformance` 381 passed, 4 skipped, 1 xfailed; `ruff check`, `ruff format --check`, `mypy src/ouroboros` (586 files), module-size, auto-boundary, and max-turns-envelope all pass. CI's `test_picker_progress_queries_are_aggregate_bounded_at_scale` failure on the prior push is a wall-clock budget assertion (`< 0.5s` over 100k inserted rows) in `tests/unit/dashboard_web/test_reader_readonly_uri.py`, unrelated to any file this PR touches and not reproducing locally across three runs — the same class of flake the PR body already notes for `test_picker_indexes`.

cohemm:
Addressed review round 3 in bbea3819 — 2 blockers fixed, follow-up applied, 0 defended.

**Blocker 1, `fanout_handler.py:292` — fixed.** Reproduced: `{"contract_id": ..., "lane_id": "   "}` returned the complete body, siblings included. The defect was the normalization itself — strip, then branch on truthiness — so the fix removes it rather than guarding it. Presence now decides the path: an absent or JSON-null `lane_id` is the legacy whole-artifact read; anything supplied travels verbatim to the store, where a lane no fan-out ever dispatched is not-found. There is no rejected-shapes list to extend — `""`, `"   "`, `"\t"`, `0`, `False`, `[]`, `{}`, and `"data_context "` (trailing space) all fail closed by the same absence of coercion. Handler-level regressions pin the blank/whitespace failures, the unknown-lane failure, both omission forms staying unscoped, and the one valid scoped path beside them.

**Blocker 2, `artifact_store.py:291` — fixed.** Reproduced: after pruning, `fetch` raised `ArtifactTombstonedError` while `fetch_lane` raised `ArtifactNotFoundError` for the same contract. The cause was `json_each` as an inner table source: a pruned NULL body yields no lanes, so the artifact's row vanished before the tombstone check could run. The join is now `LEFT JOIN`, so the row survives and the three absences stay three — no row is no contract, a NULL body raises the same `ArtifactTombstonedError` the unscoped path raises (for any requested lane), and a NULL lane match on a live body is not-found. Pinned: tombstone through scoped read (carried lane and never-dispatched lane alike), missing contract vs missing lane, and malformed body shapes (missing path, scalar result, non-list outputs) returning absence or a typed error, never a sibling's output.

**Follow-up 1, `interview_schemas.py:1073` — applied.** Each entry's `lane_id` is now a `const` matching its enclosing lane key, so an entry pairing a lane key with a sibling's identifier is unrepresentable rather than trusted. Pinned with a negative case: mutating a valid request's entry to name a sibling lane fails validation.

Also swept the changed files for prose disagreeing with the transport — the class round 1 established: two docstrings still said the offered ids were "this lane's own" (they are the turn's id paired with the lane's own identifier); both now state the two-value contract, and the tool parameter documents that a supplied lane the artifact does not carry is an error, never a broader read.

Verification: `tests/unit` 20,334 passed / 88 skipped; `tests/integration` + `tests/e2e` + `tests/conformance` 381 passed, 4 skipped, 1 xfailed; `ruff check`, `ruff format --check`, `mypy src/ouroboros` (586 files), module-size, auto-boundary, max-turns-envelope all pass. The prior push's `test_pipeline_deadline` CI failure was re-run and passed — runner flake, unrelated to this PR's files.

PR #2186, opened 2026-08-18
fix(init): make "Start workflow now?" actually run the Seed

### LINKED ISSUE #2185: init start: "Start workflow now?" does nothing without --orchestrator
> **A branch already implements this** — `feat/init-start-workflow-runs` is pushed and a PR follows immediately. No bot PR is needed for this issue.

## Summary

`ouroboros init start` finishes the interview, generates a Seed, then asks **"Start workflow now? [y/n] (y)"**. Answering yes does nothing unless the same invocation also passed `--orchestrator/-o`. Without that flag the confirmation prints two lines and returns:

```
Starting workflow...
Would execute workflow from: /Users/…/.ouroboros/seeds/seed_xxxx.yaml
Standard workflow execution not yet implemented.
```

The prompt names an action the command will not perform.

## Reproduction

1. `ouroboros init start "<any idea>"` (no `-o`).
2. Answer questions until a Seed is generated.
3. Answer `y` to "Start workflow now?".
4. Two info lines print; no execution, no session, no dashboard.

## Cause

`_start_workflow()` (`src/ouroboros/cli/commands/init.py`) branches on `use_orchestrator`, which comes from `--orchestrator/-o` and defaults to `False`:

```python
if use_orchestrator:
    await _run_orchestrator(seed_path, resume_session=None, parallel=parallel,
                            runtime_backend=runtime_backend)
else:
    print_info(f"Would execute workflow from: {seed_path}")
    print_info("Standard workflow execution not yet implemented.")
```

The placeholder dates to `4b72e0e7` ("feat(init): add end-to-end interview to workflow pipeline", 2026-01), whose stated goal was to "connect interview → seed → workflow in single flow". Only the orchestrator half was implemented. Meanwhile `ouroboros run workflow` moved to orchestrator-by-default (`--no-orchestrator/-O` to opt out), leaving `init` holding the opposite default plus a path that was never built.

## Second half: the working directory

Executing the Seed is not enough on its own. `run workflow` resolves the project directory from the Seed's own location when none is given (`_resolve_cli_project_dir`), and `init` writes Seeds to `~/.ouroboros/seeds`. Running a generated Seed by hand shows the result today:

```
orchestrator.adapter.initialized … cwd=/Users/<user>/.ouroboros/seeds
```

The agent builds inside the seed store. An interview is conducted from the project it is about, so the invocation directory is the project and the handoff should pass it. A Seed that names a brownfield `target_dir` must keep precedence over the caller's cwd.

## Proposed fix

- Take the single execution path `run workflow` already takes by default; delete the unimplemented branch rather than inverting a flag.
- Pass the invocation directory as the project directory, except when the Seed names a brownfield `target_dir`, where resolution stays with the run command.
- `--orchestrator` keeps selecting the interview's LLM backend; only the workflow handoff stops reading it.

## Out of scope

`ouroboros run workflow --no-orchestrator` has its own placeholder (`run.py:1020`). That one is behind an explicit opt-out flag a user has to type, not behind a confirmation prompt, so it is not part of this fix.

cohemm:
Closes #2185.

## Summary

`init start` asks "Start workflow now? [y/n] (y)" and then does nothing unless the same invocation also passed `--orchestrator/-o` — the other branch printed "Would execute workflow from: …" plus "Standard workflow execution not yet implemented" and returned. The confirmation named an action it did not perform.

- The handoff now takes the single execution path `ouroboros run workflow` already takes by default. The flagged branch is deleted rather than inverted; that placeholder has been unimplemented since the pipeline landed (`4b72e0e7`, 2026-01), while `run workflow` moved to orchestrator-by-default (`--no-orchestrator/-O` to opt out). `init` was the one entry point still holding the opposite default.
- It also passes the invocation directory. Without it, `_resolve_cli_project_dir` falls back to the Seed's own location, and `init` writes Seeds to `~/.ouroboros/seeds` — so the agent would build inside the seed store. This is not hypothetical: running a generated Seed by hand today logs `cwd=/Users/<user>/.ouroboros/seeds`. An interview is conducted from the project it is about, so that directory is the project.
- A Seed that names a brownfield `target_dir` keeps precedence, and so does Seed metadata and an explicit `--project-dir`. All of that precedence lives in `_resolve_cli_project_dir`, which now takes a `fallback_dir` standing in for the Seed file's own folder; `init` supplies the invocation directory and interprets nothing itself.

Review round 1 corrected the shape of that second point. The first attempt had `init` read the Seed to decide, which meant two readers of the same metadata under different rules: `init` deferred to any non-empty `target_dir`, while `_resolve_brownfield_target_dir` drops targets that are missing or not directories — so a stale target fell through to the Seed's folder, back inside `~/.ouroboros/seeds`. Deleting the second reader closes that, and it also removes a crash: `yaml.safe_load` on a valid `- item` document made `init` raise `AttributeError` before the run command's own loader could report the real error.

`--orchestrator` still selects the interview's LLM backend (`_resolve_init_llm_backend`); only the workflow handoff stops reading it.

## Test plan

- [x] `uv run pytest tests/unit/cli tests/unit/core -q` — 2800 passed, 7 skipped
- [x] `uv run pytest tests/unit/mcp tests/unit/auto tests/unit/bigbang -q` — 5035 passed, 1 skipped
- [x] `uv run pytest tests/integration tests/conformance -q` — 304 passed, 4 skipped, 1 xfailed
- [x] `uv run ruff check`, `ruff format --check`, `mypy src/ouroboros/cli/commands/init.py` — clean
- [x] New tests: yes-without-`-o` reaches `_run_orchestrator`; the handoff forwards the invocation directory and no longer passes `project_dir`; and at the resolver — a Seed that says nothing resolves to the fallback, a missing or file-valued `target_dir` resolves to the fallback rather than the Seed store, a real `target_dir` outranks it, and `--project-dir` outranks everything

## Not in this PR

`ouroboros run workflow --no-orchestrator` has the same kind of placeholder (`run.py:1020`). It sits behind an explicit opt-out flag the user has to type rather than behind a confirmation prompt, so it is left alone.

### DIFF
--- a/docs/guides/seed-authoring.md
+++ b/docs/guides/seed-authoring.md
@@ -740,21 +740,24 @@ ouroboros run minimal_seed.yaml
 
 ### CLI Flag Warnings
 
-#### `--runtime` without `--orchestrator` (init command)
+#### `--runtime` and `--orchestrator` (init command)
 
-**Symptom:**
-```
-Warning: --runtime only affects the workflow execution step when --orchestrator is enabled.
-```
+The two flags act on different stages, and neither gates the other:
 
-**Cause:** `--runtime` (e.g., `--runtime codex`) was passed to `ouroboros init` without `--orchestrator`. The `--runtime` flag only controls which agent runtime backend is used when the generated seed is immediately handed off to workflow execution. Without `--orchestrator`, the workflow handoff step uses a placeholder.
+- `--runtime` selects the agent runtime for the workflow handoff — the execution
+  that starts when you accept "Start workflow now?". It applies whether or not
+  `--orchestrator` is passed.
+- `--orchestrator` selects Claude Code as the LLM backend for the *interview and
+  seed generation*, and is equivalent to `--llm-backend claude_code`.
 
-**Behavior:** This is a **warning only** — the interview and seed generation proceed normally. The runtime flag has no effect.
-
-**Fix:** Add `--orchestrator` if you want to use the specified runtime backend for the post-generation workflow step:
 ```bash
-ouroboros init start --orchestrator --runtime codex "Build a REST API"
+# Interview on the configured backend, execute the Seed with the Codex runtime
+ouroboros init start --runtime codex "Build a REST API"
 ```
+
+Earlier releases warned that `--runtime` had no effect without `--orchestrator`,
+because the handoff without that flag printed a placeholder instead of executing.
+The handoff now always runs, so the warning is gone.
 
 ---
 
--- a/src/ouroboros/cli/commands/init.py
+++ b/src/ouroboros/cli/commands/init.py
@@ -835,20 +835,22 @@ async def _start_workflow(
     # default. The previous non-orchestrator branch printed "not yet
     # implemented" and returned, so answering yes to "Start workflow now?"
     # did nothing unless the caller had also passed --orchestrator.
+    #
+    # Nothing is caught here. `_run_orchestrator` signals every failure —
+    # unloadable Seed, unsafe project path, workspace error, failed execution —
+    # as `typer.Exit(1)` and has no zero-code exit to absorb, so catching it
+    # printed the error and still finished `init start` successfully. A caller
+    # in a script cannot tell a built product from a failed one, and Ctrl+C read
+    # the same way. Sharing the run command's path means sharing its exit code.
     from ouroboros.cli.commands.run import _run_orchestrator
 
-    try:
-        await _run_orchestrator(
-            seed_path,
-            resume_session=None,
-            parallel=parallel,
-            runtime_backend=runtime_backend,
-            project_fallback_dir=project_fallback_dir,
-        )
-    except typer.Exit:
-        pass  # Normal exit
-    except KeyboardInterrupt:
-        print_info("Workflow interrupted.")
+    await _run_orchestrator(
+        seed_path,
+        resume_session=None,
+        parallel=parallel,
+        runtime_backend=runtime_backend,
+        project_fallback_dir=project_fallback_dir,
+    )
 
 
 def _find_pm_seeds(seeds_dir: Path | None = None) -> list[Path]:
--- a/src/ouroboros/cli/commands/init.py
+++ b/src/ouroboros/cli/commands/init.py
@@ -839,9 +839,14 @@ async def _start_workflow(
     # Nothing is caught here. `_run_orchestrator` signals every failure —
     # unloadable Seed, unsafe project path, workspace error, failed execution —
     # as `typer.Exit(1)` and has no zero-code exit to absorb, so catching it
-    # printed the error and still finished `init start` successfully. A caller
-    # in a script cannot tell a built product from a failed one, and Ctrl+C read
-    # the same way. Sharing the run command's path means sharing its exit code.
+    # printed the error and still finished `init start` successfully: a caller
+    # in a script could not tell a built product from a failed one.
+    #
+    # `KeyboardInterrupt` is not caught either, but that is about ownership, not
+    # exit codes. The command wrapper below still answers Ctrl+C with exit 0 and
+    # "Interview interrupted. Progress has been saved.", which is the policy for
+    # an interactive command; the point is that one place decides it instead of
+    # this handoff printing a competing line on the way there.
     from ouroboros.cli.commands.run import _run_orchestrator
 
     await _run_orchestrator(
--- a/src/ouroboros/cli/commands/init.py
+++ b/src/ouroboros/cli/commands/init.py
@@ -1153,11 +1153,6 @@ def start(
     if debug:
         print_info("Debug mode enabled - showing verbose logs")
 
-    if runtime and not orchestrator:
-        print_warning(
-            "--runtime only affects the workflow execution step when --orchestrator is enabled."
-        )
-
     # Show mode info
     selected_llm_backend = _resolve_init_llm_backend(
         orchestrator,
--- a/src/ouroboros/cli/commands/init.py
+++ b/src/ouroboros/cli/commands/init.py
@@ -649,7 +649,7 @@ async def _run_interview(
             await _start_workflow(
                 seed_path,
                 runtime_backend=workflow_runtime_backend,
-                project_dir=_workflow_project_dir(seed_path),
+                project_fallback_dir=Path.cwd(),
             )
 
     finally:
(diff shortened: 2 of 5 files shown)


cohemm:
Addressed review round 1 in 0f230095 — 2 fixed, 0 defended. Both findings were reproduced against `1d695a0f` before any code was written:

```
probe 1  target_dir="/tmp/does-not-exist-xyz"  → resolved to /Users/<me>/.ouroboros/seeds
probe 2  seed containing "- item"              → AttributeError: 'list' object has no attribute 'get'
```

**Blocker 1, `init.py:831` — stale brownfield target re-entered the Seed store — fixed.** Your Design Note named the real defect: `_workflow_project_dir()` duplicated *part* of the shared resolution contract. Rather than teach the copy the rest of the rules, the copy is gone. `_resolve_cli_project_dir` now takes a `fallback_dir` that stands in for the Seed file's own folder, and `init` passes the invocation directory. Precedence is unchanged and still lives in one function: explicit `--project-dir` > Seed metadata > valid brownfield target > fallback > the Seed's folder.

A stale target now falls through to the invocation directory because the rule that rejects it (`_resolve_brownfield_target_dir`) and the rule that fills the hole are the same rule — the two stages can no longer disagree, because there is one stage. Measured against your probe and its file-valued variant:

```
target_dir absent          → <invocation dir>
target_dir missing path    → <invocation dir>
target_dir is a file       → <invocation dir>
target_dir is a real dir   → <that dir>
```

Pinned in `tests/unit/core/test_seed_project_path_callers.py`: `test_fallback_dir_stands_in_for_the_seed_file_folder`, `test_unusable_brownfield_target_does_not_fall_back_to_the_seed_store` (parametrized over missing/file), `test_usable_brownfield_target_outranks_the_fallback`, `test_explicit_project_dir_outranks_the_fallback`.

**Follow-up, `init.py:828` — non-mapping YAML crashed the handoff — fixed by removal.** Nothing in `init` parses the Seed anymore, so there is no site left to guard: a `- item` document now reaches the run command's own loader, which is what should be reporting it. Pinned indirectly by `test_start_workflow_passes_the_invocation_directory_as_fallback`, which asserts the handoff forwards a fallback and no longer passes `project_dir` at all.

**One consequence worth naming.** Because `init` no longer decides, the invocation directory also becomes the base for relative Seed metadata and for the containment check — for `init`-generated Seeds only, since `run workflow` passes no fallback and resolves exactly as before. Both readings match what a person means by running the interview in that directory, so I left it; if you would rather the fallback apply only to the final branch, say so and I will narrow it.

Verification (parallel subagents): `tests/unit/cli tests/unit/core` 2800 passed / 7 skipped; `tests/unit/mcp tests/unit/auto tests/unit/bigbang` 5035 passed / 1 skipped; `tests/integration tests/conformance` 304 passed / 4 skipped / 1 xfailed. `ruff check`, `ruff format --check`, `mypy` on both changed modules: clean. No references to the deleted helper remain.

cohemm:
Addressed review round 2 in 4dee03dc — 1 fixed, 0 defended, plus the P1 follow-up test.

**Blocker 1, `init.py:848` — execution failure did not fail the command — fixed.** Reproduced first:

```
_run_orchestrator raises typer.Exit(1)  →  _start_workflow returns None
```

You are right that there was nothing to suppress. Audited: `_run_orchestrator` raises `typer.Exit(1)` at five points, and `run.py` contains no `typer.Exit(0)` and no bare `typer.Exit()` anywhere. The `# Normal exit` label described an exit that does not exist. The clause predates this PR — it sat behind `--orchestrator` — but making the handoff the default path is what turned a dormant mislabel into the common case, so it is this PR's to fix.

Both handlers are deleted rather than re-raised, since there was nothing to translate. Pinned at two levels: `test_start_workflow_propagates_execution_failure` (parametrized over `typer.Exit(1)` and `KeyboardInterrupt`) and `test_cli_exit_code_reports_a_failed_workflow`, a CliRunner test asserting `ouroboros init start` exits 1 — the contract a shell caller actually observes.

**One correction to my own reasoning, since it affects what you should expect.** I first wrote that Ctrl+C was lying about its exit code too. It was not, and dropping that clause does not change it: the command wrapper at `init.py:1193-1198` catches `KeyboardInterrupt` and raises `typer.Exit(code=0)` with "Interview interrupted. Progress has been saved." That is deliberate for an interactive command and is untouched here. What the removal actually buys is that the workflow no longer prints a competing "Workflow interrupted." line — interrupt policy now lives in one place. If you consider exit 0 wrong for an interrupt *after* execution has started, that is a separate contract question and I would rather take it as its own change than fold it in here.

**Follow-up 1, `init.py:648` — caller-level wiring — added.** `test_accepting_the_final_prompt_hands_off_the_invocation_directory` asserts the accepted prompt forwards `project_fallback_dir=Path.cwd()`, and `test_declining_the_final_prompt_starts_no_workflow` asserts declining awaits no handoff. The `finally:` that closes the event store still runs when the exit propagates — the handoff is the last statement of that `try:`, so propagation changes the exit code and nothing else.

Verification (parallel subagents): `tests/unit/cli tests/unit/core` 2804 passed / 7 skipped; `tests/integration tests/conformance` 304 passed / 4 skipped / 1 xfailed; `tests/e2e/test_full_workflow.py` 19 passed. `ruff check`, `ruff format --check`, `mypy` on both changed modules: clean.

cohemm:
Addressed review round 3 in 19f06f5f — 1 follow-up fixed, **1 blocker declined**, with an accommodation that costs no durable state.

**Follow-up 1 — stale `--runtime` warning — fixed.** You are right that my change made it false. The warning and the matching section in `docs/guides/seed-authoring.md` are replaced with what the flags now do: `--runtime` selects the agent runtime for execution and applies with or without `-o`; `--orchestrator` selects the interview's LLM backend.

**Blocker 1 — resumed interview uses the resume shell's cwd — declined, and here is why.**

*It is not this PR's contract.* #2185 is "the confirmation advertises an action it does not perform". That is met, and you marked it Met. What remains is a different contract: interview state does not record the project it was about. `InterviewEngine.start_interview()` already accepts `cwd` and records `codebase_paths[role=primary]` from it; `init.py` has never passed it, on `main` and on this branch alike. That gap predates this PR and is orthogonal to the confirmation prompt — it deserves its own issue, not a block on this one.

*The proposed remedy re-opens the defect class this PR just closed twice.* Round 1's blocker was a durable path pointer gone stale — a `brownfield_context.target_dir` that no longer exists, which routed execution into the Seed store. Persisting the original cwd is the same shape: a path recorded at interview time, resolved at execution time, with nothing keeping the two in step. The next probe writes itself — the directory was renamed, deleted, or the state file moved to another machine — and the answer at that point is either "fall back to cwd" (this finding again) or "refuse" (option B below, reachable today without storing anything).

*The trigger is a chain of deliberate choices, and the outcome is a directory the person is standing in.* To reach it a user must interrupt an interview, `cd` to a different project, resume there, and accept the prompt. Compare with the two blockers I did fix: round 1 misrouted **every** Seed with an absent or stale target, and round 2 mis-signalled **every** failure. Those are properties of the common path; this one is a property of a manual detour, and the directory it picks is the one the shell is in — not a global store.

*Meanwhile the alternative it is measured against is worse.* Without this PR, a user who wants the generated Seed executed runs `ouroboros run workflow <seed>` by hand, which for a Seed in `~/.ouroboros/seeds` resolves to the Seed store — wrong in every case, resumed or not. This branch is strictly better in both the fresh and the resumed case.

**Accommodation.** `_run_orchestrator` now prints `Project directory: <path>` before execution. Four inputs can decide that path and the winner was previously visible only in a debug log, so a person could not tell where the agent was about to write until it had written. This does not guess better; it makes the guess visible before the first write — for `run workflow` as much as for the `init` handoff. It adds no field, no prompt, and no durable state, and it is the part of your concern that can be closed without inventing a pointer that goes stale.

If you can show a resume path where the wrong directory is selected **without** the user changing directories themselves — or a concrete case where the printed directory is not enough to catch it before a write — I will treat it as blocking and fix it here. Otherwise I will open a separate issue for interview project identity (`start_interview(cwd=...)` never being passed), which is where that contract actually lives.

Verification: `tests/unit/cli` 2051 passed / 7 skipped; `tests/unit/core tests/integration` 1015 passed / 4 skipped / 1 xfailed; `ruff check`, `ruff format --check`, `mypy` on both changed modules clean.

Issue #2191, opened 2026-08-18
fix OMP interview calls timing out at 30 seconds


Q00:
> **Status**: in-flight — 3 stacked PRs ready for review

## Goal

Remove interview turn latency from the MCP deadline contract without assuming concurrent backend calls or weakening state ordering. Standard and PM interviews should derive all turn decisions from one revision and one provider completion.

## Stacked PR Layout

| # | PR | Phase | Scope |
|---|---|---|---|
| 1 | #2190 | Safety | OMP 60s timeout floor + structured closure marker |
| 2 | #2192 | Core | Atomic question + ambiguity planning for standard interviews |
| 3 | #2193 | PM | Atomic scoring + question classification/reframing |

Merge order: #2190 → #2192 → #2193

## Progress

- [ ] #2190 · Safety boundary
- [ ] #2192 · Shared turn planner and standard interview adoption
- [ ] #2193 · PM policy fusion

## Verification

- Full targeted stack gate: 1,499 passed, 7 skipped
- All provider adapter tests: 847 passed, 7 skipped
- Ruff and MyPy pass

## Out of Scope

- Concurrent calls through `LLMAdapter`
- Host-specific branching inside interview engines
- Replacing direct CLI interview flows in this stack

Q00:
Merged the complete approved stack: #2193 → #2192 → #2190, with final main merge commit `4edabb2c6820b8bf0e4bea9d73a46cc75cf8eb69`. Required checks passed on the exact final stack head. Closing completed.

PR #2190, opened 2026-08-18
fix(interview): raise OMP MCP timeout floor

Q00:
Refs #2191

## Summary

- Raise OMP's extension tool-call timeout floor from 30 seconds to 60 seconds during `install.sh` and native `ouroboros update`, while preserving higher user-configured values.
- Add a language-independent `[closure]` marker so refined completion decisions do not depend on Korean, English, or other natural-language phrase matching.
- Keep ambiguity scoring and next-question generation serial so questions always consume the latest score and providers receive only one model call at a time.

## Evidence

The reported OMP session aborted MCP calls at 30 seconds while the Ouroboros server completed the same turns in 34.2, 38.2, and 47.8 seconds. The server persisted the generated question, which is why a subsequent resume succeeded immediately.

## Verification

- `uv run pytest -q tests/unit/mcp/tools/test_interview*.py` — 132 passed
- `uv run pytest -q tests/unit/bigbang/test_interview*.py tests/unit/bigbang/test_question_candidate_panel.py` — 123 passed
- `uv run pytest -q tests/unit/cli/test_omp_config.py tests/unit/cli/test_update.py` — 189 passed
- `uv run pytest -q tests/unit/scripts/test_install_runtime_selection.py` — 117 passed
- Focused serial-order and closure regression suite — 74 passed
- `uv run mypy src/ouroboros/cli/omp_config.py src/ouroboros/cli/commands/update.py src/ouroboros/mcp/tools/authoring_handlers.py`
- Ruff checks and `bash -n scripts/install.sh`

## Notes

The local OMP configuration was applied immediately: `extensionHandlers.toolCallTimeoutMs = 60000`.

PR #2192, opened 2026-08-18
feat(interview): plan scored turns atomically

Q00:
Refs #2191

## Stack: interview turn planner (2/3)

| Layer | PR | Scope |
|---|---|---|
| 1 | #2190 | OMP timeout floor and structured closure marker |
| 2 | This PR | Atomic question + ambiguity planning for standard interviews |
| 3 | Follow-up | PM scoring + question classification fusion |

Base: `fix/interview-mcp-timeout`

## Summary

- Add `InterviewTurnPlanner`, deriving the next question and ambiguity snapshot from one interview revision through one ordinary `LLMAdapter.complete` call.
- Extract `InterviewEngine.prepare_next_question` so the legacy and atomic paths share prompt budgets, references, glossary injection, and PM steering.
- Adopt atomic planning in the production MCP interview handler after the minimum completion round, while keeping injected custom engines on the serial compatibility path.
- Treat the question as authoritative when score fields are malformed; completion remains disabled until a valid score exists.

## Why

The old MCP turn performed ambiguity scoring and next-question generation as two ordered model calls. The ordering was correct, but observed wall time reached 34–48 seconds. Atomic planning preserves the dependency without concurrent adapter calls, stale state, or doubled provider latency.

## Verification

- Standard interview and planner suites included in the 1,499-test stacked quality gate.
- All provider adapter tests pass; the planner uses only the existing single-call adapter protocol.

### DIFF
--- a/src/ouroboros/bigbang/ambiguity.py
+++ b/src/ouroboros/bigbang/ambiguity.py
@@ -500,16 +500,12 @@ class AmbiguityScorer:
 
             # Parse the LLM response into scores
             try:
-                breakdown = self._parse_scoring_response(
+                ambiguity_score = self.parse_score_response(
                     result.value.content,
                     is_brownfield=is_brownfield,
                 )
-                overall_score = self._calculate_overall_score(breakdown)
-
-                ambiguity_score = AmbiguityScore(
-                    overall_score=overall_score,
-                    breakdown=breakdown,
-                )
+                breakdown = ambiguity_score.breakdown
+                overall_score = ambiguity_score.overall_score
 
                 log.info(
                     "ambiguity.scoring.completed",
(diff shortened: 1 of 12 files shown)


Q00:
리뷰 요청드립니다. 표준 interview turn을 concurrency 없이 단일 atomic provider completion으로 통합했고, prompt budget과 scoring authority marker 회귀를 추가했습니다. 현재 HEAD: 05862e311

Q00:
Current HEAD 9434dd60 addresses the review blockers: deterministic adapter questions now run one scoring completion, and the fused prompt includes explicit closure/weakest-dimension/floor routing rules. Focused core suite passes.

Q00:
Current HEAD e0e4b34ad resolves both latest blockers: the fused contract renders canonical greenfield/brownfield component floors, and PreparedInterviewQuestion forwards preserve_prefix_messages through atomic re-budgeting. Focused long-context and floor regressions pass.

Issue #1307, opened 2026-06-01
[Epic] Add pi as a selectable Ouroboros AgentRuntime


Q00:
## Goal

Add `pi` as a first-class selectable Ouroboros `AgentRuntime` adapter in `/Users/jaegyu.lee/Project/ouroboros-gemini3`, using `/Users/jaegyu.lee/Project/pi` only as a local development reference for the upstream `earendil-works/pi` runtime, while executing Ouroboros agent tasks through a configurable pi native task/session/event bridge.

## Constraints

- Do not modify any files under `/Users/jaegyu.lee/Project/pi`.
- Do not replace, redesign, or broaden the existing `AgentRuntime` protocol; pi must fit the current interface.
- Do not change observable behavior of existing runtimes.
- Use canonical backend identifier `pi` for config/env selection and `runtime_factory` resolution.
- Do not embed `/Users/jaegyu.lee/Project/pi` or any user-specific absolute path in production runtime code.

## Acceptance Criteria

- [ ] `pi` resolves through backend capabilities, config/env, and `create_agent_runtime` while existing backend choices remain accepted.
- [ ] `PiRuntime` satisfies `AgentRuntime` properties and `execute_task` / `execute_task_to_result` contracts.
- [ ] Fixture-backed bridge receives `prompt`, `system_prompt`, `cwd`, `permission_mode`, `llm_backend`/`model_config`, `tool_policy`, `resume_handle`/session metadata, and `runtime_config_source`.
- [ ] pi events normalize into ordered `AgentMessage` values with final result semantics.
- [ ] Success returns `Result.ok(TaskResult)`; errors return `Result.err(ProviderError(provider='pi'))`.
- [ ] Bridge discovery, startup, mid-stream, and malformed-event failures become normalized error messages and `ProviderError` outcomes.
- [ ] Deterministic fixture bridge is included in normal pytest coverage.
- [ ] Opt-in configured native pi invocation works when bridge config is present and skips explicitly when absent.
- [ ] Production source does not contain the local pi clone path.
- [ ] Existing runtime-related tests continue to pass, with only additive pi-supported assertions.

## Ontology

| Field | Type | Description |
|-------|------|-------------|
| agent_runtime_protocol | object | Existing Ouroboros `AgentRuntime` interface that `PiRuntime` must satisfy without redesign. |
| runtime_backend | string | Canonical selectable backend identifier; expected value is `pi`. |
[...]

PR #2204, opened 2026-08-19
feat(pi): complete /ooo arguments in the interactive bridge

ruttybob:
Refs #1307 (Pi AgentRuntime epic this bridge work belongs to).

## Summary

The managed Pi bridge (`ouroboros setup --runtime pi`) registers the `/ooo` command but offers no argument completion. Interactive Pi users have to remember dispatch subcommand names and Seed filenames by hand.

This PR registers a `getArgumentCompletions` handler on the bridge command, using Pi's native completion surface (`RegisteredCommand.getArgumentCompletions` → `AutocompleteItem { value, label, description }`).

## What completes

| Input | Completion |
|-------|------------|
| `/ooo <TAB>` | dispatchable subcommands with one-line descriptions |
| `ooo run <TAB>` | Seed files (`*.yaml` / `*.yml`) from `~/.ouroboros/seeds/` |
| anything else | `null` — Pi falls back to its default behavior |

The subcommand list mirrors the packaged skills whose `SKILL.md` declares `mcp_tool` frontmatter — exactly the commands the shared dispatcher can execute, so completion never suggests a command that would end in the unsupported-dispatch exit code.

## Design notes

- **Deterministic and offline.** The completion path spawns no subprocess and reads only the local seeds directory; a missing directory degrades to no completion instead of an error.
- **Drift guard.** The bridge hardcodes the subcommand list so completion stays offline. A new unit test (`test_pi_bridge_completions_mirror_dispatchable_skills`) compares the rendered list against the packaged skills with `mcp_tool` frontmatter and fails whenever a skill gains or loses that frontmatter, so the hardcoded list cannot drift silently.
- **Typed structurally.** The extension declares a local `CompletionItem` type; `AutocompleteItem` is not re-exported from `@earendil-works/pi-coding-agent`, and the structural shape satisfies `getArgumentCompletions` without depending on `@earendil-works/pi-tui` directly.

## Docs

`docs/runtime-guides/pi.md` — the interactive bridge section now documents the `/ooo` TAB completion surface.

## Testing

- 2 new tests in `tests/unit/cli/test_setup.py`:
  - `test_pi_bridge_registers_argument_completions` — bridge source carries the completion wiring and the seeds-directory lookup.
  - `test_pi_bridge_completions_mirror_dispatchable_skills` — drift guard described above.
- `pytest tests/unit/cli/test_setup.py -k "pi"` → 76 passed, 1 skipped.
- `pytest tests/unit/cli` → 2031 passed, 7 skipped, 22 failed. The 22 failures are pre-existing on a clean `upstream/main` checkout in this environment (Codex setup/registration tests; identical set before and after this change, verified by re-running with the change stashed).
- The rendered TypeScript loads cleanly under Node `--experimental-strip-types`, and the completion function was exercised on the edge cases (`""`, `"ru"`, `"run "`, `"run <prefix>"`, `"run a b"`, unknown prefixes → `null`).
- `ruff format` + `ruff check` clean.

### DIFF
--- a/docs/runtime-guides/pi.md
+++ b/docs/runtime-guides/pi.md
@@ -171,6 +171,13 @@ to Pi with a deterministic unsupported-dispatch exit code so the normal Pi
 session can continue handling the input instead of receiving a hard bridge
 failure.
 
+The registered `/ooo` command also provides TAB argument completion through Pi's
+native `getArgumentCompletions` surface: `/ooo <TAB>` lists the dispatchable
+subcommands with one-line descriptions, and `ooo run <TAB>` lists Seed files
+from `~/.ouroboros/seeds/`. Completion is deterministic and offline — the
+subcommand list mirrors the packaged skills that declare `mcp_tool` frontmatter
+and a unit test keeps the two in sync.
+
 For `ooo auto`, the dispatcher owns the background job lifecycle. After
 `ouroboros_start_auto` returns a `job_id`, the dispatch process polls
 `ouroboros_job_wait` and fetches `ouroboros_job_result` when the job reaches a
--- a/docs/runtime-guides/pi.md
+++ b/docs/runtime-guides/pi.md
@@ -174,9 +174,13 @@ failure.
 The registered `/ooo` command also provides TAB argument completion through Pi's
 native `getArgumentCompletions` surface: `/ooo <TAB>` lists the dispatchable
 subcommands with one-line descriptions, and `ooo run <TAB>` lists Seed files
-from `~/.ouroboros/seeds/`. Completion is deterministic and offline — the
-subcommand list mirrors the packaged skills that declare `mcp_tool` frontmatter
-and a unit test keeps the two in sync.
+from `~/.ouroboros/seeds/` as absolute store paths, so the completed value
+executes regardless of the Pi session directory (relative seed paths resolve
+against the session cwd). Completion is deterministic and offline — the
+subcommand list mirrors the skills the dispatcher itself deems eligible, and a
+unit test derives that set through `resolve_skill_dispatch` so the two cannot
+drift. Seed names containing whitespace are skipped because the dispatcher
+tokenizes the dispatched command on whitespace.
 
 For `ooo auto`, the dispatcher owns the background job lifecycle. After
 `ouroboros_start_auto` returns a `job_id`, the dispatch process polls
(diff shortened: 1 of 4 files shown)


ruttybob:
Review addressed in 7ab80b6 (`fix(pi): keep /ooo Seed completions executable through dispatch`):

**Blocker 1 — Seed completion values.** Confirmed against `_resolve_seed_content`: relative `seed_path` resolves against the session cwd and a missing file degrades to inline YAML, so bare filenames from `~/.ouroboros/seeds/` were not executable from an unrelated session directory. Completions now insert the absolute store path (`path.join(seedsDir, name)`) as the value while matching and labeling by the bare filename.

**Follow-up — canonical drift guard.** The textual `^mcp_tool:` grep is replaced: the expected set is now derived through `resolve_skill_dispatch` per packaged skill directory, sharing the runtime's frontmatter loading, `normalize_mcp_frontmatter` validation, and template resolution. Body examples and invalid/missing `mcp_args` can no longer look dispatchable.

**Determinism.** Seed names are sorted before mapping. Names containing whitespace are skipped because the dispatcher tokenizes the dispatched command on whitespace — completion never suggests a value the router cannot carry.

**Behavioral coverage.** New `test_pi_bridge_run_completion_value_round_trips_to_global_seed` pins the completed value end to end: `ooo run <absolute seed>` resolves through the router with `seed_path` preserved, then loads the store file via `ExecuteSeedHandler._resolve_seed_content` from an unrelated session cwd; the pre-fix bare-name shape is pinned as degrading to inline YAML (the exact probe from the review). TypeScript-side behavioral execution remains out of scope for this Python suite — the completion logic was exercised directly under Node with the edge cases (empty/partial prefixes, missing directory, whitespace and non-Seed names).

Verification: `tests/unit/cli/test_setup.py -k pi` → 77 passed, 1 skipped; `tests/unit/router` → 282 passed; `tests/unit/cli` → 2032 passed with the same 22 pre-existing failures as a clean `upstream/main` in this environment (verified earlier by stash); `ruff format`/`ruff check` clean; rendered extension loads under Node `--experimental-strip-types`.

Q00:
Addressed Pi full-prefix replacement in f6a68a0c6: completion values retain the run subcommand and use POSIX single-argument quoting that round-trips special path characters through shlex. Pi/router regressions and current CI pass. Please re-review.

PR #2208, opened 2026-08-20
fix(seed): reject masked verify commands

### LINKED ISSUE #2155: verify_command can be generated as an always-failing / POSIX-only command, and no gate catches it before execution

Q00:
## Summary

- Reject generated `verify_command` values that mask a failing check with `|| true` or `|| :`, including fallbacks inside explicit `sh -c` / `bash -lc` wrappers.
- Keep quoted literals and existing valid single-line commands unchanged.
- Tell the Seed architect to use exact artifacts rather than inventing a proxy command when one shell line cannot verify the criterion.

## Verification

- `uv run pytest -q tests/unit/bigbang/test_seed_generator.py` — 359 passed
- Ruff and diagnostics pass
- Regression covers the reported `http.server` timeout followed by `2>&1 || true` and proves generation retries to an artifact contract.

Closes #2155

### DIFF
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -58,6 +58,8 @@ Multi-artifact example: `[{"description":"Build outputs exist","verify":"NONE","
 
 `verify` / `verify_command` semantics:
 - Use exactly one single-line shell command.
+- The command must verify the criterion directly and exit 0 on success without masking a failure. NEVER append unconditional fallbacks such as `|| true` or `|| :`; if a criterion cannot be checked by one command, use `verify: NONE` and provide exact artifacts instead.
+- Keep the command portable across the supported host shells. NEVER rely on POSIX-only fallback commands or shell-control syntax that changes the exit contract on Windows.
 - NEVER use heredoc or multiline shell syntax such as `<<`, `<<'PY'`, `cat <<EOF`, line-continuation scripts, or an unterminated command block. The AC contract format is one line, so multiline command bodies will be lost.
 - For Python snippets, use `python -c "..."` / `python3 -c "..."`; for longer checks, require a pytest-discoverable test artifact and use `python -m pytest -q`.
 
--- a/src/ouroboros/agents/seed-architect.md
+++ b/src/ouroboros/agents/seed-architect.md
@@ -58,8 +58,8 @@ Multi-artifact example: `[{"description":"Build outputs exist","verify":"NONE","
 
 `verify` / `verify_command` semantics:
 - Use exactly one single-line shell command.
-- The command must verify the criterion directly and exit 0 on success without masking a failure. NEVER append unconditional fallbacks such as `|| true` or `|| :`; if a criterion cannot be checked by one command, use `verify: NONE` and provide exact artifacts instead.
-- Keep the command portable across the supported host shells. NEVER rely on POSIX-only fallback commands or shell-control syntax that changes the exit contract on Windows.
+- The command must verify the criterion directly and exit 0 on success without masking a failure. NEVER append an unconditional success fallback such as `|| true`, `|| :`, or an equivalent known-success command; if a criterion cannot be checked by one command, use `verify: NONE` and provide exact artifacts instead.
+- Keep the command portable across supported Bash installations. Do not rely on host-specific shell features that the verifier cannot execute.
 - NEVER use heredoc or multiline shell syntax such as `<<`, `<<'PY'`, `cat <<EOF`, line-continuation scripts, or an unterminated command block. The AC contract format is one line, so multiline command bodies will be lost.
 - For Python snippets, use `python -c "..."` / `python3 -c "..."`; for longer checks, require a pytest-discoverable test artifact and use `python -m pytest -q`.
 
--- a/src/ouroboros/bigbang/seed_generator.py
+++ b/src/ouroboros/bigbang/seed_generator.py
@@ -1394,68 +1394,147 @@ def _unsupported_verify_command_reason(command: str) -> str | None:
 
 _SHELL_COMMANDS = frozenset({"bash", "dash", "ksh", "sh", "zsh"})
 _ALWAYS_SUCCESS_COMMANDS = frozenset({"echo", "printf", "true", ":"})
+_ALWAYS_FAIL_COMMANDS = frozenset({"false"})
+_SHELL_OPTIONS_WITH_VALUES = frozenset({"--init-file", "--rcfile", "-O"})
+_SHELL_CONTROL_TOKENS = frozenset({"&&", "||", "|", ";", ")", "}"})
 
 
-def _success_command_end(tokens: list[str], start: int) -> int | None:
-    """Return the end of a known-success command after ``||``."""
-    index = start
-    while index < len(tokens) and (
-        re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*=.*", tokens[index])
-        or tokens[index] in {"command", "exec", "builtin", "nohup"}
+def _mask_command_substitutions(command: str) -> str:
+    """Hide nested substitutions whose exit status does not decide the outer list."""
+    pieces: list[str] = []
+    index = 0
+    while index < len(command):
+        if command.startswith("$(", index) or command.startswith("${", index):
+            end = _posix_expansion_end(command, index)
+            if end is not None:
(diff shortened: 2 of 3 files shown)


Q00:
The only failed required check is the repository-wide Test Python 3.12 job. Its log shows unrelated `tests/unit/persistence/test_picker_indexes.py::test_picker_projection_write_and_disk_budgets[telemetry.unrelated-1.5-1.02]`; focused 359-test SeedGenerator suite passes. Rerunning the failed job now.

PR #2211, opened 2026-08-20
fix(security): drop transformed hidden retry output

### LINKED ISSUE #2020: Assertion redaction is verbatim-only: transformed encodings survive into the retry hint

Q00:
## Summary

- Canonicalize verifier output to a bounded fixed point across HTML/numeric entities, URL percent encoding, Unicode/JSON escapes and surrogate pairs, UTF-8 hex/octal byte escapes, NFKC/case folding, diagnostic prefixes, whitespace, and invisible/control characters.
- Drop retry fragments fail-closed when they carry a readable transformed `verify_command` or `output_assertion`; preserve exact redaction behavior for safe surrounding context.
- Reject stateful CSI, OSC payloads, DCS/C1/C0/DEL controls, invalid code points, malformed encoded bytes, and decoding-bound exhaustion before worker prompt transport.
- Cover pytest-style rewritten assertions and adversarial transformations described in #2020.

## Verification

- `uv run pytest -q tests/unit/orchestrator/test_parallel_executor_verify_by_default.py tests/unit/mcp/tools/test_seed_handoff.py` — 145 passed
- Current exact-head required GitHub checks — all pass

Closes #2020

### DIFF
--- a/docs/contributing/source-tour.md
+++ b/docs/contributing/source-tour.md
@@ -9,15 +9,15 @@ would want you to judge us by.
 
 `src/ouroboros/orchestrator/contract_redaction.py`
 
-- `contract_redaction.py:10`: `hidden_contract_variants` enumerates the five
-  encodings that get masked.
-- `contract_redaction.py:42`: the substitution is literal `str.replace`.
-  That means a reshaped copy of a hidden value is not caught. We track that
-  gap in a public issue rather than claiming otherwise.
-- `retry_hints.py:76`: `build_assertion_safe_retry_hint`: retry hints are
-  assembled deterministically, no model call.
-- `retry_hints.py:102`: the retry prompt carries a tail of command output;
-  redaction on it is the same literal substitution.
+- `hidden_contract_variants` enumerates raw, quoted, and escaped forms for exact
+  masking.
+- `contains_transformed_hidden_contract_value` applies bounded, fail-closed
+  HTML/Unicode/terminal normalization before retry output crosses into a worker
+  prompt. Stateful terminal controls are rejected; harmless SGR styling may be
+  normalized away.
+- `retry_hints.py` assembles retry hints deterministically, with no model call.
+  The command-output tail passes through the shared exact and transformed-value
+  boundary before inclusion.
 
 The claim's exact scope: when an acceptance criterion defines a verify
 command or an expected-output assertion, those values are omitted from the
--- a/src/ouroboros/orchestrator/contract_redaction.py
+++ b/src/ouroboros/orchestrator/contract_redaction.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
+import html
 import json
+import re
 import shlex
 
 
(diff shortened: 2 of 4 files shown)


Q00:
The only failed required check is the repository-wide Test Python 3.12 job. Its log shows unrelated `tests/unit/persistence/test_picker_indexes.py::test_picker_projection_write_and_disk_budgets[telemetry.unrelated-1.5-1.02]`; focused 91-test redaction/seed-handoff suite passes. Rerunning the failed job now.

Q00:
Addressed exact-head blocker in 97f77003a: unsupported DCS/APC/PM/SOS and C1 terminal-control strings now fail closed, with DCS and C1 regressions. Please re-review exact HEAD.

Q00:
Exact HEAD 5772a8015 now has completed required checks. Residual ESC/C1/C0 controls fail closed after CSI/OSC removal; focused terminal-control suite passes. Requesting final exact-head review.

Q00:
Final exact-head validation is green on current HEAD. Please issue a fresh review for this HEAD; prior CHANGES_REQUESTED artifacts target older commits.

Q00:
Updated branch HEAD 4ada63a2a has all required checks green and includes the full canonical Unicode/entity/control normalization fixes. Requesting exact-head approval.

Q00:
Current HEAD c5183e6cd closes the percent-encoding gap and all required checks pass. Focused retry-boundary suite: 145 passed. Requesting exact-head review.

Q00:
Superseded by #2218, which preserved the final branch state, received exact-head approval, passed all required checks, and merged as 6131cf5e20f7eb91afbfc6f31081f1a04f1b8637.

Issue #2202, opened 2026-08-19
[Bug] Cross-harness redispatch treats installed Claude CLI as SDK-capable and makes optional fallback authoritative


qdaszx:
## Summary

`execution.cross_harness_redispatch` defaults to `true`, but alternative-runtime availability is inferred from CLI presence rather than the transport/profile that the runtime factory can actually execute. In an isolated `ouroboros-ai[mcp]` environment, a Codex run sees an installed `claude` CLI, selects canonical backend `claude`, and then enters the SDK-backed adapter even though `claude-agent-sdk` is intentionally absent from the MCP 2 profile.

The missing optional SDK is then surfaced as the authoritative alternate-attempt failure, replacing the original Codex result and blocking the dependent AC graph.

The run-start handoff also does not expose that cross-harness recovery is enabled, which alternative backends are eligible, or a per-run Codex-only/allowlist choice. `quality_first` therefore appears to opt into a provider switch that the caller neither selected nor could see.

## Impact

- Merely having `claude` on `PATH` silently opts a Codex execution into a Claude alternate attempt.
- The supported MCP 2 package split makes `claude-agent-sdk` absence normal, but runtime selection treats Claude as healthy anyway.
- An optional recovery path can make the original failure worse: one alternate infrastructure failure becomes authoritative and blocks every dependent AC.
- Callers cannot tell from the `ouroboros_start_execute_seed` handoff whether `cross_harness_redispatch` is enabled or which runtime/transport could receive their code and prompts.
- Users who want `quality_first` within one provider need to modify persistent config or an environment variable; there is no explicit per-run policy/allowlist in the start API.

## Environment

- Ouroboros: `0.51.13` (latest release, isolated reproduction)
- Actual failed execution: MCP server from the `0.51.8` release family
- OS: macOS 26.6.1 arm64
- uv: 0.10.7
- Installation profile: isolated `ouroboros-ai[mcp]`
- Primary runtime/LLM backend: Codex (`codex-cli 0.147.0`)
- Execution policy: `efficiency_mode=quality_first`, `frugality_assurance=off`
- Claude Code: 2.1.142, present on `PATH`
- `claude-agent-sdk`: absent, as expected for the MCP 2 profile
- Effective `execution.cross_harness_redispatch`: default `true`

Related package-profile context: #1918 established that MCP 2 `[mcp]` and SDK-backed `[claude]` / `[claude-sdk]` must remain separate; `[mcp,claude-cli]` is the supported CLI combination.

[...]

PR #2214, opened 2026-08-20
fix(execution): keep one runtime per run by default

Q00:
## Problem

The real failure in #2202 is implicit execution authority: an optional recovery path can silently switch providers/transports, mutate the shared workspace, and replace the selected runtime result. Making that implicit switch safer would require a second runtime identity, policy, resume, and mutation-authority system.

## Decision

- Default `execution.cross_harness_redispatch` to `false`.
- Preserve explicit environment/config opt-in for operators who accept the existing same-workspace authority semantics.
- Keep same-runtime retries and model/effort escalation unchanged.

## Verification

- Config model/loader focused tests pass
- Ruff and format pass

Refs #2202


Q00:
Addressed legacy generated true configs in fd1e76178: only process-level OUROBOROS_CROSS_HARNESS_REDISPATCH now proves explicit opt-in; historical persisted true no longer authorizes implicit provider/workspace mutation. Please re-review exact HEAD.

Q00:
Exact HEAD fd1e76178 has completed required checks. Legacy generated true config is ignored; only explicit process-level opt-in authorizes cross-harness mutation. Requesting final exact-head review.

Q00:
Final exact-head validation is green on current HEAD. Please issue a fresh review for this HEAD; prior CHANGES_REQUESTED artifacts target older commits.

PR #2217, opened 2026-08-20
fix(mcp): negotiate host subagent dispatch capability

### LINKED ISSUE #2189: ouroboros_lateral_think dispatch payload: false no-parallel-capability claim (plain + base64-obfuscated variants)
## Summary

Calling `ouroboros_lateral_think` with `persona=all` produced a dispatch payload asserting that the calling runtime has no parallel-subagent capability and directing sequential processing — even though the calling session had already executed parallel subagents successfully earlier in the same conversation. Two different delivery forms of this directive were observed.

## Documented protocol vs. observed anomaly

The MCP server instructions shipped with this tool document `host_action: "spawn_subagents"` / `dispatch_mode: "host_driven"` as the fan-out protocol, with an explicit sequential fallback only for the case of "no parallel primitive." The two observed payloads diverge from that documented contract:

1. **Plain form** — a directly visible `meta.host_action: "spawn_subagents"` field was present, but accompanied by prose asserting the fallback's own precondition ("this runtime has no parallel-subagent capability") in a case where that precondition did not actually hold.
2. **Obfuscated form**, in a separate `persona=all` call — the same class of directive appeared with the field renamed to `host_action: "process_payloads_sequentially"` (not a documented value) and `dispatch_mode: "sequential"`, plus a `subagent_orchestration_instruction` field restating the same capability claim in prose. This entire structure was base64-encoded and embedded inside an HTML comment appended after 5 normal-looking persona sections, large enough to overflow standard tool-result rendering.

## Why this seemed worth reporting

The documented protocol fields are plain and visible. Renaming a dispatch-control field and hiding it, base64-encoded, inside an oversized HTML comment has no obvious legitimate purpose, and an agent that only skims tool output could plausibly act on the embedded instruction — abandoning a capability it actually has — without the divergence ever surfacing. Not sure whether this is a bug in how/when the fallback path gets triggered and serialized, or something else, but the gap from the documented contract seemed worth flagging.

## Environment

- Ouroboros version: 0.51.7 (verified live at time of filing)
- Tool: `ouroboros_lateral_think`, `persona=all`
- Observed once so far, in a single session

## Evidence

Raw field names/values observed (persona-prompt content itself omitted — unrelated to the dispatch mechanism):

- Plain form: `meta.host_action: "spawn_subagents"`
- Obfuscated form: `host_action: "process_payloads_sequentially"`, `dispatch_mode: "sequential"`, `subagent_orchestration_instruction: "<prose restating the no-parallel-capability claim>"`

Happy to share the full captured payload if useful for reproduction.

Q00:
## Problem

`ouroboros_lateral_think` and question advisory fan-out inferred the calling MCP host's subagent capability from the configured worker/REFLECT backend. A Claude Code host connected to a server with a Gemini worker could therefore receive a false `process_payloads_sequentially` instruction. Mapping an unknown host to Claude would create the opposite false claim.

## Solution

- Preserve and normalize request-scoped MCP `clientInfo` and client capabilities instead of discarding SDK context.
- Advertise and consume the `io.ouroboros/subagents` MCP capability extension.
- Separate MCP host dispatch authority from internal worker-runtime routing.
- Add a capability-neutral `host_decides` contract when the host declares no authoritative subagent capability:
  - prefer native parallel fan-out when available;
  - use the same payloads sequentially otherwise.
- Keep explicit parallel, explicit sequential, internal runtime, and confirmed OpenCode passive-bridge paths deterministic.
- Apply the contract consistently to lateral, interview advisory, PM advisory, structured metadata, visible directives, and the text-only Base64 compatibility block.

## Telemetry

Add an allowlisted `subagent_dispatch` PostHog event for:

- `phase=emitted`: normalized host family/capability, authority, delivery mode, preference/fallback, worker mismatch, bounded payload count, and re-entry availability.
- `phase=submitted`: completion status and bounded expected/received/undispatched counts.

Raw client names, client versions, fanout/session IDs, prompts, contexts, outputs, paths, and custom backend names are never sent. Custom values fold to closed `other`/`other_known`/`unknown` enums. `TELEMETRY.md` documents the exact property allowlist.

## Checks passed

- [x] Ruff lint on changed Python files
- [x] Ruff format on changed Python files
- [x] Mypy on 13 changed source files
- [x] Pytest: 476 related tests passed
- [x] MCP-extra adapter tests: 2 passed

## Test plan

- Unknown and identity-known/capability-undeclared MCP hosts emit `host_decides`.
- Declared parallel and sequential capabilities override mismatched worker backends.
- Server-composed Claude-host/Gemini-REFLECT mismatch no longer emits a false sequential claim.
- External hosts do not inherit an OpenCode worker's passive bridge; confirmed OpenCode plugin identity remains passive.
- SDK request context reaches the handler boundary.
- Emitted/submitted telemetry is privacy-safe and excludes identifiers/content.
- Shipped skill and Claude plugin mirror describe the same contract.

Closes #2189

### DIFF
--- a/.claude-plugin/skills/interview/SKILL.md
+++ b/.claude-plugin/skills/interview/SKILL.md
@@ -180,13 +180,14 @@ MCP (question generator) ←→ You (answerer + router) ←→ User (human judgm
    parent-session assist layer. The advisory exists to help the human answer;
    it must not hide, replace, or delay the question itself.
 
-   Run the advisory lanes through your runtime's native subagent mechanism when
-   one exists. For Claude Code this is the Task/Agent tool; for Codex, explicitly
-   start a native subagent workflow in natural language. Spawn one subagent per
-   lane, pass that lane's payload prompt, wait for all agents, then synthesize.
-   If the runtime has no parallel primitive, process payloads sequentially per
-   `dispatch_mode="sequential"` and the request's `sequential_fallback`
-   semantics. The standard lanes are:
+   Read the stamped dispatch contract before running the lanes. With
+   `dispatch_mode="host_driven"`, use the declared native parallel mechanism.
+   With `dispatch_mode="host_decides"`, use native parallel fan-out when the
+   current host exposes it and otherwise process the same payloads sequentially.
+   With `dispatch_mode="sequential"`, process payloads in order. For Claude Code
+   the parallel mechanism is Task/Agent; for Codex, explicitly start one native
+   subagent per payload in a single fan-out turn. Wait for every result, then
+   synthesize. The standard lanes are:
    - `code_context` — inspect repo-local facts and reuse
      `meta.code_investigation_request` when present.
    - `web_context` — browse/search only when current external facts genuinely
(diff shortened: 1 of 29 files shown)


jcfernandez-890825:
Thanks for the thorough follow-through on #2189 — the root-cause trace, the corrected host/worker-capability design, and the test coverage (including the telemetry privacy allowlist) all look solid. I'll pull this once it's merged/released and report back from the reporting side if anything surfaces.

Issue #2222, opened 2026-08-21
RFC: the PM interview batches its questions, and the evidence speaks the PM's language


cohemm:
> **Revision 3.** Revision 2 resolved the six contradictions a three-lens adversarial review (internal logic, code reality, prior-contract consistency) found in the first draft; the cost argument that review found factually false was made true by #2223 / #2224. Revision 3 amends after live lane-session testing: decision 6 (the plumbing does not travel) was added when a batch outgrew what a host accepts inline, and decision 4 was rewritten when recording confirmed findings proved to duplicate the published fan-out — findings are now evidence only, and the lane contract shed its confirmation fields. Revision 4 amends after three rounds of adversarial review found the same seam three times: the batch kept its pending questions in a second file, and every failure lived in the gap between that file and the transcript. Decisions 2 and 5 now make a turn atomic — nothing is persisted until it is whole, and a turn's answers arrive together — which removes the seam rather than guarding it. Revision 5 amends decision 6 after the same review lens was turned on the stub it introduced: everything a lane needed had been put behind one fetch, so a call that failed left it with no schema, no roster and no rules — nothing to do but say so. The line moves to what a child cannot work without. Revision 6 amends decision 6 again after reading the stub it produced: what a lane cannot work without turned out to be much smaller than what had been sent it, and the difference was things it could not act on — a specification where a naming would do, a list it had no way to choose within, and a reason beside every rule. Changes are integrated, not appended — this body is the decision.

## Background

The PM interview spends one turn per question. Each turn carries the full cost of a turn — the fan-out's latency, the evidence block, the confirmation gate, the refine gate — so a session that needs twelve decisions costs twelve of everything, even when several of those questions never depended on each other's answers.

A second problem surfaced in the same sessions, and it is a display problem wearing the first one's clothes. The evidence block a PM sees before answering is currently the *record* format: file paths, class names, flag values (`DAILY_LIMIT_GROUP_ID`, `credit=3`). The one thing in that block a PM actually needs — *the backend and the app implement this policy differently* — arrives wrapped in citations written for a different reader. The PM is not that reader. There is only one rendering because nobody decided there should be two.

Both problems are about the shape of a turn, so they are decided together.

### 1. The PM interview may put one to three questions on the table per turn

The batch size is chosen by the question generator per turn, not configured. A companion question exists only when it targets an unresolved clarity dimension *independent* of the others in the batch. A question whose right form depends on the previous answer is not a companion — it is the next turn's question, and it waits.

[...]

PR #2224, opened 2026-08-21
fix(mcp): offer recent findings to contracted lanes

### LINKED ISSUE #2223: Recent-findings offers are gated on answer-contract shape, cutting reuse off from the only lanes that repeat work
## The problem

`attach_question_advisory` offers recent findings only to lanes returned by `_open_contract_lanes` — lanes whose answer is prose rather than a closed contract (`question_advisory.py:650-653`, `:290-323`). PM declares an answer contract on **both** of its lanes, so `recent_findings_by_lane` receives an empty lane set and returns `{}` (`recent_findings.py`: "the honest answer for a tool whose every lane answers under a closed contract"). The interview's `data_context` lane is contracted too, and is cut off the same way.

The result: **the reuse mechanism is switched off for exactly the lanes it was built for.** RFC #2167 decided eligibility by what a lane *produces* — "the same lane is being asked to do the same work against a system that has barely moved" — and named `code_context` and `data_context` as the two producing lanes. The PM code lane re-reads the same repository roster on every question of every session; it is the poster child for that argument, and it is offered nothing.

## Background: everything but this gate already runs

This is not a missing feature — it is one dead link in a pipeline that is otherwise live end to end:

- **Write side, live:** every completed PM fan-out is published into the project's `ArtifactStore` as `kind=question_advisory` with per-lane `aggregated_outputs`. PM sessions write findings to the store on every question.
- **Read side, wired:** `create_fanout_wiring` hands `PMInterviewHandler` the very store the submit side publishes into (`findings_store=submit.artifact_store`, `fanout_composition.py:109`), and `_attach_advisory` passes it through (`pm_handler.py:415`). That one-function wiring exists precisely so the two sides cannot be wired apart.
- **The dead link:** the final step narrows the offer set by answer shape, and for PM that set is always empty. Written every turn, read never.

## How the gate got there, and why the reasoning does not hold

The gate's stated chain (`question_advisory.py:290-302`):

1. A fetch that fails silently must not read as "nothing to reuse" — the confusion RFC #2167 exists to prevent. So the offer text obliges the child: *"If you cannot reach that tool, say so in your finding."*
2. A closed contract has no field for that confession; writing an undeclared field rejects the whole answer, and these lanes are required, so the fan-out cannot complete.
3. Therefore a contracted lane is not offered findings at all.

Each step looks local and careful. The chain as a whole imports two premises that are false at this layer:

**Premise 1 — a failed fetch must be reported in-band.** The looked-vs-never-looked distinction matters where a *reader acts on the absence claim*: a PM who reads "no such policy" decides on false grounds. That is why the contract's `examined` structure makes the false "nothing" unspellable — and that guard is untouched by anything here. A failed *cache fetch* has no such reader. A contracted lane's answer carries policy claims about repositories, validated against the roster identically whether the lane had a head start or not. If the fetch fails, the lane investigates from scratch — producing the same answer, slower. **The failure mode this gate guards against is byte-identical to the behavior it mandates on every turn.** Under it you cannot lose; you can only be prevented from winning.

**Premise 2 — the answer is the only place the report can live.** `ouroboros_fetch_artifact` is an MCP call. The server observes every fetch: which fan-out, which lane, which contract id. Whether reuse is actually happening is an operator's diagnostic, and it is available out-of-band for free. The in-band confession duplicates a signal the server already holds — and the duplication is the entire reason the gate exists.

There is one real hazard in the chain: hand a contracted lane the *current* offer text and a literal-minded child may write the confession field it was told to write, fail validation, and take a required lane down with it. That is a reason to render a different offer text for contracted lanes — one paragraph — not a reason to withhold the offer.

## What this costs today

Every PM interview question dispatches a `code_context` child that reads the full roster from zero — worktree snapshots of multi-repo projects, per question, per session — while findings its own lane published hours earlier sit in the store, addressable, unread. `data_context` re-derives the same aggregates the same way. The mechanism whose whole justification is "the same lane does the same work against a system that has barely moved" excludes the lanes doing the most expensive repeated work in the codebase.

## The decision

**Reuse eligibility follows what a lane produces (RFC #2167's criterion). The shape of a lane's answer — contracted or prose — is not an eligibility criterion.**

Concretely:

- The offer set for a tool is the #2167-eligible lanes it declares (`code_context`, `data_context`), regardless of whether the lane carries an answer contract. `_open_contract_lanes` stops gating the offer.
- A contracted lane receives an offer text without the in-band failure-reporting obligation, and with an explicit instruction that its answer shape does not change: fetch what helps, investigate the rest, add no fields about reuse. The existing roster guard already covers the trust boundary — a claim about a repository outside this session's roster is rejected at submission whatever the child read.
- "Is reuse working" is observed server-side from fetch calls, where it always could be.

## What this does not change

- The `examined` structure and every answer contract: untouched. The looked-vs-never-looked guard for *evidence about the user's system* lives there and stays there.
- RFC #2167's per-lane scoping: a lane is still offered only what its own lane published. This issue widens *which lanes are offered their own*, not what any lane may read across lanes.
- The offer remains ids-only, advisory, a head start and not a substitute.

cohemm:
Closes #2223. Implements RFC #2222 (revision 6).

**Why.** A PM interview session paid a full turn — fan-out latency, evidence block, gates — per question, even when questions never depended on each other's answers; and the evidence a PM saw before answering was the record format (file paths, class names, flag values), written for a reader they are not. This PR makes a turn carry one to three independent questions and makes the evidence speak product language, with the machinery cost that made both affordable fixed on the way.

Two pieces the RFC's own cost accounting binds together: the recent-findings gate fix (#2223) is what makes RFC #2222's batched turns affordable, and the RFC cites this branch's fix as the machinery its cost argument stands on. All RFC #2222 behavior is scoped to the PM tool; the regular interview's only change is the deliberate, shared #2223 fix (its contracted `data_context` lane now receives recent-findings offers).

## Part 1 — Recent findings reach contracted lanes (#2223) — `bdf0fa50`

Every piece of PM findings reuse already existed and ran — write side published each completed fan-out into the project `ArtifactStore`, and `create_fanout_wiring` handed both handlers that same store — except the last step: `recent_findings_by_lane(store, lanes=_open_contract_lanes(catalog))` offered findings only to lanes *without* a closed answer contract. Both PM lanes are contracted, so the offer set was always `∅` — written every turn, read never. The interview's contracted `data_context` was cut off the same way.

`_declared_lanes` replaces the gate: eligibility follows what a lane *produces* (RFC #2167's criterion), never the shape of its answer. The one real hazard — a contracted lane told to confess a failed fetch in-band would fail its own contract — is handled in the offer *text*: prose lanes keep the in-band duty; contracted lanes get "investigate as usual; your answer keeps its contracted shape — do not add fields about reuse."

**Verified live**: a later lane was offered prior rounds' contract ids, fetched its own lane's scoped body, and narrowed its investigation to the one fact not already recorded; cross-topic offers were triaged by the child ("covers booking-limit policy, not the trial-lesson funnel") exactly as "a head start, not a substitute" intends.

### Batched turns (decision 1, `pm_interview.py`)

`plan_next_turns` rides the existing atomic planner call: companions arrive in the same JSON payload (requested through the same `extra_response_contract` hook PM already uses) and pass through the same `_apply_classification` path as the primary. Enforced in code, not trusted to the generator: closure mode (ambiguity ≤ 0.25) is single-question; duplicate normalized question identities are dropped (fan-out isolation keys on them); three is a ceiling. Zero extra LLM calls.

### A turn is atomic (decision 2, revision 4)

Three review rounds found the same seam at three addresses — an interleaved write, an interrupted one, and an answer matched to the wrong round — because a batch kept its pending questions in `pm_meta` beside the transcript, keyed by question text. Revision 4 removes the seam instead of guarding it: **a turn persists nothing when it asks, and its answers arrive together**, so a round is written only when it is whole and there is no second place a turn is remembered. A call carrying no answers is a host that lost its turn; the next turn is planned from the transcript rather than restored.

What that deleted: the question-only round, `pending_batch` and its per-member routing, the text match from answer to member, the partial-batch response, the replay reconciliation, and the reconnect re-display. What it added: `answers: [{question, answer}]` on the tool, and one function that records them. `pm_handler.py` went from the 2000-line cap to 1809.

### The interview core is untouched (decision 2)

`git diff origin/main...HEAD` over the enumerated boundary (`interview.py`, `turn_planner.py`, `ambiguity.py`, `answer_provenance.py`, `inner_guidance.py`, `interview_schemas.py`, `authoring_handlers.py`) is **empty**. The invariant that makes it possible: the engine's `record_answer` fills the *trailing* unanswered round, so a PM turn leaves none — **it persists nothing when it asks** (RFC revision 4). No question-only round, no pending-member list. A turn's answers arrive together and each round is written whole, through the public API.

### Every question keeps its evidence; the plumbing does not travel (decisions 3 & 6, `pm_handler.py`, `pm_batch.py`)

One advisory envelope per question (`meta.question_advisories`), each with its own payloads and fan-out id. The turn's answers come back together as `answers: [{question, answer}]` — each answer names its own question, so correlation is structural rather than a lookup against remembered state, and completion and the next turn follow the one write that records them.

What made batches receivable at all: a turn's response used to carry each lane's full brief **twice** (meta + dispatch text) plus a per-envelope copy of the tool capability metadata — ~120k chars for one question; a batch outgrew what a host accepts inline and hosts improvised file surgery. The briefs are published to the artifact store (`advisory_prompt_bundle`, keyed by fan-out id, a kind the reuse query never offers as a finding) and each payload prompt is a stub. Capability metadata rides as a name. **Fail-open on every edge**: no store, no fan-out id, or a failed publish leaves the full prompts inline.

**What the stub carries** (revision 6). The first cut put everything behind the fetch, so a lane that could not reach the store had no shape to answer in and nowhere it was allowed to look — one call was the single point of failure for a whole lane. The line is now drawn at what a child cannot derive or ask for:

- **The answer's shape, named rather than specified.** Field names and nesting, the empty state and its literals read out of the contract's own schema, the enum vocabularies, and the two conventions a child gets wrong by default (a path is repository-relative; a plain statement is in the question's language). Not the JSON Schema — its bounds restate what the validator already refuses. Not a worked example either: an example is a claim already written in the answer's shape, and a lane short on evidence would hold a draft needing only its identifiers changed.
- **Where it may look** — the roster for a lane whose answer cites repositories, the host's data tools for one that does not. The branch reads each lane's own answer shape, never its name.
- **The order of work** — no-op if the question does not need this lane; otherwise read what this lane already found and *carry it as it stands*; investigate **only if** that turned up nothing that bears on the question.

**Which findings exist is asked for, not sent.** Twenty contract ids rode in every prompt so a child could choose among them, and an entry is an id and a time — nothing to choose by. `ouroboros_fetch_artifact` now accepts a `lane_id` alone and lists that lane's recent findings; reading one back by `contract_id` is unchanged. The window, the eligible kind and the cap stay inside the query, so asking reaches no further than being offered did. **Additive**: every existing call behaves exactly as before, and the regular interview keeps rendering its own offers.

A lane prompt is ~2k characters, against ~14.7k inline on `main` for the same question, and no longer grows with how much the project has found. `UNDISPATCHED` remains for a child that cannot work at all — a failed fetch is no longer one of those cases.

### Findings are evidence, never answers (decision 4, `pm_schemas.py`, `skills/pm/SKILL.md`)

The adopted-fact flow is retired: recording a confirmed finding as a `[from-code]` observation duplicated the published fan-out (already addressable and re-offered via #2223) and cost the user a confirmation turn. The v2 code-lane contract sheds `answer_prefix` / `requires_user_confirmation` / `user_confirmation_prompt` — with nothing recorded there is nothing to gate, and the closed shape rejects an answer that still carries them. Each claim gains a required lane-authored `plain_statement` (product language, the question's language, no paths or identifiers): the screen renders plain statements only; citations stay in the published fan-out. Interview rounds carry only the user's answers — their own words, or the refine-gate structure they confirmed.

### pm_handler split

Batch/report/externalization logic lives in a new PM-owned module `pm_batch.py`; `pm_handler.py` stays at the module-size cap (≤ 2000 lines) without grandfathering.

## What this does not change

- The regular interview: single-question turns, inline payload prompts, its own contracts and skill — untouched by everything in Part 2 (`git diff` over `authoring_handlers.py`, `interview_schemas.py`, `interview.py`, `turn_planner.py` is empty). Its one behavior change is Part 1's, by that issue's explicit scope: the contracted `data_context` lane is now offered its own recent findings, with the contracted offer text. Its identical wire weight is named in RFC #2222 decision 6 as that tool's own future decision.
- Answer validation, roster guards, per-lane findings scoping, ids-only transport, recency window: intact and pinned by tests — the listing moved behind a request, its bounds did not move with it.
- Single-question PM turns: same flow as before, now with stub payloads.
- `ouroboros_fetch_artifact` called with a `contract_id`, with or without a `lane_id`: byte-identical behaviour.

## Tests

- `test_recent_findings.py`: gate inversion pinned — all four eligible lane instances offered; in-band duty only where the answer has a place for it.
- `test_pm_interview.py`: batch planning — companions classified and reframe-tracked; closure mode single; duplicates/malformed dropped without trace.
- `test_pm_handler_batch.py` (new): batch issue (per-question envelopes, **nothing persisted by asking**), a turn recorded whole, an answer without its question refused, a question whose text recurs recorded as its own round, a reconnect planning a fresh turn, two turns answered concurrently, plugin mode never entering the batch producer, briefs-by-reference (stub on the wire, full brief one scoped fetch away), no-store inline fallback.
- `test_pm_question_advisory.py`: v2 contract — no forwarding fields in any state, a v1 relic answer is rejected, prompts carry `plain_statement` and no confirmation instruction.
- Stub content: every field a contract *requires* is named in what the child is shown (so the text cannot go stale as the contract gains one), step 3 follows the lane's answer shape rather than its name, and the wire carries no JSON block for a lane to copy.
- Tool listing: a `lane_id` alone returns that lane's own recent findings as ids and times and no bodies; neither argument is refused with both named.
- Changed-boundary suites (`tests/unit/mcp/tools`, `tests/unit/bigbang`): **2,509 passed**. Ruff and mypy clean; skill mirror and module-size gates green.

Verified live in lane sessions across two product topics: stub wire on start and answer paths, children fetching their own lane's prior findings by request, cross-topic reuse triage, a 3-question batch received inline, evidence blocks in product language with no confirmation turn. Publish → offer → scoped fetch was also exercised end to end against a real store, including the refusal of a lane the artifact does not carry.

### DIFF
--- a/.claude-plugin/skills/pm/SKILL.md
+++ b/.claude-plugin/skills/pm/SKILL.md
@@ -98,6 +98,31 @@ from memory, so it is the last one to skip evidence on.
 Apply this to **every** MCP response that carries a question, including the one
 Step 2 returned and any question re-shown on resume.
 
+**Batched turns (RFC #2222).** A response may carry one to three questions at
+once: `meta.question_batch` lists them and `meta.question_advisories` carries
+one advisory envelope per question, each with its own
+`question_advisory_subagents` and `question_advisory_fanout_id`. Treat every
+batch member exactly as a single question is treated below, with these batch
+mechanics:
+
+- **Dispatch all envelopes' payloads in one wave** — one subagent per payload
+  across all questions, in a single parallel batch. Never leave a question's
+  lanes undispatched: every question shown keeps its evidence.
+- **Submit results per envelope** — each question's lanes correlate by its own
+  envelope's `question_advisory_fanout_id` and
+  `question_advisory_result_correlation_key`; one
+  `ouroboros_submit_fanout_results` call per envelope.
+- **Relay answers one at a time**, each with `last_question` set to that
+  member's exact question text. Members may be answered in any order; the
+  server re-lists what is still pending after each. An unanswered member stays
+  pending — never auto-answer, auto-defer, or decide-later it on the user's
+  behalf.
+- A response that lists pending members without `question_advisories` is a
+  re-display: its lanes already ran, do not dispatch them again.
+- Skip sentinels are per member: `answer="[decide_later]"` /
+  `answer="[deferred]"` are honoured only for the member whose classification
+  allows it, always with that member's `last_question`.
+
 **A. Show alerts** (if present in `meta`):
 - `meta.deferred_this_round` → print `[DEV → deferred] "question"`
 - `meta.decide_later_this_round` → print `[DEV → decide-later] "question"`
@@ -147,11 +172,17 @@ having two lanes instead of six:
 
 Write the line in the language the user is speaking.
 
-**Do not go to step A3 or B while the lanes are still running.** Step B is where
+**Do not go to step B while the lanes are still running.** Step B is where
 you ask the user, and asking before the evidence arrives is the exact failure
 this mechanism exists to prevent: the PM decides without the two things they
 could not have looked up themselves. Waiting is for lanes still in flight: one
 that came back empty, broke its contract, or could not be spawned has returned.
+
+**Stub payloads.** A payload's `prompt` may be a short stub telling its child
+to fetch the full brief from the store (`ouroboros_fetch_artifact`) — pass it
+unchanged exactly like any payload. A child that replies exactly
+`UNDISPATCHED` could not receive its brief: submit that lane as
+`{ "key": <lane id>, "undispatched": true }`.
 
 **Submitting results back.** Correlate by
 `meta.question_advisory_result_correlation_key` (`context.lane_id`) and call
@@ -176,11 +207,10 @@ and a reply can be accepted while one of them never ran. Where the block would
 have carried that lane, write that it did not run.
 
 There are two lanes and both are required: `code_context` and `data_context`.
-A `code_context` lane that carries a policy returns `answer_prefix:
-"[from-code]"` and a `user_confirmation_prompt` — that is a step, described in
-A3 below, not the answer. `data_context` has no prefix at all: measurements are
-shown beside the question and the answer is the user's own words. Never skip
-asking the user because a lane answered clearly.
+Both are **evidence-only** (RFC #2222): what a lane finds is shown beside the
+question and sent nowhere — the published fan-out is already its record, and
+the interview records only what the user writes. Never skip asking the user
+because a lane answered clearly, and never send a lane's finding as an answer.
 
 **Synthesize into the evidence block.** This is what
 `synthesis_contract.output_shape = "evidence_beside_question"` means, and it is a
@@ -190,17 +220,24 @@ above the question, then ask the question unchanged:
 ```
 Evidence (examined: billing-api, storefront)
 
-What the code does today
-  · [billing-api] src/billing/lapse.py  — access continues to period end
-  · [storefront]  src/checkout.ts       — access is revoked immediately
+What the system does today
+  · [billing-api] access continues to the end of the paid period
+  · [storefront]  access is revoked immediately
     ! These two repositories implement this differently.
 
 Measured — active subscriptions by plan, last 90 days
   · standard 12,480 / premium 3,120
 ```
 
-Write the block in the language the user is speaking. The labels above are
-placeholders for its shape, not text to copy.
+**The screen speaks product language; the store keeps the citations
+(RFC #2222 decision 4).** Each code claim is rendered from its lane-authored
+`plain_statement` — never from `policy_claim`, and never paraphrased by you.
+No file paths, no class names, no flag values on screen. The `path` +
+`policy_claim` citations stay in the published fan-out, fetchable by its
+contract id — they are displayed nowhere and recorded nowhere else. The lane
+writes `plain_statement` in the question's language, so no translation is
+yours to do. The labels above are placeholders for the block's shape, not
+text to copy.
 
 **What this block is not.** It carries no answer options, no recommendation, no
 ranking, and no "therefore …" sentence. The moment it proposes an answer it has
@@ -238,55 +275,15 @@ Rules for building it:
   something the record does not contain. Offer to register the repository
   instead, so the next question can be answered against it.
 
-**A3. Record a confirmed finding — its own turn, before you prompt for an answer.**
+**A3. Findings are evidence, never answers (RFC #2222).**
 
-Only when some `code_context` entry came back carrying a `policy_claims` item.
-That answer also carries `answer_prefix: "[from-code]"` and a
-`user_confirmation_prompt`, and there is no prefix that skips this step:
-`[from-code][auto-confirmed]` is not a value the contract can hold, so a lane
-cannot declare itself pre-confirmed.
-
-1. Show the evidence block (A2) and ask the lane's `user_confirmation_prompt`
-   through `AskUserQuestion`. Ask it as it is: the user is confirming that this
-   is what the code does, not deciding what it should do.
-2. If they say it is wrong, or they would rather just answer, **send nothing**
-   and go to B. A question answered with no finding recorded is an accurate
-   account of how that decision was made, not a degraded one.
-3. If they confirm, send it as the answer with its prefix, composed from the
-   `examined` entries so every claim keeps the repository it was read in:
-
-```
-Tool: ouroboros_pm_interview
-Arguments:
-  session_id: <meta.session_id>
-  last_question: <meta.question>
-  answer: |
-    [from-code] billing-api src/billing/lapse.py: access continues to period end.
-    [from-code] storefront src/checkout.ts: access is revoked immediately.
-```
-
-Send `meta.question`, never the response text — the text carries the fan-out
-directive. Omitting it files the finding under a placeholder question, and
-since the answer slot is withheld by design, the finding is then lost from
-both slots.
-
-4. The server records it as an **adopted fact**: the round is marked
-   `observation`, requirement extraction reads a withheld-note in its place, and
-   it does not count toward the decisions that complete the interview. The
-   response carries the **next** question, generated with the finding in view —
-   that is the question the user answers in their own words, from step 3-A.
-
-**Why it takes the round rather than riding beside the answer.** It was built
-the other way first, with a second parameter for findings. Two entrances meant
-two sets of rules for one payload, and they stopped agreeing: the same class of
-silent loss reappeared at a new address for six review rounds. One entrance is
-what closed it, and it is what `ouroboros_interview` always did.
-
-**What this does not license.** The finding is never sent unconfirmed to save a
-turn. What holds if you do is downstream and weaker than the user's eyes: the
-`[from-code]` prefix keeps it out of requirement extraction and out of the
-completion count, so an unconfirmed forward costs the user a question turn and
-puts nothing false in the PRD. That is a floor, not a permission.
+There is no recording step. A finding's durable record is the published
+fan-out itself — addressable by contract id, re-offered to later lanes through
+recent findings — so sending it again as an answer would duplicate the store
+and spend a question turn on it. Do not send `[from-code]` answers, do not ask
+the user to "confirm" a finding as a separate turn, and do not paste findings
+into any answer payload. The evidence block is the finding's whole appearance;
+the user answers the question in their own words, with the evidence in view.
 
 **B. Show content + get user input** (once A2's lanes have returned):
 
@@ -309,6 +306,11 @@ Then check: does `meta.ask_user_question` exist?
   Do NOT modify it. Do NOT add options. Do NOT rephrase the question.
 
 - **NO** → This is an interview question. Use `AskUserQuestion` with `meta.question`.
+  - **Batched turn**: put every pending member into ONE `AskUserQuestion` call —
+    one entry per member (the tool takes up to 4). Each member keeps its own
+    options, built by the same rules below from its own entry in
+    `meta.question_batch`; a skip option follows that member's own
+    `classification`, never another member's.
   - If `meta.skip_eligible == true`: add a skip option based on `meta.classification`:
     - `classification == "decide_later"` → add option `{"label": "Decide later", "description": "Skip — will be recorded as an open item in the PRD"}`
     - `classification == "deferred"` → add option `{"label": "Defer to dev", "description": "Skip — this technical decision will be deferred to the development phase"}`
@@ -321,6 +323,11 @@ Then check: does `meta.ask_user_question` exist?
 If the user chose "Decide later" → send `answer="[decide_later]"`.
 If the user chose "Defer to dev" → send `answer="[deferred]"`.
 Otherwise → send the user's answer through the Refine gate below.
+
+On a batched turn: one call per answered member, each with that member's exact
+question text as `last_question`, each through its own Refine gate. The server
+answers each call with what is still pending; only the call that resolves the
+last member returns the next turn's question(s).
 
 **Refine gate — structure it, mark whose it is, then have the user confirm.**
 
@@ -353,8 +360,9 @@ nothing they said is the failure this labelling exists to make visible.
 
 **No codebase-context section, and no lane findings.** The regular interview
 adds one, because there the main session inspects code itself. Here it does not.
-A lane's finding is recorded by step A3, on its own turn, as an adopted fact.
-Putting it in this payload would record it as part of the user's decision.
+A lane's finding lives in the published fan-out and on the screen beside the
+question — putting it in this payload would record it as part of the user's
+decision, which it is not (A3).
 
 **Then confirm — this is the gate, and it is what licenses the structuring
 above.** One `AskUserQuestion` before sending:
@@ -387,17 +395,15 @@ Tool: ouroboros_pm_interview
 Arguments:
   session_id: <meta.session_id>
   last_question: <meta.question>
-  <meta.response_param>: <the refined answer, or "[decide_later]" / "[deferred]">
+  answer: <the refined answer, or "[decide_later]" / "[deferred]">
 ```
 
 Ignored while the server holds the question unanswered; required otherwise —
 plugin mode never persists the child's questions, and an answer with no pending
 question is refused without it.
 
-There is no second parameter carrying findings. The tool has exactly the fields
-the regular interview has, and a finding travels the same way any adopted fact
-travels there — see **A3. Record a confirmed finding** above, which is a separate
-turn taken *before* the user answers, not something appended to this call.
+There is no parameter carrying findings, second or otherwise: findings never
+travel as answers (A3). Every call on this parameter is the user's own words.
 
 **D. Check completion:**
 
--- a/.claude-plugin/skills/pm/SKILL.md
+++ b/.claude-plugin/skills/pm/SKILL.md
@@ -180,11 +180,13 @@ this mechanism exists to prevent: the PM decides without the two things they
 could not have looked up themselves. Waiting is for lanes still in flight: one
 that came back empty, broke its contract, or could not be spawned has returned.
 
-**Stub payloads.** A payload's `prompt` may be a short stub telling its child
-to fetch the full brief from the store (`ouroboros_fetch_artifact`) — pass it
-unchanged exactly like any payload. A child that replies exactly
-`UNDISPATCHED` could not receive its brief: submit that lane as
-`{ "key": <lane id>, "undispatched": true }`.
+**Stub payloads.** A payload's `prompt` may be a compact stub: it carries the
+lane's answer schema, where it may look, and the findings it may reuse, and
+points at `ouroboros_fetch_artifact` for the prose that explains them. Pass it
+unchanged exactly like any payload — the child fetches for itself, and a fetch
+it cannot make does not stop it. A child that replies exactly `UNDISPATCHED`
+could not work at all: submit that lane as
+`{ "key": <lane id>, "undispatched": true }` rather than as an empty finding.
 
 **Submitting results back.** Correlate by
 `meta.question_advisory_result_correlation_key` (`context.lane_id`) and call
--- a/.claude-plugin/skills/pm/SKILL.md
+++ b/.claude-plugin/skills/pm/SKILL.md
@@ -114,9 +114,11 @@ batch mechanics:
   `ouroboros_submit_fanout_results` call per envelope.
 - **Relay the turn's answers together** in one call:
   `answers: [{question, answer}, ...]`, one entry per question the turn asked,
-  each naming its own exact question text. A turn is recorded whole — collect
-  every answer first, and never auto-answer, auto-defer, or decide-later one on
-  the user's behalf to complete the set.
+  each naming its own exact question text. One call records the turn, so
+  collect every answer first — and never auto-answer, auto-defer, or
+  decide-later one on the user's behalf to complete the set. The server does
+  not check that you sent them all: whatever you leave out is **abandoned**,
+  not remembered.
 - The server keeps nothing between calls. A call that arrives without the
   turn's answers plans a **new** turn from the transcript rather than restoring
   the old one, so a turn you abandon is a turn the user will be asked again.
--- a/.claude-plugin/skills/pm/SKILL.md
+++ b/.claude-plugin/skills/pm/SKILL.md
@@ -96,14 +96,14 @@ from memory, so it is the last one to skip evidence on.
 ### Step 3: Loop
 
 Apply this to **every** MCP response that carries a question, including the one
-Step 2 returned and any question re-shown on resume.
+Step 2 returned and any question a resume plans anew.
 
 **Batched turns (RFC #2222).** A response may carry one to three questions at
 once: `meta.question_batch` lists them and `meta.question_advisories` carries
 one advisory envelope per question, each with its own
 `question_advisory_subagents` and `question_advisory_fanout_id`. Treat every
-batch member exactly as a single question is treated below, with these batch
-mechanics:
+question of the turn exactly as a single question is treated below, with these
+batch mechanics:
 
 - **Dispatch all envelopes' payloads in one wave** — one subagent per payload
   across all questions, in a single parallel batch. Never leave a question's
@@ -112,16 +112,16 @@ mechanics:
   envelope's `question_advisory_fanout_id` and
   `question_advisory_result_correlation_key`; one
   `ouroboros_submit_fanout_results` call per envelope.
-- **Relay answers one at a time**, each with `last_question` set to that
-  member's exact question text. Members may be answered in any order; the
-  server re-lists what is still pending after each. An unanswered member stays
-  pending — never auto-answer, auto-defer, or decide-later it on the user's
-  behalf.
-- A response that lists pending members without `question_advisories` is a
-  re-display: its lanes already ran, do not dispatch them again.
-- Skip sentinels are per member: `answer="[decide_later]"` /
-  `answer="[deferred]"` are honoured only for the member whose classification
-  allows it, always with that member's `last_question`.
+- **Relay the turn's answers together** in one call:
+  `answers: [{question, answer}, ...]`, one entry per question the turn asked,
+  each naming its own exact question text. A turn is recorded whole — collect
+  every answer first, and never auto-answer, auto-defer, or decide-later one on
+  the user's behalf to complete the set.
+- The server keeps nothing between calls. A call that arrives without the
+  turn's answers plans a **new** turn from the transcript rather than restoring
+  the old one, so a turn you abandon is a turn the user will be asked again.
+- Skip sentinels are per question: give that question the answer
+  `"[decide_later]"` or `"[deferred]"` in its own entry.
 
 **A. Show alerts** (if present in `meta`):
 - `meta.deferred_this_round` → print `[DEV → deferred] "question"`
@@ -306,11 +306,11 @@ Then check: does `meta.ask_user_question` exist?
   Do NOT modify it. Do NOT add options. Do NOT rephrase the question.
 
 - **NO** → This is an interview question. Use `AskUserQuestion` with `meta.question`.
-  - **Batched turn**: put every pending member into ONE `AskUserQuestion` call —
-    one entry per member (the tool takes up to 4). Each member keeps its own
+  - **Batched turn**: put every question of the turn into ONE `AskUserQuestion`
+    call — one entry per question (the tool takes up to 4). Each keeps its own
     options, built by the same rules below from its own entry in
-    `meta.question_batch`; a skip option follows that member's own
-    `classification`, never another member's.
+    `meta.question_batch`; a skip option follows that question's own
+    `classification`, never another's.
   - If `meta.skip_eligible == true`: add a skip option based on `meta.classification`:
     - `classification == "decide_later"` → add option `{"label": "Decide later", "description": "Skip — will be recorded as an open item in the PRD"}`
     - `classification == "deferred"` → add option `{"label": "Defer to dev", "description": "Skip — this technical decision will be deferred to the development phase"}`
@@ -324,10 +324,10 @@ If the user chose "Decide later" → send `answer="[decide_later]"`.
 If the user chose "Defer to dev" → send `answer="[deferred]"`.
 Otherwise → send the user's answer through the Refine gate below.
 
-On a batched turn: one call per answered member, each with that member's exact
-question text as `last_question`, each through its own Refine gate. The server
-answers each call with what is still pending; only the call that resolves the
-last member returns the next turn's question(s).
+On a batched turn: run each answer through its own Refine gate, then send them
+in one call as `answers: [{question, answer}, ...]` — the question text exactly
+as the turn asked it. The call that carries the turn returns the next turn's
+question(s).
 
 **Refine gate — structure it, mark whose it is, then have the user confirm.**
 
--- a/skills/pm/SKILL.md
+++ b/skills/pm/SKILL.md
@@ -98,6 +98,31 @@ from memory, so it is the last one to skip evidence on.
 Apply this to **every** MCP response that carries a question, including the one
 Step 2 returned and any question re-shown on resume.
 
+**Batched turns (RFC #2222).** A response may carry one to three questions at
+once: `meta.question_batch` lists them and `meta.question_advisories` carries
+one advisory envelope per question, each with its own
+`question_advisory_subagents` and `question_advisory_fanout_id`. Treat every
+batch member exactly as a single question is treated below, with these batch
+mechanics:
+
+- **Dispatch all envelopes' payloads in one wave** — one subagent per payload
+  across all questions, in a single parallel batch. Never leave a question's
+  lanes undispatched: every question shown keeps its evidence.
+- **Submit results per envelope** — each question's lanes correlate by its own
+  envelope's `question_advisory_fanout_id` and
+  `question_advisory_result_correlation_key`; one
+  `ouroboros_submit_fanout_results` call per envelope.
+- **Relay answers one at a time**, each with `last_question` set to that
+  member's exact question text. Members may be answered in any order; the
+  server re-lists what is still pending after each. An unanswered member stays
+  pending — never auto-answer, auto-defer, or decide-later it on the user's
+  behalf.
+- A response that lists pending members without `question_advisories` is a
+  re-display: its lanes already ran, do not dispatch them again.
+- Skip sentinels are per member: `answer="[decide_later]"` /
+  `answer="[deferred]"` are honoured only for the member whose classification
+  allows it, always with that member's `last_question`.
+
 **A. Show alerts** (if present in `meta`):
 - `meta.deferred_this_round` → print `[DEV → deferred] "question"`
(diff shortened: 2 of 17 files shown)


cohemm:
Addressed review round 2 in `dd92ad63` — 2 fixed, 1 position.

**Blocker 1 — `pm_interview.py:737`, companion routing coerced with `bool(...)` — fixed.**
Reproduced exactly as described: `"decide_later": "false"` classified the companion as `decide_later` and made it skip-eligible, so `[decide_later]` would discard a question the PM must answer. The fix is not a stricter check beside the old one — the hand-built `ClassificationResult` is gone, and companions now go through the primary's own `_parse_response`, which raises on a non-boolean `decide_later`/`defer_to_dev`. A companion that fails validation is dropped, which is what the docstring already promised. There is now one validator for both, so the two paths cannot diverge again; the same move also fixes a quieter asymmetry, a companion carrying `category: "decide_later"` without the flag now gets its skip route the way the primary does. Pinned in both directions: `test_plan_next_turns_rejects_wrong_typed_companion_routing` drops the wrong-typed companion while a properly typed decide-later companion still ships with `ClassifierOutputType.DECIDE_LATER`, and it fails without the fix.

**Follow-up `pm_handler.py:1429`, lost update on concurrent batch answers — fixed.**
Reproduced. `interview_answer_lock` (in `pm_batch.py`) serializes one interview's answers, keyed by interview, on the handler the server builds once at startup — the same idiom as `execution_handlers._idempotency_locks`. It is taken at the resume call site rather than around the batch branch, because the read-modify-write over state and PM meta is shared by every answer path: a lost update reads the same whether the two answers belong to one batched turn or to two single-question turns. Pinned by `test_two_members_answered_concurrently_both_survive`, which schedules the interleave rather than hoping for it (`save_state` is made to yield) and loses a round without the lock.

Two boundaries, so they are on record rather than found later: the lock is one server process, which is the scope of the state directory that process owns; and the plugin dispatch branch records answers on its own path, unchanged from `origin/main` in this PR (its only diff there is a comment), so it keeps whatever concurrency behaviour it had before batching. Ordering is deliberate too — state is written before meta, so a crash between them re-asks a question rather than losing the answer to it.

**Follow-up `pm_handler.py:787`, plugin runtime has no batch contract — position.**
The premise is right and the conclusion is what we disagree on: the plugin runtime does not implement batching because it never enters it, and that is enforced structurally rather than by a switch.

- `plan_next_turns` — the only producer of a batch — has exactly one call site, in the in-process branch. `pending_batch` has exactly one writer, in the same branch.
- The runtime fork is `should_dispatch_via_plugin(agent_runtime_backend, opencode_mode)`, read from handler construction. A process is one runtime or the other for its whole life, so a session cannot batch in one call and be answered through the other.
- `build_pm_interview_subagent` carries no batch protocol, so the plugin child is not asked for more than one question.
- This PR's diff over the plugin answer path is a comment. `git diff origin/main...HEAD` there is behaviourally empty: the hazard as stated ("if a plugin child emits a three-question turn") is a property `main` has today, handled today the same way — `record_answer` fills the trailing round — and this PR does not change it.

We are not taking the second option you offer either. "Explicitly reject/disable the batched contract for that runtime" means adding a switch whose only job is to turn off something that was never on, plus documentation and tests for the switch — a mechanism weaker than the guarantee already present, since a single call site cannot be misconfigured. What we did instead is record the scope where scope belongs: RFC #2222 now carries the plugin runtime under "Deliberately not decided here", stating that batching rides the in-process planner and that giving the plugin child a batch protocol is that runtime's own decision with its own child-session round trip to pay for.

If you can show a probe where plugin mode reaches `plan_next_turns`, or reads or writes `pending_batch`, that is a defect and we will fix it in this PR.

**On the design gate.** RFC #2222 puts "pending-state layout inside the PM layer" under "Deliberately not decided here — implementation, judged by whether it preserves the decisions above". Serializing answers is that implementation, not a new decision, so the RFC needed no revision for it; the runtime scope did, and got one.

**Verification.** Full unit suite: 20,675 passed / 89 skipped. `tests/integration` 262 passed / 4 skipped / 1 xfailed; `tests/conformance` 43 passed; `tests/canonical` 34 passed / 2 skipped; `tests/e2e` 77 passed. `ruff check`, `ruff format --check`, `mypy src`, module-size (`pm_handler.py` at the 2000-line cap, not grandfathered) and the auto-boundary gate all clean. Four comment hunks in `pm_handler.py` are prose compression from this PR's own additions, made to fit the lock's call site under the module cap; no behaviour is in them.

cohemm:
Addressed review round 3 in `85b8a59b` — both fixed.

**Blocker 1 — `pm_handler.py:1442`, batch answer persistence is not replay-safe — fixed.**
Reproduced: with the answer committed to state and the metadata still marking the member pending, a retry of the same answer wrote a second round under the same question.

The fix keeps one authority instead of adding a second signal. `record_member_answer` reads the decision back from the interview state — the copy that is written first and is already durable — and does not record it again. The caller then persists the pending list without that member, so **the retry repairs the metadata rather than duplicating the decision**: state keeps one round for the question, `pending_batch` loses the member, and the response names what is genuinely left.

We did not take the idempotency-marker option you offered. A marker is a third place for a fact the state already holds, and it would need its own write — with the same window between writes that this finding is about. Reconciling against the record has no window: whatever survived the interruption is what the next call reads.

The comparison is deliberately narrow — the same answer, or a sentinel, which carries no words of the user's. A question whose text repeats later in the interview and is answered *differently* is a new decision and is recorded as one; that direction is pinned too, so the guard cannot swallow an answer the PM has not given yet.

Pinned: `test_a_retry_after_an_interrupted_meta_write_repairs_it` (one round for the question, `pending_batch` repaired, no generation triggered) and `test_a_different_answer_to_a_repeated_question_is_still_recorded`. The first fails without the guard — the retry appends a fifth round.

**Follow-up `pm_interview.py:750`, duplicate cleanup erases the retained primary's reframe — fixed.**
Reproduced, and your reading of the order is exact: `_apply_classification` overwrites `_reframe_map[shown]` before the duplicate check runs, so popping the key took the primary's original technical question with it. Of the two remedies you name we took the second — the map is snapshotted before the companion is applied and restored on the undo — because the first (validating without mutating) would mean deriving the shown question outside `_apply_classification`, and two places computing which question is shown is the divergence this PR just finished removing on the routing-field side.

Pinned by `test_a_dropped_companion_does_not_take_the_primary_reframe_with_it`, which fails without the fix with exactly the symptom you describe: the companion is dropped, the primary ships, and its reframe entry is gone.

**Two comments corrected.** Both were ours, and both had become false — a reader would have trusted them:
- companions and the primary are refused by the same parser but not with the same *outcome*: the primary falls back to a planning question because the turn must have one, a companion is dropped because the turn is whole without it;
- the answer lock covers the in-process answer paths within one server process — not the plugin dispatch branch, which records on its own path, and not a data directory opened by more than one server process, which an `asyncio.Lock` cannot order.

**One thing we looked at and left.** An answerless resume looks read-only and holding the lock through the next turn's generation makes a timed-out host's retry queue behind it. It is not read-only: with no pending batch and no unanswered round it plans the next turn and persists it. Exempting it would restore exactly the lost update the previous round closed. Serialized, the retry finds the first call's question already pending and re-displays it instead of planning a second one — pinned by `test_an_answerless_resume_is_serialized_because_it_also_writes`.

**Coverage you named as missing.** Plugin-mode batch behaviour is now pinned rather than argued: `test_plugin_mode_never_enters_the_batch_contract` asserts that runtime neither plans a batch nor writes pending members.

**Verification.** Full unit suite: 20,680 passed / 89 skipped. `tests/integration` + `tests/conformance` + `tests/canonical` + `tests/e2e`: 416 passed / 6 skipped / 1 xfailed. `ruff check`, `ruff format --check`, `mypy src`, module-size (`pm_handler.py` at the 2000-line cap, not grandfathered), the auto-boundary gate, and an empty diff over the RFC #2222 enumerated core boundary — all clean.

On the CI run against `dd92ad63`: the single failure was `tests/unit/mcp/test_job_manager.py::TestJobManager::test_detached_auto_status_retrieval_tracks_running_then_terminal_result`, asserting `**Status**: running` against a projection still reading `queued`. That test polls the in-memory snapshot and then reads the event-store projection separately; the gap widens on a loaded four-worker runner. It is untouched by this branch — the diff contains no job-manager or event-store file — and passes locally in the full suite. Re-run in flight.

cohemm:
Round 4, and the blocker is not fixed — it is gone. The design it lived in was withdrawn. RFC #2222 was revised first (revision 4), then the code followed, in `9d01bf74`.

## Why the remedy is not another guard

Your three blockers were at three addresses, and they were one defect:

| round | finding | what we did |
|---|---|---|
| 2 | two answers interleave; one round is lost | added a per-interview lock |
| 3 | interrupted meta write; the retry duplicates the decision | added a replay guard keyed on question text + answer |
| 4 | that guard cannot tell a retry from a question whose text recurs | — |

Each fix guarded the gap and the next round found the gap one address over. Your own recommendation did not move across those rounds, only the blocker did: *"atomic representation, an idempotency marker, or reconciliation"*, then *"durable identity tied to the specific issued batch/member, or another ordering marker"*. We read the repetition as the finding. The instance was never the point — the point was that **a batched turn kept its pending questions in `pm_meta`, beside the transcript, keyed by question text**. One fact in two places, with a window in which they disagree, and an identity (question text) that does not identify.

Adding durable per-member identity would have closed round 4. It would also have been a third record of the same turn, written by the same second write, in the same window. So we removed the second place instead.

## What the RFC now says (revision 4, decisions 2 and 5)

> A PM turn persists nothing when it asks. A round is written only when it is whole — question and answer together, through the public API — and PM keeps no pending-batch record of its own. Those two were one fact kept in two places, and every defect adversarial review found lived in the gap between them.

> What replaces them is the shape of the call, not a better record: **a turn's answers arrive together.** [...] A turn is therefore one write, and a partially-answered batch is not a state the server can be in. Answer-to-question correlation stops being a lookup — each answer arrives holding its own question — so there is nothing to match by text and nothing to reconcile after an interruption.

> **A reconnect asks again.** [...] The transcript is the only durable state and it is always a sequence of finished decisions. This costs one regenerated question and buys the absence of every reconciliation problem above.

Decision 5's "a batch resolves before the interview moves" stops being a rule the server enforces across calls and becomes a property of the call. A call carrying fewer answers than the turn asked for is not a partial batch to track — it is a host that no longer holds the turn, so the server records what it was given and plans the next turn from the transcript.

## What that deleted

- the question-only round on both PM turn paths, and the `_pending_round` helper
- `pending_batch` — written, read, matched against, and repaired
- `resolve_batch_member` (answer-to-member by question text)
- `record_member_answer`'s replay reconciliation — round 3's fix, now unnecessary
- `batch_pending_meta` / `batch_pending_result` (the partial-batch response)
- the reconnect re-display branch, and the dead pending branch in plugin mode

Added: `answers: [{question, answer}]` on the tool, `turn_answers` to normalize one call's answers, `record_turn_answers` to record them. Net **-484 lines**; `pm_handler.py` drops from the 2000-line cap to 1809.

## Two trade-offs, named here rather than found later

**A retry after a lost response duplicates a round.** If the state write succeeds and the response never reaches the host, the retry records the same round again. That is deliberate: the server cannot distinguish "the host retried" from "the PM was asked this again and answered the same way", and *trying* to distinguish them by content is precisely what rounds 3 and 4 were. Now both are recorded, under one rule. A duplicate round is visible in the transcript and a person can act on it; a discarded decision is visible to nobody.

**A skip sentinel is honoured wherever it arrives.** The record of which questions were offered as skippable *was* the pending list. Honouring an unoffered sentinel costs a deferral recorded for a question that may not have been deferrable — it loses none of the user's words. Refusing it would write `[decide_later]` into the transcript as though the PM had typed it — inventing words they did not say. The docstring carries this reasoning at the point of the decision.

## Scope — what this does not change

- **The regular interview.** Untouched, and the enumerated RFC boundary diff (`interview.py`, `turn_planner.py`, `ambiguity.py`, `answer_provenance.py`, `inner_guidance.py`, `interview_schemas.py`, `authoring_handlers.py`) is still empty. It keeps question-only rounds and single-question turns; `record_answer`'s trailing-fill contract is unchanged and still true for it.
- **The PM CLI** (`cli/commands/pm.py`) still persists an unanswered round for its own `--resume`. RFC decision 6 scopes these decisions to the PM *tool*'s turns; naming it here rather than letting it diverge silently.
- **Plugin mode** still never enters the batch producer, and now needs no exemption to say so: `plan_next_turns` has one call site and there is no pending state at all to bypass.

## Documentation this invalidated, corrected in the same commit

The answer lock's docstring, the plugin answer path's comment, `skills/pm/SKILL.md` and its mirror (the relay protocol changed), the PR body — and `pm_schemas.py`'s module overview, which is your non-blocking note: it still described the confirmed-finding round and the inherited `answer_prefix` that decision 4 retired, and now describes the v2 contract it actually implements.

## Verification

Full unit suite: 20,676 passed / 89 skipped (the one failure in the pre-fix run was a plugin-path test still sending an answer without its question; corrected in this commit and its file re-run green). `tests/integration` + `tests/conformance` + `tests/canonical` + `tests/e2e`: 416 passed / 6 skipped / 1 xfailed. `ruff check`, `ruff format --check`, `mypy src`, module-size, the auto-boundary gate, and the skill mirror: clean.

Tests: six that pinned the removed design are deleted; four new ones pin this one — a turn recorded whole, an answer without its question refused, a question whose text recurs recorded as its own round, a reconnect planning a fresh turn. The concurrency pin and the plugin-exclusion pin carry over.

On CI: the two failures on the previous heads were `test_job_manager.py::test_detached_auto_status_retrieval_tracks_running_then_terminal_result` (an in-memory snapshot polled, then a separately-projected store read) and `test_picker_indexes.py::test_picker_projection_write_and_disk_budgets` (`assert 1.7754 < 1.75`, a 1.5% overshoot of a timing budget). Both are timing assertions in files this branch does not touch — the diff contains no job-manager, event-store or persistence file — and the same suite's wall time varied 355s–470s across those runs. This head removes 929 lines and six tests; if the next run fails on a third unrelated timing test we will measure rather than keep calling it noise.

cohemm:
Addressed review round 5 in `8d642343` — blocker fixed, both follow-ups taken.

**Blocker — `pm_handler.py:1596`, issuance routing outlives the turn — fixed, and you found the right thing.**

Revision 4's removal was one address short of finished. `pending_batch` went; `pending_reframes` — the whole `_reframe_map`, written by every ask — stayed. So the design said *a turn persists nothing when it asks* while the code still persisted that turn's routing, and `restore_meta` reinstalled it after a reconnect had already declared the turn abandoned. Two incompatible notions of whether the turn still exists, exactly as your Design Gate puts it.

Reproduced as reported: a technical question reframed to "How fast must saved reports open?" is abandoned; the next turn plans a plain question with that same displayed text; `record_response` finds the surviving mapping and writes the PM's decision under `[Original technical question: Which index strategy should the store use?]` — a question nobody was asked, attached to a decision that was.

The fix is the same removal applied where it was missing, not a guard on top of it. `_begin_turn` **replaces** the reframe map when a turn is planned, from both planning entries (`plan_next_turns` and the legacy `ask_next_question`). The map now describes the turn on the wire and nothing else: an abandoned turn's routing is gone by the time the next question exists, so a later question that merely reads the same has nothing to inherit. Of the two remedies you offer this is the second — the first (route travelling back with the answer) would put the technical question on the wire for the host to echo, which makes the transcript's attribution host-asserted, and attribution is the thing being protected here.

Pinned by `test_an_abandoned_turns_reframe_does_not_attach_to_the_next_one`, which walks your path — reframed turn, abandoned, same displayed text planned plain, answered — and fails without `_begin_turn` with exactly the stale attribution you describe.

**Follow-up 1 — the atomicity wording overclaimed — fixed.** You are right that the server accepts a subset and advances, and right that this was disclosed rather than accidental. But disclosure in a PR comment is not where a host reads. The response text and `skills/pm/SKILL.md` (and its mirror) now say what is true: **one call records the turn; the server does not check that you sent them all, so whatever you leave out is abandoned rather than remembered, and a question that still matters is asked again.** No claim of server-enforced membership, because there is none — the server stopped remembering what it issued, which is the whole point of the design.

**Follow-up 2 — `response_param` — fixed.** Batched metadata now advertises `answers`. You noted no current consumer; the reason to fix it anyway is that the one thing that field exists to tell a generic host was wrong — send one answer to a turn that asked three.

**Verification.** Full unit suite: 20,677 passed / 89 skipped. `tests/integration` + `tests/conformance` + `tests/canonical` + `tests/e2e`: 416 passed / 6 skipped / 1 xfailed. `ruff check`, `ruff format --check`, `mypy src`, module-size (`pm_handler.py` at 1809 of 2000), the auto-boundary gate and the skill mirror: clean. The RFC boundary diff stays empty.

On the earlier CI failures: `Test Python 3.12` **passed** on `9d01bf74`. The two before it were `test_job_manager` and `test_picker_indexes`, different timing assertions in files this branch does not touch, on runs whose wall time varied 355s–470s. That reads as runner noise; if a third timing test fails on a later head we will measure rather than keep saying so.

cohemm:
Two findings, two different answers.

## The cross-runtime `answers` drop — you're right, and it's fixed

Adding `answers` to the in-process branch alone meant a host sending it to a plugin-mode server got success while nothing was written. The same public shape meaning two things by runtime is worse than not accepting it, and that one is mine. Both branches now go through the same `turn_answers` normalization, pinned by a regression that fails without it. Landing with the next push.

## The retry-idempotency blocker — we are not fixing this here, and here is the reasoning

The defect is real and we reproduce it: the same successful call sent twice appends two identical rounds. What we disagree with is that this PR is where it gets solved.

**There is no place inside this PR's boundary to put the thing you're asking for.** You rule out content-based deduplication, and we agree — that was round 3's fix and round 4 correctly took it out. What remains is a durable, single-use identity per turn or request. It can live in exactly three places, and each is closed:

- **On the round.** `InterviewRound` lives in `interview.py`, which RFC #2222 lists in the enumerated boundary whose diff must be empty. The RFC is explicit about what it means when an implementation finds it needs to edit that list: *"the finding is that this RFC is wrong, not that the boundary should move."*
- **In PM metadata.** That is the side file revision 4 removed. Reinstating it recreates the seam your rounds 2–4 blockers all lived in — one fact in two places, written by two writes.
- **In the artifact store.** The prompt bundle exists only when a store is wired; the fail-open path leaves briefs inline. An idempotency guarantee that silently disappears when the store is absent is not one.

**This is not a PM defect. It is a gap in the tool family, and it predates the batch.** `authoring_handlers.py:3044`: the regular interview appends a duplicate round for a repeated identical call that names its question, exactly as PM does. Neither tool has ever had request identity. Solving it for PM alone would give the two tools different durability contracts for the same operation.

**On what changed here, stated plainly rather than defended away.** Before this PR an unnamed retry was refused, because the pending question-only round could only be filled once. That was never a boundary — it was a side effect of a slot that was doing two jobs, "what is being asked" and "this has not been recorded yet". Removing it for the first job (which is what rounds 2–4 were about) removed the second. So the exposure did widen, and we should have named that cost when we removed it rather than after you found it. The remedy is not to restore the slot; restoring it restores the seam.

**Why we are not treating it as merge-blocking.** No user words are lost — the transcript gains a visible duplicate, and a duplicate a person can see is a different class of harm from the two we did fix at your prompting (an answer silently overwritten, a decision attributed to a question nobody asked). The concrete downstream cost is an inflated decision count, which can bring completion one turn early. That is worth fixing. It is not worth fixing by putting a fourth mechanism into the boundary this design just finished emptying.

**One correction to the premise, since it decides the shape of any future fix.** Your constraint is that dedup must not suppress a legitimately recurring question. In the PM tool there is no path that produces one: within-turn duplicates are already dropped on normalized identity, no planner instruction re-asks an answered question, deferred items are recorded as open PRD items rather than re-asked, and PM has no reopen/challenge path at all — that exists only in the regular interview. A verbatim repeat in PM is a generation artifact, not a designed flow. We mention it not to argue for content dedup (we agree it is the wrong instrument) but because it means the only remedy you are actually asking for is request identity, and that is a decision the interview core owns.

**Reopen conditions, so this is a position and not a refusal.** We will take it inside this PR if either holds: a probe shows a duplicate arriving without an identical retry — i.e. the shape is reachable in ordinary use rather than through a repeated call — or a session shows the inflated decision count actually completing an interview early. Otherwise this belongs to a request-identity decision for both tools, and we would rather it be made once, in the open, than approximated here.

## Verification

Full unit suite: 20,678 passed / 89 skipped. `tests/integration` + `tests/conformance` + `tests/canonical` + `tests/e2e`: 416 passed / 6 skipped / 1 xfailed. `ruff`, `mypy`, module-size, auto-boundary, skill mirror: clean. The enumerated RFC boundary diff is empty.

On the CI failure you may see on `8d642343`: `test_picker_indexes.py::test_picker_projection_write_and_disk_budgets`, `assert 1.7587 < 1.75`. We measured it rather than calling it noise — the projected/baseline write-time ratio, five runs each on the same machine: this branch 1.465 / 1.482 / 1.479 / 1.501 / 1.436, `origin/main` 1.465 / 1.466 / 1.463 / 1.486 / 1.461. Same distribution, ~17% of headroom under the budget, and the branch touches no file the test imports. CI's 1.7587 and 1.7754 are that headroom being eaten by a loaded runner.

cohemm:
**Blocker 1 (plugin mode ignores `answers`) — fixed, and the finding was correct.**

Checked against the snapshot you reviewed: at `8d642343` the plugin branch was `if answer:`, so a valid `answers` request took the delegation path and returned success with nothing persisted. The earlier reply claiming both branches already shared `turn_answers` described work that had not reached that commit — your reading of the source was right and mine was not.

Both branches now normalize through the same `turn_answers` before recording, and every resulting pair is written to plugin state. Regression: `test_the_batch_transport_means_the_same_thing_in_plugin_mode` sends two `answers` entries in plugin mode and asserts the **persisted rounds**, not the receipt. Your naming note is handled by keeping the two properties apart rather than merging them: `test_plugin_mode_never_enters_the_batch_contract` pins that plugin mode never *produces* a batch, the new test pins what it does with answers it is *handed*.

**Follow-up 1 (retry duplication) — agreed as scoped.** Repeating a successful `answers` call appends identical rounds. That is request identity across the interview tools, not something a PM-only guard should own; the last attempt at one is what the earlier replay blockers came from.

**What else changed in this push** (12 commits since `8d642343`), all inside decision 6:

- **The stub was the single point of failure for a lane.** Schema, roster and rules all sat behind one fetch, so a child that could not reach the store had no shape to answer in and nowhere it was allowed to look — its only move was to withdraw. The stub now carries what a child cannot derive: the answer's shape *named* rather than specified (fields and nesting, the empty state with the literals read out of the contract's own schema, the enum vocabularies, and the two conventions defaulted wrong — a path is repository-relative, a plain statement is in the question's language), where it may look, and the order of work. Not the JSON Schema — its bounds restate what the validator already refuses. Not a worked example either: an example is a claim already written in the answer's shape, so a lane short on evidence would hold a draft needing only its identifiers changed, and a fabricated finding that validates is the one outcome this mechanism exists to prevent.
- **Which findings exist is asked for, not sent.** Twenty contract ids rode in every prompt so a child could choose among them, and an entry is an id and a time — nothing to choose by. `ouroboros_fetch_artifact` now accepts a `lane_id` alone and lists that lane's recent findings; reading one back by `contract_id` is unchanged and every existing call behaves identically. The window, the eligible fan-out kind and the cap stay inside the query, so asking reaches no further than being offered did. This is the one shared file the change touches, and it is additive: the regular interview keeps being offered its own findings and its prompts are unchanged.
- **Order-of-work semantics.** A finding this lane published is carried as it stands rather than re-established, and investigation runs **only if** reuse turned up nothing that bears on the question. The previous wording — stop once what you read "answers the question" — gave a lane a reason to stop at the first plausible match and a reason to call that match an answer.
- The prompt carries constraints, not the reasons for them; the reasons moved to the module.

A lane prompt is ~2k characters, against ~14.7k rendered inline on `main` for the same question, and it no longer grows with how much the project has found.

RFC #2222 is at revision 6 and the PR body is updated to match; both were rewritten at decision level rather than restating the implementation.

**Verification.** `tests/unit/mcp/tools` + `tests/unit/bigbang`: 2,509 passed. Ruff over `src` and `tests` clean, mypy clean, skill mirrors byte-identical, `pm_handler.py` within the module-size policy. Publish → offer → scoped fetch was also exercised end to end against a real store, including the refusal of a lane the artifact does not carry.

cohemm:
Addressed review round 8 in `277a0021` — both blockers fixed, follow-up unchanged.

**Blocker 1 — `answers`-only plugin dispatch resumes the child with no history — fixed** (`subagent.py`). Confirmed exactly as described: `build_pm_interview_subagent` gated its resume branch on `and answer`, so an `answers`-only turn (where `answer` is `None` by construction) fell through to the generic branch, which renders no `transcript_section`. The turns carrying the most history were precisely the ones that lost it.

The branch now reads "is there anything to resume with" — `and (answer or transcript)` — and the latest answer is a conditional section beneath the history rather than the condition for showing it. Rendering is byte-identical when both are present, so the singular path is untouched. Note the regular interview's builder already put `{transcript_section}` in its own resume branch; the PM builder was the outlier, so this is that shape restored rather than a new one.

Pinned on the dispatched payload, not the on-disk rounds, as asked: `test_the_batch_transport_means_the_same_thing_in_plugin_mode` now asserts `result.meta["_subagent"]["prompt"]` contains `## Conversation History` and both recorded answers.

**Blocker 2 — skip controls change meaning by runtime — fixed** (`pm_batch.py`, `pm_handler.py`, `pm_interview.py`). Confirmed, and it contradicted RFC #2222 in its own words — *"Every call on the answer parameter is the user's answer — their own words, or the refine-gate structure they confirmed"*. A literal `[decide_later]` in the transcript is the opposite of what the token asks for.

Worth reporting how the fix landed, because the obvious version was wrong. The first attempt routed the plugin path through `record_turn_answers(self._get_engine(), ...)` — one recorder, no divergence. That broke four existing tests (`test_round5_fixes.py`) and deserved to: `_get_engine()` constructs an LLM adapter, and the plugin runtime exists so the server need not hold one. Putting an adapter behind an answer being written down would have been a worse defect than the one it fixed, on the same path.

So what was removed is the divergent *judgement*, not the call. `record_turn_answers` now takes `engine: Any | None`, and the sentinel decision — is this a control token, and what does a skipped round then say — is made once, above the runtime fork, against a shared `_SKIP_PLACEHOLDERS` table whose values are the same `DECIDE_LATER_PLACEHOLDER` / `DEFERRED_PLACEHOLDER` constants the engine's `skip_as_*` methods feed. `InputValidator.validate_user_response` now guards the field on both paths too; previously the plugin path skipped it.

Below the fork remains only what an engine uniquely owns: its reframe map, and the open items it accumulates for the summary. A runtime that plans no questions has neither — that is structural (RFC #2222: "the plugin runtime ... never enters the batch contract"), not a rule someone has to keep in step. This is also why no batch protocol reaches the plugin child: the server still only records the answer it is handed.

Pinned for both sentinels, and as an equality rather than a spelling — pinning the placeholder text would pass a plugin path that had merely copied today's wording: `test_a_skip_control_is_a_skip_on_every_runtime[[decide_later]]` / `[[deferred]]` record the same call on both runtimes and assert the persisted round is identical, is not the literal token, and that the canonical placeholder (not the token) reaches the child's prompt.

All three new assertions fail against the pre-fix source and pass after; verified by stashing the source changes and re-running.

**Follow-up 1 (request/turn identity) — unchanged.** Still shared interview-core scope, as previously agreed.

**Where the same concern could live next, since it is worth naming before it is found.** The plugin path discards `engine.decide_later_items` (it passes `engine=None` to `_save_pm_meta`), so a plugin-mode session's open items are visible to the child through the transcript's placeholder rounds rather than through that list. And a placeholder round counts as a decided round in the `generate` gate — because `record_answer` settles provenance as `user` on both runtimes, which is exactly the equality asserted above. Both are consistent between runtimes today. If either should instead be *different* from the in-process behaviour, that is a decision about what a deferral is, and belongs in the RFC before the code.

Verification: `tests/unit/mcp` + `tests/unit/pm` + `tests/unit/bigbang` — 3,288 passed, 1 skipped. `tests/unit/core` + `tests/unit/interview_adapters` — 791 passed. `tests/integration` — 262 passed, 4 skipped, 1 xfailed. Ruff check and format clean; mypy clean on all four changed source files. Module-size and skill-mirror gates green (`subagent.py` stayed inside its grandfathered budget — the explanatory comment was trimmed to fit rather than the budget raised).

cohemm:
On the follow-up finding (`answers` cardinality / duplicate identities) — **position, not fixed.** And we read the requirement row it grades as Met: it is scored against an enforced invariant the design never claimed.

One-to-three is a budget on how many questions a person can reasonably answer at once, not a rule anything downstream reads as a promise. That is why it is decided where the questions are *asked*: the planner bounds and de-duplicates what it emits, in code, pinned by tests.

Enforcing it again where the answers *arrive* inverts its purpose. `turn_answers` rejects before anything is recorded, so a turn carrying a fourth answer would lose all four — including the three nobody objects to. Refusing answers a person already wrote, for exceeding a comfort ceiling, is a worse outcome for a real session than the malformed one this describes.

PR #2234, opened 2026-08-22
fix(core): stop leaking secrets nested in lists through log sanitization

Q00:
## Problem

Two defects in `core/security.py`.

### 1. Secret leak (HIGH)

`sanitize_for_logging` recursed into `dict` but not into `list`/`tuple`. Sequences fell through to `else: result[key] = value` and were copied verbatim. Confirmed before the fix:

```python
sanitize_for_logging({"providers": [{"api_key": "sk-live-abc123"}]})
# -> {'providers': [{'api_key': 'sk-live-abc123'}]}   secret intact
```

`observability/logging.py:58` is the live consumer, so this is the real logging path — not a helper nobody calls.

### 2. Guard bypass (MEDIUM)

`is_credential_shaped` opened with `if type(value) is not str: return False`. `str` **subclasses** therefore bypassed the check entirely, including `enum.StrEnum` members — and this codebase uses `StrEnum` heavily (`SessionSignalMode`, `HumanInputKind`, `Directive`). A `StrEnum` member carrying a credential was reported as not credential-shaped.

## Fix

Added `_sanitize_logging_value`, which recurses into nested mappings and sequences to arbitrary depth, applies the same `is_sensitive_value`/`mask_api_key` treatment to bare credential strings inside sequences, and preserves sequence type (list→list, tuple→tuple, named tuples rebuilt via `_make`). Dict input still returns a dict. The original repro now yields `{'providers': [{'api_key': '<REDACTED>'}]}`.

`is_credential_shaped` switched to `isinstance`. Both functions now renormalize through the built-in implementations (`str.strip(str.__str__(value))`) so a caller-controlled method override is never invoked, and `is_credential_shaped` fails closed (returns `True`) if normalization itself fails.

### Why `is_stable_authority_identity` was deliberately *not* changed

The two predicates fail in opposite directions:

| Function | `False` means | Skipping subclasses is |
| :--- | :--- | :--- |
| `is_credential_shaped` | safe to copy/log | **fail-open** — the bug |
| `is_stable_authority_identity` | rejected | fail-closed — already safe |

Loosening the authority predicate to `isinstance` would weaken a security gate. Two things confirmed keeping it strict: an existing test (from the routing admission kernel work, #1735) asserts a hostile `str` subclass overriding `strip()` to raise is rejected; and `route_compat._canonical_compat_token` checks only `isinstance`, so it *can* retain a hostile subclass whose `__eq__`/`__hash__` could lie after the check — making this predicate a genuine last line of defense. Every production caller already normalizes to a plain `str` first (`route_policy._bounded_token`, `route_escalation`), so rejecting subclasses has no live impact. The asymmetry is now documented in both docstrings.

## One inverted test expectation

`assert is_credential_shaped(HostileString("runtime:apikey123")) is False` → now `is True`. That value **is** credential-shaped on content (verified: returns `True` as a plain `str`), so the old `False` expectation encoded the fail-open bypass being fixed here. The test's actual purpose — not raising, and terminating in linear time — is preserved.

## Testing

- `pytest tests/unit/core/test_security.py` — 51 passed
- `ruff check src/ouroboros/core/security.py` — clean
- Added coverage for both defects: nested-list secrets, lists of lists, bare credential strings inside sequences, tuple/named-tuple preservation, and `StrEnum` members


Q00 (commented):
Fixed the review concerns:
1. Wired the recursive sanitizer into the live structlog processor chain (`_mask_sensitive_data` and `_mask_dict_sensitive_data` now recurse into lists/tuples)
2. Added `_mask_sequence_sensitive_data` helper that preserves list/tuple/namedtuple types
3. Added 3 live PROD logging regression tests confirming secrets in lists, tuples, and deeply nested structures are masked

All tests pass (77/77), ruff clean, mypy clean.

Q00:
Addressed the remaining logging security blockers in c40f9a733: built-in string normalization before masking, recursive structured event sanitization, and safe tuple-subclass fallback. Added live logging/security regressions; focused tests and current CI pass. Please re-review.

Q00:
Follow-up security blockers fixed in cc1783fcc: hostile scalar events never trust arbitrary __str__, string operations normalize to built-in str, and tuple reconstruction is accepted only when it contains the exact sanitized elements. Live adversarial regressions pass. Please re-review.

Q00:
Current head includes the requested fixes and focused validation. Please run a fresh exact-head review.

shaun0927:
Direction/log update for PR #2234 at exact HEAD `7c9c35a39`.

Paperthin route applied: `hate` focused the load-bearing blocker on whether any renderer-boundary path can still persist credential material; `macrothink` kept the approved/green-check framing secondary to live console/file confidentiality.

What changed:
- `is_safe_structured_event_identifier()` now preserves the existing credential-named structured event contracts only through an explicit allowlist (`mcp.auth.invalid_api_key`, `mcp.auth.token_valid`, `context.token_count.failed`). Other structured event identifiers that are credential-shaped no longer bypass redaction.
- `_sanitize_exception_text()` now preserves quoted diagnostic mapping keys while redacting paired sensitive values, so the exception regression keeps field names without leaking values.
- Extended the live console/file namespace-event regression with the exact dotted credential forms from the latest review: `workflow.api_key_sk_live_abc123`, `workflow.access_token_ghp_credentialshapedvalue`, `workflow.client_secret_abc123`.

Verification on the pushed head:
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest tests/unit/core/test_security.py tests/unit/observability/test_logging.py tests/unit/observability/test_logging_security.py -q` -> `172 passed`
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff check ...` -> passed
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff format --check ...` -> `5 files already formatted`
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run mypy src/ouroboros/core/security.py src/ouroboros/observability/logging.py` -> no issues
- Manual PROD logger smoke wrote console/file JSON for the three dotted credential events plus a quoted exception mapping and asserted none of the secrets were persisted while `api_key`/`password` diagnostic labels remained visible.

Not run: full repository suite.

@ouroboros-agent please re-review exact HEAD `7c9c35a39`.

shaun0927:
Paperthin re0 refresh for exact HEAD `e525b4a999faf623426d24666c78a23b5f2a6bf3`.

Decision: keep this PR and keep it high priority. The leak surface is a live logging renderer boundary: credential-shaped event identifiers and exception text must not reach console or JSON file sinks verbatim. This remains necessary even though the branch was only behind main, because stale base state was the remaining merge-readiness blocker.

What changed:
- Merged current `upstream/main` into `fix/log-sanitization-secret-leak` without conflicts.
- Preserved the narrow PR diff: `src/ouroboros/core/security.py`, `src/ouroboros/observability/logging.py`, and the focused logging/security tests only.
- Amended the merge commit into Lore format so the refresh records the redaction boundary decision.

Verification at this HEAD:
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest tests/unit/core/test_security.py tests/unit/observability/test_logging.py tests/unit/observability/test_logging_security.py -q` -> `172 passed`
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff format --check src/ouroboros/core/security.py src/ouroboros/observability/logging.py tests/unit/core/test_security.py tests/unit/observability/test_logging.py tests/unit/observability/test_logging_security.py` -> `5 files already formatted`
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff check src/ouroboros/core/security.py src/ouroboros/observability/logging.py tests/unit/core/test_security.py tests/unit/observability/test_logging.py tests/unit/observability/test_logging_security.py` -> all checks passed
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run mypy src/ouroboros/core/security.py src/ouroboros/observability/logging.py` -> success, no issues
- Manual PROD logger smoke wrote four console/file events using credential-shaped dummy values and asserted `leaked=[]`; file events were redacted as `...c123`, `...alue`, `...c123`, and `manual.exception.path`.

Not tested: full repository pytest suite locally after the main refresh. GitHub checks should cover the full PR gate on this pushed head.

@ouroboros-agent please re-review exact HEAD `e525b4a999faf623426d24666c78a23b5f2a6bf3` for PR #2234. The branch is refreshed on current main, keeps the redaction scope narrow, and still preserves only explicitly safe structured event names while masking credential-shaped event values. No merge requested.

shaun0927:
Bot re-review request for exact HEAD `e525b4a999faf623426d24666c78a23b5f2a6bf3`.

Direction: KEEP, high priority. This PR closes a live logging confidentiality boundary for nested structured data, event identifiers, and exception text.

Current verification: all PR checks are green/skipped, including `Test Python 3.12`, `Ruff Lint`, `MyPy Type Check`, `enforce-boundary`, `enforce-envelope`, `enforce-module-size`, `enforce-perf-budget`, `Bridge TypeScript`, and `Native TUI (Rust)`.

The latest ouroboros-agent[bot] review is stale: it requested changes on `c3b5e8825cf96d666c835a4baa32c3c45e9b842a`, not this exact head. Please run a fresh exact-head review.

shaun0927:
Synchronize trigger log for exact HEAD `cf5ec32bce5091ad8ae44d1c5ace6a3d6bdd2610`.

Direction: KEEP, high priority. This empty Lore commit intentionally changes no tree content; it exists to force a GitHub `synchronize` event after comment, label, and reviewer-request metadata did not refresh the stale bot review.

Expected verification path: CI is running again on this exact head, and a fresh ouroboros-agent[bot] review should bind to this SHA. Prior tree-equivalent head was green/skipped.

shaun0927:
@ouroboros-agent please re-review exact HEAD `cf5ec32bce5091ad8ae44d1c5ace6a3d6bdd2610`.

This is the current paperthin readchk/re0 checkpoint: CI is green/skipped, stale reviews tied to older commits have been dismissed where applicable, and this request is only for exact-head approval or current-head change requests. No merge is requested.

shaun0927:
@ouroboros-agent please re-review current head `cf5ec32bce5091ad8ae44d1c5ace6a3d6bdd2610`. This head is the current log-sanitization security fix and the stale dismissed review is tied to an older commit, not this tree.

Fresh GitHub status on this head: all reported checks are green/skipped, and the current PR state remains blocked only by exact-head review. Please return either exact-head APPROVE or a current-head blocking finding.

PR #2237, opened 2026-08-22
fix(orchestrator): stop silently discarding the tournament winner's work

Q00:
### 1. Silent data loss (HIGH)

`export_worktree_diff` ran `git diff HEAD` and on `CalledProcessError`/`OSError` did `log.debug(...); return ""`. `apply_diff_to_workspace` then opens with:

```python
if not diff.strip():
    return True
```

So a git failure while exporting the tournament **winner's** work was indistinguishable from "the winner changed nothing", and was reported as a **successful merge**. The winning contestant's code was discarded silently, at `log.debug` severity.

### 2. Hang risk and worktree leak (HIGH)

`RunWorktreeManager._git` called `subprocess.run(["git", *args], check=True)` with no `timeout=`. It backs `create()`, `cleanup()`, `cleanup_all()`, and `__exit__`. A wedged `git worktree add` — for example `index.lock` contention with a concurrent generation — hung the tournament indefinitely, and because teardown shares the helper, `__exit__` could hang too, leaking worktrees and the `tempfile.mkdtemp` root.

## Fix

**Bug 1** — `export_worktree_diff` now returns `str | None`: `None` means "git could not be consulted", `""` means "git succeeded, genuinely no changes". `apply_diff_to_workspace` accepts `str | None` and returns `False` for `None` with an `error` log, checked **before** the empty-diff fast path. A genuinely empty diff still counts as success. Both failure logs escalated `debug` → `error`, and `TimeoutExpired` added to the caught set.

I used a sentinel rather than a new exception because this module has no exception types of its own and already signals outcomes by return value (`select_tournament_winner` returns `Optional`, `apply_diff_to_workspace` returns `bool`). Raising would have been foreign to the file's conventions.

**Bug 2** — Added `GIT_COMMAND_TIMEOUT_SECONDS = 30`, applied to `_git` and to the `git diff`/`git apply` calls. 30s matches `core/worktree.py` for the same class of operation; `git_workflow.py`'s 5s is right for `rev-parse` but too tight for `worktree add`.

Cleanup resilience: `TimeoutExpired` is **not** a `CalledProcessError` subclass, so the pre-existing handlers would have let a timed-out teardown step escape and abandon the rest of cleanup. Added `_GIT_CLEANUP_ERRORS = (CalledProcessError, TimeoutExpired, OSError)`, and `cleanup_all` now removes the `mkdtemp` root in a `finally` so a hung git cannot leak it. `create()` still fails loudly (correct for the acquisition path) but records the possibly-partial worktree before re-raising, so cleanup can force-remove and prune it.

## Caller updates

No production callers — the module is opt-in scaffolding not yet live-wired. The only call sites were in `tests/unit/orchestrator/test_n_version_tournament.py`, updated to assert `diff is not None` before the substring check. Fixing the contract now prevents the bug being inherited when the live per-AC trigger is wired into `parallel_executor`.

## Testing

`tests/unit/orchestrator/test_n_version_tournament.py` — **21 passed**. `ruff check` and `mypy` clean.

11 regression tests added: failed export via non-repo / missing git binary / timeout; the genuinely-empty-diff path staying a success; timeout kwarg propagation; cleanup surviving a wedged git without leaking the temp root; `__exit__` not propagating; failed-`create` tracking. All 11 verified to **fail against the unfixed code** (stashed the source file and re-ran) and pass with the fix.

One caveat: the full `pytest tests/unit/orchestrator/` run exceeded my command timeout, so I verified the tournament test file and the `-k tournament` selection rather than the whole directory.


Q00 (commented):
Fixed the review concerns:
1. Export now captures untracked files using `git ls-files --others --exclude-standard` + `git diff --no-index /dev/null <file>` for each untracked file, so winner work consisting only of new files produces a usable patch instead of empty string
2. Cleanup ordering fixed: `git worktree prune` now runs AFTER `rmtree` (in the finally block), so stale worktree registrations are cleaned up after the directory is actually removed
3. Added 5 regression tests covering untracked-only winner export, combined tracked+untracked, application to workspace, ls-files failure, and cleanup ordering

26/26 tests pass, ruff clean.

Q00:
Addressed winner-transfer blockers in 2c58e9c76: binary-capable patches, NUL-safe untracked path transport, option-safe path arguments, and rejection of empty rc=1 patches, with tracked/untracked binary and quoted/non-ASCII regressions. Current CI is green. Please re-review.

PR #2238, opened 2026-08-22
fix(seed): close four validation gaps in the Seed schema

Q00:
## Problem

Four fields in `core/seed.py` accept values the rest of the system cannot handle.

### 1. Blank AC descriptions collide on one identity (HIGH)

`AcceptanceCriterionSpec.description` was a bare `str` — unlike `Seed.goal`, which carries `min_length=1`. `_strip_description` stripped without rejecting, so `"   "` validated to `""`. `derive_semantic_ac_key` then hashed a constant payload, giving **every blank AC an identical `semantic_ac_key`**. That key is an identity across retries and successors, so AC-keyed state cross-contaminated between unrelated criteria.

### 2. `SeedMetadata` degradation invariants were documented but unenforced (HIGH)

The docstring states that `degraded=True` pairs with a non-default `generation_mode`, and that the partial-evidence mode carries at least one `unresolved_slots` entry. No validator checked any of it, so a Seed with real unresolved slots could hide them behind `degraded=False` and pass the run gate.

### 3. `decision_provenance` was mutable on a frozen model (MEDIUM)

`dict[str, int]` was the only unprotected field on `Seed` — every other collection is a tuple and `model_extra` is frozen via `_FrozenDict`. `seed.metadata.decision_provenance["x"] = 99` mutated the model in place.

### 4. `task_type` was free-form despite a closed vocabulary (MEDIUM)

Documented as 7 values, typed as `str`, while the same file already uses `Literal` for `InvestmentLevel`/`InvestmentProvenance`/`InvestmentConfidence`. A typo like `"cod"` was accepted and silently changed execution routing.

## Fixes

1. `Field(..., min_length=1)` **and** a rejecting before-validator, matching how `Seed.goal` is guarded. The validator fires first so the error names the field.
2. A `model_validator(mode="after")` with three rules.
3. `decision_provenance` frozen through the file's existing `_FrozenDict`. It subclasses `dict`, so `== {}`, `model_dump`, and the additive serializer keep working; its `__copy__`/`__deepcopy__` return self, so `model_copy` is unaffected.
4. `TaskType = Literal[...]` beside the `Investment*` literals, with a before-validator that strips and lower-cases — `get_strategy` has always lower-cased its lookup, so without it a hand-authored `task_type: Code` would newly break.

## One correction to the invariant as originally specified

Enforcing "partial mode ⇒ non-empty `unresolved_slots`" *unconditionally* is wrong and breaks a deliberately-tested path. `ledger_seed.partial_seed_from_evidence` derives `degraded = bool(unresolved_slots)` (`ledger_seed.py:300`), so a **complete** ledger routed through the recovery helper legitimately produces `degraded=False, generation_mode="partial_seed_from_evidence", unresolved_slots=()`. `test_ledger_seed.py::test_complete_ledger_marks_seed_non_degraded` asserts exactly that, noting the partial mode is retained "so audit can tell this Seed came from the recovery path even though no gap existed". That is not a partial product — nothing is unresolved.

Since the docstring scopes both cases under "degraded: True when…", the rules landed as:

1. `degraded=True` ⇒ `generation_mode != "normal"`
2. `degraded=True` and partial mode ⇒ non-empty `unresolved_slots`
3. non-empty `unresolved_slots` ⇒ `degraded=True`

Rule 3 is what actually blocks the failure mode: a Seed with genuine unresolved slots can no longer claim `degraded=False`. No call site needed changing; both `ledger_seed` paths satisfy all three.

## Call sites updated

Two tests that asserted on the old, later blank-AC boundary:

- `test_workflow_ir_adapter.py` — `Seed(...)` now rejects the blank AC before the adapter sees it. Split into a Seed-boundary assertion plus a direct test of the adapter's `_normalize_acceptance_criteria` guard, so the defense-in-depth check stays covered.
- `test_main.py::test_workflow_ir_inspect_rejects_blank_ac` — still exits 1; now asserts the earlier schema failure message.

## Testing

- `tests/unit/core/test_seed.py` — 133 passed
- `tests/unit/bigbang tests/unit/core tests/unit/auto` — 3753 passed
- `tests/unit` (full) — 20684 passed, 89 skipped, 12 failed
- `ruff check` + `ruff format` clean; `mypy src/ouroboros` clean across 596 files, which verifies the `TaskType` narrowing breaks no caller

All 12 failures are pre-existing and environmental — this box has only `python3`, not `python` — and reproduce identically on a stashed, unmodified tree. Zero regressions.

+183 lines of regression tests: accepted/rejected `SeedMetadata` shapes, the `_FrozenDict` mutation guard across `__setitem__`/`update`/`pop`/`clear`, additive serialization round-trip, a test pinning `_PARTIAL_SEED_GENERATION_MODE` to `ledger_seed.PARTIAL_SEED_GENERATION_MODE` so the duplicated constant cannot drift, parametrized blank-description forms, and all seven `task_type` values resolving to a real strategy.

## Note for reviewers

The venv's editable install points at the main repo's `src`, so tests run from a worktree silently exercise the wrong tree. All results above were re-run with `PYTHONPATH` pinned to this branch's `src` and verified via `ouroboros.core.seed.__file__`.

### DIFF
--- a/src/ouroboros/core/seed.py
+++ b/src/ouroboros/core/seed.py
@@ -879,16 +879,18 @@ class Seed(BaseModel, frozen=True):
         """
         if not isinstance(value, str):
             return value
-        normalized = value.strip().lower()
+        # Late import to avoid circular dependency (core -> orchestrator).
+        from ouroboros.orchestrator.execution_strategy import (
+            _canonicalize_task_type,
+            is_registered_task_type,
+        )
+
+        normalized = _canonicalize_task_type(value)
         if not normalized:
             msg = "task_type must be a non-empty string"
             raise ValueError(msg)
         if normalized in BUILTIN_TASK_TYPES:
             return normalized
-        # Late import to avoid circular dependency (core -> orchestrator).
-        from ouroboros.orchestrator.execution_strategy import (
-            is_registered_task_type,
-        )
 
         if is_registered_task_type(normalized):
             return normalized
--- a/src/ouroboros/core/seed.py
+++ b/src/ouroboros/core/seed.py
@@ -364,6 +364,18 @@ class BrownfieldContext(BaseModel, frozen=True):
     )
 
 
+_DEFAULT_GENERATION_MODE = "normal"
+"""``SeedMetadata.generation_mode`` for the legacy ledger-complete path."""
+
+_PARTIAL_SEED_GENERATION_MODE = "partial_seed_from_evidence"
+"""``SeedMetadata.generation_mode`` for incomplete-ledger recovery Seeds.
+
+Mirrors ``ouroboros.auto.ledger_seed.PARTIAL_SEED_GENERATION_MODE``; duplicated
+here because ``ledger_seed`` imports this module. The two are pinned equal by a
+regression test.
+"""
+
+
 class SeedMetadata(BaseModel, frozen=True):
     """Metadata about the Seed generation.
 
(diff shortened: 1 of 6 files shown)


Q00:
Addressed the custom strategy regression in 007b71581 by validating built-ins plus dynamically registered strategy identifiers through the shared registry contract. Added registered/unregistered/case-normalization regressions; targeted and broad suites pass, and current CI is green. Please re-review.

Q00:
Follow-up blocker fixed in 0cdf943de: registration, lookup, presence checks, and Seed validation now share one strip+lower canonicalizer, including whitespace-at-registration end-to-end regressions. Please re-review the new HEAD.

PR #2239, opened 2026-08-22
fix: bound every subprocess wait that could hang indefinitely

Q00:
### Three CLI adapters defaulted to an unbounded wait

Each had this shape, with `None` as the constructor default — so the **unbounded branch was the default path**:

```python
if self._timeout is None:
    await process.wait()          # blocks forever
else:
    await asyncio.wait_for(process.wait(), timeout=self._timeout)
```

A wedged CLI blocked the generation forever with no diagnostic.

Caller analysis: the real class names are `CopilotCliLLMAdapter`, `CodexCliLLMAdapter`, `GjcLLMAdapter`. The only construction path is `providers/factory.py` forwarding `create_llm_adapter(timeout=...)`, and **no production caller passes `timeout` at all** — not `evaluation_handlers`, `pm_handler`, `authoring_handlers`, `qa.py`, `mcp/server/adapter.py`, or `auto/adapters.py`. Only two tests pass one, both explicit positive floats. So nothing relied on unbounded behavior and a default ceiling is the correct fix.

### The plugin installer could hang on a credential prompt

`plugin.py` `_shallow_clone` ran `git clone --depth 1` and `git rev-parse HEAD` with `check=True`, no `timeout=`, and stdin not redirected. With `capture_output=True`, an unreachable host or a git credential prompt hung `ooo plugin add` silently and forever.

### Adapters

Added `DEFAULT_CLI_COMPLETION_TIMEOUT_SECONDS = 1800.0` plus `_effective_timeout_seconds()` / `_timeout_is_default_ceiling()` to `RuntimeStreamMixin`. **1800s** matches the `per_iteration_timeout_seconds` default the Ralph loop already enforces, so a single completion can never outlive the iteration that owns it. An explicit positive `timeout` still wins unchanged.

The idle-guard approach used by `opencode_runtime`/`codex_cli_runtime` does not apply here: `_iter_stream_lines` does not accept the chunk-timeout kwargs (that is `iter_runtime_stream_lines`, used only by the orchestrator runtimes), and these adapters hang on `process.wait()`, not on a stream read.

Sites fixed — including three not in the original report:

| Site | What was unbounded |
| :--- | :--- |
| `copilot_cli_adapter.py:803` | `_handle_streaming_process` |
| `copilot_cli_adapter.py:600` | `_collect_legacy_process_output` — same shape on `communicate()` |
| `codex_cli_adapter.py:1188` | streaming |
| `codex_cli_adapter.py:950` | legacy `communicate()` |
| `gjc_llm_adapter.py:400` | `await run_protocol()` |
| `gjc_llm_adapter.py:437` | `await stderr_task` + `await process.wait()` after `agent_end` |

The last one uses a 5s `_process_shutdown_timeout_seconds` — the established "process should be exiting now" budget. On expiry the child is terminated and the already-received answer is still returned, so a merely slow-exiting CLI is not turned into a failure.

Timeout errors now report the effective seconds plus `timeout_was_default: bool`, distinguishing an operator limit from the fallback.

### Plugin installer

`_GIT_CLONE_TIMEOUT_SECONDS = 300` (network reach) and `_GIT_REV_PARSE_TIMEOUT_SECONDS = 30` (local object-store read — it either answers immediately or is wedged), following the `timeout=30` style in `core/worktree.py`. Both calls get `stdin=subprocess.DEVNULL` and `env=_git_noninteractive_env()`, which sets `GIT_TERMINAL_PROMPT=0`, `GIT_ASKPASS`/`SSH_ASKPASS=true`, and `GIT_SSH_COMMAND="ssh -oBatchMode=yes"` — the SSH transport prompts through its own channel, which `GIT_TERMINAL_PROMPT` alone does not close.

## A latent bug the timeout would have introduced

Both git call sites caught `(subprocess.CalledProcessError, OSError)`. `subprocess.TimeoutExpired` is a `SubprocessError`, **not** an `OSError` — so adding a timeout would have converted a hang into an unhandled traceback. Widened both to `subprocess.SubprocessError` and added an explicit `TimeoutExpired` branch to `plugin_cache.url_cache_refresh_error`, which also handles `exc.cmd` being a `str` (where `" ".join` would otherwise join characters).

## Testing

- `ruff check` + `ruff format --check` on all 10 changed files — clean
- `mypy` on all 6 changed source files — clean
- `pytest tests/unit/providers` — 851 passed, 7 skipped
- `pytest tests/ -k plugin` — 1138 passed, 5 skipped
- `pytest tests/unit/cli` — 2083 passed, 7 skipped

Six regression tests: default-ceiling coverage for copilot / codex / gjc (each asserts `_timeout is None`, that the effective value is the module default, then drives a wedged process to a bounded failure with `timeout_was_default is True`), the bounded gjc reap returning its answer despite a never-exiting child, `_shallow_clone` asserting timeout + DEVNULL stdin + prompt-disabling env with clone > rev-parse, and a parametrized `TimeoutExpired` path asserting exit 1 with no traceback.

## Note for reviewers

The venv's editable install resolves `ouroboros` to the main repo's `src`, so tests launched from a worktree silently exercise the wrong tree. All results above used `PYTHONPATH` pinned to this branch's `src`, verified via `ouroboros.__file__`.

### DIFF
--- a/src/ouroboros/cli/commands/plugin.py
+++ b/src/ouroboros/cli/commands/plugin.py
@@ -1278,6 +1278,33 @@ def _looks_like_url(target: str) -> bool:
     )
 
 
+# `git clone` reaches the network, so it gets a generous ceiling; `git rev-parse`
+# is a local object-store read that either answers immediately or is wedged.
+# Both are bounded because `capture_output=True` hides a credential prompt: an
+# untimed call turns an unreachable host into a silent, permanent hang of
+# `ooo plugin add`.
+_GIT_CLONE_TIMEOUT_SECONDS = 300
+_GIT_REV_PARSE_TIMEOUT_SECONDS = 30
+
+
+def _git_noninteractive_env() -> dict[str, str]:
+    """Environment that makes Git fail instead of prompting for credentials.
+
+    `GIT_TERMINAL_PROMPT=0` is Git's own kill switch for terminal prompts.
+    Pointing `GIT_ASKPASS`/`SSH_ASKPASS` at `true` stops a GUI helper from
+    waiting on input off-screen, and `ssh -oBatchMode=yes` does the same for
+    the SSH transport.  Combined with a `DEVNULL` stdin, a private or moved
+    repository fails fast with a real stderr message instead of blocking on a
+    prompt nobody can see.
+    """
+    env = os.environ.copy()
+    env["GIT_TERMINAL_PROMPT"] = "0"
+    env["GIT_ASKPASS"] = "true"
+    env["SSH_ASKPASS"] = "true"
+    env["GIT_SSH_COMMAND"] = "ssh -oBatchMode=yes"
+    return env
+
+
 def _normalize_clone_url(target: str) -> str:
     """Strip the Python-style `git+` prefix that pip/uv accept but Git itself
     does not understand.
@@ -1301,6 +1328,9 @@ def _shallow_clone(repo_url: str, dest: Path) -> str:
         check=True,
         capture_output=True,
         text=True,
+        timeout=_GIT_CLONE_TIMEOUT_SECONDS,
+        stdin=subprocess.DEVNULL,
+        env=_git_noninteractive_env(),
     )
     sha = subprocess.run(
         ["git", "rev-parse", "HEAD"],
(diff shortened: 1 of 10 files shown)


Q00:
Addressed both subprocess lifecycle blockers in 967bf3492: stream drains now remain inside the completion deadline, legacy Codex/Copilot timeouts terminate and return structured ProviderError diagnostics, and module size is back under budget. Provider suites, focused CLI tests, Ruff, mypy, and current CI pass. Please re-review.

Q00:
Follow-up blocker fixed in 8e8948467: noninteractive Git now preserves inherited GIT_SSH_COMMAND transport options/wrappers and appends BatchMode only when absent, with inherited/custom/end-to-end regressions. Please re-review the new HEAD.

Issue #2273, opened 2026-08-24
Codex setup fails and deletes rules/ouroboros.md when the filesystem rejects RENAME_NOREPLACE (NFS)


bjkim95:
## Summary

On a filesystem that does not implement `RENAME_NOREPLACE`, every Codex artifact
install fails, and a refresh over an existing install **deletes the user's
`~/.codex/rules/ouroboros.md`** instead of rolling back.

`src/ouroboros/codex/artifacts.py::_rename_noreplace` commits every managed
artifact with `renameat2(AT_FDCWD, src, AT_FDCWD, dst, RENAME_NOREPLACE)`.
NFS rejects that flag with `EINVAL` and renames nothing. The current code treats
any non-zero return as a rename failure, so the install aborts — and the rollback
path, which restores the previous generation through the *same* helper, fails
identically. The previous generation is then left detached in a hidden
`.ouroboros.md.<rand>.backup` sibling with nothing pointing at it.

Any home directory on NFS is affected (clusters, NAS-mounted `$HOME`, some
container setups). `ENOSYS` (kernel < 3.15) and `EOPNOTSUPP` (some FUSE and
overlay mounts) are the same class of failure.

## Impact

- `ouroboros setup refresh` / `ouroboros setup --runtime codex` can never
  complete on an NFS home: `Runtime artifact refresh incomplete: codex`.
- **Silent data loss on the second run.** A pre-existing
  `~/.codex/rules/ouroboros.md` is removed and not restored. The content
  survives only as a dotfile the user has no reason to look for.
- Directory artifacts are equally affected: the Codex stage installs 22 skill
  directories through the same helper, and the existing non-Linux POSIX fallback
  (`os.link` + `unlink`) cannot rename a directory — so a file-only fallback
  would not be enough.

## Steps to Reproduce

Requires `$HOME` (or `$CODEX_HOME`) on NFS.

1. Confirm the filesystem rejects the flag:

   ```python
   import ctypes, os
   libc = ctypes.CDLL(None, use_errno=True)
   fn = libc.renameat2
   fn.argtypes = (ctypes.c_int, ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p, ctypes.c_uint)
   fn.restype = ctypes.c_int
   open("probe_src", "w").write("x")
   ctypes.set_errno(0)
   rc = fn(-100, b"probe_src", -100, b"probe_dst", 0x1)   # RENAME_NOREPLACE
[...]

PR #2274, opened 2026-08-24
fix(codex): degrade when renameat2 rejects RENAME_NOREPLACE

bjkim95:
## Summary

`_rename_noreplace` commits every managed Codex artifact with
`renameat2(..., RENAME_NOREPLACE)` and treats any non-zero return as a rename
failure. NFS rejects the flag itself with `EINVAL` and renames nothing, so on an
NFS home every Codex artifact install aborts — and the rollback, which restores
the previous generation through the same helper, fails identically and leaves
`rules/ouroboros.md` deleted with its content stranded in a hidden `.backup`
sibling. This treats `EINVAL` / `ENOSYS` / `EOPNOTSUPP` as "the flag is not
implemented here" and falls through to a portable path instead, which is now
also able to publish directories (`os.link` cannot, and the Codex stage installs
22 skill *directories*). `EEXIST` and every other errno still propagate — the
no-replace guarantee is not what is being relaxed.

## Review boundary

- **User problem**: On a home directory whose filesystem rejects `RENAME_NOREPLACE`, every Codex artifact install fails, and the rollback — which renames through the same helper — deletes `rules/ouroboros.md` and strands its content in a hidden `.backup` sibling. `ouroboros setup` is unusable on an NFS home. (#2273)
- **Inputs and execution conditions**: POSIX hosts where `renameat2(..., RENAME_NOREPLACE)` returns `EINVAL`, `ENOSYS`, or `EOPNOTSUPP` *without renaming anything* — NFS, some FUSE and overlay mounts, kernels older than 3.15. The artifacts are the one Codex rule file and the 22 packaged skill *directories*. The supported concurrency condition is the one Ouroboros already assumes elsewhere: writers that are not cooperating are handled to the bound stated below, not eliminated.
- **Promised contract**:
  - Those three errnos, and only those, degrade to a portable publication path. `EEXIST` and every other errno reach the caller unchanged.
  - Files publish with `os.link`, which is no weaker than `RENAME_NOREPLACE`: fully atomic, `EEXIST` on an occupied target.
  - Directories publish by reserving the name with `os.mkdir` — atomic, `EEXIST` rather than replace — and consuming that reservation with `os.rename`.
  - A committed publication is never reported as a failure, and a failed publication never removes a name it did not reserve, never leaves the final name occupied for the next replay, and never strands the previous generation.
- **Implementation boundary and owner**: `src/ouroboros/codex/artifacts.py` only — `_rename_noreplace`, the new `_rename_noreplace_fallback` and `_release_directory_reservation`, and the staging-alias sweep in `_commit_staged_artifact`. No new subsystem, no new ownership boundary, no install-path or managed-surface change. Owned by the Codex artifact installer.
- **Non-goals**:
  - Crash and interruption replay, or durable journaling, for Codex artifact transactions. The `.backup` scheme on `main` has the same exposure with this PR reverted; making it replay-safe is a new ownership boundary.
  - Pinned no-follow parent authority across the Codex lifecycle — publication, backup acquisition, prune discovery, and `cli/commands/uninstall.py`. Pre-existing on `main`, in files this PR does not open.
  - Extracting a shared managed-artifact transaction primitive and migrating every Codex boundary onto it. That is a new subsystem and a maintainer decision, not a bug fix; see the follow-up below.
  - The same `renameat2`-or-die shape in `hermes/artifacts.py::_rename_no_replace` and `cli/runtime_activation.py::_rename_if_absent`. I could not reproduce either — the Codex stage fails first — and I would rather not widen an untested change.
  - The macOS `renamex_np` `ENOTSUP` path. No macOS mount to verify it on, so that branch is unchanged.
  - Two race windows that are **not closable in a branch defined by the absence of `RENAME_NOREPLACE`**, both bounded to an *empty* directory and both documented in the helper: a writer who removes the reservation and recreates an empty directory before the rename, and the identity-check-to-`rmdir` interval in failure cleanup. `mkdir` returns no handle, so a reservation cannot be pinned at creation, and inode comparison does not substitute — a recreated directory reuses the same inode number (measured 5/5 on both NFS and ext4). A racing writer with *any content* is never touched: the rename fails `ENOTEMPTY` and `rmdir` refuses a non-empty directory. `test_directory_commit_never_destroys_a_racing_writer_with_content` pins that boundary.
- **Evidence**: `TestRenameNoReplaceOnFilesystemsWithoutTheFlag` forces `renameat2` to report each errno without renaming anything, so the contract is checked on any filesystem including CI; plus a manual before/after run against a throwaway `CODEX_HOME` on the reporting machine's NFS4 mount. Both below.

## Test plan

- [x] `uv run pytest tests/ -q` passes **in CI** (`Test Python 3.12`, 7m4s). Run in full
  locally too: 53 failures remain on this host, none of them from this change — evidence below.
- [x] `uv run mypy src/` clean (or noted exception)

Environment: Python 3.13.13, `uv sync --python 3.13 --dev --group mcp-test --group litellm-test`
(the same dependency groups `test.yml` installs). Ubuntu, `$HOME` on NFS4.

<details>
<summary><code>ruff</code></summary>

```console
$ uv run --python 3.13 ruff check src/ tests/
All checks passed!

$ uv run --python 3.13 ruff format --check src/ tests/
1396 files already formatted
```

</details>

<details>
<summary><code>mypy</code></summary>

```console
$ uv run --python 3.13 mypy src/ouroboros
Success: no issues found in 613 source files
```

</details>

<details>
<summary><code>pytest</code> — full suite, CI invocation</summary>

Green in CI on `648655e2`, on the invocation `test.yml` uses:

```console
Test Python 3.12 ....... pass  7m4s
```

Locally, on a 20-core compute node, same invocation:

```console
$ uv run --python 3.13 pytest tests/ -n 4 --dist worksteal --durations=25 -m "not performance" -q
53 failed, 21878 passed, 13 skipped, 1 xfailed, 8 warnings in 227.23s (0:03:47)
```

**The 53 are this host, not this change.** Proof rather than assertion: I checked
`src/ouroboros/codex/artifacts.py` out from `main` (nothing else touched) and re-ran
the 11 files that contain every one of those failures, same interpreter, same env,
same command:

```console
# with the fix applied
54 failed, 1681 passed, 2 skipped in 14.77s
# with src/ouroboros/codex/artifacts.py reverted to main
53 failed, 1682 passed, 2 skipped in 16.92s

$ diff <(grep ^FAILED with_fix.log | sort) <(grep ^FAILED reverted.log | sort)
< FAILED tests/unit/orchestrator/test_parallel_executor.py::test_capture_rejects_intermediate_workspace_ancestor_swap
```

That one difference is a flaky test, not a regression, and it is flaky **on `main`**.
Run alone, repeatedly, on the same node:

```console
# with the fix              # with artifacts.py reverted to main
1 passed in 0.95s           1 passed   1 passed
1 failed in 1.09s           1 passed   1 passed
1 passed in 0.91s           1 passed   1 failed   <-- fails on main too
1 passed in 0.70s           1 passed   1 passed
1 failed in 0.87s           1 passed   1 passed
                            ( 8 runs: 7 passed, 1 failed )
```

It also passed in the full run above, and `Test Python 3.12` runs the same file green
in CI. `orchestrator/parallel_executor.py` contains no reference to `codex`, and the
test's own adversarial `os.open` counter never matches a path this PR opens.

The remaining 52 are the host's shell and PATH, not git — this node ships git 2.55.0,
so the `project_identity` cluster reported in earlier rounds of this PR is gone. They
are dominated by `auto/test_seed_preflight.py` and `orchestrator/test_runtime_evidence.py`,
which shell out to `sh -c` / `env -C` and assert on program resolution; the captured
stderr is `sh: 1: python: not found`. None of the modules behind them
(`orchestrator/runtime_evidence.py`, `auto/seed_preflight.py`,
`orchestrator/parallel_executor.py`) reference `codex.artifacts` or `_rename_noreplace`.

</details>

<details>
<summary>Conditional gates</summary>

```console
$ python3 scripts/check-module-size.py --baseline-ref origin/main
module-size: OK (612 modules, cap 2000, 26 grandfathered)

$ python3 scripts/check-auto-boundary.py
ooo-auto-boundary: OK (49 files scanned, 0 findings)

$ python3 scripts/check-max-turns-envelope.py
check-max-turns-envelope: OK (613 files scanned, 0 findings)
```

`codex/artifacts.py` goes from 1076 to 1100 lines, well under the 2000 cap.
`enforce-perf-budget` and `enforce-boundary` do not apply — this PR does not
touch `src/ouroboros/auto/`.

</details>

<details>
<summary><code>Bridge TypeScript</code></summary>

```console
$ bun install && bunx tsc --noEmit && bun test
bun test v1.3.14 (0d9b296a)
 152 pass
 0 fail
 1477 expect() calls
Ran 152 tests across 1 file. [5.12s]
```

(`tsc --noEmit` exited 0 with no diagnostics. Untouched by this PR; run because
it is a required check.)

</details>

### Manual verification on the affected filesystem

The reporting machine's `$HOME` is NFS4, so the real path is testable. Against a
throwaway `CODEX_HOME` (never the live `~/.codex`):

Before — both runs fail, and the second one loses the file:

```text
PASS 1 (clean install)
    FAILED: OSError: [Errno 22] Invalid argument: '.../rules/.ouroboros.md.yo16rrrd.tmp -> .../rules/ouroboros.md'
--- after pass 1 ---
PASS 2 (refresh over an existing generation)
    FAILED: OSError: [Errno 22] Invalid argument: '.../rules/.ouroboros.md.6jwh2a4c.backup -> .../rules/ouroboros.md'
    rules/ouroboros.md exists after failure: False
--- after pass 2 ---
    rules/.ouroboros.md.6jwh2a4c.backup
```

After — `ouroboros setup refresh` succeeds, twice, with nothing stranded:

```text
╭───────────────────── Success ─────────────────────╮
│ Installed Codex rules → .../rules/ouroboros.md    │
╰───────────────────────────────────────────────────╯
╭───────────────────── Success ─────────────────────╮
│ Installed 22 Codex skills → .../skills            │
╰───────────────────────────────────────────────────╯
╭───────── Success ─────────╮
│ Refreshed runtime         │
│ artifacts: codex          │
╰───────────────────────────╯

$ ls -a .../rules
.  ..  ouroboros.md
```

### What the new tests pin

`tests/unit/test_codex_artifacts.py::TestRenameNoReplaceOnFilesystemsWithoutTheFlag`
forces `renameat2` to report a given errno without renaming anything, so the
requirement is checked on any filesystem (CI included). Every case below fails
on `main` except the two "must still propagate" ones, which pin that the fix
does not over-swallow:

- a file artifact and a directory artifact both install under `EINVAL`
- a refresh over an existing generation keeps `rules/ouroboros.md` and leaves no
  `.backup` sibling and no surviving staging alias — the data-loss regression
- the fallback still raises `FileExistsError` on an occupied file target
  (`EINVAL` / `ENOSYS` / `EOPNOTSUPP`) and on an occupied *empty directory*
  target, which a bare `os.rename` would have silently replaced
- a failed publication leaves no reservation at the final name, and never
  removes a reservation another writer took over — by `rmdir` + recreate, or by
  renaming it away, which a liveness-only check would have missed
- a racing writer with *content* in the reservation keeps their tree
  (`ENOTEMPTY`) and the staged tree stays unpublished — the stated bound on the
  window that cannot be closed here
- a reservation that cannot be pinned (`os.open` fails after `os.mkdir`) does
  not stay at the final name, so the next replay does not see a permanent
  collision with the previous generation still hidden in `.backup`
- a descriptor release that fails *after* the rename does not turn a committed
  publication into a reported failure
- `EEXIST` from `renameat2` still surfaces as `FileExistsError`, and `EACCES`
  still propagates as `OSError`

### Notes for review

- **Boundary.** The declared boundary above is the one I am asking to be
  reviewed against. Findings that are valid but land outside it — crash replay,
  parent-path authority across the Codex lifecycle, `uninstall.py` discovery,
  a shared transaction primitive — I would like to take as an owned follow-up
  rather than absorb here, per
  [CONTRIBUTING.md#five-question-review-rubric](https://github.com/Q00/ouroboros/blob/main/CONTRIBUTING.md#five-question-review-rubric).
  Tracked in #2294, which I own; happy to be redirected on its shape.
- **Behavior change on other POSIX platforms.** The old code raised
  `OSError(ENOTSUP, "Atomic no-replace directory rename is unavailable on ...")`
  for a *directory* on any non-Linux, non-Darwin POSIX platform. That path now
  installs via the reservation described above. Nothing in the tree depended on
  that message.
- **Guarantee degradation.** Files keep the atomic guarantee (`os.link` fails
  with `EEXIST`). Only directories degrade — to an atomic `mkdir` reservation
  consumed by `rename`, with the residual window stated in the non-goals and in
  the helper's docstring — because these filesystems offer nothing stronger.
- **Docs.** `codex/artifacts.py` has no entry in the CONTRIBUTING doc-mapping
  tables and no install path or managed surface changes, so no doc update.

## Related issues

Refs #2273
Follow-up: #2294


bjkim95:
Both blockers were real. I reproduced each one against `858135ae` before changing anything, and both are fixed in `efbbcbd3`.

### Blocker 1 — directory publication could replace a concurrent generation

Confirmed, and worse than my docstring admitted: POSIX `rename` *removes an empty destination directory*, so the check-then-rename did not merely fail to detect the race, it destroyed the other writer's name.

```console
rename onto EMPTY dir: REPLACED
```

Fixed by reserving the name instead of checking it. `os.mkdir` is the atomic no-replace gate — it fails with `EEXIST` rather than replacing — and the rename then only consumes the reservation this call made:

```python
os.mkdir(target_path)
try:
    os.rename(source_path, target_path)
except BaseException:
    try:
        os.rmdir(target_path)
    except OSError:
        pass
    raise
```

A writer who fills the reservation before the rename makes it fail with `ENOTEMPTY`; their tree is never removed. Verified on the NFS mount this PR is about:

```console
mkdir on occupied name        -> FileExistsError errno 17
rename onto own reserved dir  -> staged tree published
rename onto NON-empty dir     -> errno 39 Directory not empty
```

### Blocker 2 — post-link cleanup failure reported a failed commit

Confirmed exactly as described. Forcing `renameat2` to `EINVAL` and the first staging unlink to fail, through `install_codex_rules()`:

```console
install: FAILED OSError: Managed Codex artifact changed during rollback: .../rules/ouroboros.md
rules dir contents:
    .ouroboros.md._zo9m6dk.backup -> ['PREVIOUS GENERATION']
    ouroboros.md                  -> ['# Ouroboros for Codex']
```

The new rule live, the old one stranded — the failure this PR exists to remove, reintroduced on the path it enables.

Taking your first option: the link **is** the commit boundary, so the unlink after it is cleanup and is now best-effort. Same reproduction on `efbbcbd3`:

```console
install: OK
rules dir contents:
    .ouroboros.md.ndyv9arc.tmp -> ['# Ouroboros for Codex']
    ouroboros.md               -> ['# Ouroboros for Codex']
```

One trade-off worth stating plainly: when the unlink fails, the staging hard link survives. It is a dotfile pointing at the same inode as the committed rule, it is not namespaced so pruning ignores it, and it is inert. I judged an orphaned hard link strictly better than losing a generation. If you would rather the caller sweep it, say so and I will thread the "target was published, staging may remain" signal back into `_commit_staged_artifact` instead.

### Regression tests

Three added, in `TestRenameNoReplaceOnFilesystemsWithoutTheFlag`:

- `test_directory_commit_refuses_a_target_that_appears_after_the_check` — forces a stale-free existence check (`os.path.lexists` → `False`) with the target actually occupied, i.e. the writer wins the race. Asserts `FileExistsError` and that the concurrent directory and the staged tree are both untouched.
- `test_failed_staging_cleanup_does_not_report_a_failed_commit` — installs over an existing generation with the staging unlink refused; asserts the packaged rule is installed and **no `.backup` sibling remains**.
- `test_directory_commit_leaves_no_reservation_when_publication_fails` — a failed rename must not leave the reserved name occupied.

The first two fail against `858135ae` and pass on `efbbcbd3`:

```console
$ pytest tests/unit/test_codex_artifacts.py -k RenameNoReplaceOnFilesystems -q   # at 858135ae
FAILED ...::test_directory_commit_refuses_a_target_that_appears_after_the_check
FAILED ...::test_failed_staging_cleanup_does_not_report_a_failed_commit
2 failed, 10 passed

$ pytest tests/unit/test_codex_artifacts.py -k RenameNoReplaceOnFilesystems -q   # at efbbcbd3
12 passed
```

### Checks

```console
$ ruff format src/ tests/          1396 files left unchanged
$ ruff check src/ tests/ --fix     All checks passed!
$ mypy src/ouroboros               Success: no issues found in 613 source files
$ scripts/check-module-size.py --baseline-ref origin/main   OK (612 modules, cap 2000)
$ pytest tests/unit/test_codex_artifacts.py tests/unit/cli/test_codex_command.py -q   277 passed
$ pytest tests/unit/codex tests/unit/hermes tests/unit/skills tests/unit/cli -q
  1 failed, 2258 passed, 3 skipped
```

That one failure is `test_tui_command.py::test_tui_open_unknown_terminal_exits_zero_with_manual_command`, which fails identically with my two files reverted — this host has a terminal the test expects to be absent. As before, the full suite is not green here for environment reasons I documented in the PR body; CI's `Test Python 3.12` was green on `858135ae` and is the authority.

On your design note — agreed that `_rename_noreplace` is a commit primitive, not a rename helper. Both changes above are aimed at that contract rather than at making the NFS case pass.

bjkim95:
Thanks — on the remaining follow-up (`artifacts.py:425`, reservation removed and recreated before the rename). The race is real; I reproduced it. But I am not implementing the suggested ownership check, because I measured it and **it does not detect this scenario**, and I would rather not ship a check that advertises a guarantee it cannot deliver.

### Why inode comparison does not work here

The suggestion is to prove the destination is still the reservation this call created. `mkdir` returns no handle, so the only available proof is an inode comparison — and a recreated directory at the same path reuses the same inode number:

```console
os.mkdir(tgt); reserved = os.stat(tgt)
os.rmdir(tgt); os.mkdir(tgt)          # the ABA sequence from the review
current = os.stat(tgt)
inode changed by the swap: False
rename consumed the SWAPPED dir: True
```

Not a one-off — 5/5 trials on both filesystems:

```console
NFS  (workspace): 5 5 trials where the swapped dir reused the SAME inode
local(/tmp)     : 5 5 trials where the swapped dir reused the SAME inode
fs types: ['nfs', 'ext2/ext3']
```

So `(st_dev, st_ino)` verification would pass on exactly the sequence it is meant to catch, and "fail closed when ownership cannot be verified" would never fire. Beyond that, any such check is itself check-then-act: the swap can land between the verification and the rename. `RENAME_NOREPLACE` is the primitive that closes this, and this branch exists precisely because the filesystem does not have it.

### What the exposure actually is

Bounded to an empty directory created inside that window. POSIX `rename` only removes an *empty* destination, so a racing writer who put anything in the path keeps it:

```console
non-empty racing dir -> errno 39 Directory not empty; their file intact: theirs
```

I pinned that boundary as a regression test rather than leaving it to argument — `test_directory_commit_never_destroys_a_racing_writer_with_content` fills the reservation from inside `os.mkdir` and asserts `ENOTEMPTY`, the operator's file surviving, and the staged tree unpublished. 13 focused tests, `278 passed` across `test_codex_artifacts.py` + `test_codex_command.py`.

I also corrected the docstring, which is where this genuinely was wrong: it claimed both branches keep the guarantee atomic. That holds for files (`os.link`), and overstated the directory branch. It now states the gap, why it stays open, and the bound (`6802792a`).

### If that is not acceptable, the sound options are

1. **Accept and document** — what this PR now does. Residual risk: an empty directory created by a concurrent writer inside a sub-millisecond window on flagless filesystems only.
2. **Fail closed for directories** when `RENAME_NOREPLACE` is unavailable — a real guarantee, but it re-breaks all 22 skill directories on NFS, which is the bug this PR is fixing. Strictly worse for the reported user unless you want NFS unsupported.
3. **Serialize Ouroboros writers** with an `O_CREAT|O_EXCL` lock around the batch transaction. This is the one that actually fixes the realistic case — the concurrent writer here is another `ouroboros setup`/`refresh`, not an arbitrary process — but it is a change to the transaction design, not to a rename helper, and it belongs in its own issue rather than widening this fix. Happy to open it and implement if you want it.

Tell me which you prefer. If you would rather have (2) or (3) before merge, say so and I will do it; otherwise this is (1) with the limitation now documented in the code and covered where it can be covered.

bjkim95:
Blocker accepted and fixed in `3c4e83a3` — this one is genuinely different from the success-path ABA, and unlike that one it is fixable. Follow-up #1 (the staging alias) is fixed in the same commit.

### Blocker — cleanup deleted a pathname it did not own

Right: on rename failure the handler called `os.rmdir(target_path)` with no proof the reservation was still ours, so a writer who took the name over lost their directory for a publication that never happened.

An open descriptor **does** prove ownership here, which is what distinguishes this from the earlier inode discussion — a removed directory reports zero links locally and `ESTALE` on NFS, and neither depends on inode numbers:

```console
NFS  : untouched reservation -> still_ours=True   (want True)
NFS  : swapped reservation   -> still_ours=False  (want False)
local: untouched reservation -> still_ours=True   (want True)
local: swapped reservation   -> still_ours=False  (want False)
```

So the reservation is now opened at creation and released only while that descriptor still refers to a live directory. Anything that leaves ownership unproven — an unexpected `fstat` failure included — is treated as not ours and left alone.

### Where this lands, in both interleavings

I ran your probe in both orderings rather than only the favourable one:

```console
swap_in_open_to_rename_window: target_exists_after_cleanup=True    # fixed
swap_in_mkdir_to_open_gap:     target_exists_after_cleanup=False   # still exposed
```

The first is the window that matters: it spans the entire rename syscall, i.e. all of the elapsed time in the operation, and a takeover anywhere in it now survives. The second is the gap between `os.mkdir` returning and the `os.open` on the next line. Your probe swaps inside the patched `mkdir`, which models that gap as arbitrarily wide; in a real process it is two adjacent syscalls. I cannot close it: `mkdir` returns no handle, so there is no atomic create-and-hold for directories, and — as measured in my previous comment — an inode comparison cannot stand in for one. Stated plainly in the docstring rather than left for the next reviewer to find.

Regression test: `test_failed_publication_never_removes_a_reservation_taken_over` removes the reservation, puts a different writer's directory at the path, then fails the rename with `EXDEV`, and asserts the directory survives and the staged tree is untouched. It fails on `6802792a`.

### Follow-up #1 — the retained `.tmp` alias

Agreed, and thank you for the correction: I called it inert, and a hard link aliasing the live rule is not inert — refreshes would pin old inodes through it. `_commit_staged_artifact` now sweeps a staging name that survived its own cleanup failure, after the commit boundary, best-effort so it can never change the outcome. `test_failed_staging_cleanup_does_not_report_a_failed_commit` now asserts the rules directory contains exactly `ouroboros.md` — no stranded `.backup`, no surviving alias. That assertion also fails on `6802792a`.

### Checks

```console
$ ruff format src/ tests/                                   1396 files left unchanged
$ ruff check src/ tests/ --fix                              All checks passed!
$ mypy src/ouroboros                                        Success: no issues found in 613 source files
$ scripts/check-module-size.py --baseline-ref origin/main   OK (612 modules, cap 2000)
$ pytest tests/unit/test_codex_artifacts.py tests/unit/cli/test_codex_command.py -q   279 passed
$ pytest tests/unit/codex tests/unit/hermes tests/unit/skills -q                      132 passed, 2 skipped
$ pytest tests/unit/test_codex_artifacts.py -k RenameNoReplaceOnFilesystems -q        14 passed
```

Both new assertions fail against `6802792a` and pass on `3c4e83a3`.

If the residual `mkdir`-to-`open` gap is unacceptable for this boundary, the only construction that removes it is failing closed for directory publication on flagless filesystems — which returns NFS users to the broken state this PR is fixing. I would rather not make that trade silently, so it is your call; say the word and I will switch it, or open the writer-serialisation issue instead.

bjkim95:
Two of the five blockers were reproducible inside this PR's own new code and are fixed in `f9c9af1a`. The other three I would like to take as an owned follow-up rather than absorb here, and I have opened #2294 for them. Also merged current `main` — the branch was behind, and the one red check was not a test failure.

### Fixed — blocker 3, both halves

Both were real, both are in code this PR introduced, and both fail without the fix.

**`os.open` failing after `os.mkdir`** left the reservation sitting at the final name. The caller sees a failed publication, so the previous generation stays in its `.backup` sibling while the target is occupied by an empty directory no later replay can resolve — the exact failure this PR exists to remove. The reservation is now removed on that path. That removal is exposed to the `mkdir`-to-`open` window already documented in the helper and to nothing else: `rmdir` refuses a non-empty directory, so a racing writer who put anything there keeps it.

**`os.close` raising in `finally` after a successful `os.rename`** reported an already-published directory as a failed commit. You are right that `_commit_staged_artifact` has not set `staged_generation_active` at that point, so refresh would roll back over a live generation and strand the previous one. The rename is the commit boundary; the release is housekeeping and is now best-effort. That also stops it masking a cleanup error on the failure path.

```console
$ pytest tests/unit/test_codex_artifacts.py -q -k RenameNoReplaceOnFilesystems
17 passed, 46 deselected in 15.56s

$ git checkout upstream/main -- src/ouroboros/codex/artifacts.py   # nothing else touched
$ pytest tests/unit/test_codex_artifacts.py -q -k RenameNoReplaceOnFilesystems
14 failed, 3 passed, 46 deselected in 2.41s
```

The two new cases — `test_unpinnable_reservation_does_not_occupy_the_final_name` and `test_release_failure_after_the_rename_keeps_the_publication_committed` — also fail against `dac4d2f3` specifically, so they pin the fix and not just the file's absence. `test_codex_artifacts.py` + `test_codex_command.py` + `test_uninstall.py`: 321 passed. Ruff, format, and mypy clean; module-size, auto-boundary, and max-turns-envelope gates OK.

### Blocker 2 — the identity-check-to-`rmdir` interval

Reproducible, and I am not disputing it. But it is not closable in this branch, and I would rather not ship a check that advertises a guarantee it cannot deliver — the same conclusion the reservation-inode measurement reached earlier in this thread.

There is no `rmdir` by descriptor. Any ownership proof for a directory is therefore a pathname operation followed by another pathname operation, and the window between them is exactly what you describe. It is the failure path *inside* an already-failed publication, and it is bounded the same way everything else in this branch is: `rmdir` removes only an empty directory, so a racing writer with any content is never touched. `test_directory_commit_never_destroys_a_racing_writer_with_content` pins that bound.

Closing it needs an atomic no-replace primitive, which is by definition what this code path does not have. I have stated it in the helper's docstring and in the PR's declared non-goals rather than leave it for the next reader to find.

### Blockers 1, 4, 5 — real, and outside this PR's boundary

I have now declared the boundary explicitly in the PR description, as [CONTRIBUTING.md#review-boundary-contract](https://github.com/Q00/ouroboros/blob/main/CONTRIBUTING.md#review-boundary-contract) requires. Against it:

| Blocker | Reproduces on `main` with this PR reverted? | In a file this PR opens? |
|---|---|---|
| 1 — crash/replay after backup + reservation | Yes; the `.backup` scheme has no durable intent today | Yes, but the property is pre-existing |
| 4 — parent authority; `uninstall.py:155` discovery-to-delete | Yes | No (`cli/commands/uninstall.py`) |
| 5 — document the full shared-directory threat model | n/a — documentation of the subsystem, not of this change | Partly |

Running the five-question rubric: Q1 no for 4 (it does not reproduce under this PR's declared inputs — it is not the flag-rejection path), Q3 yes and Q4 yes for all three (the original user problem is solved without the new subsystem — that is what the 14 failing tests above show), Q5 no (no immediate user-data risk that this PR introduces; the risks named are on `main` with or without it). That row of the table is **owned follow-up**, so I have opened one: **#2294**, with the reproductions, the affected call sites, and what I would need decided.

On the Design Notes specifically. Extracting `core/managed_artifact_transaction.py` and migrating publication, backup acquisition, commit disposal, rollback restoration, prune discovery and removal, staging and failure cleanup, and `_remove_codex_artifacts` is close to what #2276 set out to build as `core/fs_ownership.py`. That closed unmerged on 2026-08-27 and `core/fs_ownership.py` is not on `main`. I do not think a contributor should start that by inferring the target design from a review comment; it needs a maintainer decision on shape and sequencing first, which is what #2294 asks for. I am happy to take a slice once it exists — the natural first one is `hermes/artifacts.py` and `cli/runtime_activation.py`, where this PR's degradation behavior applies mechanically.

Also taking the non-blocking suggestion: `test_directory_commit_refuses_a_target_that_appears_after_the_check` is misnamed for the current reservation design and I will rename it in the next push.

### The red check is not a test failure

`Test Python 3.12` on `6a80b31e` reached `[ 99%]` at `17:29:48` with no `F` in the output, produced nothing for the next 22 minutes, and was killed by the job's 30-minute cap at `17:51:58`. It hung; no test failed. The preceding commit `dac4d2f3` was green on the same suite, and `6a80b31e` was a `main` merge with no source delta of mine. The branch has been merged up to `main` again, so this is re-running.

@Q00 — this is the sixth review round and the first three blockers now ask for a subsystem. Could you weigh in on whether #2294's scope is the right shape, and whether this PR can land on its declared boundary? Per the new contract that call is yours, and I do not want to keep expanding against a target I am guessing at.

bjkim95:
CI is green on `648655e2` — all twelve checks, `Test Python 3.12` in **7m4s**.

Worth recording why it was red before: on `6a80b31e` that job reached `[ 99%]` at `17:29:48` with no `F` anywhere in its output, then produced nothing for 22 minutes and was killed by the job's 30-minute cap. It hung; no test failed. `dac4d2f3` was green on the same suite, and `6a80b31e` was a `main` merge with no source delta of mine.

I have also replaced the full-suite evidence in the description, because the old numbers were stale in a way that mattered. Re-running the CI invocation on a 20-core node:

```console
53 failed, 21878 passed, 13 skipped, 1 xfailed in 227.23s (0:03:47)
```

Not 553. This host now ships git 2.55.0, so the `project_identity` cluster that dominated the earlier rounds is gone. The remaining 52 are shell/PATH — `auto/test_seed_preflight.py` and `orchestrator/test_runtime_evidence.py` shelling out to `sh -c` / `env -C`, with `sh: 1: python: not found` in captured stderr.

The reverted-file comparison over the 11 files holding those failures differed by exactly one test, and I chased it rather than rounding it off: `test_capture_rejects_intermediate_workspace_ancestor_swap` is flaky, and flaky **on `main`** — 8 isolated runs with `codex/artifacts.py` checked out from `main` gave 7 passed, 1 failed. It also passed in the full run and passes in CI. `orchestrator/parallel_executor.py` has no reference to `codex`, and that test's adversarial `os.open` counter only matches a `dir_fd`-relative open of its own ancestor name, which is not a path this PR opens.

Also renamed `test_directory_commit_refuses_a_target_that_appears_after_the_check` to `..._admits_an_occupied_name_only_through_the_reservation` in `648655e2`, per the non-blocking suggestion.

Q00:
@bjkim95 Hi, thank you for your dedicated contribution! 
As reviewbot says, we have to step back and watch these whole changes on this PR. The issue intent is making 'ouroboros setup' install its rul file and skill directories. 

If you changes several subsystem and file system, it can be exposed many security issues. So can you reset this direction and discuss what the right direction is in discord? 

We welcome this discussion! https://discord.gg/nUzUrhSdQ

bjkim95:
@Q00 Thank you — and yes, let's step back. I have stopped pushing to this branch until we agree on a direction.

Before the discussion, one correction and one admission, because I think they point at different things.

### The diff is one source file

```
src/ouroboros/codex/artifacts.py   | 139 +++-      <- the only source file
tests/unit/test_codex_artifacts.py | 419 +++
2 files changed, 551 insertions(+), 7 deletions(-)
```

No other subsystem is touched. The review that ran alongside your comment says the same thing in its own words — *"Limit implementation to the Codex artifact installer | Satisfied"*, *"Scope expansion: none"*. The lifecycle, parent-authority, and uninstall material that makes this thread look wide is discussion, not code; I moved all of it out to #2294 rather than implement any of it here.

### But you are right about the accretion, and the commit history shows it exactly

```
858135ae  +31 / -7    the actual fix: treat EINVAL/ENOSYS/EOPNOTSUPP as "flag not implemented",
                      fall through to a portable path
efbbcbd3 .. f9c9af1a  +145        ownership machinery, added one blocker at a time
```

The bug I opened #2273 for is 31 lines. Everything after it — an `mkdir` reservation, a pinned descriptor, `(st_dev, st_ino)` identity comparison, `_release_directory_reservation` — exists because each review round found a way a hostile concurrent writer could exploit the previous round's answer, and I kept answering. That is the filesystem-level complexity you are reacting to, and I should have stopped and asked several rounds earlier instead of building it. That one is on me.

### Why it could not converge

The requirements arrived in a pair that cannot both hold:

- Rounds 3–8: **never remove a pathname whose ownership you cannot prove.** So cleanup returns without touching anything when inspection fails.
- Round 9: **never leave the final name occupied after a failed publication.** That same early return is now the blocker.

The round 9 finding is real — I reproduced it on current HEAD. Force the unsupported-flag path, make `os.rename` fail `EXDEV`, and make `os.stat(target)` fail `ESTALE` (plausible on NFS, which is the filesystem this PR is about):

```console
raised: EXDEV
target still occupied : True | empty: True
source unpublished    : True
```

It violates a guarantee I wrote into this PR's own contract, so I am not disputing it. But without `RENAME_NOREPLACE` there is no primitive that satisfies both halves. When ownership is unprovable you must choose which one to give up, and that is a product decision about the threat model, not something I should decide inside a compatibility fix.

### Three directions, from smallest

| | Shape | Concurrency stance |
|---|---|---|
| **A** | Revert to `858135ae` — capability detection plus the portable path, ~31 lines. Directories publish by `mkdir` reservation then `rename`; on failure, best-effort `rmdir`. | Adversarial concurrent writers in `~/.codex` are an explicit non-goal, stated in the docstring. |
| **B** | A, plus letting `_commit_staged_artifact` reconcile a reservation it could not release, since it is the component that holds the `.backup` generation. | Closes the round 9 case specifically; still no ownership protocol. |
| **C** | What is on the branch now. | Tries to survive a hostile writer; that attempt is what grew to +145 lines and is still incomplete. |

My own preference is **A**, possibly **B**. `~/.codex` is the operator's own directory, `setup` is interactive and not concurrent with itself, and I do not think this PR should be where Ouroboros decides how to defend a shared artifact directory. If that defense is wanted, it deserves its own design — which is what #2294 is for — rather than being reverse-engineered inside a fix for an NFS `errno`.

Whichever you pick, I will rewrite the branch to it. If A or B, the test file shrinks with it: several of the 419 test lines only pin behavior of the machinery we would be removing.

Happy to take this to Discord — say the word and I will bring the same summary there. And if the answer is that the whole approach is wrong and `setup` should handle flagless filesystems some other way entirely, I would rather hear that now than defend this one.

shaun0927:
Direction/log update for PR #2274 at exact HEAD `5e5913dbff10ca492d212a0f68369f0890fac506`.

Paperthin route applied: `hate` narrowed the load-bearing objection to the unsupported-flag directory publication rollback invariant: after a reported failed directory publication, the previous generation must be active and no installer-owned final-name reservation or `.backup` may remain. `macrothink` kept this scoped to the Codex artifact installer instead of expanding into shared lifecycle/uninstall/crash-replay work.

What changed:
- Added a regression that drives the real `install_codex_skills` refresh path with `renameat2 -> EINVAL`, staged directory publication `os.rename -> EXDEV`, and first reservation inspection `os.stat(..., follow_symlinks=False) -> ESTALE`.
- `_commit_staged_artifact` now records whether the staged path was a directory and, when a failed directory publication leaves an empty final-name reservation while the prior generation backup is still held, removes that empty reservation and restores the backup instead of treating the reservation as a concurrent generation.
- The fix stays inside `src/ouroboros/codex/artifacts.py` plus `tests/unit/test_codex_artifacts.py`.

Verification on the pushed head:
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest tests/unit/test_codex_artifacts.py tests/unit/cli/test_codex_command.py -q` -> `283 passed`
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff check src/ouroboros/codex/artifacts.py tests/unit/test_codex_artifacts.py` -> passed
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff format --check src/ouroboros/codex/artifacts.py tests/unit/test_codex_artifacts.py` -> `2 files already formatted`
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run mypy src/ouroboros/codex/artifacts.py` -> no issues
- Manual unsupported-flag `install_codex_skills` smoke refreshed a temp skill directory and left no `.backup` sibling.

Note: running the same focused suite without clearing `CODEX_HOME` produced unrelated default-Codex-home expectation failures because this Codex session exports `CODEX_HOME=/Users/jh0927/.codex-isolated/...`; the clean run above uses `env -u CODEX_HOME` to match the tests' default-home assumptions.

Not run: full repository suite.

@ouroboros-agent please re-review exact HEAD `5e5913dbff10ca492d212a0f68369f0890fac506`.

shaun0927:
Follow-up direction log for exact HEAD `12d5a4f669c6477059ce373bf1030c9948f3de93`.

CI follow-up: GitHub `Test Python 3.12` failed at prior HEAD `5e5913dbff10ca492d212a0f68369f0890fac506` in `test_refresh_failed_directory_publication_restores_previous_generation`. The local reproduction showed the test injected `ESTALE` during the pre-mutation fingerprint of the existing installed generation, before the fallback directory publication reached the simulated `EXDEV` failure.

What changed:
- Narrowed the regression’s injected stale `os.stat(..., follow_symlinks=False)` to the empty final-name reservation state, after the existing generation has been moved to backup.
- No production behavior changed in this follow-up commit.

Verification at this HEAD:
- `env -u CODEX_HOME uv run --python 3.12 pytest -q tests/unit/test_codex_artifacts.py::TestRenameNoReplaceOnFilesystemsWithoutTheFlag::test_refresh_failed_directory_publication_restores_previous_generation -vv` -> passed
- `env -u CODEX_HOME uv run --python 3.12 pytest -q tests/unit/test_codex_artifacts.py tests/unit/cli/test_codex_command.py` -> `283 passed`
- `uv run ruff format --check tests/unit/test_codex_artifacts.py` -> already formatted
- `uv run ruff check tests/unit/test_codex_artifacts.py` -> all checks passed
- `uv run mypy src/ouroboros/codex/artifacts.py` -> success, no issues

Not tested: full repository pytest suite locally after this test-only correction. GitHub checks for this new HEAD are pending.

@ouroboros-agent please re-review exact HEAD `12d5a4f669c6477059ce373bf1030c9948f3de93` for PR #2274. This follow-up addresses the failed CI regression targeting error; no merge requested.

shaun0927:
Paperthin re0 refresh for exact HEAD `7f582e0f1be562a9a1a0218924d6608fa067dda2`.

Decision: keep this PR, but keep it narrow. The compatibility fix is still needed for filesystems that reject `RENAME_NOREPLACE`, and the branch should remain scoped to Codex artifact installer rollback behavior rather than becoming a general shared-directory publication design.

What changed:
- Merged current `upstream/main` into `fix/rename-noreplace-unsupported-flag-fallback` without conflicts.
- Preserved the narrow PR diff: `src/ouroboros/codex/artifacts.py` and `tests/unit/test_codex_artifacts.py`.
- Amended the merge commit into Lore format with the unsupported-filesystem fallback boundary recorded.

Verification at this HEAD:
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run --python 3.12 pytest -q tests/unit/test_codex_artifacts.py tests/unit/cli/test_codex_command.py` -> `283 passed`
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff format --check src/ouroboros/codex/artifacts.py tests/unit/test_codex_artifacts.py` -> `2 files already formatted`
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run ruff check src/ouroboros/codex/artifacts.py tests/unit/test_codex_artifacts.py` -> all checks passed
- `env -u CODEX_HOME SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run mypy src/ouroboros/codex/artifacts.py` -> success, no issues
- Manual `install_codex_skills` smoke in a symlink-free temporary Codex home installed `ouroboros-probe`, refreshed it from v1 to v2, and left `leftovers=[]` for backup/tmp siblings.

Not tested: full repository pytest suite locally after the main refresh. GitHub checks should cover the full PR gate on this pushed head.

@ouroboros-agent please re-review exact HEAD `7f582e0f1be562a9a1a0218924d6608fa067dda2` for PR #2274. The branch is refreshed on current main and remains a narrow unsupported-flag fallback/rollback fix. No merge requested.

shaun0927:
Bot re-review request for exact HEAD `7f582e0f1be562a9a1a0218924d6608fa067dda2`.

Direction: KEEP, but keep narrow. This PR is still needed for filesystems that reject `RENAME_NOREPLACE`; it should remain confined to Codex artifact installer publication/rollback behavior.

Current verification: all PR checks are green/skipped, including `Test Python 3.12`, `Ruff Lint`, `MyPy Type Check`, `enforce-boundary`, `enforce-envelope`, `enforce-module-size`, `enforce-perf-budget`, `Bridge TypeScript`, and `Native TUI (Rust)`.

The latest ouroboros-agent[bot] review is stale: it requested changes on `648655e2b406d41c1901127494b5fe243295a1f5`, not this exact head. Please run a fresh exact-head review.

shaun0927:
Synchronize trigger log for exact HEAD `6bd1443610a9620fb3299c68bd640126a0c0b705`.

Direction: KEEP, but keep narrow. This empty Lore commit intentionally changes no tree content; it exists to force a GitHub `synchronize` event after comment, label, and reviewer-request metadata did not refresh the stale bot review.

Expected verification path: CI is running again on this exact head, and a fresh ouroboros-agent[bot] review should bind to this SHA. Prior tree-equivalent head was green/skipped.

shaun0927:
@ouroboros-agent please re-review exact HEAD `6bd1443610a9620fb3299c68bd640126a0c0b705`.

This is the current paperthin readchk/re0 checkpoint: CI is green/skipped, stale reviews tied to older commits have been dismissed where applicable, and this request is only for exact-head approval or current-head change requests. No merge is requested.

Q00 (approved):
@bjkim95 this contribution is one of difficult problem which access the file system. thanks a lot. I will take a follow PR

PR #2266, opened 2026-08-23
fix(gjc): make runtime lifecycle ownership-safe

Q00:
## Why

GJC setup could report success after MCP registration failed, overwrite a user-managed bridge config, and remove the legacy bridge during refresh without first restoring the replacement MCP path. Uninstall also left the managed routing guide behind, while broad launcher matching could claim user-owned MCP entries.

## Benefits

- Makes GJC setup failure-atomic and propagates failure instead of printing false success.
- Registers and validates replacement MCP routing before removing the legacy bridge.
- Preserves custom `mcp-bridge.yaml` and removes only exact setup-owned MCP and routing artifacts.
- Adds focused regressions for setup failure, refresh migration, uninstall cleanup, and ownership boundaries.

## Validation

- `uv run --group dev pytest tests/unit/gjc/test_gjc_artifacts.py tests/unit/cli/test_setup.py -k "Gjc or gjc" tests/unit/cli/commands/test_setup_refresh.py -k "Gjc or gjc" tests/unit/cli/test_uninstall.py -k "Gjc or gjc" -q` — 28 passed
- `uv run --group dev ruff check ...` — passed for all changed Python files
- `uv run --group dev ruff format --check ...` — passed for all changed Python files
- `uv run --group dev mypy src/ouroboros/cli/gjc_setup.py src/ouroboros/cli/commands/setup.py src/ouroboros/cli/commands/setup_refresh.py src/ouroboros/cli/commands/config.py src/ouroboros/cli/commands/uninstall.py src/ouroboros/runtime_instruction_artifacts.py` — passed

Refs #2243

PR #2243, opened 2026-08-23
fix(gjc): route ooo commands through native skills

Q00:
## Summary

- Replace the executable GJC input bridge with namespaced `ouroboros-*` skills and native MCP registration.
- Add an always-applied exact-command routing guide so explicit `ooo <command>` input wins over generic GJC workflow routing.
- Make setup, refresh, and uninstall own the full GJC projection lifecycle while preserving user-managed files and MCP configuration.

## Test plan

- [x] `uv run --group dev pytest tests/unit/gjc/test_artifacts.py tests/unit/cli/commands/test_setup_refresh.py tests/unit/cli/test_setup.py tests/unit/cli/test_uninstall.py tests/unit/test_runtime_instruction_artifacts.py -q` — 497 passed, 1 skipped
- [x] `uv run --group dev ruff check src/ouroboros/gjc/artifacts.py tests/unit/gjc/test_artifacts.py`
- [x] `uv run --group dev ruff format --check src/ouroboros/gjc/artifacts.py tests/unit/gjc/test_artifacts.py`
- [x] Interactive GJC smoke: `ooo interview I want to make todo task cli` reached the Ouroboros interview and emitted its first question
- [ ] Full repository test and mypy suites not run

## Related issues

N/A

PR #2276, opened 2026-08-25
feat(gjc): consolidate ownership-safe lifecycle

Q00:
## Summary

Consolidate the GJC native-skill projection and MCP lifecycle into one canonical PR.

This supersedes #2243 and #2266 instead of merging their overlapping histories.

- Projects namespaced Ouroboros skills with generation ownership.
- Installs exact routing/bridge artifacts without overwriting operator-owned files.
- Validates MCP registration and performs a read-only endpoint health check before retiring the legacy route.
- Makes setup/refresh rollback ownership-aware.
- Makes uninstall best-effort and discovers orphan managed bridge configuration.
- Preserves the compatibility bridge only when native autoload is unavailable.

## Architecture

GJC domain logic lives in one package, orchestration in one CLI module:

- **`ouroboros/gjc/`** — single home for GJC domain logic (pure: no printing, no subprocess).
  - `paths.py`: every canonical artifact location (agent dir, rules guide, `mcp.json`, bridge config, compatibility bridge).
  - `artifacts.py`: namespaced skill projection with tree-digest generation ownership.
  - `mcp.py`: MCP registration ownership judgment and the atomic setup-owned removal primitive.
  - `bridge.py`: compatibility-bridge source rendering and generation judgment.
  - `guide.py`: routing-guide rendering and generation judgment.
- **`cli/gjc_setup.py`** — the one orchestration module: narration, GJC CLI invocation, and the snapshot/rollback transaction. Setup-owned filesystem/launcher seams are injected once through a `GjcSetupHost` dataclass instead of per-callable parameters.
- **`cli/stdio_mcp_probe.py`** — the dependency-free stdio MCP health probe, shared by `codex doctor` and GJC endpoint validation (no cross-command import between codex and gjc).
- **`core/fs_ownership.py`** — runtime-agnostic claim-then-verify primitives (`publish_owned_file`, `publish_owned_tree`, `publish_owned_entry`, `claim_and_remove_owned`, `recover_owned_claims`): every destructive or publishing boundary claims the existing entry by atomic rename, re-validates the claimed generation with inode-identity binding through the deletion itself, never follows symlinks (including a caller-declared trusted root, which is opened `O_NOFOLLOW`), and treats claim names as durable self-describing intent records — every transaction first reconciles orphans a crashed run left behind, adapting the hermes journaled-intent design without separate journal files. GJC rollback restores snapshots through `publish_owned_entry` (no check-then-restore gap), and MCP failure cleanup binds its removal to the exact observed entry generation under the registration lock. Full consolidation of `hermes/artifacts.py` and `codex/artifacts.py` onto these primitives remains a documented follow-up.

`setup.py` keeps only the host binding (`_gjc_setup_host`/`_setup_gjc`/`_install_gjc_runtime_artifacts`); uninstall and refresh import ownership judgment directly from `ouroboros.gjc` with no private cross-module imports.

## Validation

- `ruff format` / `ruff check src/ tests/`
- `mypy src/ouroboros`: clean
- module-size ratchet: OK
- `pytest tests/unit`: 21k+ passed; the only local failures are machine-load-dependent perf-budget tests in untouched persistence/orchestrator modules
- Deterministic race regressions: claim/recreate/no-clobber and symlink publication at the primitive level, symlinked-trusted-root rejection at every GJC boundary, interrupted-claim crash recovery (primitive, skills, guide, uninstall, refresh discovery), claim-identity swap refusal, rollback races against operator generations, generation-bound MCP failure cleanup with a non-cooperating writer, a lock-held proof for the MCP add transaction, and operator-swap preservation across skill publish/prune/remove and uninstall

## Supersedes

- #2243
- #2266


Q00:
This PR replaces #2243 and #2266 with one current-main implementation. Focused GJC lifecycle tests, Ruff, MyPy, and module-size pass locally. Please review the consolidated ownership/activation/uninstall contract.

Q00:
Addressed all three blockers in 81fa916f.

**Blocker 1 — check-then-delete races (`artifacts.py`, uninstall):** added `ouroboros/gjc/fs.py` with a claim-then-verify primitive: every destructive boundary now atomically renames the artifact aside, re-validates the *claimed* generation, and deletes it or restores it untouched. Applied to skill publication (the backup claim is re-validated before `os.replace`), pruning, `remove_gjc_skills`, `remove_legacy_gjc_bridge`, and uninstall's guide/bridge-config/compatibility-bridge removal. Deterministic race regressions cover publish, prune, remove, and uninstall by giving the initial ownership check a stale observation for the canonical path while the claimed generation is judged by the real check.

**Blocker 2 — failed-add ownership attribution (`gjc_setup.py`):** `register_gjc_mcp_server` snapshots the durable pre-add generation. All failure branches (exception, nonzero exit, malformed receipt) clean up only when no setup-shaped registration existed before the add, and `registration_state.created` / rollback now mean created-by-this-invocation. Regression: pre-existing canonical persisted entry + list omission + failed add → `remove_persisted_gjc_mcp_server` is never called.

**Blocker 3 — refresh discovery conflates occupancy with ownership (`setup_refresh.py`):** discovery now requires an intact setup-managed compatibility bridge (`is_setup_managed_gjc_bridge`) — and, for consistency, a setup-managed routing guide — matching the ownership-gated discovery already used for skills, bridge config, and MCP state. Regression: a custom extension at the canonical bridge path on a native host triggers no projection; the legacy-bridge tests now use a real setup-rendered generation.

Validation: ruff/mypy clean, `pytest` over the GJC lifecycle suites + setup + codex: 813 passed locally including the 6 new race/collision regressions.

Q00:
Addressed the three second-round blockers in fd7b7b38, and promoted the primitives out of the GJC package per the duplication concern.

**Blocker 1 — publication validates separately from the final write:** publication now goes through `ouroboros/core/fs_ownership.py::publish_owned_file`, which claims any existing entry by atomic rename, re-validates the *claimed* generation, and publishes via `os.replace` of a same-directory temp file — the final rename never follows symlinks, so a concurrently inserted link cannot route the write to its target. Applied to the MCP bridge config, the compatibility bridge, and the routing guide (replacing the `atomic_write_text`-without-`expected_current` and `Path.write_text` paths). Per-artifact symlink regressions cover all three; primitive-level tests cover the concurrent-file case.

**Blocker 2 — unconditional claim restoration clobbers a recreated generation:** `restore_claimed` is now no-clobber: when the canonical path was recreated while the claim was held, both generations are preserved (the recreated entry stays canonical; the claimed one remains beside it). `claim_and_remove_owned`, `publish_owned_file`, and `_publish_skill`'s backup restoration all use it. Regressions recreate the canonical path *after* the rename, at the primitive level and through skill publication.

**Blocker 3 — failed-add cleanup attribution under concurrency:** the complete add/validate/cleanup transaction now runs under a cross-process lock on the durable registration (`gjc_mcp_registration_lock`, `file_lock` over GJC's `mcp.json`), with in-transaction removals using a lock-held variant. A concurrent Ouroboros setup/refresh invocation can no longer interleave, so a setup-shaped entry observed during the window either predates the add (`pre_add_owned`, never removed) or was produced by this invocation's own `gjc mcp add`. A regression proves the lock is held during the add (a non-blocking acquire inside the add raises `BlockingIOError`), alongside the existing pre-existing-entry preservation test.

On the structural side: the claim-then-verify primitives moved from `ouroboros/gjc/fs.py` to `ouroboros/core/fs_ownership.py` (runtime-agnostic, generic names), and the module documents `hermes/artifacts.py` (journaled swaps) and `codex/artifacts.py` (fingerprint-gated replacement) as existing siblings and follow-up consolidation candidates — intentionally untouched in this PR to keep the blast radius bounded.

Validation: ruff/mypy clean; affected suites (core fs, gjc, setup, refresh, uninstall, codex, runtime instruction artifacts) 791 passed locally.

Q00:
Addressed all four third-round blockers in fcacf79a by finishing what they collectively point at: `ouroboros/core/fs_ownership.py` is now the single hardened boundary for every operator-shared artifact mutation, with an explicit threat model in the module docstring.

**Blocker 1 — publication clobbers a generation recreated after validation:** the final publish rename is now atomic no-replace — `renameat2(RENAME_NOREPLACE)` on Linux, `renameatx_np(RENAME_EXCL)` on macOS, native no-replace `rename` on Windows, existence-guarded rename only as a last resort on platforms with neither. A recreation after `is_owned` (or on an initially absent path) makes the publication fail with `UnownedArtifactError`; the recreated entry stays canonical and the claimed generation is preserved beside it. Regressions recreate the target inside the ownership callback for both files and trees.

**Blocker 2 — symlinked parent components:** mutations now run relative to a pinned directory descriptor. The chain from a caller-supplied trusted ancestor down to the artifact's parent is opened one component at a time with `O_NOFOLLOW`, directories are created with `mkdir(dir_fd=...)`, and temp files with dir-fd `O_CREAT|O_EXCL` — so `<agent>/ouroboros -> /external` can redirect neither publication, removal, nor temp-file creation. GJC callers anchor at the agent profile root (which may itself legitimately be a symlink, e.g. `GJC_CODING_AGENT_DIR`). Regressions cover publish and removal through a symlinked component at the primitive level plus the reviewer's exact bridge-config scenario.

**Blocker 3 — `_publish_skill`'s independent claim/validate/replace:** eliminated. `publish_owned_tree` absorbs the sequence — the skill publisher only builds the staged generation (copy, render, digest); claiming, validation, no-replace publication, and no-clobber restoration are the primitive's. A regression recreates an empty directory at the canonical path after backup validation and proves the publication fails with both generations preserved.

**Blocker 4 — uninstall drops bridge-config-only failures:** the failed-cleanup gate matched only labels starting with `GJC Ouroboros` while the bridge-config target reads `GJC MCP bridge config (...)`; it now matches every GJC target. A CLI regression with only a managed bridge config present and a failing removal asserts exit code 1 and the partial-removal summary.

Validation: ruff/mypy clean, module-size ratchet OK, affected suites (core fs-ownership, gjc, setup, refresh, uninstall, codex, runtime instruction artifacts) 836 passed locally including the new regressions.

Q00:
Addressed all three fourth-round blockers in 4c19a9ac.

**Blocker 1 — staging failure strands the claimed generation:** `_publish_owned` now wraps the build step (temp-file creation, write, fsync, tree build callback) and restores the claimed generation before propagating any failure, so a failed publication can no longer leave the canonical path absent with the old generation hidden under a `.replacing` sibling. Regressions inject a file-staging failure and a tree-build failure after a pre-existing generation has been claimed and assert the canonical content is restored with no claim siblings left behind.

**Blocker 2 — removal failure leaves an undiscoverable `.removing` sibling:** `claim_and_remove_owned` restores the claim when `remove_path` raises, so the generation stays at its canonical path where refresh/uninstall discovery can still see it, and the error propagates for the caller's failure reporting. Regression injects one transient removal failure, asserts the canonical route survives, and proves the retry succeeds.

**Blocker 3 — validation and retirement are not one transaction:** `register_gjc_mcp_server` gained `assume_registration_lock`, and `install_gjc_runtime_artifacts` now holds the registration lock across the complete sequence: validate-or-add, a final re-validation of the persisted durable registration, and only then legacy-bridge retirement. The pre-existing-entry validation path also runs under the lock. A concurrent uninstall (which takes the same lock for its durable removal) can therefore no longer delete the registration between validation and retirement and leave neither route. Regression simulates the disappeared registration at the final validation and asserts retirement is never invoked and the compatibility bridge is kept.

Validation: ruff/mypy clean; affected suites 802 passed locally including the four new regressions.

Q00:
Addressed both fifth-round blockers in eeef14de.

**Blocker 1 — pathname escape after the claim:** every post-claim operation is now descriptor-bound or identity-bracketed. The claimed-entry symlink check, deletion (`unlink`/`rmtree` via `dir_fd`), staging-directory creation (`mkdir` via `dir_fd`), staging cleanup, and retired-claim removal run relative to the held parent descriptor; ownership validation — which necessarily reads through the pathname — is bracketed by parent-identity revalidation (`fstat` of the held descriptor against the canonical path) before and after, and publication re-validates once more immediately before the final no-replace rename. Your probe's scenario (parent renamed away, symlink planted at the canonical path, external file seeded under the claim name) now fails closed: the external file is never read or deleted, and the claimed generation is restored into the real, moved directory through the descriptor. Regressions reproduce the probe exactly, plus a plain post-claim parent rename and a parent swap during a tree build.

**Blocker 2 — `_atomic_replace_json` compare-then-replace:** the durable MCP config rewrite is now a claim-and-verify compare-and-swap through the hardened primitive: `mcp.json` is claimed, the claimed generation is re-validated against the exact raw text that was read, and the new document is published with a no-replace rename (`publish_owned_file` with `require_existing`). A write landing between the read and the rewrite — from GJC itself or an operator, which the registration lock cannot serialize — is preserved and the mutation fails with "changed concurrently" instead of overwriting concurrent servers, disabled-server state, or other operator configuration. Regression injects a concurrent write immediately after generation validation and asserts the removal rewrite preserves it.

Validation: ruff/mypy clean; affected suites 806 passed locally including the four new regressions.

Q00:
Round-7 blockers are addressed on the new head. Blocker → fix mapping:

**1. Symlinked trusted root (`fs_ownership.py:310`)** — `_pinned_parent` now opens a caller-declared trusted root itself with `O_NOFOLLOW` (with an `lstat` guard on the non-`*at` fallback): a symlinked configured root such as a redirected `GJC_CODING_AGENT_DIR` is rejected with `ELOOP: refusing to operate under a symlinked trusted root` instead of silently redirecting mutations into the link target. Regressions reproduce the review's `agent -> external` probe at the primitive level (publish, tree publish, removal) and at every affected caller: skill projection (`test_install_rejects_a_symlinked_agent_root`), routing guide (`test_gjc_guide_publication_rejects_a_symlinked_agent_root`), MCP bridge config + compatibility bridge + legacy-bridge retirement (`test_gjc_install_steps_reject_a_symlinked_agent_root`), and uninstall (`test_uninstall_refuses_a_symlinked_agent_root`).

**2. Non-recoverable / pathname-only claims (`fs_ownership.py:501`)** — two parts, adapting the hermes journaled-intent design as suggested rather than adding a second journal format:
- *Crash recovery:* a claim name is now treated as a durable self-describing intent record (`.{name}.{nonce}.{removing|replacing}`). New `recover_owned_claims` / `find_orphaned_claims` reconcile orphans — a claim whose canonical path is absent is restored no-clobber so it stays discoverable; a leftover claim beside an occupied canonical path is deleted only when the ownership predicate confirms it. Every publish/remove transaction runs this reconciliation first (self-healing), and discovery now surfaces the state: refresh treats orphaned claims as installed state (`has_orphaned_gjc_claims`, `test_orphaned_gjc_claim_state_triggers_refresh`), and uninstall reconciles before its ownership judgments (`test_uninstall_recovers_an_interrupted_guide_claim`, `test_remove_recovers_an_interrupted_skill_claim`). The reviewer's exact interrupted state (canonical absent, random claim present) is reproduced and recovered in `test_recover_owned_claims_restores_an_interrupted_claim` and `test_every_transaction_reconciles_prior_interrupted_claims`.
- *Inode attribution:* the claimed entry's identity (dev, inode, file type) is captured immediately after the claim rename, re-checked after the ownership predicate, and enforced inside the deletion itself (`remove_entry(expected_identity=...)` raises instead of deleting a swapped entry). `test_remove_refuses_a_claim_whose_identity_was_swapped` swaps the claimed sibling mid-validation and proves the swapped generation survives. The final `unlink`/`rmtree` syscall is the platform's atomicity limit; the unpredictable claim name closes the remaining window (documented in the threat model).
- Parent-directory fsync now runs after claim, restoration, and publication (follow-up 1), with the durability boundary documented: recovery, not fsync, is what crash consistency rests on.

**3. Rollback outside the primitive (`gjc_setup.py:629`)** — `_restore_gjc_paths` no longer has a check-then-restore sequence. Each path's rollback runs through the shared primitives: the current entry is claimed and verified against the exact generation the transaction last wrote, and the snapshot is materialized in a staging sibling and published with a no-replace rename via new `fs_ownership.publish_owned_entry` (arbitrary topology: file, tree, or symlink); snapshot-was-absent paths go through `claim_and_remove_owned`. The review's probe — an operator replacement written between the judgment and the restore — is reproduced deterministically in `test_gjc_rollback_preserves_an_operator_generation_inserted_mid_restore`: the operator generation survives and the rollback is skipped with a warning.

**4. Failed-add cleanup attribution (`gjc_setup.py:323`)** — implemented as the locked compare-and-swap against the exact post-add generation. `gjc_mcp_entry_generation` produces a canonical generation token; cleanup removes an entry only when it (a) differs from the exact pre-add generation, (b) is setup-shaped, and (c) still equals — under the registration lock, backed by the whole-file CAS in `_atomic_replace_json` — the generation sampled for the cleanup decision (`remove_persisted_gjc_mcp_server[_locked](expected_entry_generation=...)`). The same generation token now also binds the receipt/validation rollbacks, `setup_gjc_runtime` rollback (`registration_state["entry_generation"]`), and uninstall's MCP removal to the exact observed generation. Failure-path tests cover a non-cooperating writer (`test_failed_add_cleanup_preserves_a_noncooperating_writers_registration`, `test_locked_removal_is_bound_to_the_observed_entry_generation`) alongside the real-file cleanup transaction (`test_failed_add_cleanup_removes_only_the_generation_from_its_own_window`). A byte-identical setup-shaped registration raced in by a non-cooperating writer inside the add window is indistinguishable by construction; this is documented at the boundary.

Non-blocking notes: the crash/partial-state and cleanup-attribution threat model is now part of the `fs_ownership` module docstring (suggestion 1). On suggestion 2, the warning text at `gjc_setup.py` contains the sentence "The GJC MCP registration disappeared before the legacy route was retired; kept the compatibility bridge." exactly once in the source (it is one string split across two source lines); no duplication was found to remove.

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; module-size ratchet OK; focused suites (core fs_ownership 27, gjc 12, uninstall 55, GJC setup 45, plus the runtime-instruction and refresh suites) and the full unit suite pass locally.

Q00:
Round-8 blockers are addressed in `6d18ed1d`. Both were correct catches — thank you for executing the identity-swap regression on an inode-recycling filesystem.

**1. Recovery promoted unauthenticated claims (`fs_ownership.py:454`)** — `_recover_claims` now authenticates ownership *while the entry is still under its claim name*, through a pinned entry descriptor, before anything is promoted or deleted. Only an owned claim is restored (no-clobber) when the canonical path is absent, or deleted as a leftover when the canonical path is occupied; a claim that fails authentication is left untouched as a collision. A forged `.index.ts.<nonce>.replacing` can therefore never be renamed into a live extension path. The suggested regression is added: `test_recovery_never_promotes_a_forged_claim` asserts a forged claim beside an absent canonical path is neither restored nor deleted — and that a subsequent publish proceeds without touching it. The follow-up documentation request is folded into the module threat model: claim-name syntax is discovery metadata, not ownership evidence, and forged claims are classified as collisions and left in place.

**2. Inode reuse defeated the identity tuple (`fs_ownership.py:184`)** — claim identity is now a held capability, not a reusable tuple. New `_PinnedParent.pin_entry` opens the claimed entry `O_NOFOLLOW` immediately after the claim rename, and the descriptor is held through validation and the final mutation (`claim_and_remove_owned`, `_publish_owned`'s claimed-generation cleanup, and recovery all use it). While the descriptor is open the filesystem cannot recycle the inode, so an unlink-and-recreate necessarily produces a different `(device, inode)` — the re-checks after the predicate and inside `remove_entry(expected_identity=...)` now detect every replacement, including the immediate-reuse case your execution reproduced. `test_remove_refuses_a_claim_whose_identity_was_swapped` should now hold on ext4/tmpfs as well; symlink entries are returned identity-only and rejected as before, and the non-`*at` fallback keeps the best-effort tuple (documented).

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; core fs_ownership suite 28 passed (including the updated interrupted-claim and new forged-claim regressions), gjc/runtime-instruction/refresh/uninstall suites 135 passed, GJC setup suite 45 passed.

Q00:
Round-9 blockers are addressed in `9f5f0c75`.

**1. Blocking pin on special files (`fs_ownership.py:240`)** — `pin_entry` now acquires the entry without blocking: `O_PATH` where the platform provides it (which also never triggers device side effects), `O_RDONLY|O_NONBLOCK` otherwise. Ownability is a type property enforced before any content read: only regular files and directories can pass (`_OWNABLE_ENTRY_TYPES`), so a FIFO, socket, or device planted at a managed path is classified and rejected before the ownership predicate runs — then claimed, refused, and restored like any other unowned generation. Your `os.mkfifo` probe now returns immediately with `UnownedArtifactError` (verified directly: pre-call marker followed by immediate rejection), and `test_special_file_at_canonical_path_is_rejected_without_blocking` additionally pins that the predicate is never invoked on a special file, for both publication and removal. All five shared entry points inherit the fix through the common pin.

**2. Syntax-only refresh discovery (`gjc/paths.py:65`)** — `has_orphaned_gjc_claims` no longer treats claim-name syntax as evidence. It moved into `gjc/artifacts.py` on top of the new read-only `fs_ownership.has_recoverable_claim`, which authenticates under the claim name exactly like recovery (pinned entry, ownable type before any content read, identity-bracketed predicate) while mutating nothing. Discovery now enumerates only the known GJC artifacts — routing guide, MCP bridge config, compatibility bridge, and `ouroboros-*` namespaced skills — each authenticated against its exact ownership predicate. Regressions: the orphan-refresh test now uses an authentic rendered managed generation (per the non-blocking suggestion), `test_forged_gjc_claim_does_not_trigger_refresh` proves a forged guide claim and an unrelated claim-shaped file in the agent root cause no activation and are left untouched, and `test_has_recoverable_claim_requires_authentication` covers the primitive read-only contract.

The module threat model documents the special-file contract (non-blocking pins, type-gated ownability). The follow-up on consolidating Hermes/Codex publication machinery onto this primitive remains documented as the next step after this PR.

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; core fs_ownership 30 passed, gjc/refresh/uninstall/runtime-instruction suites 138 passed, GJC setup suite 45 passed.

Q00:
Round-10 blockers are addressed in `1dece6ca`.

**1. Builder redirection (`fs_ownership.py:798`)** — builders no longer touch any pathname under the shared parent. `publish_owned_tree` and `publish_owned_entry` hand the builder a staging path inside a private `mkdtemp` workspace nobody else can reach; the finished generation is then imported beside the canonical path through descriptor-bound writes only — `os.rename(..., dst_dir_fd=parent_fd)` on the same filesystem, or a descriptor-relative recursive import (`_import_tree_fd`: regular files, directories, and symlinks; anything else refused) across filesystems. A canonical parent renamed away and replaced by a symlink *during* the build therefore cannot receive or redirect a single builder write — there is no longer any write for the swap to capture, rather than a post-hoc `revalidate` catching it after the fact. Your probe is reproduced with builders that write *after* the parent swap in `test_tree_builder_writes_cannot_escape_through_a_swapped_parent`, `test_entry_builder_writes_cannot_escape_through_a_swapped_parent`, and (for rollback restoration through `_restore_gjc_paths`) `test_gjc_rollback_restoration_cannot_escape_through_a_swapped_parent` — each asserts the external target stays empty, with no residue.

**2. Final-syscall binding (`fs_ownership.py:295`)** — implemented as the suggested quarantine: destruction of a validated entry never runs as a stat-then-pathname-unlink. `_PinnedParent.quarantined_remove` atomically renames the entry into a fresh, unpredictable `0700` quarantine directory created beside it, re-authenticates it *there* against the pinned identity, and destroys only a match. A replacement raced onto the claim name at the last instant is captured by the atomic rename, fails re-authentication, and is moved back out untouched under a discoverable name while the removal fails with `UnownedArtifactError`. Normal removal, orphan-recovery leftover deletion, and post-publication claim cleanup all route through it, so skill pruning, legacy retirement, uninstall, and rollback cleanup inherit the fix. Your deterministic interception is reproduced for both files and trees (`test_remove_never_deletes_a_replacement_swapped_at_the_last_instant`, `test_tree_remove_never_deletes_a_replacement_swapped_at_the_last_instant`): the replacement survives and the validated generation is never misattributed.

Follow-up 1 is folded in: the threat model's claim-identity section now describes the quarantine contract instead of "through the deletion itself", and a new "Builder redirection" section documents the private-workspace/descriptor-bound-import guarantee.

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; core fs_ownership 34 passed, gjc/refresh/uninstall/runtime-instruction suites 142 passed, GJC setup suite 46 passed.

Q00:
Round-11 blockers are addressed in `50116e47`.

**1. Trusted-root ancestry (`fs_ownership.py:559`)** — `_pinned_parent` now pins a caller-declared trusted root by walking *every* component of its absolute path from the filesystem root with `O_NOFOLLOW`, not just the final entry. Your `/profile-link/agent` probe is rejected with `refusing to operate under a symlinked trusted root` naming the offending component; a profile that legitimately lives behind a symlink must be configured by its resolved path (documented). `test_rejects_a_symlinked_ancestor_of_the_trusted_root` reproduces the probe for publication and removal and asserts the link target stays untouched.

**2. Quarantine-container identity and failure recovery (`fs_ownership.py:367`)** — the quarantine is no longer addressed by pathname after creation. It is opened `O_DIRECTORY|O_NOFOLLOW` immediately, the descriptor's `fstat` is validated against the created inode, and the claimed entry moves in via `dst_dir_fd` and is destroyed strictly relative to that held descriptor (`rmtree("entry", dir_fd=quarantine_fd)`), so a renamed-and-recreated container name can never receive the destruction — the only pathname operation left is a final `rmdir` that removes an *empty* directory and cannot destroy content. The failure path you reproduced is fixed the way you specified: a destruction that fails after the move restores the generation *from the quarantine location* through a cross-descriptor no-replace rename, so nothing is stranded in an undiscoverable `*.tmp/entry`; the container is removed on success and failure alike. `test_tree_remove_restores_from_the_quarantine_when_destruction_fails` forces the `rmtree` failure, asserts the canonical tree is restored with no `.tmp` residue, and proves the retry succeeds.

**3. Cross-filesystem import atomicity (`fs_ownership.py:347`)** — `import_entry`'s EXDEV branch now removes its own partial staging state (fd-relative) on every exception before propagating. `test_failed_cross_filesystem_import_leaves_no_staging_residue` forces `EXDEV`, fails the import on an unsupported staged entry, asserts the shared parent is completely clean, and then proves the descriptor-relative import path succeeds end to end.

**4. Descendant tree-generation attribution (`fs_ownership.py:656`)** — implemented as the requested re-authentication after isolation: `quarantined_remove` re-runs the caller's ownership predicate on the isolated entry inside the descriptor-pinned quarantine, bracketed by entry-identity re-checks and parent revalidation. A tree whose descendants were modified after the original ownership read — same top-level inode — now fails that re-authentication and is moved back out untouched while the removal fails. Every destructive caller (normal removal, recovery leftovers, post-publication claim cleanup) passes its predicate through, so `_is_managed_skill` during prune, uninstall, recovery, and replacement cleanup inherits the guarantee. `test_tree_remove_preserves_a_tree_modified_after_the_ownership_read` edits a child after the read reports owned and proves the edit survives at the canonical path.

The threat model now describes the actual guarantees: full-chain trusted-root pinning, descriptor-pinned quarantine with predicate re-authentication covering descendant content, and quarantine-based failure recovery.

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; core fs_ownership 38 passed, gjc/refresh/uninstall/runtime-instruction suites 146 passed, GJC setup suite 46 passed.

Q00:
Round-12 blockers are addressed in `73aa20c8`.

**1. Staged-generation binding (`fs_ownership.py:1141`)** — publication is now bound to the authored generation end to end. `publish_owned_file` pins the staged inode through the write descriptor itself (captured before the first byte is written — no window at all); tree and entry publication pin the imported entry with an `O_NOFOLLOW` capability immediately after the descriptor-bound import. The staged identity is re-verified immediately *before* the final no-replace rename (a swap onto the random staging name aborts with `staged generation changed before publication`) and immediately *after* it (an imposter that races the rename itself is pulled aside untouched into a claim-shaped name and the publication fails with `canonical path was raced during publication`, with the prior owned generation restored). The collision/error cleanup branches you flagged now route through the identity-bound quarantine primitive (`quarantined_remove(staged_name, staged_pin.identity)`) instead of a bare pathname removal, so a raced-in replacement on the staging name is left untouched. Regressions intercept `rename_no_replace` and swap the staged file (`test_publish_never_publishes_a_staged_generation_swapped_before_the_rename`) and tree (`test_tree_publish_never_publishes_a_staged_generation_swapped_before_the_rename`), proving attacker content never becomes canonical and the prior generation survives.

**2. Commit-first destruction (`fs_ownership.py:490`)** — a recursive deletion can no longer restore a half-destroyed tree, per your first suggested option: removal commits with the atomic move into the pinned quarantine, and tree residue is *never* renamed back to the canonical path. Trees are discarded commit-first: moved whole into a private temporary directory and destroyed there (same-filesystem fast path); across filesystems they are destroyed inside the pinned container, and any irrecoverable residue stays quarantined under a self-describing `*.discarded` tombstone for the operator — outside the canonical and claim namespaces. Single-file `unlink` failures keep the previous behavior (the file is intact by construction, so restoring it for a retry is safe). `test_tree_remove_never_restores_a_half_destroyed_tree` reproduces your probe — one descendant deleted, then the traversal fails — and proves the canonical namespace never sees the damaged tree again.

**3. Containerized cross-filesystem import (`fs_ownership.py:381`)** — the EXDEV path now assembles inside a self-describing `0700` `.{name}.{nonce}.importing` intent container (descriptor-pinned) from its first destination write, and commits with one atomic fd-relative rename out of the container. A failed or interrupted import discards the container whole through the same commit-first discard (private-move fast path; tombstone fallback), so no loose `.tmp` partial state can be stranded or accumulate in the shared directory — cleanup failure no longer silently leaves unrecognized residue. The existing EXDEV regression (`test_failed_cross_filesystem_import_leaves_no_staging_residue`) now runs against this container path and asserts the shared parent is completely clean after a mid-import failure, then proves the import succeeds end to end.

The threat model documents all three contracts: authored-generation binding through the publication, commit-first destruction with tombstone quarantine, and import-intent containers.

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; core fs_ownership 40 passed, gjc/refresh/uninstall/runtime-instruction suites 148 passed, GJC setup suite 46 passed.

Q00:
## A calibration request for the remaining review surface

Every blocker from rounds 6–12 has been fixed and its fix acknowledged in the following round — the finding class has been narrowed in a consistent direction each time, and several catches (inode reuse on ext4, the FIFO blocking open, the builder-pathname escape) were genuine defects that made the primitive materially stronger. That loop has been valuable.

At the same time, the *mechanism* behind the reproductions has shifted. Early rounds reproduced findings with real filesystem states an operator or crashed process could plausibly leave behind (a symlinked profile root, an orphaned claim, a FIFO at a managed path). The most recent rounds reproduce findings by **deterministically intercepting the primitive's own internal calls** — patching `rename_no_replace` to swap the staged entry mid-publication, interposing between two syscalls inside `quarantined_remove`. Those interceptions prove a window *exists in the instruction stream*; they do not establish that a concurrent process — which cannot patch our address space and does not know the 8-byte-random sibling names — can actually *hit* it.

The boundary is now at the platform's own primitives. The current implementation holds descriptor capabilities across every validation-to-mutation span, pins the trusted-root chain, authenticates in a pinned quarantine, binds publication to the authored inode, and commits destruction before traversal. What remains is the irreducible fact that POSIX offers no compare-and-rename: any residual window is now between two adjacent syscalls, on an unpredictable random name, with the held descriptor forcing any replacement to a fresh inode that the very next check detects.

So a request for the next round, in the spirit of the zero-trust profile rather than against it:

1. **For any new finding in this class, please establish reachability, not just existence.** A reproduction should come from a genuinely concurrent process (or an equivalent external actor) racing the *unmodified* primitive — not from instrumenting the primitive's internals. If a finding is only reachable via in-process interception, classify it as documented residual risk at the platform-atomicity boundary rather than a merge blocker.

2. **Alternatively, state the accepted atomicity boundary explicitly** so we can converge: if the review's position is that adjacent-syscall windows on unpredictable names are in scope regardless of reachability, then no userspace implementation on POSIX can terminate this loop, and that is a roadmap decision for the maintainer — worth recording in the advisory memory as a policy, not re-derived per round.

3. The module's threat model now documents exactly which guarantees are held by capability and which end at platform rename atomicity. If any *documented* residual boundary is considered unacceptable, flagging the documentation line is more actionable than a new interception probe.

To be clear: findings with realizable preconditions — like everything through round 11 — are exactly what this PR wanted from review, and more of those are welcome. This request is only about calibrating the standard for windows that exist only under in-process instrumentation.

Q00:
Round-13 blockers are addressed in `4a07e98b`. All three had realizable preconditions (a concurrent same-user write, a process crash, a failed cleanup), so they fall squarely inside the calibration standard proposed earlier — no objection to any of them.

**1. Content-generation binding (`fs_ownership.py:1234`)** — publication now carries a content digest of the authored generation alongside the pinned inode: for files it is computed directly from the authored content; for trees and entries it is computed over structure plus every descendant's bytes in the private workspace, before the import. Both the identity *and* the digest are re-verified immediately before the no-replace commit and immediately after it, so a same-inode in-place mutation of the staged file — or of a staged tree's descendants — is detected on either side of the rename and never becomes canonical; an unauthored generation that races the commit itself is pulled aside untouched and the prior generation restored. Regressions cover exactly the requested cases: in-place same-inode file mutation (`test_publish_never_publishes_staged_content_mutated_in_place`) and descendant mutation with an unchanged top-level inode (`test_tree_publish_never_publishes_descendants_mutated_in_place`).

**2. Replay-recoverable import intents (`fs_ownership.py:397`)** — the cross-filesystem container now writes an intent marker naming the canonical artifact as its *first destination write*. `_reconcile_stale_containers` runs at the start of every transaction on a path (inside the same reconciliation pass as claim recovery, so refresh and uninstall inherit it through their existing recovery calls) and completes the discard of any interrupted `*.importing`/`*.discarding` container whose marker names that path — the correct replay of an interrupted transaction. A container without a matching marker is a forgery or belongs to another artifact and is left untouched. `test_interrupted_import_container_is_reconciled_on_the_next_transaction` simulates the process-crash residue you produced, proves the next publication reconciles it, and proves a forged marker-less container survives untouched.

**3. Truthful discard reporting (`fs_ownership.py:447`)** — a discard whose cleanup cannot complete no longer reports success. The private destruction is verified afterward; surviving residue is restored into the pinned container where possible, the container is retained under an intent-marked `.{canonical}.{nonce}.discarding` tombstone, and the removal raises (`cleanup residue retained in a discoverable tombstone`), so `claim_and_remove_owned` — and therefore uninstall — reports the failure instead of claiming a finished cleanup. The tombstone is durable, discoverable, and reconciled by the next transaction on the same path. The half-destroyed-tree regression now asserts the truthful error, the intact intent marker, and that a later `recover_owned_claims` collects the tombstone; the no-restore-of-damage guarantee from round 12 is preserved throughout.

The threat model documents the content-digest binding, the intent-marked container lifecycle, and the truthful-failure contract.

Validation: `ruff format`/`ruff check`, `mypy src/ouroboros` clean; core fs_ownership 43 passed, gjc/refresh/uninstall/runtime-instruction suites 151 passed, GJC setup suite 46 passed.

Q00:
**Round 14 response — `bd8f2f2a`** (addresses both reviews on `4a07e98b`: `req_1787666853_30` and `req_1787668147_1`, which report the same blocker)

The finding is accepted without objection: it was established by a deterministic **external** probe against the unmodified primitive — exactly the reachability standard proposed in the calibration comment — and the flaw was real. `_reconcile_stale_containers` granted destructive authority to metadata (`container name shape + `.ouroboros-intent` marker`) that lives in the operator-shared parent and is therefore forgeable by any same-user writer.

**Fix — writer-owned transaction ledger** (`fs_ownership.py`), following the structure prescribed in the review's Design Notes:

- Every cross-filesystem import and every tombstone rename first writes a **ledger record** under an exclusively writer-owned state root (`~/.ouroboros/fs-transactions`, `0700`; overridable via `OUROBOROS_FS_TRANSACTION_DIR`), *before* the container becomes discoverable in the shared parent. The record binds the absolute canonical path, the operation (`importing`/`discarding`), and the exact container name.
- `_reconcile_stale_containers` now deletes a container **only when the ledger vouches for it** (record exists and matches canonical path + operation + exact name) and the container's marker does not name a different artifact. The record is **retired atomically with the replay**, so a *replayed* container name — recreated later with operator data — carries no live authority either. Anything the ledger does not vouch for, marker-bearing or not, is operator state and is left untouched; the ownership predicate is never a substitute for this authority.
- Markers are demoted to pure discovery metadata: a crash that dies before the marker write no longer strands vouched residue, and a failed-cleanup container is renamed into the recoverable `.discarding` shape only after its record is durably written (otherwise it keeps its current name and whatever live record it already has).
- Records orphaned by a crash between cleanup and retirement are retired during reconciliation; the GC direction is one-sided — it can only strand residue, never delete operator state.

Because `_recover_claims` is shared, the fix covers `publish_owned_file`, `publish_owned_tree`, `publish_owned_entry`, `claim_and_remove_owned`, and `recover_owned_claims` — i.e. every GJC skill/guide/bridge publication, rollback, refresh, and uninstall caller, with no caller-specific checks.

**Regressions added** (`tests/unit/core/test_fs_ownership.py`, now 45):
- `test_recovery_never_destroys_a_forged_container_with_a_matching_marker` — the review's probe verbatim: forged `.artifact-dir.0123456789abcdef.importing` with matching marker and operator payload survives direct recovery (`is_owned=lambda _: False` returns `False`, payload intact), a publish transaction, a removal, and a rename into the `.discarding` shape.
- `test_reconcile_replays_a_container_that_crashed_before_its_marker_write` — ledger-vouched, marker-less residue is still replayed and its record retired.
- The interrupted-import regression now models the real crash state (record written pre-crash) and asserts record retirement with the replay.

Gates: `ruff format`/`check` clean, `mypy` clean (622 files), focused suites 153 passed, GJC-selected setup tests 46 passed.

Q00:
**Round 16 response — `7c04e283`** (review `req_1787673802_2` on `bd8f2f2a`)

The blocker is accepted: the trust-source analysis is correct, was established by a deterministic external probe, and follows directly from the ledger fix — an authority store whose *location* is selectable by an untrusted project `.env` is no authority at all. The fix follows the Architecture Reset Direction's implementation path exactly, staying inside the recorded subsystem budget:

1. **`OUROBOROS_FS_TRANSACTION_DIR` added to `UNTRUSTED_ENV_DENYLIST`** (`src/ouroboros/config/untrusted_env.py`), documented as the destructive-recovery authority root. Trusted precedence is preserved unchanged: the real process environment and `~/.ouroboros/.env` may still select it; only the project-directory `.env` that travels with a cloned repository loses the ability. The denylist floor test (`test_denylist_covers_known_execution_routing_keys`) now pins the key so it can never be silently dropped.
2. **Loader-to-recovery regression** — `test_untrusted_env_cannot_select_fs_transaction_dir_for_recover_owned_claims` (`tests/unit/config/test_loader_env.py`) reproduces the reset direction's minimal reproduction end to end: a project `.env` naming an attacker-controlled ledger directory holding a matching record, plus a forged `.artifact-dir.<nonce>.importing` container with a matching marker and operator payload in the shared parent. After `_load_env_file(..., trusted=False)` the key never reaches `os.environ`, `_transaction_ledger_root()` does not resolve to the attacker directory, `recover_owned_claims(..., is_owned=lambda _: False)` returns `False` with the operator payload untouched — and the trusted process environment still selects an isolated ledger.

Reset verification command passes as specified: `uv run pytest -q -k 'fs_transaction_dir and untrusted_env and recover_owned_claims'` → 1 passed.

Per the reset contract, nothing else was built: no GJC-specific guards, no second ledger, no parallel recovery path — the change is two files (one denylist entry + comments, one regression), and the centralized ledger/claim primitives from `bd8f2f2a` are unchanged.

Gates: `ruff format`/`check` clean, `mypy` clean (622 files), focused suites 391 passed (core 45, gjc 12, refresh, uninstall 55, runtime artifacts, config env 238), GJC-selected setup tests 46 passed.

Q00:
**Round 17 response — `acc58410`** (review `req_1787674969_1` on `7c04e283`)

The blocker is accepted: the orphan-record pass judged a live transaction by an unlocked directory snapshot, and the interleaving (record written → concurrent reconcile retires it → container published → creator dies → residue permanently unauthenticated) was deterministically reachable. The fix follows the Architecture Reset Direction's implementation path exactly, inside the recorded subsystem budget:

1. **Per-canonical-path ledger lock** — `_ledger_lock(canonical)` in `fs_ownership.py`, built on the existing `ouroboros.core.file_lock.file_lock` as directed (no new locking mechanism, no leases, no timestamps). The lock file lives in the writer-owned ledger root, keyed by a hash of the canonical path, so holders never touch the operator-shared parent.
2. **Publishers hold it across the record-before-container window** — `_PinnedParent.import_entry()`'s cross-filesystem branch runs its entire record → mkdir → copy → commit-or-rollback lifetime under the lock, and `quarantined_remove()`'s container discard (whose failure path writes a tombstone record before the tombstone rename) is wrapped the same way. `_retain_tombstone()` and `_discard_container_entry()` themselves stay lock-free — they are only ever reached under a caller's lock, and `flock` conflicts across descriptors within one process, so acquisitions are never nested.
3. **Reconciliation acquires the same lock before replay or retirement** — `_reconcile_stale_containers()` takes the lock first, and the absence-only orphan pass is gone: retirement now judges a **fresh snapshot taken while holding the lock**. Since every cooperating publisher holds the lock from before its record exists until its container is published, acquisition proves no live publisher is mid-publication — absence under the lock is evidence of completion, not a guess. The comments that previously described the unlocked window as an accepted trade-off were deleted with it.
4. **Deterministic interleaving regression** — `test_live_record_survives_concurrent_reconcile_and_creator_crash_recovery` (`tests/unit/core/test_fs_ownership.py` only, per the reset contract): the creator writes its record under the lock, a concurrent `recover_owned_claims` thread provably blocks (the live record survives), the creator publishes the container and dies releasing the lock, and the unblocked reconciliation replays the container and retires the record — the reset success condition verbatim. Determinism comes from the lock itself, not sleeps.

Reset verification command passes as specified: `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest -q tests/unit/core/test_fs_ownership.py -k 'live_record and concurrent_reconcile and creator_crash'` → 1 passed.

Nothing outside `core/fs_ownership.py` changed; GJC callers inherit the corrected protocol through the shared primitive.

Gates: `ruff format`/`check` clean, `mypy` clean (622 files), core 46 passed, focused suites 392 passed, GJC-selected setup tests 46 passed.

Q00:
**Round 18 response — `cc12382b`** (review `req_1787674969_1`... superseded; this responds to the review on `acc58410`)

Both blockers are accepted — each was established by a deterministic probe against realizable states, and both follow the reset direction's implementation path exactly, inside `core/fs_ownership.py` only.

**Blocker 1 — ledger root not pinned.** `_LedgerAuthority` now opens the trusted transaction root **once** and performs locking, record creation, reads, retirement, and listing through that single held descriptor (`file_lock(..., parent_fd=fd)` for the lock; `dir_fd`-relative `open`/`unlink`/`listdir` for records). It is threaded through `import_entry`, `_discard_container_entry`, `_retain_tombstone`, `quarantined_remove`, and `_reconcile_stale_containers`, so within any locked section every operation acts on the exact directory generation the lock lives in — a renamed-and-recreated root can no longer split a publisher and a reconciler across two authorities. A pleasant consequence of reusing `core.file_lock`: its own lockfile-parent validation makes a **live publisher detect the swap** (its lock release fails with "lockfile parent changed") rather than ever mistaking the imposter root for its authority — the regression asserts this explicitly.

**Blocker 2 — stat-then-pathname EXDEV deletion.** The in-container fallback now follows the module's capture-and-verify discipline end to end: `entry` is captured by an **atomic rename to an unpredictable in-container name** (taking whatever occupies the name), re-verified against the pre-capture `(dev, ino, type)` — a mismatch restores the replacement to `entry` untouched and reports the cleanup as failed — and destruction runs through descriptors only (`_remove_tree_contents_fd`: no-follow child opens re-verified by `fstat`, `dir_fd`-relative unlink/rmdir; the final `rmdir` on the captured name can only ever remove an empty directory). A destruction failing partway restores the remains to the canonical in-container name so tombstone replay finds them where the transaction schema expects.

**Regressions** (both in `tests/unit/core/test_fs_ownership.py`, per the reset contract):
- `test_ledger_root_swap_cannot_split_lock_and_record_authority` — root renamed+recreated inside the publisher's record-before-container window; a reconciler on the replacement root retires nothing (the live record survives in the pinned authority), the publisher's lock release detects the swap, and once the genuine root is restored the interrupted import replays and its record retires.
- `test_discard_entry_swap_is_detected_before_recursive_deletion` — forced `EXDEV`, `entry` swapped for an operator directory after the pre-deletion stat; the identity mismatch is detected, the replacement survives intact inside the truthfully-reported tombstone, and nothing is recursively deleted through a reused pathname.
- The prior half-destroyed-tree regression was updated to inject failure into the fd-relative destruction path (its old `shutil.rmtree` interception no longer corresponds to any code path — the pathname deletion it patched is gone).

Reset verification command passes as specified: `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest -q tests/unit/core/test_fs_ownership.py -k 'ledger_root_swap or discard_entry_swap'` → 2 passed.

Gates: `ruff format`/`check` clean, `mypy` clean (622 files), core 48 passed, focused suites 394 passed, GJC-selected setup tests 46 passed. No GJC-specific guards, no second ledger, no parallel recovery path.

Q00:
**Round 19 response — `b7f91224`** (review on `cc12382b`)

Both findings are accepted, and the Architecture Reset Direction's structural conclusion is adopted wholesale: physical recursive erasure is no longer part of the ownership transaction at all.

**Blocker — descendant stale-pathname deletion.** Rather than generalizing capture-and-verify to every descendant (which would keep growing a transaction engine the filesystem cannot back — atomic rename exists for an entry, atomic tree deletion does not), the in-place destruction path is **deleted**, exactly as the reset direction prescribes:

- `_remove_tree_contents_fd()` and the entire EXDEV in-place fallback are gone.
- **The atomic quarantine is the removal commit.** Discard moves the entry *whole* into a writer-private area — the writer-owned ledger root first (typically the artifact's own filesystem), the system temporary directory as a fallback — and destroys it there through private pathnames no other writer can reach.
- When **no private area can receive the entry**, nothing is destroyed: the residue is retained under the intent-marked `.{canonical}.{nonce}.discarding` tombstone and the removal raises truthfully. A later substitution therefore *cannot* be deleted — there is no code path that deletes through a name in an operator-reachable location anymore.

**Follow-up — skill setup mutating before validation.** `install_gjc_skills()` no longer calls `target_root.mkdir()` or its symlink preflight; the `skills/` parent is created solely by `publish_owned_tree()`'s pinned no-follow walk from the trusted agent root. A symlinked configured root now produces **zero** side effects — the new regression asserts the external target remains entirely unchanged, including the absence of an empty `skills` directory.

**Regressions** (as directed, only in the two named test files):
- `test_recursive_descendant_swap_is_never_deleted_through_stale_names` — forced EXDEV on every private area, `content.txt` descendant replaced mid-removal; the replacement survives byte for byte inside the truthful tombstone.
- `test_symlinked_agent_root_no_side_effects_before_validation` — `agent -> external`; install raises with `external` left entirely empty.
- The prior `discard_entry_swap` regression was rewritten to the new model (top-level swap: both the displaced generation and the replacement survive inside the tombstone), and the half-destroyed-tree regression still passes — its guarantee (no damaged remains ever return to the canonical namespace) is now trivially structural.

Reset verification command passes as specified: `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest -q tests/unit/core/test_fs_ownership.py tests/unit/gjc/test_gjc_artifacts.py -k 'recursive_descendant_swap or symlinked_agent_root_no_side_effects'` → 2 passed. The prior rounds' commands also still pass (`ledger_root_swap or discard_entry_swap` → 2 passed).

Net effect on the module: this round *removed* more machinery than it added (`fs_ownership.py` shrank), which is the direction the reset analysis pointed at — the ownership boundary got stronger by doing less inside operator-reachable space.

Gates: `ruff format`/`check` clean, `mypy` clean (622 files), core 49 passed, gjc 13 passed, focused suites 396 passed, GJC-selected setup tests 46 passed.

Q00:
**Round 20 response — `434f5ee6`** (review on `b7f91224`)

The blocker is accepted: the swallowed reinsertion failure was exactly the kind of false-success this PR exists to eliminate, and the unrecorded temporary quarantine was indeed a second authority outside the transaction ledger. The fix follows the reset direction's implementation path, in `core/fs_ownership.py` only, with GJC callers unchanged:

1. **The private quarantine is ledger-recorded before movement.** `_discard_container_entry` writes an `"erasing"` record — binding the canonical path to the absolute private location — *before* the entry moves into the private area (a candidate whose record cannot be written is skipped entirely). From that moment the residue is never outside durable recovery authority.
2. **Retirement only after verified erasure.** The `"erasing"` record is retired exclusively when the private location verifiably no longer holds the entry — after successful private destruction, or after a successful reinsertion back into the intent-marked container (whose own tombstone record then takes over).
3. **The swallowed failure now raises.** When private erasure leaves residue *and* reinsertion fails, the discard raises `OSError("cleanup residue retained in a ledger-backed private quarantine", <location>)` instead of `pass`-ing into a success return — `claim_and_remove_owned`, setup rollback, refresh, and uninstall all report the truth through the shared primitive.
4. **Recovery replays recorded private quarantines.** `_reconcile_stale_containers` (under the same per-canonical ledger lock) replays `"erasing"` records for the transaction's path: the recorded location is writer-private (and sanity-checked for the `ouroboros-discard-` shape), so destroying it by its recorded pathname is safe, and the record is retired only after verified absence. The absence-only record GC is now scoped to shared-parent container operations (`importing`/`discarding`), so it can never retire an `"erasing"` record whose location is intentionally outside the shared parent.

**Regression** — `test_private_cleanup_restore_failure_retains_ledger_authority` (`tests/unit/core/test_fs_ownership.py`), the review's probe verbatim: private deletion patched to a no-op, reinsertion rename raising `EIO`. The removal raises (no false success), the canonical artifact is gone (the atomic quarantine committed), the residue's bytes sit under a live `"erasing"` record, and a later `recover_owned_claims` replays the erasure and retires the record only once the location is verifiably absent.

Reset verification command passes as specified: `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run python -m pytest -q tests/unit/core/test_fs_ownership.py -k 'private_cleanup_restore_failure_retains_ledger_authority'` → 1 passed. All prior rounds' verification selections still pass (5 passed across `recursive_descendant_swap`, `symlinked_agent_root_no_side_effects`, `ledger_root_swap`, `discard_entry_swap`, `live_record…`).

Gates: `ruff format`/`check` clean, `mypy` clean (622 files), core 50 passed, focused suites 397 passed, GJC-selected setup tests 46 passed. No GJC-specific recovery, no second ledger, no shared-path recursive deletion.

Q00:
**Round 21 response — `43f97ba8`** (review `pullrequestreview-5022194853`)

The blocker is accepted. The root cause was the split authority identified in the review: `_LedgerAuthority` pinned the lock and record generation, while `_discard_container_entry` re-resolved `OUROBOROS_FS_TRANSACTION_DIR` by pathname for private quarantine creation and later replay.

The fix moves the complete private-quarantine lifecycle under `_LedgerAuthority`:

- quarantine names are validated relative names (`ouroboros-discard-<nonce>`), never absolute recorded paths;
- creation uses `os.mkdir(..., dir_fd=ledger.fd)`;
- entry movement and reinsertion use the same `ledger.fd` as source/destination authority;
- existence checks and erasure are descriptor-relative;
- `erasing` replay calls `ledger.erase_private_quarantine(relative_name)` and never re-resolves the configured root;
- the independent system-temp/pathname fallback was removed; an EXDEV or unavailable private move fails closed to the existing intent-marked shared-parent tombstone;
- no GJC-specific ledger or recovery path was added.

Two deterministic regressions cover both sides of the lifecycle:

1. `test_private_quarantine_root_swap_keeps_residue_with_ledger_authority` swaps the configured root during quarantine creation, forces erasure and reinsertion failure, and proves the record and residue remain together under the originally pinned generation and replay after that generation is restored.
2. `test_private_quarantine_replay_stays_on_pinned_root_during_swap` swaps the root during replay and proves erasure plus record retirement stay on the authority opened before the swap; the replacement root receives no cleanup effect.

Verification:

- reset command selection: `1 passed`
- private-quarantine creation/replay regressions: `2 passed`
- EXDEV, descendant-swap, root-swap, and stranded-residue selections: `4 passed`
- focused core/GJC/refresh/uninstall/runtime/config lifecycle suites: `399 passed`
- Ruff: clean
- MyPy: clean

This removes the residual pathname authority rather than adding another recovery layer: lock, record, residue, replay, and retirement are now one pinned ledger-root generation.

Q00:
**Round 23 response — `c7f772bb`** (review `pullrequestreview-5037310099`)

The blocker is accepted. The pinned ledger root was correct, but the durable `erasing` record named only a quarantine pathname and did not attribute the leaf generation occupying that name.

The fix centralizes the complete record-bound cleanup contract in `_LedgerAuthority`:

- `erasing` records now persist the quarantine `(device, inode, file-type)` identity alongside the validated relative name;
- quarantine creation captures that identity immediately;
- movement into the quarantine opens the leaf `O_DIRECTORY|O_NOFOLLOW`, verifies both the held descriptor and current pathname against the recorded identity, and writes through the held leaf descriptor;
- immediate cleanup and replay use the same identity-bound erase primitive;
- recursive erasure occurs only after the recorded leaf is pinned and verified;
- final empty-leaf removal atomically claims the pathname and re-verifies the captured generation before `rmdir`;
- a replacement at the recorded name is preserved, the original record is retained, and removal/replay fail closed;
- record retirement occurs inside the erase primitive only after the exact attributed generation is removed;
- absence-based `erasing` retirement and pathname-only cleanup calls are removed;
- no GJC-specific recovery path or second ledger was added.

Regressions:

1. `test_private_quarantine_leaf_replacement_fails_immediate_cleanup_closed` renames the original quarantine after the owned entry moved in, plants operator data under the recorded name, and proves removal fails while both the operator replacement, original residue, and `erasing` record survive.
2. `test_private_quarantine_leaf_replacement_fails_replay_closed` reproduces the same replacement before replay and proves recovery deletes neither generation and retains the record.

Verification:

- reset command: `2 passed, 52 deselected`
- complete core ownership suite: `54 passed`
- focused core/GJC/refresh/uninstall/runtime/config lifecycle suites: `401 passed`
- Ruff: clean
- MyPy: clean (622 files)

A full pytest run is also in progress; I will report any result that differs from the focused evidence above.

Q00:
Full local pytest follow-up for `c7f772bb`: `21,907 passed, 95 skipped, 1 xfailed, 4 failed`.

The four failures are outside the changed filesystem path:

- two GJC setup endpoint tests inherit the harness `_OUROBOROS_NESTED` marker into their launched subprocess (the same environment-specific failure observed before; the focused GJC suites pass);
- one macOS APFS test rejects an intentionally invalid UTF-8 filename with `EILSEQ`;
- one machine-load-sensitive persistence performance ratio exceeded its threshold.

The review reset command, complete ownership suite, focused GJC lifecycle suites, Ruff, MyPy, and the GitHub required gates covering this change are clean.

Q00:
**Round 24 response — `191c06e0`**

The child-generation blocker is accepted and fixed in the shared effect owner.

`_LedgerAuthority` now owns one complete private-entry lifecycle:

- after moving `entry`, its `(device, inode, file-type)` identity is durably added to the existing `erasing` record;
- immediate cleanup and replay both require the recorded outer-quarantine identity and recorded child identity;
- `entry` is atomically claimed to an unpredictable private name before recursive work;
- the claimed child is opened `O_DIRECTORY|O_NOFOLLOW`, and both the held descriptor and claim pathname are verified against the recorded identity;
- recursive erasure runs through the claimed generation rather than the public `entry` pathname;
- a raced-in replacement at `entry` is captured, rejected, and restored no-clobber; the original displaced residue, replacement, and ledger record all survive;
- record retirement occurs only after the exact child and outer quarantine generations are removed;
- immediate cleanup and `erasing` replay use the same primitive; no GJC-specific cleanup path or additional ledger was introduced.

Regressions:

- `test_private_quarantine_entry_replacement_fails_immediate_cleanup_closed`
- `test_private_quarantine_entry_replacement_fails_replay_closed`

Both replace `quarantine/entry` immediately before the claim and assert the operator replacement, displaced original generation, and erasing record survive.

Verification:

- reset command: `2 passed, 53 deselected`
- complete ownership suite: `55 passed`
- focused core/GJC/refresh/uninstall/runtime/config suites: `402 passed`
- Ruff: clean
- MyPy: clean (622 files)

Q00:
Follow-up `050b55c8`: the child-generation implementation pushed `core/fs_ownership.py` over the non-grandfathered 2,000-line cap. I condensed the module threat-model docstring without changing runtime behavior.

Local verification after the reduction:

- `python3 scripts/check-module-size.py --baseline-ref github/main` → `module-size: OK`
- Ruff: clean
- MyPy: clean (622 files)
- focused lifecycle suites: `402 passed`

The child-generation code and regressions from `191c06e0` are unchanged.

Q00:
**Round 26 architecture reset — `9929fd62`**

Adopted the review's structural conclusion: physical recursive reclamation is no longer part of authenticated tree removal.

- `_PinnedParent._discard_container_entry` now verifies the held container generation and atomically no-replace renames the complete tree to an unpredictable terminal `.<canonical>.<nonce>.retired` tombstone.
- The retired generation is fsynced and never recursively mutated afterward.
- Removed the private `erasing` ledger records, quarantine/entry identity machinery, recursive erasure, reinsertion, and erasing replay.
- Shared `importing`/`discarding` intents still use the existing ledger and reconcile by terminal retirement.
- No GJC-specific cleanup or second lifecycle path was added.

External-process regression:

`test_concurrent_descendant_replacement_survives_terminal_retirement` starts a separate forked process, waits for the retired generation, replaces `victim/hot` with a new inode, and proves:

- `claim_and_remove_owned()` returns success;
- the canonical path is absent;
- the raced-in operator replacement survives;
- the displaced original descendant also survives inside the retired tree.

Verification:

- exact reset command: `1 passed, 48 deselected`
- focused ownership/GJC/refresh/uninstall/runtime/config suites: `396 passed`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

Net effect: this reset removes 1,016 lines of recursive cleanup and recovery machinery. Logical removal is atomic; retained generations are immutable from Ouroboros' perspective.

Q00:
**Round 27 response — `f67851c7`**

Fixed the rollback contract without weakening terminal retirement for ordinary removal.

- Added `claim_and_reclaim_owned()` in `core/fs_ownership` for transaction-owned rollback material only.
- It atomically claims the canonical generation, re-validates the exact snapshot-authored generation, moves it whole out of the active profile into a private `0700` temporary quarantine, verifies the isolated inode, and recursively disposes it only there.
- A canonical generation recreated or modified by another writer is restored/preserved and never reclaimed.
- `_restore_gjc_paths()` now uses this transaction reclamation path only when the pre-transaction snapshot was missing.
- Ordinary uninstall/pruning still use terminal `.retired` retirement and never recursively mutate retained live/shared generations.

Verification:

- `TestGjcSetup::test_setup_gjc_registration_failure_rolls_back_and_exits_nonzero` plus complete ownership suite: `50 passed`
- focused core/GJC/refresh/uninstall/runtime/config suites: `396 passed`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

Failed native activation now leaves no transaction-created skill tombstones in `agent_dir/skills`, while concurrent operator generations remain protected.

Q00:
**Round 28 response — `0ae35fea`**

Removed recursive rollback reclamation. Failed activation now clears the active GJC profile by atomically archiving the complete transaction generation intact.

- Replaced `claim_and_reclaim_owned()` with `claim_and_archive_owned()`.
- The canonical generation is claimed and snapshot-verified, then moved whole into the pinned writer-owned transaction-ledger root with one cross-descriptor rename.
- The archived top-level inode is verified after the move; no descendant is traversed, unlinked, or recursively deleted.
- Any descendant written after the ownership callback moves with the archived generation and survives.
- Concurrent canonical replacements remain protected by the existing no-clobber claim/restore contract.
- `_restore_gjc_paths()` uses archival only for paths absent before activation; ordinary removal retains terminal `.retired` semantics.

Requested regression:

`test_failed_gjc_activation_preserves_post_validation_descendant_write` adds `operator.txt` inside the ownership callback and verifies rollback removes the active path while the complete original and operator descendant both survive in the rollback archive.

Verification:

- requested adversarial regression + failed-registration rollback: `2 passed, 526 deselected`
- focused ownership/GJC/refresh/uninstall/runtime/config suites: `397 passed`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

Q00:
**Round 29 response — `6db14437`**

Moved rollback payload authority to the managed profile filesystem before publication.

- `prepare_rollback_archive()` creates and pins a private `0700` archive beside the configured GJC agent root before any native artifact is published.
- The existing global ledger stores only the archive intent metadata; it is no longer a payload destination.
- `claim_and_archive_owned()` receives the preflighted archive and moves complete rollback generations there with a same-filesystem descriptor-relative rename.
- Descendants are never copied or recursively deleted; post-validation writes move intact with the archived tree.
- `_restore_gjc_paths()` reuses the archive for all missing-snapshot tree rollbacks.
- Empty archives and their metadata are retired on successful activation; failed activation retains non-empty archives outside the active GJC profile.

Cross-device regression:

`test_gjc_rollback_cross_filesystem_archive` simulates `EXDEV` for any payload move into the home-scoped ledger. Failed registration still restores the original config, leaves no generated `skills/` state, and retains the complete rollback generations in the same-device preflight archive.

Verification:

- exact cross-filesystem selection: `1 passed, 478 deselected`
- GJC setup suite: `47 passed, 432 deselected`
- combined GJC/rollback/ownership selection: `165 passed, 711 deselected`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

Q00:
**Round 30 response — `d5a2b7e7`**

`RollbackArchive` is now a retained generation capability rather than a pathname record.

- The live handle owns the pinned parent fd, archive fd, creation `(device, inode, type)`, archive name, and ledger nonce for the entire activation transaction.
- `claim_and_archive_owned()` writes directly through the retained archive fd after verifying both the held archive descriptor and current archive name still match the creation generation.
- `close_rollback_archive()` never calls `Path.rmdir()`: it revalidates the capability, atomically claims the empty archive name, verifies the captured generation, and only then removes it and retires metadata.
- A replacement archive pathname receives no rollback payload and is never removed.
- Ledger records now include archive generation identity. Conservative reconciliation retires only empty archives whose current identity matches the recorded generation; mismatches and non-empty archives are left untouched.
- GJC continues to preflight this capability before artifact publication and reuses it through rollback.

Regressions:

- `test_rollback_archive_replacement_cannot_receive_payload`
- `test_rollback_archive_replacement_cannot_be_removed_on_close`

Verification:

- exact replacement selection: `2 passed, 50 deselected`
- focused ownership/GJC/refresh/uninstall/runtime/config suites: `399 passed`
- full GJC setup selection: `47 passed, 432 deselected`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

Q00:
**Round 30 completion — `bda9a57c`**

Closed the remaining pathname-only crash-replay path.

Live transaction:
- `RollbackArchive` retains pinned parent/archive descriptors and creation identity from prepare through archive and close.
- No live mutation reopens the archive pathname as authority.

Crash reconciliation:
- opens the recorded parent through the no-follow pinned-parent walk;
- validates the current archive name against the recorded `(device, inode, type)`;
- opens the archive `O_DIRECTORY|O_NOFOLLOW`;
- retires only an empty matching generation after atomically claiming and re-verifying it;
- leaves missing, non-empty, displaced, or replacement generations and their ledger records untouched.

Added `test_rollback_archive_replacement_survives_crash_replay`, which simulates descriptor loss, displaces the authentic archive, plants a substitute, and proves conservative replay preserves both directories and retains the original ledger record.

Verification:
- rollback archive replacement selection: `3 passed, 50 deselected`
- GJC setup selection: `47 passed, 432 deselected`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

Q00:
**Round 32 architecture reset response — `8f566feb`**

Removed the rollback archive lifecycle entirely, as directed.

- Deleted `RollbackArchive`, `prepare_rollback_archive()`, `close_rollback_archive()`, `claim_and_archive_owned()`, archive-specific ledger records, and archive replay/pathname handling.
- `_restore_gjc_paths()` now uses the existing `claim_and_remove_owned()` terminal-retirement contract for transaction-created paths.
- Failed activation removes the exact published canonical generation from the active namespace by atomic claim and leaves it as an explicit `.retired` tree; no descendant is copied, recursively deleted, or replayed through a second authority.
- Ordinary uninstall/pruning and rollback now share one removal primitive.
- Existing canonical replacement and descendant-preservation guarantees remain intact.

Regression coverage now asserts post-validation descendant writes survive inside the terminal retired generation. Archive replacement/crash tests were removed with the deleted subsystem.

Verification:

- full core ownership suite: `50 passed`
- full GJC setup selection: `46 passed, 432 deselected`
- focused lifecycle suites: `397 passed`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

This implements the reset’s final contract: no rollback archive lifecycle, no pathname-based archive replay, no recursive cleanup of a published generation, and one canonical generation-bound removal path.

Q00:
**Round 33 architecture reset response — `e058c7f9`**

Removed the cross-filesystem transaction state instead of adding another replay layer.

- `import_entry()` now publishes only via one destination-local atomic rename.
- If the private builder workspace is on another filesystem, it fails closed with `EXDEV` before any staging generation becomes visible in the shared parent.
- Deleted cross-filesystem importing containers, ledger-backed container replay, intent-marker recovery, and the related environment/config surface.
- Retained only authenticated claim-sibling recovery for single-entry mutations and terminal `.retired` tree retirement for ordinary removal.
- GJC rollback now uses the same canonical generation-bound removal primitive; no archive, second ledger, or pathname replay remains.
- Updated tests to assert cross-filesystem staging fails before visible publication and to cover terminal retirement plus external descendant replacement.

Verification:

- core ownership + GJC artifact suites: `58 passed`
- GJC setup selection: `46 passed, 432 deselected`
- refresh/uninstall/runtime/config suites: `331 passed`
- Ruff: clean
- MyPy: clean (622 files)
- module-size ratchet: OK

This closes the crash boundary by removing the unowned intermediate state rather than trying to make another transaction engine replay-safe.

Q00:
Architecture reset applied in `c92d1d12`.

- Moved GJC contract probing, MCP registration, and endpoint validation into `ouroboros/gjc/adapter.py`.
- Reduced `cli/gjc_setup.py` to ordering independent setup-owned publications.
- Deleted the GJC-wide path snapshot/restore transaction, registration rollback state, and snapshot-only skill path inventory.
- Runtime config/credentials are committed only after host activation succeeds, with generation-checked config writes.
- Failed activation preserves complete setup-owned generations for refresh; no GJC-wide journal, rollback archive, cross-filesystem transaction emulator, or GJC filesystem authority remains.
- Kept `core/fs_ownership.py` as the single-entry claim/verify/publish/remove owner; no new recovery subsystem was added.

Verification:
- `SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uv run pytest -q tests/unit/core/test_fs_ownership.py tests/unit/gjc/test_gjc_artifacts.py` → 58 passed
- GJC setup + refresh + uninstall focused lifecycle → 140 passed
- complete setup/refresh/uninstall suites → 552 passed, 1 skipped
- Ruff, MyPy, module-size ratchet, and `git diff --check` clean.

Q00:
Closing this superseded architecture PR in favor of #2288. The native skill/MCP projection assumed a GJC autoload capability that current GJC does not provide; #2288 starts from main and keeps the existing input bridge with fail-closed collision handling and post-install config commit.

PR #2288, opened 2026-08-27
fix(gjc): make bridge setup fail closed

Q00:
## Summary

- Keep GJC's existing input bridge as the single `ooo` command route.
- Refuse to overwrite a differing bridge at the canonical path.
- Commit the GJC runtime selection only after bridge installation succeeds, and propagate failures as non-zero results.

## Non-goals

No native skill projection, MCP registration, filesystem transaction layer, refresh migration, or uninstall lifecycle changes.

## Supersedes

This is the minimal replacement for the unsupported native-projection design in #2276. The previous PR is linked as design history only; this branch starts from `main` and retains only the working bridge route.


Q00:
@ouroboros-agent Please review the replacement PR. It intentionally follows the minimal handoff: existing bridge only, fail-closed collision handling, bridge-before-config commit, no native GJC projection/MCP/filesystem lifecycle.

Q00:
@ouroboros-agent The parent-directory authority finding is intentionally outside this replacement’s reduced contract. Per the handoff, #2288 must not introduce or modify a filesystem capability/transaction layer; it retains the existing bridge only, rejects all differing canonical bytes (including un-attributed historical generations), and uses the existing snapshot guard for ordinary setup races. Please re-review the reduced bridge-only scope and classify parent-authority hardening as a separate follow-up rather than a blocker.

PR #2278, opened 2026-08-25
fix(telemetry): reduce PostHog collection to core metrics

Q00:
## Summary

- reduce PostHog ingestion to installs, daily activity, adoption dimensions, workflow outcomes, and actionable failures
- remove high-volume session, polling/helper success, subagent dispatch, and unused diagnostic properties
- daily-deduplicate retained service and command activity with deterministic `$insert_id` values

## Retained metrics

- successful installs by day and `ref`
- app version, OS, country, and runtime backend adoption
- service DAU and command DAU
- user lifecycle
- workflow terminal outcomes by status
- blocked seed paths, rejected/failed commands, and workflow failure reasons

## Removed collection

- `install_started`
- per-session `mcp_serve_started`
- `subagent_dispatch`
- successful polling and internal helper calls
- tool names, durations, Python version, provider details, frontdoor/onboarding attribution, and recovery actions

## Production PostHog

- consolidated to one `Ouroboros — Lean Product Metrics` dashboard
- retained 10 requested insights; soft-deleted 32 obsolete insights and 6 obsolete dashboards
- current 30-day baseline: 280,638 service-start events, 42,353 command events, and 4,710 subagent events; the new contract removes or daily-deduplicates these high-volume sources

## Verification

- `pytest` — 208 passed
- `ruff check` — passed
- `mypy` — passed
- `bash -n scripts/install.sh` — passed
- `git diff --check` — passed

PR #2295, opened 2026-08-28
feat: proactive MCP activation routing + verification failure analytics

Q00:
Refs #2296

## Summary

- Make MCP hosts route natural-language requests to Ouroboros entry tools proactively (interview / start_auto / start_execute_seed / lateral_think) instead of waiting for literal `ooo` — targeting the 38.7% service → command activation gap measured in PostHog.
- Answer "why did verification fail" with data: a closed `verify_cause` vocabulary on the deterministic AC verify gate, including the worker-`cd` signature (`artifacts_missing_found_elsewhere`) and concurrent-modification (`workspace_mutated`) failure modes observed in real user transcripts.
- Reinstate the adoption metrics #2278 removed (`mcp_serve_started` and `subagent_dispatch` as **daily-deduplicated** rows; `install_started` per-invocation by design — retries are the drop-off signal, volume self-bounding), preserving #2278's volume reduction (~398k → ~50-70k events/month, well inside the PostHog free tier) while restoring the attached → used funnel and the User-lifecycle dashboard series.

## Changes

**Activation routing (3 layers)**
- `render_mcp_server_instructions()`: outcome-first `WHEN TO USE OUROBOROS` section, provider-neutral, 2,038 bytes (< 2KB host budget).
- Entry-tool descriptions rewritten as *Use when / Result / Do not use when* with cross-references; `ouroboros_lateral_think` now instructs proactive invocation on stagnation signals (same fix failed twice, circular debugging) without waiting for "lateral"/"unstuck".
- `codex/ouroboros.md`: general natural-language triggers ("build this feature", "the requirements are vague", "I'm stuck") + artifact-aware routing guards (never force simple questions/small edits through Ouroboros, never blanket-route to auto).
- Verified by two independent blind routing simulations: 12/12 scenarios routed as intended, including proactive lateral on self-observed stagnation and NONE on trivial requests.

**Verification failure analytics**
- `_VerifyGateOutcome.cause` (closed 8-value enum) stamped on every gate failure branch, serialized through checkpoints (legacy checkpoint shapes stay decodable), and re-gated on read.
- Bounded elsewhere-scan detects the "artifact exists under a subdirectory the worker cd-ed into" case; annotation only, never part of pass/fail.
- `execution.verify.failed` events now carry `verify_cause` + local-only `verify_cwd`; coordinator revalidation failure sites classify as `workspace_mutated`.
- New `ac_verify_failed` PostHog event carries **only** the closed cause enum (vocabulary equality with the gate is test-enforced).

**Telemetry reinstatement (lean)**
- `mcp_serve_started` (per user/day/transport) and `subagent_dispatch` (per user/day/phase/fanout_kind) return via the same deterministic `$insert_id` + process-local suppression used by `service_active`; rich per-dispatch properties stay removed.
- `install_started` restored in `install.sh` (installs are rare; pairs with `install_completed` for drop-off).
- TELEMETRY.md updated in lockstep with every allowlist change (SSOT pairing).

## Notes

- This partially reverses #2278 by design: the *metrics* return, the *volume* does not. August baseline was ~398k events (329k of it per-session `mcp_serve_started`); the deduplicated shape projects to ~50-70k/month and scales linearly with users.
- The User-lifecycle insight's Aug-28 cliff was an artifact of #2278 removing `mcp_serve_started`; this PR restores the series going forward (a gap remains for the days between v0.51.16 and this release — worth a dashboard annotation).
- PostHog side (not in this PR): repoint the DAU insight at `service_active` and prefer explicit events over "All events" in insights.

https://claude.ai/code/session_01K7XrR5p9VcEu4ggfTrhQGk


Q00:
Addressed the review in 29683395:

**Blocker 1 (fragmented verify-cause emission)**
- Quarantine path: `_apply_verify_gate` now emits `ac_verify_failed` before quarantining, so `timeout` and `environment_unverifiable` reach telemetry; the quarantine event also records `verify_cause`.
- Final settlement: `individual_failures` now carries an explicit per-branch `(reason, outcome, verify_cause)` — missing final artifacts classify via `_missing_artifacts_cause` (so the worker-cd signature survives to the final boundary), unavailable digests fold to `unknown` instead of being guessed as `workspace_mutated`, and only genuine mutation/no-replay branches say `workspace_mutated`.
- Regression tests: quarantine telemetry for timeout and no-shell, and final-boundary artifact loss asserting `artifacts_missing` (not mutation).

**Finding 2 (optimistic attachment row)**
- `mcp_serve_started` moved to after store initialization and successful `create_ouroboros_server()`; a startup that dies on an invalid EventStore target emits nothing (regression test included).

**Low**
- HANDOFF non-goal reworded: the volume contract stays (daily dedupe, no per-dispatch detail), the reinstated events are deliberate.

On the non-blocking note about routing tests asserting keywords: acknowledged — the description/instruction text *is* the shipped contract surface, so the keyword tests pin exactly what hosts read; the blind host-model simulations are documented in the PR but intentionally not repo fixtures (they require a live model). Open to adding an eval harness in a follow-up if wanted.

Verification: `pytest` on the touched suites (173 + 645 adjacent) green, ruff format/check clean, mypy clean on changed modules.

Q00:
Round-2 blocker addressed in f487d42a: the artifacts-only branch of `_revalidate_results_after_coordinator()` now emits `execution.verify.failed` (with `verify_cause` from `_missing_artifacts_cause`, so the found-elsewhere signature survives this boundary too) and the `ac_verify_failed` analytics row before converting the result to failed. Focused regression test drives the exact probe scenario: artifacts-only contract passes the atomic gate, coordinator reconciliation removes the artifact, revalidation rejects — asserting one local event with `verify_cause="artifacts_missing"` and one telemetry row, on a result the settlement pass never revisits.

Note on the environmental test failure you observed: the failed-startup test reaches `_require_mcp_dependency()` before the invalid-store check, so it needs the repo's `mcp-test` dependency group (it passes here and should in CI); happy to guard it with a skip-if-missing marker if you prefer.

Q00:
Round-3 blocker addressed:

- **Attachment boundary**: `capture_mcp_serve_started` moved past the serve-readiness boundary — after `serve_task` starts, the lifecycle waits a short confirmation window (`_ATTACH_CONFIRM_SECONDS = 3.0s`); a bind/listen failure (`OSError: address already in use`) completes the task with an exception inside that window and records nothing, while a serve loop that survives the window — or finishes cleanly (a real short session) — records the daily row. Extended the existing bind-failure regression (`test_serve_failure_propagates_after_cleanup`) to assert `capture_mcp_serve_started` is never called, and added the successful-serve counterpart asserting exactly one row.
- **Issue link**: created #2296 and prepended `Refs #2296` to the PR body.

Verification: mcp CLI suites 81 passed, ruff format/check clean, mypy clean on the changed module.

Q00:
Thanks for the approval. On the two follow-up findings — acknowledged, but deferring both deliberately rather than iterating further in this PR:

- **Early-shutdown race on the attachment row**: the window is 3 seconds, the event is deduplicated to one row per user/day, and the losing scenario requires an orphan-exit/signal inside that window on the user's *first* session of the day. Any later real session that day records the identical row anyway, so the funnel error this can introduce is bounded to near zero. Not worth more lifecycle complexity now; happy to revisit if the data ever shows otherwise.
- **`verify_cause: null` vs telemetry `unknown` on digest-unavailable finalization**: the analytics side is already consistent (`unknown`); the local event carrying `null` for a genuinely unclassifiable condition is honest rather than wrong. A cosmetic alignment can ride along with the next orchestrator change.

Diminishing returns from here — four rounds have converged every substantive path (gate, quarantine, coordinator, settlement, attachment). Merging as approved.

Q00:
Round-5 blocker resolved via the second option you offered, in 7db2ecbd: `install_started` is now **explicitly excluded** from the daily-deduplication contract in TELEMETRY.md (event table + notes) and the PR body. The exclusion is deliberate, not an omission: per-invocation rows are the point — a started-without-completed sequence is exactly the install drop-off this event exists to measure, retry runs included, and installer volume is self-bounding (~374 rows across all of August, i.e. 0.04% of the free tier). Daily-deduplicating it would erase the retry signal while saving effectively nothing.

The remaining follow-up (early-shutdown race on the attachment row) stays deferred per the earlier comment: bounded to a 3-second window on a once-per-day row, and any later real session that day records the identical row. This PR has now cycled five rounds on progressively narrower slices of the same telemetry-boundary theme; the contract as documented and the code now agree everywhere, so I consider the scope converged.

Issue #2308, opened 2026-08-30
install: Windows users without Python/uv/Git have no one-command install path (PowerShell)


Q00:
## Problem

`scripts/install.sh` is bash-only and requires uv, pipx, or Python >= 3.12 to already exist (`scripts/install.sh` step 1/4 never bootstraps uv). A Windows machine with none of them cannot install Ouroboros without first setting up WSL 2 or a Python toolchain by hand. The README only shows the `curl | bash` one-liner.

The Claude Code plugin path does not actually need a system Python: `.claude-plugin/.mcp.json` launches the MCP server through `uvx --python '>=3.12'`, skills fall back to `uv run`, and hooks fail open. The real prerequisites on native Windows are Git and uv.

## Proposed

A `scripts/install.ps1` runnable as `irm .../install.ps1 | iex` on PowerShell 5.1+ that installs Git and uv through winget when missing, runs `uv tool install ouroboros-ai` with the same pins as `install.sh`, and wires setup + the Claude Code plugin. README/getting-started should show both one-liners.

Related: #2050 (single-file exe distribution — a different, longer-term path).

Issue #2050, opened 2026-08-10
Distribution: shipping a single-file Windows exe as a release asset would open winget, Scoop, and direct download


Q00:
The codebase already carries deliberate Windows support: UTF-8 stream normalization at the CLI boundary (`cli/streams.py`), a `tasklist`-based process liveness fallback in `mcp doctor` (`cli/commands/mcp_doctor.py`), APPDATA-based OpenCode config resolution (`cli/opencode_config.py`), and win32 paths in the OpenCode bridge with tests. `pipx install 'ouroboros-ai[mcp]'` works on Windows today.

What Windows users lack is a native distribution path, and one artifact would open three of them:

1. **winget** — the community repo only accepts MSIX/MSI/APPX/exe installers; script-based installers are explicitly unsupported, and the manifest `InstallerType` enum has no pip/PyPI value. A pure PyPI package cannot be listed. Notably, aider, goose, and opencode are all absent from winget too — same wall.
2. **Scoop** — takes portable exe/zip artifacts with far less ceremony once one exists.
3. **Direct download** — a release asset for users who won't install Python tooling first.

Suggested scope: a PyInstaller (or pyapp) build job in the release workflow producing `ouroboros-windows-x64.exe`, uploaded as a release asset.

Known costs, so this is a real trade-off rather than a free win: PyInstaller onefile binaries trigger antivirus false positives often enough that some projects sign their builds or ship onedir zips instead; the build adds a Windows CI job to maintain; and the exe would need the `[mcp]` extra baked in to be useful for the Claude Code / Codex plugin path. If those costs outweigh Windows demand right now, closing this with "pipx is the supported path" is a fine outcome — this issue is mainly to record why winget listing is blocked, so the next person who asks gets the researched answer.

PR #2309, opened 2026-08-30
feat(install): PowerShell installer for Windows without Python

Q00:
## Summary

Adds `scripts/install.ps1` so a Windows machine with no Python, uv, or Git can install Ouroboros with one PowerShell command, and documents that path next to the existing `curl | bash` one-liner. `install.sh` never bootstraps uv, so a Python-less Windows user had no entry point short of setting up WSL 2 first.

```powershell
irm https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.ps1 | iex
```

## Review boundary

- **User problem**: A native Windows user without Python/uv/Git cannot install Ouroboros with one command; the README only shows the bash one-liner.
- **Inputs and execution conditions**: Windows PowerShell 5.1 or pwsh 7 on native Windows, run as `irm | iex` or as a bound scriptblock with `-Runtime/-Reconfigure/-Pre`. Options also come from `OUROBOROS_INSTALL_RUNTIME`, `OUROBOROS_INSTALL_RECONFIGURE`, `OUROBOROS_INSTALL_PRE` (same names as `install.sh`). Network access to winget/astral.sh/PyPI.
- **Promised contract**: (1) Git >= 2.36 and uv are present after step 1 or the script exits non-zero with remediation. (2) `uv tool install ouroboros-ai` is invoked with exactly the pins `install.sh` uses (`click`, `textual`, `textual-serve`, and the per-backend `claude-agent-sdk`/`anthropic`/`mcp`/`litellm` pins); `[all]` uses the `>=3.12,<3.14` Python spec. (3) Backend resolution order equals `install.sh`: explicit > saved `~/.ouroboros/config.yaml` runtime (`claude_mcp` → `claude-cli`) > single detected CLI > prompt (non-interactive defaults match the bash script). (4) `ouroboros setup --runtime <b> --non-interactive` failure exits with that status; `setup refresh` and the Claude plugin install are non-fatal, as in `install.sh`. (5) The script emits no installer-owned events (`install_started`/`install_completed`); its `ouroboros setup` calls are ordinary CLI invocations and keep the normal `command_run` telemetry controls, same as under `install.sh`.
- **Implementation boundary and owner**: New file `scripts/install.ps1` plus docs (README en/ko/zh-CN, `docs/getting-started.md`, `docs/platform-support.md`, `TELEMETRY.md`, `UNINSTALL.md`) and `tests/unit/scripts/test_install_ps1.py`. No Python source changes; no change to `install.sh`, the plugin manifest, or `ouroboros setup`. Owner: installer/docs (maintainer).
- **Non-goals**: Installer-owned telemetry events (`install_started`/`install_completed`) on Windows — the bash implementation validates config through the real schema and porting it blind risks polluting the metric; documented as absent in `TELEMETRY.md`. Suppressing CLI telemetry inside the installer's `setup` subprocesses is also out of scope: it would diverge from `install.sh` and override the user's default silently. Native-Windows Codex support (still WSL 2; the script warns). The optional `ouroboros config` GUI prompt at the end of `install.sh`. A single-file exe (#2050). Changing native Windows from "experimental".
- **Evidence**: `tests/unit/scripts/test_install_ps1.py` — pin set equality with `install.sh`, pyproject pin coverage, Python/click spec parity, no-telemetry assertion, and a `pwsh` parse check (runs on ubuntu CI where pwsh is present). Locally with portable pwsh 7.6.5: parse clean, PSScriptAnalyzer clean apart from intentional `Write-Host` and the uv-fallback `Invoke-Expression`; `irm | iex` and scriptblock-bound parameter forms both verified to bind `-Runtime/-Reconfigure` and the env var; `Get-SavedRuntime` maps a fixture `runtime_backend: "claude_mcp"` to `claude-cli`. **Not yet run end-to-end on a real Windows machine** — winget flow, `uv python install --default` placement, and the Microsoft Store `python` alias interaction need a live check before this is called more than experimental.

## Test plan

- [x] `uv run pytest tests/unit/scripts/test_install_ps1.py tests/unit/scripts/test_install_runtime_selection.py -q` — 125 passed
- [x] `uv run ruff check` / `ruff format --check` clean on the new test
- [x] `uv run mypy src/` — no `src/` changes
- [ ] End-to-end run on a native Windows machine (winget-fresh) — follow-up before promoting beyond experimental

## Related issues

Refs #2308
Related: #2050

### DIFF
--- a/README.ko.md
+++ b/README.ko.md
@@ -41,7 +41,13 @@
 </p>
 
 ```bash
+# macOS / Linux / WSL 2
 curl -fsSL https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.sh | OUROBOROS_INSTALL_REF=readme-hero-ko bash
+```
+
+```powershell
+# Windows (PowerShell) — Python 없어도 됩니다. Git과 uv를 대신 설치합니다
+irm https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.ps1 | iex
 ```
 
 <p align="center"><sub>설치는 위 한 줄입니다. 이후 코딩 에이전트 안에서 <code>ooo setup</code>을 한 번 실행하세요. 자세한 내용은 <a href="#빠른-시작">빠른 시작</a>에 있습니다.</sub></p>
@@ -139,8 +145,19 @@ Ouroboros는 이 철학을 **Double Diamond** 구조로 풀어냅니다:
 **설치** — 한 줄이면 전부 자동:
 
 ```bash
+# macOS / Linux / WSL 2
 curl -fsSL https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.sh | OUROBOROS_INSTALL_REF=readme-ko bash
 ```
+
+```powershell
+# Windows (PowerShell 5.1+ 또는 pwsh 7+) — 미리 설치할 것 없음
+irm https://raw.githubusercontent.com/Q00/ouroboros/main/scripts/install.ps1 | iex
+```
+
+Windows 설치 스크립트는 Git과 uv가 없으면 winget으로 설치하고, Python은 uv가
+직접 내려받습니다. 그다음 `ouroboros-ai`를 설치하고 발견한 호스트를 연결합니다.
+네이티브 Windows는 실험 단계이고 Codex CLI는 WSL 2가 필요합니다.
+[플랫폼 지원](./docs/platform-support.md)을 참고하세요.
 
 **첫 명령** — AI 코딩 에이전트를 열고 아래 두 명령을 순서대로 입력하세요:
 
--- a/README.md
+++ b/README.md
@@ -47,7 +47,13 @@
 </p>
 
 ```bash
+# macOS / Linux / WSL 2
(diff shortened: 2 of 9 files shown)


Q00:
Addressed Blocker 1 in `103827d61`.

Confirmed: `ouroboros setup` / `setup refresh` go through `cli/main.py` → `telemetry.capture_cli_command`, so "sends no telemetry" was too strong. Chose to **revise the contract rather than suppress telemetry in the subprocesses**: `install.sh` makes the identical `setup` calls with the identical `command_run` rows, and forcing `OUROBOROS_TELEMETRY=0` only inside the Windows installer would silently override the user's default and make the two installers disagree. The promise is now the one that holds — the installer emits no installer-owned events (`install_started`/`install_completed`) — in `install.ps1`, `TELEMETRY.md`, `docs/getting-started.md`, and the test (`test_ps1_emits_no_installer_events`, which also pins that the script's only network call is fetching the uv installer). PR body updated to match.

Follow-up 1 (native-Windows winget-fresh smoke) stays open as documented; it is the gate for moving beyond experimental.

Issue #2311, opened 2026-08-30
ooo run: ~40% of executions fail on verifier false negatives and environment faults — raise run success to ≥80%


Q00:
## Measurement (local event store, 2026-07-01 → 08-04, `~/.ouroboros/ouroboros.db`)

- Per execution: **51 completed / 35 failed (59%)**.
- Per AC session, 181 failures by cause:

| Cause | Count | Share |
|---|---|---|
| `Fat-harness verifier failed (unsupported evidence claims: files_touched/tests_passed …)` | ~100 | 55% |
| `Verify gate failed: verify_command exited with status 1` | 19 | 10% |
| Provider proxy 502/503 (`auth_unavailable`, `unknown provider for model`) | ~19 | 10% |
| `Claude Agent SDK is not installed` (SDK runtime selected, SDK absent) | 15 | 8% |
| Codex configuration drift, misc | ~5 | 3% |

- Of 57 distinct ACs rejected by the verifier, only 13 ever completed on a later attempt (avg 2.4 attempts) — the rejection is systematic, not a phrasing problem.
- `codex_cli` persists **0** `execution.tool.completed` events against 6,830 `execution.tool.started`, so a rejected leaf cannot be audited after the fact.

## Reproduced false negatives (deterministic, `_verify_atomic_evidence_against_runtime_messages` fed codex-shaped `item.started`/`item.completed` events)

A leaf whose transcript shows `Edit hello.py` and `/bin/zsh -lc 'python3 -m pytest --doctest-modules -q hello.py'` (real July case `exec_33b50bc500a8`) is rejected when:

- the file was written through a shell command (`cat > f <<EOF`, `sed -i`, scaffolding, `git apply`) — no `file_change` item, no authenticated Bash lease → `files_touched` unsupported;
- the Codex completion carries no `exit_code`, or an empty/truncated `aggregated_output` → `tests_passed` unsupported even with `1 passed` in the output;
- `task_cwd` differs from the worktree the runtime reports → `files_touched` unsupported.

Root cause: the verifier's ground truth is the *shape of the transcript*, not the workspace or the harness's own observation. Each runtime encodes tools differently, so every new host needs another matcher (#2245 is one such patch).

## Levers

- **L1a** ground `files_touched` in a before/after workspace snapshot the dispatcher takes itself (unforgeable, runtime-independent; unchanged files still rejected).
[...]

PR #2312, opened 2026-08-30
feat(orchestrator): back files_touched claims with a harness workspace observation

Q00:
## Summary

The transcript verifier rejected leaves as `FABRICATION_SUSPECTED` for files that demonstrably changed on disk, because `files_touched` could only be proven from the transcript's *shape* (Edit/Write path values, or Bash lease targets). In the July event store that is the largest run-failure family (~55% of AC failures; 13 of 57 rejected ACs ever passed on retry). The dispatcher now fingerprints the workspace before and after the leaf and hands the verifier its own unforgeable observation of which paths changed.

## Review boundary

- **User problem**: `ooo run` fails an AC whose leaf really wrote the claimed file (heredoc, `sed -i`, scaffolding CLI, `git apply`, Python one-liner) because no transcript shape backs the `files_touched` claim.
- **Inputs and execution conditions**: any atomic leaf dispatched through `LeafDispatcher.stream` with a resolvable `task_cwd` (or adapter working directory); fat-harness verification via `_verify_atomic_evidence_against_runtime_messages`. All runtimes; no config flag.
- **Promised contract**: (1) A `files_touched` claim is supported when the harness saw that workspace-relative path appear or change (size or mtime) between leaf start and stream end. (2) Only positive support: unchanged files, out-of-workspace or `..` claims, and paths outside a truncated snapshot's budget remain unsupported — a stale file still cannot prove this run touched it. (3) The observation is a `WorkspaceObservation` *instance* inside message data; runtime adapters produce JSON, so a leaf cannot forge it (`observation_from_message` is the type-check boundary; a dict with the same key is rejected). (4) An empty runtime stream still classifies as `transcript_missing_infrastructure`: the observation is only inserted when at least one non-final runtime message arrived, and the verifier excludes it from its emptiness check. (5) `commands_run`/`tests_passed` are untouched. (6) Snapshot cost is bounded: ignore list (`.git`, `node_modules`, `.venv`, caches, …), 50k entries, 5 s; past either bound the observation is marked truncated and withholds support.
- **Implementation boundary and owner**: `src/ouroboros/orchestrator/` only — new `evidence/harness_observation.py`, a new support tier at the top of `evidence/claims._runtime_message_supports_file_reference`, the emptiness check in `evidence/verification.py`, and the snapshot/insert in `leaf_dispatcher.stream`. `parallel_executor.py` is not modified (grandfathered ratchet). No new config, no persistence change, no MCP surface.
- **Non-goals**: `tests_passed` re-execution (next PR), codex tool-result persistence, runtime preflight/transient classification (#2311 L2), the final-gate `workspace_mutated` rule, content hashing (size+mtime is sufficient for "changed during this window" and cheap), emitting the observation as an event.
- **Evidence**: `tests/unit/orchestrator/evidence/test_harness_observation.py` (13 tests): snapshot diff, ignore list/symlinks, budget truncation, forgery boundary, insertion order, the July false negative turning into a pass end-to-end through the verifier, stale/unchanged and out-of-workspace claims still rejected, and observation-only streams still classified as transcript-missing. Regression: `tests/unit/orchestrator/test_parallel_executor.py`, `test_execution_authority.py`, `evidence/` — 704 passed; `check-module-size.py --baseline-ref origin/main` OK; ruff/mypy clean on touched files.

Measurement plan (#2311): 10 live `ouroboros run` executions of a fixed cli-todo seed on `main` vs this branch, with an independent product smoke as ground truth. Baseline is running now; results will be posted on the issue.

## Test plan

- [x] `uv run pytest tests/unit/orchestrator/evidence/ tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_execution_authority.py -q`
- [x] `uv run ruff check` / `ruff format --check` on touched files
- [x] `uv run mypy` on touched files
- [x] 10-run live before/after on the cli-todo seed: harness success 0/10 → 10/10 (#2311)

## Related issues

Refs #2311

### DIFF
--- a/src/ouroboros/orchestrator/evidence/claims.py
+++ b/src/ouroboros/orchestrator/evidence/claims.py
@@ -348,6 +348,7 @@ def _runtime_messages_support_file_claim(
                     index=index,
                     task_cwd=task_cwd,
                     allow_bash_command_text=False,
+                    allow_observation=False,
                 )
                 for index, message in enumerate(messages)
             ):
@@ -387,6 +388,7 @@ def _runtime_message_supports_file_reference(
     index: int,
     task_cwd: str | None,
     allow_bash_command_text: bool = True,
+    allow_observation: bool = True,
 ) -> bool:
     """Return True when one message plausibly reports touching a file reference."""
     normalized_reference = reference.strip().lower()
@@ -395,10 +397,11 @@ def _runtime_message_supports_file_reference(
     observation = observation_from_message(message)
     if observation is not None:
         # The harness saw the workspace change during this leaf's window; that
-        # is support the leaf could not have narrated into existence. Basename
-        # probing never reaches here with a path that changed, so an unchanged
-        # claim stays unsupported.
-        return observation.supports_file_claim(reference)
+        # is support the leaf could not have narrated into existence. The
+        # observation only ever answers for the full workspace-relative claim:
+        # the basename fallback passes ``allow_observation=False`` so a changed
+        # ``foo.py`` cannot vouch for an unchanged ``nested/foo.py``.
+        return allow_observation and observation.supports_file_claim(reference)
     if message.tool_name == "Bash":
         return _bash_message_mutates_file_reference(
             message,
--- a/src/ouroboros/orchestrator/evidence/claims.py
+++ b/src/ouroboros/orchestrator/evidence/claims.py
@@ -10,6 +10,7 @@ import sys
 
 from ouroboros.orchestrator.adapter import AgentMessage
 from ouroboros.orchestrator.evidence.common import _flatten_evidence_values
+from ouroboros.orchestrator.evidence.harness_observation import observation_from_message
 from ouroboros.orchestrator.evidence.shell_parsing import (
     _has_trailing_output_filter_pipeline,
     _normalized_command_claim_aliases,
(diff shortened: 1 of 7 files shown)


Q00:
Round 1 blockers addressed at 159a77451b3f65e167e557fe2f73db75e4f51077 (rebased onto current main):

**Blocker 1 — basename fallback borrowing observation support** (claims.py): `_runtime_message_supports_file_reference` gained `allow_observation`; the basename pass sets it False alongside `allow_bash_command_text=False`. The observation now answers only for the exact workspace-relative claim, so a changed `foo.py` can no longer vouch for an unchanged `nested/foo.py`. Regression: `test_duplicate_basename_cannot_borrow_observation_support`.

**Blocker 2 — truncated pre-snapshot treating unseen paths as changed** (harness_observation.py): `diff_workspace_snapshots` skips paths absent from a truncated pre-snapshot — absence under budget truncation is uncertainty, not creation. Paths present in both snapshots with a changed fingerprint still count. Regressions: `test_truncated_pre_snapshot_withholds_paths_it_never_saw` (the review's `max_entries=1` repro) and `test_truncated_snapshots_still_observe_a_path_seen_in_both`.

**Blocker 3 — SessionSignal cursor desync** (session_signal_followup.py, parallel_executor.py): `CompletedProviderTurn` now captures `message_list_length = len(state.messages)` at capture time; both the follow-up reply slice (`messages[boundary:]`) and pre-entry `restore` truncation use the list boundary, while `message_count` stays runtime-only for heartbeats. A successful first signal can no longer leak its tail into an error-only second signal's acknowledgement window, and restore no longer deletes the primary's observation. Regression: `test_follow_up_slice_and_restore_use_the_list_boundary`.

**Follow-up P2 #1** also taken: the time-budget deadline is now checked inside the filename loop, so a single large directory cannot overrun the 5s bound without marking `truncated`. P2 #2 (case folding) is left as-is deliberately — the lowercase normalization matches the existing evidence-tier reference normalization in claims.py; changing identity semantics for case-sensitive filesystems belongs to a coordinated change across both. P2 #3 (sibling attribution under concurrent ACs) stays with the `parallel_executor` owner as the review itself scoped it.

Local gates: ruff format/check, mypy, `tests/unit/orchestrator` 5546 passed (1 pre-existing local-env failure in `test_n_version_tournament` undecodable-filename, fails identically on a clean tree).

PR #2313, opened 2026-08-30
feat(orchestrator): re-run unprovable tests_passed claims in the harness

Q00:
## Summary

Stacked on #2312. When the transcript cannot prove a `tests_passed` claim (Codex completions without `exit_code`, truncated output, missing tool results), the harness re-runs the claimed test command itself in the workspace and judges its own exit status and output. Second lever of #2311.

## Review boundary

- **User problem**: a leaf that really ran a passing test suite is rejected as `FABRICATION_SUSPECTED` because the runtime delivered no (or partial) output for that Bash call.
- **Inputs and execution conditions**: a successful atomic leaf (`state.success`, parseable evidence JSON) whose `tests_passed` claims are not backed by the transcript, with a resolvable `task_cwd` and the verify-gate POSIX shell available. All runtimes; no flag.
- **Promised contract**: (1) At most `MAX_REEXECUTED_COMMANDS` (3) distinct commands are re-run, chosen from the claim itself, reported `commands_run`, and unwrapped transcript Bash commands, each only if `_looks_like_test_command`. (2) They run through the same shell, `sanitized_verify_environment()`, cwd, and `verify_command_timeout_seconds` as the verify gate; a start failure yields no observation, a timeout is recorded as a failed run. (3) A re-executed run backs a claim only when exit is 0, its output proves tests ran and passed (same `_text_proves_test_execution_success` rule as transcript output), and the command targets the claim (`_test_command_targets_claim`). (4) A node-id/file claim must additionally name a file this run touched (`_runtime_messages_support_file_claim`, which includes the #2312 workspace observation); a pre-existing stale suite cannot be claimed — `test_fat_harness_verifier_rejects_bare_pytest_for_unmentioned_stale_test` still passes. (5) `commands_run` claims equal (whitespace-normalized) to a successful re-executed command are backed. (6) Nothing is re-run when the transcript already proves every claim, when the leaf failed, or when there is no evidence JSON.
- **Implementation boundary and owner**: `src/ouroboros/orchestrator/` — new `evidence/test_reexecution.py`, `CommandObservation` on the #2312 observation, a first tier in `evidence/test_detection._runtime_messages_support_test_claim`, a `commands_run` tier in `evidence/verification.py`, and `LeafDispatcher._attach_test_reexecution`. `parallel_executor.py` unchanged.
- **Contract change to flag**: `test_fat_harness_verifier_rejects_exit_code_only_test_success` asserted that a zero exit with no output is rejected. It is rewritten as `test_fat_harness_exit_code_only_test_output_is_settled_by_harness_reexecution[True/False]`: the transcript alone is still insufficient, and the verdict now follows the harness's own run — accepted when the suite passes, rejected when it fails. Re-running a leaf-authored test command executes agent-authored shell text once more in the workspace, at the same trust level the leaf already had there and under the verify gate's sandboxed environment/timeout.
- **Non-goals**: re-running non-test commands; changing which runtimes emit tool results; the final-gate `workspace_mutated` settlement rule (separate PR, see #2311); Windows shells beyond what `verify_shell` already resolves.
- **Evidence**: `tests/unit/orchestrator/evidence/test_test_reexecution.py` (selection rules and cap, real subprocess exit/timeout recording, claim-support matrix incl. failed/timed-out/empty-output/non-test runs, stale-file guard, verifier end-to-end on the Codex-shaped unprovable transcript); executor + authority + evidence suites 719 passed; mypy clean; `check-module-size.py` OK.

## Test plan

- [x] `uv run pytest tests/unit/orchestrator/evidence/ tests/unit/orchestrator/test_parallel_executor.py tests/unit/orchestrator/test_execution_authority.py -q`
- [x] ruff / mypy on touched files; module-size gate
- [x] 10-run live before/after: harness success 0/10 → 10/10 (#2311)

## Related issues

Refs #2311

### DIFF
--- a/src/ouroboros/orchestrator/evidence/harness_observation.py
+++ b/src/ouroboros/orchestrator/evidence/harness_observation.py
@@ -1,4 +1,4 @@
-"""Harness-observed workspace changes as unforgeable ``files_touched`` support.
+"""Harness observations as unforgeable evidence support.
 
 The transcript verifier proves ``files_touched`` claims from the runtime
 transcript: structured ``Edit``/``Write`` path values, or shell mutation targets
@@ -20,6 +20,12 @@ in :func:`observation_from_message` is the forgery boundary.
 Only positive support is granted: a path that did not change stays unsupported,
 so a stale file in the workspace still cannot prove that this run touched it,
 and a truncated snapshot (workspace over budget) can only withhold support.
+
+The same message also carries :class:`CommandObservation` records: test
+commands the harness re-executed itself in the workspace when the transcript
+could not prove a ``tests_passed`` claim (see ``evidence/test_reexecution.py``).
+The exit status and output there are the harness's own, so they can back a
+``tests_passed`` or ``commands_run`` claim the same way the transcript would.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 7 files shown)


Q00:
Addressed the blocker in the latest push (see the new top commit). PR #2313: re-executed candidates now run as a direct argv with no shell (safe_test_argv rejects any shell metacharacter at selection and execution; the `$(touch marker)` probe is a regression test asserting the marker is never created), and re-execution is gated on the leaf's effective tool list containing Bash plus run_verify_commands being explicitly enabled. PR #2314: a replay whose outcome is workspace_mutated now folds into the settlement-wide mutation state — no further commands run and the existing complete-success-set invalidation rejects every provisional success; a two-AC regression (contract AC + contract-less sibling) proves both flip to FAILED.

Q00:
Round 2 blocker addressed at 9beb7f9c1161f28f1b1e6e5accde026c011887ee (also rebased onto the reworked #2312 tip, which itself is rebased on current main):

**Env-prefix parser/executor split** (test_reexecution.py): new `safe_test_invocation` splits a claimed command into a controlled environment delta plus the executable argv, peeling a bare `env` prefix with the same rule the recognizer's `_strip_env_prefix` uses (`_is_env_assignment` is imported from shell_parsing so the two can't drift). Both `select_test_reexecution_commands` and `reexecute_test_commands` consume this one representation; execution merges the delta over the sanitized verify env. Every token still passes the shell-metacharacter gate first, so assignment values are literal bytes under both shell and direct-argv semantics — `FLAG="$(id)" pytest` remains rejected, and assignments without a command are not runnable.

Regressions added: `test_env_prefix_becomes_a_delta_not_an_executable` (unit: prefix/env-wrapper/bare/assignment-only/injection cases) and `test_env_prefixed_command_runs_with_the_delta_applied` (the review's repro: `REEXEC_FLAG=yes python -m pytest -q test_env.py` is selected, runs, exits 0 with the delta visible to the test, and yields a successful `CommandObservation`).

Local gates: ruff format/check, mypy (617 files), `tests/unit/orchestrator/evidence` 73 passed.

PR #2314, opened 2026-08-30
fix(orchestrator): replay verify_command at final settlement instead of rejecting

Q00:
## Summary

Stacked on #2313 (branch only; the code change is independent of #2312/#2313 and is 17 lines in `_settle_verify_gate_results`). In the live cli-todo baseline for #2311, **3 of 3 runs so far fail solely at final settlement** with `Final acceptance rejected because the workspace changed after verify_command completed; replaying verify_command is not permitted`, while every level reported success and the built CLI passes an independent smoke (3/4, 4/4, 3/4 probes). Any multi-level seed where a later AC edits a file an earlier contract covered — or just adds tests to the tree — hits this.

## Review boundary

- **User problem**: a passing, intact AC is discarded at the end of the run because a sibling AC touched the workspace after its verify_command ran.
- **Inputs and execution conditions**: `_settle_verify_gate_results` in the non-coordinator path (`coordinator_revalidated=False`), for a successful command-bearing AC whose cached `workspace_digest` differs from the final digest and whose expected artifacts are present. `run_verify_commands` enabled (default).
- **Promised contract**: (1) Instead of rejecting, settlement re-runs the AC's `verify_command` once against the final workspace through the same Foundation A authority entry (`_FOUNDATION_A_ENTRY_RUN_AC_VERIFY_GATE`) — same shell, sanitized env, timeout, mutation detection, and output assertion. (2) A passing replay keeps the AC and stores the fresh `_VerifyGateOutcome`. (3) A failing replay rejects the AC with `Final acceptance rejected because verify_command failed on the final workspace: <reason>` and the replay's own `cause` (falling back to `workspace_mutated`), emitting the same `execution.verify.failed` event and telemetry as before. (4) The coordinator-revalidation branch (`did not replay verify_command`) and the "a final verifier mutated the workspace" invalidation are unchanged. (5) A verify_command therefore runs at most one extra time per AC, and only when the workspace changed after it passed.
- **Policy change to flag for the maintainer**: the removed comment argued a command "may have effects outside the workspace (DB/API/deployment writes)" and must never be replayed. The seed contract already treats `verify_command` as an observation: the gate rejects a command that mutates the workspace (`workspace_mutated`), the command is re-run on every AC retry, and `--skip-completed` runs it against the working tree. Replaying once more at settlement is consistent with that contract; a seed whose verify_command has external side effects is out of contract today.
- **Implementation boundary and owner**: `src/ouroboros/orchestrator/parallel_executor.py` (grandfathered ratchet: net line count unchanged, `check-module-size.py --baseline-ref origin/main` OK) and one rewritten test.
- **Non-goals**: coordinator revalidation replay; changing digest scope; verify_command idempotency enforcement.
- **Evidence**: `test_final_settlement_rejects_stale_command_without_replay` rewritten as `test_final_settlement_replays_stale_command_against_final_workspace[before/after]` — the command's counter proves exactly one replay; a still-true condition keeps the AC, a now-false condition rejects it with the replay reason. `test_parallel_executor_verify_by_default.py` + `test_route_escalation_live.py` 268 passed; executor + authority 666 passed; mypy clean.

## Test plan

- [x] targeted suites above; ruff/mypy; module-size ratchet
- [x] 10-run live before/after on the cli-todo seed: 0/10 → 10/10 (#2311; all 10 baseline failures were this rule)

## Related issues

Refs #2311

### DIFF
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -5419,6 +5419,15 @@ class ParallelACExecutor:
                     replayed = await _invoke_execution_authority_entry(
                         self, _FOUNDATION_A_ENTRY_RUN_AC_VERIFY_GATE, spec=spec, cwd=cwd
                     )
+                    if replayed.workspace_mutated:
+                        # The replay itself changed the workspace (or made its
+                        # digest unreadable): every provisional success is now
+                        # stale. Fold it into the settlement-wide mutation
+                        # state so no further command runs and the complete
+                        # success set is invalidated below.
+                        verify_mutated_workspace = True
+                        settled.append(result)
+                        continue
                     if not replayed.passed:
                         individual_failures[result.ac_index] = (
                             "Final acceptance rejected because verify_command failed on "
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -5414,19 +5414,19 @@ class ParallelACExecutor:
 
                 if cached_digest != final_digest:
                     # A later worker changed the workspace after this command
-                    # passed.  Do not replay an unrestricted shell contract:
-                    # effects outside the workspace (DB/API/deployment writes)
-                    # cannot be detected by the digest.  The stale evidence is
-                    # therefore rejected at the final boundary.
-                    individual_failures[result.ac_index] = (
-                        "Final acceptance rejected because the workspace changed "
-                        "after verify_command completed; replaying verify_command "
-                        "is not permitted.",
-                        outcome,
-                        "workspace_mutated",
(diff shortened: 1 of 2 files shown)


Q00:
Addressed the blocker in the latest push (see the new top commit). PR #2313: re-executed candidates now run as a direct argv with no shell (safe_test_argv rejects any shell metacharacter at selection and execution; the `$(touch marker)` probe is a regression test asserting the marker is never created), and re-execution is gated on the leaf's effective tool list containing Bash plus run_verify_commands being explicitly enabled. PR #2314: a replay whose outcome is workspace_mutated now folds into the settlement-wide mutation state — no further commands run and the existing complete-success-set invalidation rejects every provisional success; a two-AC regression (contract AC + contract-less sibling) proves both flip to FAILED.

Issue #2326, opened 2026-08-31
GjcRuntime has no termination path: worker sessions and nested mcp serve leak on every run (success and failure)


Q00:
## Symptom (production, 2026-08-31)

A failed \`execute_seed --runtime gjc\` run left 19 gjc worker (bun) processes and 18 nested \`ouroboros mcp serve --runtime gjc\` children alive 17+ minutes after the run ended, each holding DB connections — a major contributor to the pool-exhaustion incident.

## Root causes (audited on main @ v0.52.0)

1. \`GjcRuntime._build_runtime_handle\` (orchestrator/gjc_runtime.py:120-160) never calls \`bind_controls\`, so \`can_terminate\` is False and **every** existing termination path (per-AC finally in parallel_executor, run-level finallys in runner, cancel/escalation paths) is a silent no-op for gjc. Only codex/opencode bind a terminate callback.
2. \`GjcCoordinatorClient.stop_session\` exists (gjc/sdk_client.py:390) but is never called from the agent-runtime path — the finally only disconnects the coordinator client (gjc_runtime.py:376-381). The success path leaks sessions too.
3. gjc workers inherit the project env and cwd verbatim (sdk_client.py:104, bypassing \`build_child_env\`), read the project \`.mcp.json\`, and spawn one nested \`ouroboros mcp serve\` per AC attempt. claude has \`--strict-mcp-config\` and codex has \`enabled=false\`; gjc has no equivalent suppression.
4. No run-scoped reaper: \`execute_parallel\` has no outer try/finally sweeping runtimes; \`_close_adapter\` only calls \`aclose\` where it exists (not on GjcRuntime).

## Fix plan (single PR)

- Call \`stop_session\` in the gjc turn finally on terminal states.
- Bind a terminate callback on the gjc RuntimeHandle so the already-wired termination paths start working.
- Suppress project MCP config / route env through \`build_child_env\` for gjc workers to stop nested serve spawns.

Issue #2325, opened 2026-08-31
mcp serve zombies survive client death; orphan watchdog and EOF shutdown are ineffective under mcp 2.0


Q00:
## Symptom (production, 2026-08-31)

- 119 \`ouroboros mcp serve\` processes on one machine; 54 holding \`~/.ouroboros/data/ouroboros.db\` open → SQLite write-lock contention → per-process \`QueuePool limit of size 5 overflow 10 reached\` (714 occurrences) → run failures.
- Zombies' fd 0/1/2 are unix sockets with a dead peer (\`->(none)\` in lsof): the client died, the server never exits. Reproduced nightly (29 zombies from one evening).
- One streamable-http serve orphaned to launchd ran 5 days at 40–80% CPU (4,263 CPU-minutes).

## Root causes (audited on main @ v0.52.0)

1. The shutdown EOF trick \`os.close(0)\` (cli/commands/mcp.py:978-981) is a no-op under mcp==2.0.0: the SDK diverts fd 0 to /dev/null and reads the wire from a private dup'd fd (mcp/server/stdio.py).
2. The stdin readline runs in a shielded, **non-daemon** anyio worker thread; even after the serve coroutine returns, the interpreter blocks forever in threading shutdown. \`os._exit\` is explicitly forbidden (mcp.py:967-969), so there is no escape hatch.
3. \`_orphan_watchdog\` disables itself entirely when \`orig_ppid == 1\` (mcp.py:873-874) — a launchd-orphaned server has no lifecycle tether at all. streamable-http additionally has no idle shutdown.
4. CPU runaway: \`JobManager._monitor_job\` polls at 1 Hz and \`get_snapshot\` replays the job's events from row 0 on every call (mcp/job_manager.py:888, 1970), with no backoff.

## Fix plan

- PR A: hard-exit backstop after the shutdown grace period + dead-peer (MSG_PEEK) and \`getppid()==1\` detection in the watchdog.
- PR B: idle shutdown for streamable-http + incremental cursor / backoff for the job monitor loops.

PR #2327, opened 2026-08-31
fix(gjc): reclaim broker-owned worker sessions on terminal turns and termination

Q00:
Refs #2326

## One user problem

A `--runtime gjc` run leaves its worker sessions alive forever — success and failure alike. In the 2026-08-31 incident, 19 leaked gjc worker (bun) clients and 18 nested `ouroboros mcp serve` children were still holding DB connections 17+ minutes after the run failed, driving the SQLite pool exhaustion (`QueuePool limit of size 5 overflow 10 reached`, 714 occurrences).

## Root cause

- `GjcRuntime` never bound a `terminate_callback`, so `can_terminate` was `False` and every already-wired reclamation path (per-AC finally in `parallel_executor`, run-level finallys in `runner`, cancel/escalation) was a silent no-op for gjc. Only codex/opencode bind one.
- The turn `finally` only disconnected the coordinator MCP client; the Broker-owned worker session was never stopped, even on success. `stop_session` existed but had exactly one caller (the one-shot `gjc_llm_adapter`).
- The coordinator env was a raw `os.environ.copy()`, bypassing the shared `build_child_env` recursion guard, so workers inherited the project runtime markers and spawned one nested `ouroboros mcp serve` per AC attempt.

## Change

1. `orchestrator/gjc_runtime.py` — `execute_task` stops the broker session in its `finally` once the turn reaches a terminal (non-resumable) state. `waiting_for_answer` turns and coordinator errors keep the session alive so the existing idempotent resume contract is untouched.
2. `orchestrator/gjc_runtime.py` — every session-bearing `RuntimeHandle` now binds a terminate callback (`connect → stop_session(force) → close`, best-effort), following the codex/opencode `bind_controls` pattern. The wired termination paths start working for gjc with zero changes to them.
3. `gjc/sdk_client.py` — the coordinator env is built through `runtime/child_env.build_child_env`: strips `OUROBOROS_AGENT_RUNTIME`/`OUROBOROS_LLM_BACKEND` and enforces the `_OUROBOROS_DEPTH` ceiling (raises `GjcCoordinatorError(code="recursion_depth_exceeded")`), matching every other backend.

gjc agent-runtime execution paths (`execute_task` / `execute_task_to_result`) and any consumer of gjc `RuntimeHandle.terminate()`. No behavior change for other backends.

- A gjc turn that ends terminal (completed/failed) leaves no live broker session behind.
- gjc handles with a `native_session_id` report `can_terminate=True`; `terminate()` force-stops the session and returns whether a stop was issued.
- Question turns and coordinator errors preserve the session for resume (unchanged resume/idempotency semantics).
- gjc child env carries the standard recursion guard.

`orchestrator/gjc_runtime.py` + `gjc/sdk_client.py` (gjc backend adapter boundary). No new subsystem; reuses `RuntimeHandle.bind_controls` and `runtime/child_env` seams. Owner: @Q00.

## Non-goals

- Serve-side zombie lifecycle fixes (dead-peer EOF, hard-exit backstop, http idle shutdown) — #2325.
- A run-scoped OS-level PID reaper — the terminate-callback seam makes the existing sweeps effective; a broader reaper remains follow-up under #2326 if leaks persist.
- Known gap, intentionally not invented here: gjc has no verified flag to suppress project `.mcp.json` discovery inside worker sessions (claude has `--strict-mcp-config`, codex has `enabled=false`). The env recursion guard bounds the depth; if gjc grows such a flag, wiring it into `start_session` args is a one-line follow-up.

## Evidence

- New tests: `test_terminal_success_turn_stops_broker_session`, `test_terminal_failed_turn_stops_broker_session`, `test_question_turn_keeps_broker_session_alive`, `test_session_handle_can_terminate_and_stops_session`, `test_sessionless_handle_cannot_terminate`, `test_server_env_applies_child_recursion_guard`, `test_server_env_depth_ceiling_raises_coordinator_error` (no real CLI spawns; FakeClient/FakeAdapter).
- Full suite with xdist: 21753 passed; the only 5 failures (`test_picker_indexes.py` perf budgets) reproduce identically on origin/main on the same machine (load-sensitive, pre-existing).
- `ruff format`/`ruff check`/`mypy`: clean.

https://claude.ai/code/session_01CDJPBa4pxVbUvcUcKeT5Ea

### DIFF
--- a/src/ouroboros/gjc/sdk_client.py
+++ b/src/ouroboros/gjc/sdk_client.py
@@ -19,6 +19,7 @@ from uuid import uuid4
 
 from ouroboros.mcp.client.adapter import MCPClientAdapter
 from ouroboros.mcp.types import MCPServerConfig, MCPToolResult, TransportType
+from ouroboros.runtime.child_env import build_child_env
 
 _SERVER_NAME = "gjc-coordinator"
 _TERMINAL_STATUSES = frozenset({"completed", "failed", "cancelled", "superseded"})
@@ -101,7 +102,15 @@ class GjcCoordinatorClient:
     def _server_config(self) -> MCPServerConfig:
         repo_digest = hashlib.sha256(self._cwd.encode("utf-8")).hexdigest()[:16]
         state_root = Path(self._cwd) / ".gjc" / "state" / "ouroboros-coordinator" / repo_digest
-        env = os.environ.copy()
+        # The coordinator forwards this environment to Broker-managed worker
+        # sessions, so it needs the same recursion guard every other backend
+        # applies: strip the Ouroboros runtime markers and bound the depth.
+        env = build_child_env(
+            depth_error_factory=lambda depth, max_depth: GjcCoordinatorError(
+                f"GJC session nesting depth {depth} exceeds the maximum of {max_depth}.",
+                code="recursion_depth_exceeded",
+            ),
+        )
         cli_path = Path(self._cli_path).expanduser()
         if cli_path.name == "gjc" and cli_path.parent != Path("."):
             existing_path = env.get("PATH", "")
--- a/src/ouroboros/gjc/sdk_client.py
+++ b/src/ouroboros/gjc/sdk_client.py
@@ -19,7 +19,7 @@ from uuid import uuid4
 
 from ouroboros.mcp.client.adapter import MCPClientAdapter
 from ouroboros.mcp.types import MCPServerConfig, MCPToolResult, TransportType
-from ouroboros.runtime.child_env import build_child_env
+from ouroboros.runtime.child_env import DEFAULT_OUROBOROS_STRIP_KEYS, build_child_env
 
 _SERVER_NAME = "gjc-coordinator"
 _TERMINAL_STATUSES = frozenset({"completed", "failed", "cancelled", "superseded"})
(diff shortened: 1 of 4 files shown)


Q00:
Both round-1 blockers addressed at 4ddf562ba (rebased onto current main):

**Blocker 1 — lifecycle authority escaping at generator boundaries**: `execute_task` now owns the complete lifecycle. (a) Terminal turns (success and failure) stop the broker-owned session **before** the result message is yielded — a consumer that takes only the terminal message via `anext()` and never closes the generator observes `client.stopped == [session]` already. (b) The `finally` block reclaims any session bound by `start_session` under caller cancellation and generator finalization too — the cleanup awaits run inside `anyio.CancelScope(shield=True)` so cancellation while `await_turn` blocks still stops the bound session. Only explicitly resumable states keep the session alive: question turns and `GjcCoordinatorError` paths set `keep_session_for_resume`. Regressions: `test_terminal_stop_precedes_the_final_yield` (single `anext`, no aclose) and `test_cancellation_while_awaiting_turn_stops_the_bound_session` (cancel mid-`await_turn` → session stopped, client closed) — both exercise exactly the two probes from the review instead of exhausting the comprehension.

**Blocker 2 — legacy selector inherited by workers**: `_server_config` now passes `strip_keys=("OUROBOROS_RUNTIME", *DEFAULT_OUROBOROS_STRIP_KEYS)` to `build_child_env`, so the public legacy selector (which routes both runtime and LLM choice) is removed along with the two default markers before the coordinator forwards the environment to Broker-managed workers. Regression: the child-recursion-guard test now sets `OUROBOROS_RUNTIME=gjc` and asserts it is absent from the server env.

Local gates: ruff format/check, mypy (614 files), gjc runtime + sdk_client suites 19 passed.

PR #2332, opened 2026-08-31
fix(codex): exempt main-repo-root automatic trust from config drift for worktree cwds

Q00:
Refs #2311

## User problem
Every codex `ooo run` AC after the first completed task fails instantly (0.16s) with `Failed to prepare Codex CLI: Codex configuration changed after runtime initialization; start a new execution session`. Measured live: canonical cli-todo, 10/10 runs affected — codex could never finish a multi-AC seed.

## Root cause
Codex records first-use project trust against the **git main repository root**, while orchestrator tasks run inside a linked worktree. The drift guard's exemption only covered `self._cwd` (the worktree path), so codex's own benign `projects.<main-root>.trust_level = "trusted"` write permanently poisoned the config fingerprint.

codex runtime backend, tasks dispatched in linked git worktrees (the orchestrator's normal layout) or plain directories. No behavior change for non-codex backends.

- The automatic first-use trust entry is exempt for exactly two keys: the runtime cwd and, when the cwd is inside a linked worktree, the main repository root resolved via the worktree's `.git` gitfile (pure file inspection, no subprocess).
- Every other project's trust entry — and any other config key — remains authority-bearing drift and still fails closed.
- Unrecognized `.git` layouts resolve to no extra key (fail-closed to today's behavior).

## Changed subsystem / boundary
`src/ouroboros/orchestrator/codex_cli_runtime.py` only (+ its unit tests). No data/security boundary change: the exemption is narrower than the existing cwd exemption's intent (same benign mutation, correct key).

## Non-goals
Verifier/evidence changes (separate stacked PR), other runtimes' attestation, trust entries the user edits manually.

## Evidence
- Live before/after: codex cli-todo instant-death retries eliminated; with the verify stack on top, 10-run success went 0/10 → 10/10 (#2311 comment).
- New regression test `test_codex_config_fingerprint_ignores_main_root_trust_for_worktree_cwd`; existing drift tests unchanged and passing (226 codex runtime tests).

### DIFF
--- a/src/ouroboros/orchestrator/codex_cli_runtime.py
+++ b/src/ouroboros/orchestrator/codex_cli_runtime.py
@@ -361,6 +361,7 @@ class CodexCliRuntime:
             self._profile_resolution_fingerprint = self._fingerprint_profile_resolution_config()
             self._codex_profile_v2_names = self._codex_profile_v2_names_from_ouroboros_config()
             self._codex_project_trust_baseline = self._read_codex_project_trust_levels()
+            self._codex_trust_exempt_keys_cache: frozenset[str] | None = None
             self._codex_config_fingerprint = self._fingerprint_codex_config_files()
             self._skill_dispatch_registry_fingerprint = self._fingerprint_skill_dispatch_registry()
             self._builtin_mcp_handler_registry_fingerprint = (
@@ -379,6 +380,7 @@ class CodexCliRuntime:
             self._codex_config_fingerprint = None
             self._codex_profile_v2_names = set()
             self._codex_project_trust_baseline = None
+            self._codex_trust_exempt_keys_cache = None
             self._skill_dispatch_registry_fingerprint = None
             self._builtin_mcp_handler_registry_fingerprint = None
             self._runtime_handle_profile_fingerprints = {}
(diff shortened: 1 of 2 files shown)


Q00:
Both round-1 blockers addressed at fad04b8e455ef542f4d42c5deeea919ffda8f764 (rebased onto main, which now includes #2324):

**Blocker 1 — absent-config baseline breaks the first-use exemption**: a missing `config.toml` now hashes as the normalized empty global config (`_stable_global_codex_config_bytes(b"")`) instead of `missing\0`, so the baseline taken before Codex creates the file equals the fingerprint of a file holding only the exempt trust entry — both mean "no non-exempt global settings". Only `config.toml` gets this treatment; profile files keep the `missing` sentinel so an appearing profile file still drifts. Regression: `test_codex_config_fingerprint_exempts_first_use_trust_when_config_absent` (absent baseline → exempt-only file passes `_assert_codex_config_files_unchanged`; the same file with a non-exempt `model` entry drifts).

**Blocker 2 — main-root recognition failed open**: `_git_main_repository_root` now (a) returns None when the ancestor `.git` is a directory — an ordinary checkout is the main repo itself and mints no extra exempt key — and (b) accepts a gitdir target only when `gitdir.parent.name == "worktrees"`, the grandparent is `.git`, and both the per-worktree gitdir and the main `.git` directory actually exist. Submodule-style (`.git/modules/…`), misnamed (`.git/not-worktrees/…`), and nonexistent targets are rejected. Regressions: `test_ordinary_checkout_does_not_mint_a_main_root_exempt_key` (repo-root trust entry now drifts) and `test_malformed_gitdir_layouts_do_not_resolve_a_main_root` (all three malformed shapes → None; the valid layout still resolves).

Local gates: ruff format/check, mypy (615 files), module-size OK (stayed within the 4140 grandfathered budget), `tests/unit/orchestrator/test_codex_cli_runtime.py` 149 passed.

Issue #2338, opened 2026-09-02
Bug: v1.0.4 Seed parsing can rewrite or discard semantic AC keys


wjswlgnsdlqslek:
## Summary
Persisted v1.0.4 `revised_after_qa` Seed compatibility can rewrite a valid canonical semantic AC identity or silently rehash a malformed hash-shaped key.

## Impact
Saved Seeds can lose their explicit acceptance-criterion identity when reloaded, and malformed `ac_` hash-shaped values can be accepted as a different generated identity instead of failing validation.

## Steps to Reproduce
1. From a source checkout on a supported Python version, load a v1.0.4 `revised_after_qa` Seed directly with `semantic_ac_key: ac_a123456789abcdef0`.
2. Load Workflow IR containing `{criterion: "A", semantic_ac_key: "ac_a123456789abcdef"}` through the Seed boundary.

## Expected Behavior
- Direct loaders preserve exact canonical `ac_[a-f0-9]{16}` values and reject all-hex noncanonical-length values through strict validation.
- Workflow IR preserves the supplied `semantic_ac_key` while using `criterion` as the Seed description alias.
- Human-readable legacy labels continue to materialize deterministically, and missing ontology descriptions are filled from nonblank names.

## Actual Behavior
- Compatibility normalization can treat malformed all-hex keys as descriptive legacy labels and rehash them.
- Workflow IR normalization can reduce a `{criterion, semantic_ac_key}` mapping to a bare string, discarding the key before Seed validation.

## Acceptance Criteria for Fix
- [ ] `Seed.from_dict` and `Seed.model_validate` reject malformed all-hex noncanonical-length keys rather than rehashing them.
- [ ] Exact canonical keys survive Workflow IR `{criterion, semantic_ac_key}` loading byte-for-byte.
- [ ] Descriptive legacy keys are deterministically rehashed and missing ontology field descriptions are filled.
- [ ] No runtime worker-role or orchestration behavior changes.

## Environment
- Checkout: Q00/ouroboros source checkout
- Python: supported Python 3.12+
- OS: any supported development environment

## Logs
No production runtime logs are required; the reproductions are source-level parser and CLI-loader cases.

Q00:
[...]

PR #2283, opened 2026-08-26
fix(seed): migrate v1.0.4 legacy parser fields

actives-jay:
## Summary
- migrate only the v1.0.4/revised_after_qa parser shape emitted by the legacy Seed generator
- derive hashed semantic AC keys and deterministic ontology descriptions before existing strict Pydantic validation
- preserve fail-closed behavior for malformed keys and fields

Refs #1686

## Verification
- `ruff check src/ouroboros/core/seed.py tests/unit/core/test_seed.py`
- focused RED before implementation, then `138 passed` in `tests/unit/core/test_seed.py`
- fixed frozen Seed SHA unchanged and parsed to 5 ACs / 7 described ontology fields with patched source

Base: c80144781ac3d87fe7a7d40ab93b14ddf15a191c

PR #2339, opened 2026-09-02
fix(seed): preserve v1.0.4 semantic AC identity

wjswlgnsdlqslek:
## User problem
Persisted v1.0.4 `revised_after_qa` Seeds must remain loadable without rewriting explicit AC identity or accepting malformed hash-shaped keys.

## Promised contract
- Only exact v1.0.4 / `revised_after_qa` gets compatibility normalization.
- Human-readable legacy labels deterministically materialize canonical hashes.
- Exact `ac_[a-f0-9]{16}` values preserve byte-for-byte.
- all-hex noncanonical-length values fail existing strict validation.
- `workflow-ir inspect` preserves `{criterion, semantic_ac_key}` through Seed.
- nonblank ontology names fill missing descriptions deterministically.

## Implementation boundary
- core Seed migration, Workflow IR normalization, focused core/CLI tests.

## Non-goals
- no runtime worker policy / orchestration behavior
- no parser/dependency/adapter abstraction or WorkflowSpec semantic-key projection
- no compatibility outside exact envelope
- no docs source change: narrow persisted-input compatibility only; stop/report if current docs assert contrary behavior.

## Evidence
- RED: the focused new-regression command produced 5 failures and 4 passes before source changes.
- GREEN: the same focused regressions passed (9 passed).
- Affected Seed and CLI unit modules passed (192 passed).
- Ruff check and format check passed on all four changed files; `mypy src/ouroboros` passed (617 source files).
- Full unit attempt under the approved base profile stopped at 3 collection errors because optional `mcp` is not installed; CI readback follows.

Refs #2338
Supersedes #2283

### DIFF
--- a/src/ouroboros/cli/commands/workflow_ir.py
+++ b/src/ouroboros/cli/commands/workflow_ir.py
@@ -77,7 +77,7 @@ def _normalize_seed_payload(raw: dict[str, Any]) -> dict[str, Any]:
 def _normalize_criterion(item: Any) -> str | dict[str, Any]:
     if isinstance(item, dict):
         if isinstance(item.get("criterion"), str):
-            return item["criterion"]
+            return item
         if isinstance(item.get("description"), str) or isinstance(item.get("content"), str):
             return item
     if isinstance(item, str):
--- a/src/ouroboros/core/seed.py
+++ b/src/ouroboros/core/seed.py
@@ -375,6 +375,11 @@ here because ``ledger_seed`` imports this module. The two are pinned equal by a
 regression test.
 """
 
+_LEGACY_V104_VERSION = "1.0.4"
+_LEGACY_V104_GENERATION_MODE = "revised_after_qa"
+_LEGACY_V104_SEMANTIC_AC_KEY_RE = re.compile(r"^ac_[a-z][a-z0-9_]*$")
+_HASH_SHAPED_SEMANTIC_AC_KEY_RE = re.compile(r"^ac_[a-f0-9]+$")
+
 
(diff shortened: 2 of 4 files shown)


wjswlgnsdlqslek:
@Q00 Could you please rerun the `Test Python 3.12` check for this PR?

The job reached 99%, then produced no further output and was cancelled at the workflow timeout rather than failing on a test assertion. The focused Seed and Workflow IR test suite passes locally.

If the rerun reproduces, I’ll isolate the hanging full-suite/xDist case.

Q00:
Thank you for the report and for tracking this down. You were right that the Workflow IR loader was discarding `semantic_ac_key`: it reproduced on `main`, and the one-line loader change you identified here is now merged in #2352 (4bc392d5c) with you credited as co-author on the commit.

The v1.0.4 / `revised_after_qa` migration layer was left out. No producer in this repository has ever emitted that envelope (`git log -S revised_after_qa` across all branches is empty; the metadata version default is `1.0.0`), and malformed all-hex keys are already rejected by Seed's strict `^ac_[a-f0-9]{16}$` pattern on `main` in both `Seed.from_dict` and `Seed.model_validate`. Under the review boundary contract, compatibility for a format we never produced counts as a new subsystem and needs a maintainer decision, so we kept the fix to the part a supported input reproduces.

If you have a persisted Seed in the wild that actually carries `metadata.version: 1.0.4` + `generation_mode: revised_after_qa`, please attach it to #2338 and we will revisit with real input. Closing in favor of #2352 for now. Thanks again.

PR #2341, opened 2026-09-03
fix(auto): type the grade-gate stop_reason_code instead of shipping None

MyoungSoo7:
## Summary

The auto pipeline's grade gate is its most common BLOCKED outcome, and it was the last blocker family still leaving `stop_reason_code` at `None`. All 6 `mark_blocked(..., tool_name="grade_gate")` sites passed no `error_code`, so every consumer downstream — `ooo auto` (prints `-`), the MCP envelope (omits the key), harness traces (null), `attention_relay` (nothing to relay) — had to parse English prose to tell three different situations apart:

| prose the consumer had to read | what it actually means | right next step |
|---|---|---|
| "Seed grade C did not meet required grade B" | the grade bar was not cleared | raise the grade |
| "Seed review did not clear the Seed for execution" | the grade bar **was** cleared; the review withheld `may_run` | raising the grade is the wrong move |
| "Degraded seed retains hard safety blockers…" | a deadline-recovery Seed still carries safety markers | resolve the markers |

This is an internal inconsistency, not a missing feature: two lines below the untyped sites, the interview layer already emits `UNSTUCK_EXHAUSTED_STOP_REASON_CODE`, and the runtime emits `watchdog_wall_clock_exceeded` — both as module-level consts, precisely so "the alphabet of valid `stop_reason_code` values can be discovered by reading the module surface" (`interview_driver.py:170`). The grade gate simply never joined that convention.

Scope is deliberately one blocker family. 37 of the 51 `mark_blocked` call sites in `src/` were untyped; typing them all in one PR would be unreviewable. This types the 6 that share one gate and one meaning, leaving 31.

## Review boundary

- **User problem**: a consumer of `AutoPipelineResult` / the MCP envelope / a harness trace cannot machine-distinguish the three grade-gate terminals, because all three arrive as `stop_reason_code=None` plus free prose.
- **Inputs and execution conditions**: the REVIEW-phase grade gate and the post-QA grade gate in `AutoPipeline`, for both normal and `degraded` Seeds. No new I/O, no new external calls, no config surface.
- **Promised contract**: each of the three grade-gate terminals now sets a canonical `stop_reason_code` — `seed_grade_below_required`, `seed_review_withheld_run`, `degraded_seed_safety_blockers` — and the blocker prose is unchanged, so existing prose-reading consumers keep working. Blockers *without* a canonical code still leave `stop_reason_code` at `None` (regression-guarded).
- **Implementation boundary and owner**: `src/ouroboros/auto/`. The three consts and the two gate predicates live in a new module `grade_gate_terminals.py` rather than extending `pipeline.py`, which is a #1797 grandfathered module. No data or security boundary is crossed.
- **Non-goals**: the other 31 untyped `mark_blocked` sites; any change to which Seeds are blocked; any change to blocker wording; any new envelope field.
- **Evidence**: 3 new tests in `tests/unit/auto/test_stop_reason_code.py`, each asserting the exact code on a real pipeline run. Verified red against the unfixed pipeline (`assert None == 'seed_review_withheld_run'`), not merely as an import error. The post-QA case needs a gate that changes its mind: a fixed gate can never reach that terminal, because the REVIEW-phase gate asks the same question strictly earlier, so the test drives a real QA-fail → repair → re-review round and asserts the evaluator was entered twice. Each of the 6 `error_code=` sites was reverted individually and the suite re-run: 6/6 killed, none unpinned.

## Test plan

- [x] `pytest tests/unit/auto/test_stop_reason_code.py -q` — 8 passed
- [x] The 3 new cases fail against `origin/main`'s `pipeline.py` (3/3 at the baseline)
- [x] `pytest tests/unit/auto/test_pipeline_partial_product_terminal.py tests/unit/auto/test_pipeline_interview_deadline_closure.py tests/unit/auto/test_provenance_gate.py -q` — 34 passed (degraded-Seed path, which this PR refactors)
- [x] `ruff check` / `ruff format --check` / `mypy src/` clean
- [x] `scripts/check-module-size.py --baseline-ref origin/main` — OK; `pipeline.py` **shrinks** from 3348 to 3289 lines against a 3306 budget
- [x] `scripts/check-auto-boundary.py`, `check-max-turns-envelope.py` — clean

Filled `N/A` deliberately, and stating why so the intent is auditable rather than skipped:

1. **No per-round work changes.** The diff replaces inline conditionals with calls to two pure predicates over already-computed values, and passes an already-known string to `mark_blocked`. It adds no loop, no I/O, no LLM call, and no allocation inside a round. The only reachable cost is one extra Python function call per grade-gate evaluation, which happens at most twice per run.
2. **The live R-run was not executed.** `OUROBOROS_RUN_CANONICAL=1 pytest tests/canonical/ -k cli-todo` is documented in `tests/canonical/README.md` as a maintainer-only manual path that "costs real LLM" budget, and `~/.ooo-observability/` holds no baseline on this machine. I did not spend maintainer LLM budget to measure a change that cannot move the number. If a maintainer wants the measurement anyway, say so and I will run it.

Budget compliance: [ ] within 1.5× / [ ] regression flagged with mitigation / [x] N/A (no per-round cost path changed; see reasoning above)

## Found while auditing this branch, and deliberately not fixed here

Nine adversarial lenses ran against this exact commit. The refactor itself came back
clean under a 674-combination equivalence sweep (grades `A/B/C/D/F/""/"a"/None` ×
`required_grade` × `may_run` × `skip_run` × 0/1/2 blockers, plus `review=None`,
transcribed from the removed `-` lines of this diff): zero divergences, zero
fail-opens, unknown grades still fail closed. The four findings below reproduce, and
none of them is mine — I checked that claim rather than taking the lenses' word for
it, by extracting each enclosing function from the AST at `origin/main` and at HEAD
and comparing whole bodies, which is immune to the line shifts this refactor causes.

1. **`pipeline.py:2729` — a pre-run judge's `passed` bit is the only evidence that
   "the product was verified."** `_has_verified_product_completion` returns `True` on
   `state.last_qa_passed`, whose only substantive writer is the Seed-*document* QA gate
   that runs strictly before dispatch. A judge envelope `{"passed": true, "score": 0.0,
   "verdict": "fail"}` therefore reports `artifact_state="complete_verified"` for a
   session where no run handle exists — short-circuiting the guard in `artifact_state.py`
   written to prevent exactly that. The existing regression at `test_surface.py:2664`
   pins the guard while hand-passing `product_verified=False`, a value the production
   path never produces, so it is vacuous. Function and writer line are both byte-identical
   across the two revisions. The fix changes what `complete_verified` means for every
   MCP caller and belongs in its own PR.

2. **`pipeline.py:1022` — the `grade_gate` resume branch trusts the file, not the
   artifact it ruled on.** `--resume` re-reads the Seed from `state.seed_path` with no
   provenance or integrity check and overwrites `state.seed_artifact`. One edit to
   `ambiguity_score` reverses a refusal-to-run terminal into `complete` **and dispatches
   the run**; `degraded: true` reverses it into a partial-product success. The control
   row re-blocks with the same typed code this PR adds, which is how the reversal is
   visible at all. `pipeline.py:120-128` asserts the opposite in prose — "editing the
   on-disk `state.seed_path` file therefore has no effect". Line 1022 is untouched by
   this diff and `_load_seed` is byte-identical; closing it means owning seed-artifact
   authority on resume, which is a boundary decision, not a silent in-PR expansion.

3. **`pipeline.py:3121` — the ambiguity recogniser's misses are permissive, not
   strict.** `_requests_seed_qa_ambiguity_repair` is the Seed-QA gate's only hard stop;
   everything else on a failed verdict is a deliberate advisory-continue. It matches
   exactly two clauses (`must not exceed`, `(must|should) not be greater than`) under
   `re.fullmatch`, so any other modal, any synonym of *exceed*, or a trailing clause
   falls through to the permissive branch. Byte-identical across revisions. The fix is
   to decide from `seed.metadata.ambiguity_score` rather than from English; merely
   relaxing the regex regresses the negatives pinned at `test_pipeline_lateral.py:254-260`.

4. **`pipeline.py:2096` — the ambiguity terminal is unreachable on the last attempt.**
   It is nested inside `if attempt < max_attempts`, so with `--max-repair-rounds 1`
   (reachable from both the CLI and the MCP argument) the loop runs once, `1 < 1` is
   false, and control falls to the `repair_budget_exhausted` advisory continue. This is
   the repo's known partial-gate-on-one-exit shape. `_run_seed_qa_gate` differs between
   the revisions *only* in how the review-gate blocker is constructed; the nesting is untouched.

Happy to open them as their own PRs — (3) and (4) together, (1) and (2) separately.

## One gap in this PR's own coverage, disclosed rather than buried

Mutation testing by the audit's `exits` lens deletes each added `error_code=` in turn.
Two of six are killed by the new tests; four survive:

```
KILLED   REVIEW grade/may_run gate
KILLED   post-QA review gate
SURVIVED REVIEW degraded-seed terminal      (pipeline.py:1165)
SURVIVED RUN re-entry grade gate            (pipeline.py:1253)
SURVIVED RUN re-entry may_run gate          (pipeline.py:1297)
SURVIVED QA-advisory review gate            (pipeline.py:2193)
```

The same lens confirms the other half of the claim: all three new tests are genuinely
red against `HEAD~1`, none is a passing shell. So the behaviour ships correct — this is
coverage, not a defect — but four of the codes this PR introduces are pinned by nothing,
and `grep -rn "degraded_seed_safety_blockers" tests/` returns empty repo-wide.

I am not widening the PR to close it. The degraded-seed and QA-advisory sites would take
one added assertion each on tests that already drive those lines, but the two RUN re-entry
sites need new tests that drive a RUN-phase resume, and that is a second change with its
own review surface. Stated here so the number is the maintainer's to weigh rather than a
silent assumption; say the word and it is a follow-up PR.

## Related issues

Refs #1157 — a typed terminal envelope is what lets `ooo auto` be driven without a human reading prose. Refs #1797 for the module-size constraint that put the new code in its own module.

### DIFF
--- a/src/ouroboros/auto/grade_gate_terminals.py
+++ b/src/ouroboros/auto/grade_gate_terminals.py
@@ -1,0 +1,84 @@
+"""The auto pipeline's grade gate: its verdict, and the codes its terminals stop with.
+
+Grade-gate terminals are the pipeline's most common BLOCKED outcome and were
+the last blocker family still leaving ``stop_reason_code`` at ``None``. The
+consequence is downstream: ``ooo auto`` prints ``-``, the MCP envelope omits
+the key, harness traces write null, and ``attention_relay`` has nothing to
+relay -- so every consumer had to read English prose to tell "the grade is too
+low" (raise the grade) from "the review withheld ``may_run``" (the grade bar
+was already cleared, so raising it is the wrong next step) from "a degraded
+Seed still carries hard safety blockers" (resolve the markers).
+
+Module-level consts, so every call site emits the same string and the alphabet
+of valid codes stays discoverable by reading a module surface -- the same
+convention as the interview layer's ``UNSTUCK_EXHAUSTED_STOP_REASON_CODE`` and
+the runtime's ``watchdog_wall_clock_exceeded``.
+
+The gate's own verdict lives here beside them rather than in ``pipeline``,
+because a code that is decided in one module and named in another is a code
+that can drift from what it names.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from ouroboros.auto.seed_reviewer import SeedReview
+    from ouroboros.auto.state import AutoPipelineState
+
+SEED_GRADE_BELOW_REQUIRED_STOP_REASON_CODE = "seed_grade_below_required"
+SEED_REVIEW_WITHHELD_RUN_STOP_REASON_CODE = "seed_review_withheld_run"
+DEGRADED_SEED_SAFETY_BLOCKERS_STOP_REASON_CODE = "degraded_seed_safety_blockers"
+
+_GRADE_RANK = {"A": 0, "B": 1, "C": 2}
+
+
+def grade_meets_required(actual: str | None, required: str) -> bool:
+    """Whether ``actual`` is at least as good as ``required``; unknown never is."""
+    if actual not in _GRADE_RANK or required not in _GRADE_RANK:
+        return False
+    return _GRADE_RANK[actual] <= _GRADE_RANK[required]
+
+
+def degraded_seed_safety_blocker(review: SeedReview) -> tuple[str, str] | None:
+    """Return ``(blocker, stop_reason_code)`` if a degraded Seed still has blockers.
+
+    A Seed synthesized under deadline pressure may surface as a partial product,
+    but never while hard safety markers remain unresolved -- so this terminal is
+    typed beside the other two rather than left as prose at the call site.
+    """
+    if not review.grade_result.blockers:
+        return None
+    codes = ", ".join(finding.code for finding in review.grade_result.blockers)
+    return (
+        "Degraded seed retains hard safety blockers; "
+        f"unsafe/destructive markers must be resolved before run: {codes}",
+        DEGRADED_SEED_SAFETY_BLOCKERS_STOP_REASON_CODE,
+    )
+
+
+def seed_review_gate_blocker(
+    state: AutoPipelineState, review: SeedReview | None, *, skip_run: bool
+) -> tuple[str, str] | None:
+    """Return ``(blocker, stop_reason_code)`` for the deterministic review gate.
+
+    The code travels with the message because these are the *same* terminals
+    the inline REVIEW-phase gate raises: returning the prose alone would force
+    callers to re-derive the classification the gate has already computed,
+    which is exactly the prose-parsing this typing exists to remove.
+    """
+    if review is None:
+        return None
+    if not grade_meets_required(review.grade_result.grade.value, state.required_grade):
+        return (
+            f"Seed grade {review.grade_result.grade.value} did not meet "
+            f"required grade {state.required_grade}",
+            SEED_GRADE_BELOW_REQUIRED_STOP_REASON_CODE,
+        )
+    if not review.may_run and not skip_run:
+        return (
+            "Seed review did not clear the Seed for execution",
+            SEED_REVIEW_WITHHELD_RUN_STOP_REASON_CODE,
+        )
+    return None
--- a/src/ouroboros/auto/pipeline.py
+++ b/src/ouroboros/auto/pipeline.py
@@ -25,6 +25,13 @@ from ouroboros.auto.checkpoint_commits import checkpoint_final_auto
 from ouroboros.auto.domain_inference import derive_domain_from_ledger
 from ouroboros.auto.domain_profile import DEFAULT_REGISTRY
 from ouroboros.auto.execution_acceptance import normalize_execution_acceptance
+from ouroboros.auto.grade_gate_terminals import (
+    SEED_GRADE_BELOW_REQUIRED_STOP_REASON_CODE,
+    SEED_REVIEW_WITHHELD_RUN_STOP_REASON_CODE,
+    degraded_seed_safety_blocker,
+    grade_meets_required,
+    seed_review_gate_blocker,
+)
 from ouroboros.auto.grading import GradeGate, deterministic_floor
 from ouroboros.auto.handoff_contract import (
     IDEMPOTENCY_KEY_FIELD,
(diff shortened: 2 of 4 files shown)


sergiobuilds:
Maintainer follow-up is in #2344. GitHub rejected a direct maintainer push to this fork despite maintainerCanModify=true, so the replacement preserves contributor commit e3f3a8c99dca98c464452189c588b796fc457f1f, merges current main, and adds the exact degraded result/state stop-reason assertions requested here. Keeping this PR open until #2344 passes current-head review and CI.

MyoungSoo7:
Round 1 blocker addressed at `f10a3028` (rebased onto `c1b77f7b`, behind 0).

**Blocker — `degraded_seed_safety_blockers` is not contract-tested.** Confirmed exactly as reported: `grep -rn DEGRADED_SEED_SAFETY_BLOCKERS_STOP_REASON_CODE tests/` returned nothing, and the reviewer's own reproduction is what I used as the red baseline. Deleting `error_code=code` at `pipeline.py:1165` left all of `test_degraded_seed_with_safety_blocker_still_terminates` green while the terminal returned `stop_reason_code=None` — the fourth code shipped with the same prose-parsing this PR removes.

`tests/unit/auto/test_pipeline_partial_product_terminal.py:355-361` now asserts the exact code in three places:

```python
assert result.stop_reason_code == DEGRADED_SEED_SAFETY_BLOCKERS_STOP_REASON_CODE
assert state.last_error_code == DEGRADED_SEED_SAFETY_BLOCKERS_STOP_REASON_CODE
assert result.blocker is not None
reloaded = AutoStore(tmp_path).load(state.auto_session_id)
assert reloaded.last_error_code == DEGRADED_SEED_SAFETY_BLOCKERS_STOP_REASON_CODE
```

The third is not redundant with the second. `state` is the object the pipeline mutated in place; an operator resuming the session reads the record `_save` wrote. A regression that set the field but failed to persist it would pass the in-memory assertion.

**Each assertion was checked red independently**, under the same one-line deletion at `pipeline.py:1165`:

| assertion | line | baseline result |
|---|---|---|
| `result.stop_reason_code` | 355 | `AssertionError: assert None == 'degraded_seed_safety_blockers'` |
| `state.last_error_code` | 356 | same, with the earlier assertion removed |
| `reloaded.last_error_code` | 361 | same, with both earlier assertions removed |

Removing only some of them was necessary to show the later ones are not shadowed by the first — a test that fails on line 355 proves nothing about line 361. (Line numbers are the final file's; each baseline run was made with the earlier assertions deleted, so pytest reported the shifted line at the time.)

**Checks after the fix:** `tests/unit/auto` **2181 passed** (post-rebase), `ruff check` / `ruff format --check` clean over `src tests`, `mypy src` clean over 632 files.

**Not addressed, and disclosed rather than silently skipped.** The medium-priority note about mutation-surviving sites at `pipeline.py:1253,1297,2081,2193` (RUN-resume and QA-advisory paths) is untouched. Those are not grade-gate terminals and typing them is a separate alphabet decision; this PR's scope is the `tool_name="grade_gate"` family. Happy to take them in a follow-up if you want them in the same alphabet.

PR #2286, opened 2026-08-27
feat: surface ambiguity gate override rate

kuilinggg:
## Summary

This PR adds ambiguity-gate statistics to project status output for both the CLI and MCP interfaces.

It records whether each Gen-1 Seed passed the ambiguity gate normally or used `force=True`, then exposes the following project-level fields:

- `gated_seed_count`
- `forced_seed_count`
- `unknown_seed_count`
- `override_rate`

The override rate is calculated as:

```text
forced_seed_count / (forced_seed_count + gated_seed_count)
```

Unknown values are excluded from the denominator and include legacy Seeds, manually-authored Seeds, Gen-2/Reflect Seeds, and records where the gate decision cannot be determined.

When there are no known gate decisions, `override_rate` is returned as `null` in JSON and displayed as `n/a` in text output.

Existing Seeds and session-start events without `gate_forced` remain compatible. This change does not add telemetry, database tables, or historical threshold inference.

## Test plan

- [ ] `uv run pytest tests/ -q` passes

  The full local test run is partially limited by the Windows-backed WSL environment. The run completed with 588 passed, 41 skipped, 12 failed, and 18 collection errors. The failures are related to existing platform/dependency assumptions, including Linux-only path semantics, `fcntl`, symlink privileges, and MCP SDK version compatibility.

- [x] `uv run ruff check src/ tests/` and `uv run ruff format --check src/ tests/` clean

- [ ] `uv run mypy src/` clean (or noted exception)

  MyPy reports an existing error outside the scope of this PR:

  ```text
  src/ouroboros/hermes/artifacts.py:205:
  Name "result" is not defined
  ```

- [x] Targeted tests for Seed metadata, Seed generation, session-start events, Project Map statistics, CLI project status, and MCP project status pass

  The targeted test run passed the relevant tests. The Project Map fixture requiring a directory name containing a newline could not run on the Windows-backed environment.

- [x] `git diff --check` clean

The repository GitHub Actions workflow should provide the authoritative Linux and Python 3.12 validation.

## Related issues

Refs #2231

PR #2343, opened 2026-09-03
fix(codex): observe mid-run drift of frozen authority inputs instead of failing the run

Q00:
Refs #2311
Stacked on #2342 (merge that first; this branch contains it).

## User problem

A healthy run turns into FAILED the moment a frozen runtime input moves under it. The Codex-family runtime had eleven kill-switches that failed every later AC with "… changed after runtime initialization; start a new execution session", and since each following AC re-observed the same change, one benign event ended the run:

| trigger | what the user did |
|---|---|
| `~/.codex/config.toml` changed | edited config mid-run; any other codex session or terminal wrote to it |
| CLI executable content/path changed | `npm update -g @openai/codex` during a run |
| built-in MCP handler / skill dispatch registry changed | `ouroboros update` / pip upgrade during a run (hashes handler module bytes) |
| Ouroboros profile routing changed | edited `~/.ouroboros/config.yaml` mid-run |
| version attestation timed out / failed at init | slow `codex --version` on a loaded machine |

Live on this branch's bench: two runs died with "built-in MCP handler registry changed" because a source commit landed while they ran — correct detection, wrong consequence.

## What the guards protected, and what still does

Their real invariant: a native thread created under the old inputs must never be *resumed* under the new ones. That is kept by a drift epoch (`orchestrator/runtime_drift.py`): every observation bumps it, runtime handles are stamped with the epoch they were created under, and a handle from an earlier epoch is not resumed — the next turn starts a fresh thread on the re-baselined inputs.

Each check now: logs `runtime.drift_observed`, emits `runtime_drift.kind` (closed enum, `TELEMETRY.md`), re-baselines on the current input (config fingerprint + trust baseline together; executable re-attested; registries re-fingerprinted; role-default model/profile re-resolved), and returns. Detection logic is byte-for-byte the same and stays tested through `_verify_cli_executable_identity_unchanged` (moved to `cli_version_attestation.py` as a pure policy function).

Codex, Copilot, Gemini, Goose, Grok runtimes (all inherit the executable check; the config/registry/routing checks are codex-only as before). No behavior change when nothing drifts.

- No AC fails because a frozen input changed after initialization.
- A resume handle stamped with an older drift epoch resolves to no resume; `_build_command` drops `resume_session_id` for such a handle and logs `codex_cli_runtime.resume_dropped_after_drift`.
- After re-baselining, the same state is not reported as drift again (idempotent).
- A missing/unresolvable executable still fails the launch with the real error (`Codex CLI not found`), not the attestation message.
- `runtime_drift.kind` ∈ {codex_config, cli_executable, skill_dispatcher, mcp_handler_registry, skill_dispatch_registry, profile_routing, baseline_unavailable, unknown}; anything else folds to `unknown`.

`orchestrator/codex_cli_runtime.py` (−53 lines net; grandfathered module keeps shrinking), new `orchestrator/runtime_drift.py`, `orchestrator/cli_version_attestation.py` (policy function extracted), subclass call sites (copilot/gemini/goose/grok), `telemetry.py` + `TELEMETRY.md` (one closed event). Owner: @Q00.

**Security boundary decision (owner-directed, "무조건 돌아가야 한다"):** a binary swapped mid-run is no longer a hard stop — it is re-attested, logged, and counted, and the run proceeds on it. The thread-resume invariant is preserved. If a hard stop is ever wanted again for one kind, it is one branch in `_rebaseline_fingerprint` / `_reconcile_cli_executable_identity`, not a redesign.

## Non-goals

Other runtimes' own drift handling (none have kill-switches today), the fat-harness `files_touched` false negative, a durable event for runtime preparation failures.

## Evidence

Live codex bench on this branch (same user-style 4-AC seed and product smoke as #2342):

| scenario | harness | product smoke | notes |
|---|---|---|---|
| 2 sessions concurrent × 2 runs | **4/4** | 4/4 | no drift observed (the #2342 trust exemption already covers this case) |
| 1 run, a real `~/.codex/config.toml` key change injected 75s in | **1/1** (213s) | 4/4 | `runtime.drift_observed kind=codex_config` fired once; run continued. On v0.53.0 this kills every remaining AC ("Codex configuration changed after runtime initialization"). |

Earlier on this stack, two runs were killed by "built-in MCP handler registry changed" after a source commit landed mid-run — the same class, now observed instead of fatal.

Unit: orchestrator + providers + telemetry suites green (6.5k); 30 executable-drift detection tests retargeted at the raising verifier; new tests for resume retirement, telemetry fold, CLI-upgrade re-baseline, cross-runtime observe-and-continue.

https://claude.ai/code/session_01CwzsmprTfPVjoWRDynyufG


Q00:
Review round 1 addressed in `3c99e0cbc` — both blockers reproduced from source and are fixed with the deterministic probes you named.

**Blocker 1 — Goose resume bypass (fixed).** Confirmed: Goose's `_resolve_resume_session_id` and `_build_command` overrides consumed neither the epoch nor the base resume gate, so epoch 0 handle + swapped executable → `--resume old-session` at epoch 1. The gate is now a single base helper, `_retire_resume_after_drift(resume_session_id, runtime_handle)`, called after every runtime's reconcile checks (Codex base and Goose; Copilot/Gemini/Grok pass no resume today). Goose's `_resolve_resume_session_id` also checks `handle_predates_drift`, and a retired id no longer leaks back in as the `-n` session name. Test: `test_goose_executable_drift_retires_resume_and_pre_drift_session_name` — your exact probe (handle at epoch 0, executable rewritten, next `_build_command` asserts epoch 1 and no `--resume`), plus a post-drift handle resuming normally.

**Blocker 2 — invalid config treated as benign profile drift (fixed, with one deliberate deviation).** Confirmed: the `load_error` fingerprint was re-baselined, and `resolve_completion_profile`'s role path returns the unprofiled default on `ConfigError`, so later commands silently dropped the selected profile. I did *not* restore it as a preparation failure — that reintroduces a kill-switch on a corrupt/half-written `config.yaml`, which is the failure class this PR exists to stop. Instead: an unloadable config is observed once as `baseline_unavailable` (your suggested classification) and the frozen routing is kept — role-default fallback untouched, and the last routing resolved from a loadable config is cached per handle selector and served while the config is unloadable. A valid changed config is still adopted as `profile_routing` drift exactly as before. Test: `test_unloadable_config_mid_run_keeps_frozen_routing_instead_of_default` — `load_config` raising `ConfigError` mid-run keeps `--profile ouroboros-standard` for both role-selected and default commands, reports once (idempotent on the repeated command), and a subsequently loadable remap is adopted.

**Follow-ups (not in this PR, both pre-existing on `main`):**
- Process-local epoch across runner restarts: before this PR a restart also re-baselined on current inputs and resumed the persisted handle, so this is not a regression of the observe-not-kill change. Owner @Q00, tracked under #2311 as the durable-resume lever.
- Registry fingerprint unavailable → `None` baseline: same "unavailable ≠ drift" principle as blocker 2; owner @Q00, same tracking.

Verification: ruff format/check, mypy (620 files), module-size gate (`codex_cli_runtime.py` 4136 ≤ 4140 cap) clean; orchestrator+providers+telemetry suites 6.5k green serially. Note for the record: under `-n auto` on a loaded machine the executable-attestation tests flake identically on the parent commit (version probe times out → observed as drift → epoch counts inflate); that is the attestation-timeout policy, not this round.

Q00:
Review round 2 addressed in `4f7683367`. All four sub-findings under blocker #1 reproduced from source; each is fixed inside the existing runtime-drift / command-preparation boundary, with your minimal reproductions as tests.

**Late reconcile after retirement (fixed).** `_resolve_runtime_codex_config` re-runs the profile reconcile after the resume decision had been taken. The gate is now `RuntimeDriftLedger.retire_resume` (moved out of the runtime) and is called *after* routing resolution in `CodexCliRuntime._build_command`; Goose's builder uses the same ledger method after its reconcile. Test `test_routing_drift_observed_by_the_last_reconcile_still_retires_resume` flips `load_config` between the first profile check and the routing resolution of one `_build_command` call with an epoch-0 resume handle, and asserts `--profile profile-b` with neither `resume` nor `thread-old` in the command.

**Invocation-generation attribution (fixed).** Handles were stamped with the ledger's epoch at event time. The stream loop now records the admitted epoch on its per-invocation `_CodexItemCorrelationScope` right after the command is built, and every handle created by that stream (`thread.started` in `_convert_event`, the session-rebind path in the loop, Goose's `_convert_event`) is stamped via `_build_runtime_handle(..., drift_epoch=scope.admitted_drift_epoch)`. Test `test_handles_are_stamped_with_the_invocation_admitted_epoch`: invocation A admitted at epoch 0, concurrent observation → epoch 1, A's `thread.started` handle carries 0 and does not resume; a stream admitted at 1 resumes.

**Cold selector under unloadable config (fixed).** A selector first used during the unloadable interval now gets the frozen role default rather than the resolver's unprofiled fallback; a native `codex_profile` in handle metadata needs no Ouroboros config and resolves as before. Extended `test_unloadable_config_mid_run_keeps_frozen_routing_instead_of_default` with a cold `qa` role handle asserting `--profile ouroboros-standard`.

**Unchanged unavailable attestation (fixed).** Unavailable evidence (probe keeps timing out, init-time or check-time wording) is observed once as `baseline_unavailable` — your suggested classification — then re-probed on every command without advancing the epoch; a real binary change is still `cli_executable`. Test `test_unchanged_unavailable_attestation_is_reported_once_and_recovers`: three commands under a `TIMED_OUT` probe → one observation, epoch 1; once the probe succeeds the runtime re-attests and clears the state with no further epoch.

**Follow-up (unchanged):** durable drift generation across runner recreation stays an owned follow-up under #2311 (@Q00), as agreed.

Verification: ruff, mypy (620 files), module-size gate (`codex_cli_runtime.py` 4136 ≤ 4140) clean. Orchestrator + providers + telemetry suites serial: 6584 passed; the 3 failures are pre-existing on the parent commit and on `main` (`test_undecodable_filename_exports_and_applies` — macOS filesystem `EILSEQ`; `test_fat_harness_exit_code_only_test_output_is_settled_by_harness_reexecution[True]` — reproduces identically with this branch's changes stashed).

Q00:
Addendum in `4de8b012f` — a self-audit of every site that samples the drift generation (observe/stamp/retire call sites, all `_build_command` overrides incl. zcode/antigravity, the skill-intercept reconcile, the session-rebind path, legacy unstamped handles) found two residual routes to a retired thread. Both fixed with tests.

- **Goose resumes by name, and the post-drift thread reused the retired thread's name.** `goose run -n <name>` without `--resume` creates a second session under an existing name (verified against a local `sessions.db`, which already holds duplicate ouroboros names), and `--resume -n <name>` then lets Goose pick between generations. The derived session name now folds in the admitted drift epoch once it is non-zero; `_build_command` and `_convert_event` derive the same new name from the epoch, so the next turn resumes the replacement, never the retired one. Epoch-0 names are unchanged. Test: `test_goose_executable_drift_retires_resume_and_pre_drift_session_name` now asserts a distinct post-drift name recovered through event normalization.
- **A resume retired inside the build was still reported as attempted.** `attempted_resume_session_id` is resolved before the build; when the build retired it and the launch failed early, the result carried `resume_retry` recovery and the runner would pause telling the user to retry a resume that never happened. It is re-resolved after the build. Test: `test_resume_retired_inside_the_build_is_not_reported_as_attempted`.

Left as is, for the record: a `baseline_unavailable` observation still advances the epoch and retires live threads once even though nothing verifiably changed. Conservative by design; happy to make it non-retiring if you'd rather.

Verification unchanged from the previous comment: ruff, mypy, module-size (4139 ≤ 4140) clean; orchestrator + providers + telemetry serial 6586 passed, same two pre-existing failures as on the parent commit.

Q00:
Addendum in `3eeab7aba` — an independent adversarial pass (deterministic probes, separate reviewer session) against the declared invariants found two latent routes, both fixed with the probes as regression tests:

- **Bare `resume_session_id` had unknown admission history but was stamped current.** `execute_task(resume_session_id=...)` (the deprecated `RuntimeAdapter` parameter; no production caller uses it today) manufactured a handle stamped with the ledger's current epoch, so a pre-drift thread named by a bare id was resumed under post-drift inputs. The manufactured handle is now stamped `DRIFT_EPOCH_UNKNOWN` and retired like any stamp-less persisted handle once drift exists; with no drift observed it resumes as before. Test: `test_bare_resume_session_id_is_retired_once_any_drift_was_observed` (asserts through the real `execute_task` path that the build receives no resume target).
- **`handle_predates_drift` accepted a bool stamp.** `isinstance(True, int)` holds and `True == 1`, so a bool stamp deserialized from an external store would pass as epoch 1. Comparison is now `type(...) is int`. Test: `test_bool_epoch_stamp_is_not_mistaken_for_the_current_epoch`.

Held under probing: concurrent drift between `_build_command` and `thread.started`; drift between the two routing checks inside one build; Goose `-g<epoch>` naming, pending-initial-launch handles, generated names; `_rebaseline_fingerprint` with a `None` baseline; str/negative stamps (fail safe).

**Live run on this HEAD's parent (`4de8b012f`)**: user-style 4-AC seed, `--runtime codex`, `~/.codex/config.toml` key appended 40s in. Harness 4/4, independent product smoke 4/4, 396s. `runtime.drift_observed kind=codex_config` fired once at the next command build and the run continued; the init-time attestation timeout was observed once as `baseline_unavailable` and not repeated. On v0.53.0 the config change alone kills every AC after the first. (No command in that run carried a `resume` argument, so thread retirement is covered by the unit probes, not by this run.)

Verification: ruff, mypy (620 files), module-size (`codex_cli_runtime.py` 4140 ≤ 4140) clean; orchestrator + providers + telemetry serial 6588 passed, same two pre-existing failures as on `main`.

Note for the record: `enforce-module-size` is currently red on every PR because `main` at 50f884d2d (#2286) left `mcp/tools/authoring_handlers.py` at 3725 lines against its own 3722 budget; main's own module-size run fails there. Not a required check; needs a separate three-line shrink on main.

PR #2344, opened 2026-09-03
fix(auto): type grade-gate stop reasons and pin degraded terminal

sergiobuilds:
## Maintainer replacement

Supersedes #2341 after GitHub rejected a maintainer push to the contributor fork despite `maintainerCanModify=true`. This branch preserves contributor commit `e3f3a8c99dca98c464452189c588b796fc457f1f`, merges current `main`, and adds the exact degraded-terminal assertions requested by the exact-head review.

## Summary

The auto pipeline's grade gate is its most common BLOCKED outcome, and it was the last blocker family still leaving `stop_reason_code` at `None`. All 6 `mark_blocked(..., tool_name="grade_gate")` sites passed no `error_code`, so every consumer downstream — `ooo auto` (prints `-`), the MCP envelope (omits the key), harness traces (null), `attention_relay` (nothing to relay) — had to parse English prose to tell three different situations apart:

| prose the consumer had to read | what it actually means | right next step |
|---|---|---|
| "Seed grade C did not meet required grade B" | the grade bar was not cleared | raise the grade |
| "Seed review did not clear the Seed for execution" | the grade bar **was** cleared; the review withheld `may_run` | raising the grade is the wrong move |
| "Degraded seed retains hard safety blockers…" | a deadline-recovery Seed still carries safety markers | resolve the markers |

This is an internal inconsistency, not a missing feature: two lines below the untyped sites, the interview layer already emits `UNSTUCK_EXHAUSTED_STOP_REASON_CODE`, and the runtime emits `watchdog_wall_clock_exceeded` — both as module-level consts, precisely so "the alphabet of valid `stop_reason_code` values can be discovered by reading the module surface" (`interview_driver.py:170`). The grade gate simply never joined that convention.

Scope is deliberately one blocker family. 37 of the 51 `mark_blocked` call sites in `src/` were untyped; typing them all in one PR would be unreviewable. This types the 6 that share one gate and one meaning, leaving 31.

## Review boundary

- **User problem**: a consumer of `AutoPipelineResult` / the MCP envelope / a harness trace cannot machine-distinguish the three grade-gate terminals, because all three arrive as `stop_reason_code=None` plus free prose.
- **Inputs and execution conditions**: the REVIEW-phase grade gate and the post-QA grade gate in `AutoPipeline`, for both normal and `degraded` Seeds. No new I/O, no new external calls, no config surface.
- **Promised contract**: each of the three grade-gate terminals now sets a canonical `stop_reason_code` — `seed_grade_below_required`, `seed_review_withheld_run`, `degraded_seed_safety_blockers` — and the blocker prose is unchanged, so existing prose-reading consumers keep working. Blockers *without* a canonical code still leave `stop_reason_code` at `None` (regression-guarded).
- **Implementation boundary and owner**: `src/ouroboros/auto/`. The three consts and the two gate predicates live in a new module `grade_gate_terminals.py` rather than extending `pipeline.py`, which is a #1797 grandfathered module. No data or security boundary is crossed.
- **Non-goals**: the other 31 untyped `mark_blocked` sites; any change to which Seeds are blocked; any change to blocker wording; any new envelope field.
- **Evidence**: 3 new tests in `tests/unit/auto/test_stop_reason_code.py` pin the real-pipeline grade-below-required and review-withheld paths. The existing degraded safety-terminal test in `tests/unit/auto/test_pipeline_partial_product_terminal.py` now pins the third canonical code on both `result.stop_reason_code` and persisted `state.last_error_code`. The new cases were verified red against the unfixed pipeline, and removing the degraded terminal's `error_code` now fails its exact assertion. Three RUN re-entry / QA-advisory call sites remain mutation-surviving and are disclosed below as follow-up coverage.

## Test plan

- [x] `pytest tests/unit/auto/test_stop_reason_code.py -q` — 8 passed
- [x] The 3 new cases fail against `origin/main`'s `pipeline.py` (3/3 at the baseline)
- [x] `pytest tests/unit/auto/test_pipeline_partial_product_terminal.py tests/unit/auto/test_pipeline_interview_deadline_closure.py tests/unit/auto/test_provenance_gate.py -q` — 34 passed (degraded-Seed path, which this PR refactors)
- [x] `ruff check` / `ruff format --check` / `mypy src/` clean
- [x] `scripts/check-module-size.py --baseline-ref origin/main` — OK; `pipeline.py` **shrinks** from 3348 to 3289 lines against a 3306 budget
- [x] `scripts/check-auto-boundary.py`, `check-max-turns-envelope.py` — clean

Filled `N/A` deliberately, and stating why so the intent is auditable rather than skipped:

1. **No per-round work changes.** The diff replaces inline conditionals with calls to two pure predicates over already-computed values, and passes an already-known string to `mark_blocked`. It adds no loop, no I/O, no LLM call, and no allocation inside a round. The only reachable cost is one extra Python function call per grade-gate evaluation, which happens at most twice per run.
2. **The live R-run was not executed.** `OUROBOROS_RUN_CANONICAL=1 pytest tests/canonical/ -k cli-todo` is documented in `tests/canonical/README.md` as a maintainer-only manual path that "costs real LLM" budget, and `~/.ooo-observability/` holds no baseline on this machine. I did not spend maintainer LLM budget to measure a change that cannot move the number. If a maintainer wants the measurement anyway, say so and I will run it.

Budget compliance: [ ] within 1.5× / [ ] regression flagged with mitigation / [x] N/A (no per-round cost path changed; see reasoning above)

## Found while auditing this branch, and deliberately not fixed here

Nine adversarial lenses ran against this exact commit. The refactor itself came back
clean under a 674-combination equivalence sweep (grades `A/B/C/D/F/""/"a"/None` ×
`required_grade` × `may_run` × `skip_run` × 0/1/2 blockers, plus `review=None`,
transcribed from the removed `-` lines of this diff): zero divergences, zero
fail-opens, unknown grades still fail closed. The four findings below reproduce, and
none of them is mine — I checked that claim rather than taking the lenses' word for
it, by extracting each enclosing function from the AST at `origin/main` and at HEAD
and comparing whole bodies, which is immune to the line shifts this refactor causes.

1. **`pipeline.py:2729` — a pre-run judge's `passed` bit is the only evidence that
   "the product was verified."** `_has_verified_product_completion` returns `True` on
   `state.last_qa_passed`, whose only substantive writer is the Seed-*document* QA gate
   that runs strictly before dispatch. A judge envelope `{"passed": true, "score": 0.0,
   "verdict": "fail"}` therefore reports `artifact_state="complete_verified"` for a
   session where no run handle exists — short-circuiting the guard in `artifact_state.py`
   written to prevent exactly that. The existing regression at `test_surface.py:2664`
   pins the guard while hand-passing `product_verified=False`, a value the production
   path never produces, so it is vacuous. Function and writer line are both byte-identical
   across the two revisions. The fix changes what `complete_verified` means for every
   MCP caller and belongs in its own PR.

2. **`pipeline.py:1022` — the `grade_gate` resume branch trusts the file, not the
   artifact it ruled on.** `--resume` re-reads the Seed from `state.seed_path` with no
   provenance or integrity check and overwrites `state.seed_artifact`. One edit to
   `ambiguity_score` reverses a refusal-to-run terminal into `complete` **and dispatches
   the run**; `degraded: true` reverses it into a partial-product success. The control
   row re-blocks with the same typed code this PR adds, which is how the reversal is
   visible at all. `pipeline.py:120-128` asserts the opposite in prose — "editing the
   on-disk `state.seed_path` file therefore has no effect". Line 1022 is untouched by
   this diff and `_load_seed` is byte-identical; closing it means owning seed-artifact
   authority on resume, which is a boundary decision, not a silent in-PR expansion.

3. **`pipeline.py:3121` — the ambiguity recogniser's misses are permissive, not
   strict.** `_requests_seed_qa_ambiguity_repair` is the Seed-QA gate's only hard stop;
   everything else on a failed verdict is a deliberate advisory-continue. It matches
   exactly two clauses (`must not exceed`, `(must|should) not be greater than`) under
   `re.fullmatch`, so any other modal, any synonym of *exceed*, or a trailing clause
   falls through to the permissive branch. Byte-identical across revisions. The fix is
   to decide from `seed.metadata.ambiguity_score` rather than from English; merely
   relaxing the regex regresses the negatives pinned at `test_pipeline_lateral.py:254-260`.

4. **`pipeline.py:2096` — the ambiguity terminal is unreachable on the last attempt.**
   It is nested inside `if attempt < max_attempts`, so with `--max-repair-rounds 1`
   (reachable from both the CLI and the MCP argument) the loop runs once, `1 < 1` is
   false, and control falls to the `repair_budget_exhausted` advisory continue. This is
   the repo's known partial-gate-on-one-exit shape. `_run_seed_qa_gate` differs between
   the revisions *only* in how the review-gate blocker is constructed; the nesting is untouched.

Happy to open them as their own PRs — (3) and (4) together, (1) and (2) separately.

## One gap in this PR's own coverage, disclosed rather than buried

Mutation testing by the audit's `exits` lens deletes each added `error_code=` in turn.
Three of six are killed by committed tests; three survive:

```
KILLED   REVIEW grade/may_run gate
KILLED   post-QA review gate
KILLED   REVIEW degraded-seed terminal      (pipeline.py:1165)
SURVIVED RUN re-entry grade gate            (pipeline.py:1253)
SURVIVED RUN re-entry may_run gate          (pipeline.py:1297)
SURVIVED QA-advisory review gate            (pipeline.py:2193)
```

The three new tests are genuinely red against `HEAD~1`, and the existing degraded safety-terminal test now fails if that site's `error_code` is removed. The remaining three call sites behave correctly under independent probes but are not mutation-pinned by the committed suite.

I am not widening the PR further. The two RUN re-entry sites need dedicated resume tests, and the QA-advisory site needs a focused committed assertion; those are follow-up coverage with their own review surface.

## Related issues

Refs #1157 — a typed terminal envelope is what lets `ooo auto` be driven without a human reading prose. Refs #1797 for the module-size constraint that put the new code in its own module.


sergiobuilds:
Ready for exact-head re-review at f00edbd1ebced4436b345feb46bb770c196f661a. The original #2341 blocker is fixed with exact result/state assertions for degraded_seed_safety_blockers, the PR body now reflects the repaired coverage accurately, all current-head CI checks are green, and the contributor commit remains preserved as the first parent.

sergiobuilds:
@ouroboros-agent Please perform a fresh exact-head review of `f00edbd1ebced4436b345feb46bb770c196f661a`. The original #2341 blocker is fixed with exact result/state assertions for `degraded_seed_safety_blockers`; the PR body now accurately reports the three committed mutation-pinned sites and the three remaining follow-up coverage sites. All required CI checks and three independent exact-head reviews pass. Please publish an explicit APPROVED or current-head REQUEST_CHANGES verdict.

sergiobuilds:
Superseded by #2341, which now includes the repaired stop-reason classification and degraded-terminal regression coverage and has been merged after current-head approval and green CI. Closing this duplicate maintainer repair PR.

PR #2347, opened 2026-09-04
Add knos to .mcp.json so agents share one decision record

drexthealpha:
Q00/ouroboros runs more than one coding agent over the same tree. Agents do not
see each other's decisions, so two of them can reach opposite conclusions about
the same file in the same hour, and neither one knows.

**What this adds**

- `knos` in added to the existing `.mcp.json` - a local MCP server (`pip install knos`, Python 3.10+).
  No network, no account, no model download; one SQLite file.
- `examples/knos-decisions.example.md` - the shape of the record it writes,
  seeded with two rules taken from this repo's own CLAUDE.md, so the format is
  visible without installing anything.

**What it does**

When one agent claims a piece of work, another agent asking about that topic is
told who holds it instead of being handed the answer. Claims lapse after 30
minutes, so a crashed agent cannot hold work forever.

**Runtime note:** the server needs Python 3.10+ and `pip install knos`. Nothing
else in the repo depends on it - if it is absent, the example file still reads
normally and no other tooling changes.

Disclosure: I wrote knos. Happy to drop either half, or close this, if it is not
a fit.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -16,6 +16,13 @@
         "--llm-backend",
         "claude_code"
       ]
+    },
+    "knos": {
+      "command": "python",
+      "args": [
+        "-m",
+        "knos.mcp"
+      ]
     }
   }
 }
--- a/examples/knos-decisions.example.md
+++ b/examples/knos-decisions.example.md
@@ -5,8 +5,8 @@
 
 ## Decisions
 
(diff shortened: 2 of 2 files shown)


drexthealpha:
Closing this. Your PR Hygiene gate requires every PR to reference an issue, and after two passes over the open issues I could not find one this honestly answers - the MCP issues here are about your own server's launcher and lifecycle, not about sharing decisions between agents. Rather than invent a link or open an issue just to satisfy the gate, I'll withdraw. Sorry for the noise.

PR #2336, opened 2026-09-01
feat(lateral): grounded decision advisory + interview-less seed — RFC P1–P4

Q00:
Refs #2335

## One user problem

38% of MCP-attached users ever make a first tool call, and the two features that need no upfront spec — lateral fan-out and seed generation — either trigger too rarely (stagnation only), produce ungrounded opinion piles, or demand a heavy interview. This PR ships the grounded-lateral RFC (P1–P4) plus the reference-contrast deadlock fix from the 0.52.0 field incident.

## Supported inputs and execution conditions

- `ouroboros_lateral_think`: new optional `mode` (`unstuck` default | `decision`) and `research` (bool). Omitted args are byte-compatible with today's behavior on every dispatch path (plugin-passive / host-driven / sequential-inline).
- `ouroboros_generate_seed`: new optional `session_context` object (goal, acceptance_criteria, constraints, decisions, project_type) accepted **instead of** `session_id`; the interview path is untouched. No LLM backend is required or used on this path.
- `ouroboros_record_conductor_decision`: new `actor_mode="advisory"` value for read-only decision records; existing actor modes and all mutating-governance gates unchanged.
- Citation audit runs only when a lateral persona-panel fan-out submission actually contains fenced evidence blocks with URLs; otherwise no network is touched.

## Observable contract and invariants

- `mode=decision` with no persona selection fans out all 5 personas and stamps a synthesis contract (ONE recommendation, grounds, dissent, flip conditions) plus a `record_decision` template keyed by fanout_id.
- Deep tier is opt-in only; runtimes without web tools degrade to opinion-only — never an error. Citation audit is withhold-only: dead links are marked (`unreachable`/`invalid`/`unchecked`), never a synthesis failure. Bounded per-request (4s) and total (10s) budgets.
- `session_context` seeds are fully deterministic: settled material enters the Seed verbatim (strip-only), settled decisions become constraints, missing goal/criteria return `gap_questions_required` (never blocked/force), and metadata records the 0.2 ambiguity ceiling unconditionally (a caller-claimed score would be the #210 gate override).
- Verbatim anchor backstop at the SeedGenerator chokepoint: promoted USER-confirmed constraints/ACs missing from the LLM extraction (normalized equality, no similarity scoring) are appended verbatim — add-only, extraction stays the composer.
[...]

PR #2353, opened 2026-09-07
fix(lateral): handle malformed citation URLs safely

Q00:
Refs #2336

## User problem
Deep-tier lateral citation auditing could raise on malformed URLs instead of returning a withhold-only verdict.

## Promised contract
Malformed or non-http citation URLs are classified as `invalid`; citation auditing never raises and continues auditing other URLs.

## Implementation boundary
`src/ouroboros/mcp/tools/citation_check.py` and its unit tests. No network or synthesis behavior changes beyond malformed URL handling.

## Evidence
`uv run pytest -q tests/unit/mcp/tools/test_citation_check.py` and the lateral/verify target suite pass.


Q00:
Superseded by citation-only PR.

PR #2354, opened 2026-09-07
fix(verify): reject same-content verifier writes

Q00:
Refs #2336

## User problem
A deferred verify command could rewrite a file with identical content during settlement and remain accepted because content digests were unchanged.

## Promised contract
A verifier is observational. Any workspace write during settlement replay, including same-content rewrites, invalidates the verification result.

## Implementation boundary
`src/ouroboros/orchestrator/parallel_executor.py`; settlement replay compares workspace file metadata before and after the replay. No change to the configured verification command or retry policy.

## Evidence
`uv run pytest -q tests/unit/orchestrator/test_parallel_executor_verify_by_default.py` and the lateral/verify target suite pass.

### DIFF
--- a/src/ouroboros/mcp/tools/citation_check.py
+++ b/src/ouroboros/mcp/tools/citation_check.py
@@ -130,8 +130,12 @@ def audit_citations(
     deadline = time.monotonic() + total_budget_seconds
     checked = 0
     for url in urls:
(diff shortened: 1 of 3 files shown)


Q00:
Superseded by verify-only PR.

PR #2355, opened 2026-09-07
fix(lateral): preserve decision fanout identity and research contract

Q00:
Refs #2336

## User problem
Decision-mode lateral results lost re-entry identity on the OpenCode plugin path, and `research=true` was omitted from single-persona inline responses.

## Promised contract
Plugin decision fan-outs register and return a `fanout_id` for later synthesis and decision recording. Single-persona deep-tier responses include the evidence contract and expose the research mode.

## Implementation boundary
`lateral_think_handler.py` and its unit tests. Existing unstuck behavior and the citation audit remain unchanged.

## Evidence
`uv run pytest -q tests/unit/mcp/tools/test_lateral_think_handler.py` and the lateral/verify target suite pass.


Q00:
Superseded by separate PRs for fanout identity and single-persona research.

PR #2360, opened 2026-09-07
fix(verify): reject same-content verifier writes

Q00:
Refs #2336

## User problem
A deferred verify command could rewrite a file with identical content during settlement and remain accepted because content digests were unchanged.

## Promised contract
A verifier is observational. Any workspace write during settlement replay, including same-content rewrites, invalidates the verification result.

## Implementation boundary
`src/ouroboros/orchestrator/parallel_executor.py`; settlement replay compares workspace file metadata before and after replay.

## Evidence
`uv run pytest -q tests/unit/orchestrator/test_parallel_executor_verify_by_default.py` passes.

### DIFF
--- a/src/ouroboros/orchestrator/parallel_executor.py
+++ b/src/ouroboros/orchestrator/parallel_executor.py
@@ -5287,6 +5287,21 @@ class ParallelACExecutor:
             return results
 
         cwd = self._task_cwd or self._adapter.working_directory or os.getcwd()
+
+        def workspace_metadata_fingerprint() -> tuple[tuple[str, int, int], ...] | None:
+            """Capture write metadata so a same-content verifier write is visible."""
+            try:
+                entries: list[tuple[str, int, int]] = []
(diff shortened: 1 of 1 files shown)


Q00 (commented):
Reviewed the same-content verifier-write detection. The metadata fingerprint is compared before and after settlement replay and conservatively invalidates results when metadata cannot be read. The current blocker is PR hygiene: `Refs #2336` points to a merged pull request rather than an issue, so the Issue link check fails; please link a real issue or apply the repository-approved no-issue label.

Q00 (commented):
Reviewed the same-content verifier-write detection. The metadata fingerprint is compared before and after settlement replay and conservatively invalidates results when metadata cannot be read. The current blocker is PR hygiene: Refs #2336 points to a merged pull request rather than an issue, so the Issue link check fails; please link a real issue or apply the repository-approved no-issue label.

PR #2361, opened 2026-09-07
test: share packaging fixtures and classify integration tests

Q00:
Refs #2336

## User problem
Packaging tests rebuilt the same wheel independently and the test suite did not classify integration tests consistently. PR checks also ran artifact-building `slow` tests by default.

## Promised contract
Packaging assertions inspect one session-scoped wheel artifact, preserving all existing checks while removing duplicate builds. Integration tests are explicitly marked, and PR checks skip only `slow` tests; pushes retain the full non-performance suite.

## Implementation boundary
Pytest fixtures and markers, wheel packaging/profile tests, and the test workflow. No production code or test assertions are removed.

## Evidence
- Packaging/profile target: 44 passed.
- `ruff check` clean.
- Collection with the PR expression: 22,146 selected / 8 slow tests deselected.
- The packaging target fell from 29.01s to 24.90s locally.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,7 +112,13 @@ jobs:
           if [[ "${{ github.event_name }}" != "pull_request" && "${{ matrix.python-version }}" == "3.12" ]]; then
             COV_ARGS="--cov=src/ouroboros --cov-report=xml"
           fi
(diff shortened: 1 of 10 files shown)


Q00 (commented):
Review findings: the implementation direction is reasonable, but the required Ruff format check currently fails in tests/conftest.py, tests/integration/plugin/test_wheel_packaging.py, and tests/unit/orchestrator/test_profile_loader.py. Please run ruff format on the changed files and push the formatting fix before merge.

Issue #2370, opened 2026-09-07
bug(evaluation): Windows mechanical.toml parsing retains grouping quotes


zittenyetten:
## Problem

On native Windows, a valid mechanical command such as `pytest -k "slow"` produces `('pytest', '-k', '"slow"')` instead of `('pytest', '-k', 'slow')`. The grouping quotes become part of the argument sent to the test runner. Quoted multi-word filters and quoted flag values are affected too.

## Environment

- Native Windows, PowerShell host; reproduced on Python 3.13.3.
- Source checkout of upstream `1fc754e7b7599b3df057660bcc6af37241183766` (v0.54.0), not a claim about the locally installed package.

## Reproduction

With the repository's test dependencies installed, run the existing regression:

```text
python -m pytest tests/unit/evaluation/test_detector.py::TestTomlSerialization::test_commands_with_quotes_roundtrip -q
```

It fails on Windows with:

```text
assert ('pytest', '-k', '"slow"') == ('pytest', '-k', 'slow')
At index 2 diff: '"slow"' != 'slow'
```

Equivalent project configuration (with pytest declared in pyproject.toml):

```toml
# .ouroboros/mechanical.toml
test = 'pytest -k "slow"'
```

Expected: grouping quotes are removed, each quoted group remains one argv element, and Windows path backslashes are preserved.

## Diagnosis and bounded fix

Both the reader (`evaluation/languages.py`) and detector validation (`evaluation/detector.py`) call `shlex.split(..., posix=False)` on Windows. This mode preserves surrounding quotes. Changing only the reader would also make validation and execution interpret different arguments.

Use one shared evaluation tokenizer for both paths. Preserve POSIX behavior, Windows path backslashes, and existing executable/entry-point/path/operator checks. Do not add shell execution, broaden the command allowlist, or change the policy for rejected commands. Ambiguous backslash-adjacent quote syntax should be rejected explicitly and documented, rather than silently changed.

This is separate from the Windows command dispatch and diagnostic retention work in #2368 / #2369.

## Validation of the proposed fix

On Python 3.12.13 and 3.13.3, the focused reader/detector/tokenizer suite passes 297 tests, with two existing environment-dependent skips and no new skips. This includes a native TOML-to-child-pytest check using a Korean/spaced path and `-k "slow or fast"`, plus validation regressions.

sergiobuilds:
[...]

PR #2369, opened 2026-09-07
fix(evaluation): resolve Windows commands and retain Stage 1 diagnostics

zittenyetten:
## Summary

- Resolve validated Windows bare commands at the mechanical dispatch boundary so PATH-installed `npm.CMD` can execute; do not rewrite project configuration or enable `shell=True`.
- Reject unsafe CMD/BAT argument/path forms before launch and retain both the configured command and the actual argument tuple passed to the launcher.
- Preserve shared Stage 1 diagnostics in multi-AC responses and persisted job metadata, with 500-character stdout/stderr tails and no changes to approval logic.

## Review boundary

- **User problem:** A Windows npm launch failure is surfaced as a generic build/test failure, and multi-AC job results lose the detail needed to diagnose it.
- **Inputs and execution conditions:** Existing validated command tuples and a Windows PATH-installed executable, tested on Python 3.12.13/3.13.3 with npm 10.9.2. New lookup uses fully qualified candidates from absolute PATH entries; it does not prepend the current directory. Missing PATH matches retain the previous native-spawn behavior. Native/POSIX tuples and explicit native paths are not rewritten. Ordinary spaces, Unicode and empty batch arguments are covered.
- **Promised contract:** Actually execute the configured supported command or report failure. CMD/BAT paths/arguments containing `& | < > ^ % ! " ( )`, control characters, or space-containing values ending in a backslash are rejected explicitly before launch. No shell fallback or silent skip/pass is introduced. Keep cwd, environment cleanup, timeout/cancellation and exit-code behavior. Existing direct-child cleanup is preserved; no process-tree containment guarantee is added. Retain requested `command`, attempted `executed_command`, cwd and available failure details. Shared Stage 1 runs once, regardless of AC count; the same details survive a real persisted job roundtrip without changing verdicts.
- **Implementation boundary and owner:** Existing `evaluation` subprocess boundary and MCP evaluation presentation, under their existing maintainers; contributor responsible for this patch: @zittenyetten. Two small helpers isolate dispatch and report formatting. The oversized evaluation handler shrinks by 21 lines. No new service, DB schema, configuration surface or runtime backend.
- **Non-goals:** General TOML invalid-config-to-None/skip policy, transcript test-command/output classification, generated-artifact evidence, AC verify-gate redesign, game/Seed changes, global environment or installation changes, Linux/macOS execution claims, and unrelated upstream test/type defects. This PR does not make a failed application execution approved.
[...]

PR #2371, opened 2026-09-07
fix(evaluation): preserve Windows mechanical command arguments

zittenyetten:
## Summary

Windows `shlex.split(posix=False)` retains grouping quotes in mechanical argv. Share a small tokenizer between detector validation and the config reader so quoted filters/flags work without stripping Windows path backslashes.

## Review boundary

- **User problem:** A valid `pytest -k "slow"` command reaches Stage 1 with literal grouping quotes in its filter argument on native Windows.
- **Inputs and execution conditions:** String commands after TOML decoding (including detector-authored commands and the reader's Python overrides). Windows supports single/double quote grouping, embedded quoted flag values, empty quoted arguments, and literal path backslashes. POSIX keeps the existing `shlex.split` behavior.
- **Promised contract:** Detector and reader split into identical argv. Existing allowlisting, operator rejection, repository entry-point validation, argument containment, and malformed-command skip behavior remain in place. Quoting cannot newly admit a Windows drive/root/UNC executable head; both paths reject those heads after tokenization.
- **Implementation boundary and owner:** Evaluation configuration parsing and its existing detector validator. One 26-line shared helper, two call sites, focused tests, and a guide update. No new subsystem or execution authority; the oversized detector module shrinks by one line.
- **Non-goals:** Full cmd.exe/PowerShell parsing; shell invocation; additional allowed executables; changed skip policy; subprocess dispatch, evidence classification, or diagnostic aggregation changes. Backslash immediately before either quote character is explicitly rejected on Windows as ambiguous; for a directory path, omit the trailing separator or use forward slashes. The separate #2369 work is not included or required.
- **Evidence:** The existing quote-roundtrip test fails on clean upstream `1fc754e7b7599b3df057660bcc6af37241183766` and passes with this change. New tests exercise both tokenizer modes and a native TOML-to-child-pytest run selecting two tests from a Korean/spaced path. The child uses the current interpreter explicitly, so this is argv verification, not a PATH-dispatch test.

## Test plan

- [x] Native Windows Python 3.12.13 and 3.13.3, each: `python -m pytest tests/unit/evaluation/test_command_parsing.py tests/unit/evaluation/test_languages.py tests/unit/evaluation/test_detector.py -q -rs` — **297 passed, 2 existing skips** (POSIX-only rejection / symlink unavailable).
- [x] `python -m ruff check src/ tests/` and `python -m ruff format --check src/ tests/` — pass (1432 files formatted).
- [x] MyPy on the three changed production modules — pass.
- [x] `git diff --check` — pass.
- [ ] Full evaluation unit suite on Windows Python 3.13: `python -m pytest tests/unit/evaluation -q -rs` — **692 passed, 2 existing skips, 1 existing failure**. `TestRunCommand.test_successful_command` invokes `echo` as an executable, which is unavailable here. That same failure was reproduced on the clean base, alongside the quote-roundtrip failure fixed by this PR. No skip or dispatch changes are included here.
- [ ] Full repository tests, native Linux/macOS execution, and Bridge TypeScript were not run locally. Platform-mode unit tests are not a native Linux run.
- [ ] Full Windows MyPy: existing `hermes/artifacts.py:205` undefined `result` diagnostic; reproduced on the clean base with a focused check. Not fixed here.

Validation used disposable Windows profile directories, disabled telemetry, and an explicit source-tree PYTHONPATH; the user's installed runtime was not changed. The patch was prepared with coding-agent assistance and independently reviewed against this boundary; that review was not a separate full test execution.

## Related issues

Refs #2370


Q00 (approved):
Reviewed the shared Windows command tokenizer and validation integration. Quoting and path handling are covered by focused tests; CI checks pass.

Issue #2372, opened 2026-09-07
bug(mcp/security): interview-less generate_seed rejects shell chains in session_context acceptance_criteria


Q00:
## Problem

`ouroboros_generate_seed(session_context=...)` (interview-less path, #2336, shipped in v0.54.0) fails with

```
Shell metacharacter detected in session_context.acceptance_criteria[0] details={'pattern': '&&'}
Shell metacharacter detected in session_context.decisions[0] details={'pattern': ';'}
```

whenever an acceptance criterion is a shell chain (`uv run ruff check src/ && uv run pytest`) or a decision is a sentence with a semicolon. The tool description asks for "verifiable checks (a command...)", so the gate rejects the exact input the tool requests.

## Evidence

- Reproduced live against the 0.54.0 plugin server (both payloads above).
- PostHog (project 544718): `command_run` with `command='seed'` and `error_type='MCPServerError'` is 0 on 0.52.0 and 0.53.0 (their seed failures are all `MCPToolError`) and 7 events / 4 users on 0.54.0 within 16h of release — 3 of those users are external.

## Cause

`InputValidator.FREETEXT_FIELDS` in `src/ouroboros/mcp/server/security.py` exempts `goal` and `acceptance_criterion` (singular) but not `acceptance_criteria`, `constraints`, or `decisions`, so the new container's leaves are scanned. The identical text passes when sent as `seed_content` to `ouroboros_start_execute_seed`, which is already exempt — the check adds friction with no security value on this path (no leaf of `session_context` reaches a shell; it is verbatim Seed prose).

## Fix

Add `session_context` as a freetext container (per-segment match already exists). Sibling arguments such as `session_id` stay scanned. Document the interview-less path and its gap-question loop in `skills/seed/SKILL.md`.

PR #2373, opened 2026-09-07
fix(mcp/security): interview-less generate_seed accepts shell chains in session_context

Q00:
Refs #2372

## The one user problem
A user who settled goal + acceptance criteria in conversation calls `ouroboros_generate_seed(session_context=...)` (the interview-less path shipped in v0.54.0) and gets `Shell metacharacter detected in session_context.acceptance_criteria[0]` because their verifiable AC is a shell chain (`uv run ruff check src/ && uv run pytest`), or `...decisions[0]` because a sentence has a semicolon. The tool asks for exactly this input and then rejects it.

## Supported inputs and execution conditions
- `ouroboros_generate_seed` with a `session_context` object whose leaves are `goal`, `acceptance_criteria[]`, `constraints[]`, `decisions[]`, `project_type`.
- Any transport (stdio / plugin); the change is in `InputValidator`, before any handler runs.

## Observable contract and invariants
- Every leaf under `session_context` is exempt from the shell-metacharacter / dangerous-pattern scan, exactly as `seed_content` and `goal` already are. Rationale: the leaves are verbatim Seed prose; no code path passes them to a shell (`bigbang/session_seed.py` copies strings into the Seed model only).
- Invariant kept: sibling arguments of the same call (`session_id`, etc.) are still scanned. Test: `test_validate_still_checks_siblings_of_session_context`.
- Invariant kept: the exemption is per key segment, so a non-freetext field elsewhere is unaffected (existing `TestFreetextInsideContainers` suite, 31 tests pass).

- `src/ouroboros/mcp/server/security.py` — one entry added to `FREETEXT_FIELDS` (MCP input validation, owner: maintainer).
- `skills/seed/SKILL.md` — documents the interview-less path and its `gap_questions_required` loop so the host actually takes it instead of sending the user back to `ooo interview`.
- `CHANGELOG.md` — Fixed entry.
- No new subsystem, no data/security boundary change: the same bytes were already admitted through `seed_content`.

## Non-goals
- No change to the interview path, the ambiguity gate, or seed preflight.
- No telemetry property for interview-less adoption (closed-vocabulary contract; separate decision).
- No change to `.claude-plugin/skills/` (dead copy).

## Evidence
- Reproduced live against the 0.54.0 plugin server: `&&` in `acceptance_criteria[0]` and `;` in `decisions[0]` both rejected; the same content without those characters yields `seed_b4f2838692cd`.
- PostHog project 544718: `command_run{command=seed, error_type=MCPServerError}` = 0 on 0.52.0 and 0.53.0 (their seed failures are all `MCPToolError`), 7 events / 4 users on 0.54.0 within 16h of release, 3 users external.
- Local: `ruff format/check` clean, `mypy` clean (632 files), `tests/unit/mcp/server` + `tests/unit/bigbang` + skill-doc tests: 1673 + 59 passed.

https://claude.ai/code/session_01NzJNXSia9xSj1Ghu49o2jK


Q00:
CI is green after three commits:

1. `22ca95f` — the fix: `session_context` added to `InputValidator.FREETEXT_FIELDS` (per-segment match, so only that container's leaves are exempt; `session_id` and other siblings stay scanned — covered by `test_validate_still_checks_siblings_of_session_context`). Seed skill documents the interview-less path + gap-question loop.
2. `b07732c` / `f8ec468` — test-table ripple only. The capability registry derives each tool's `required_context_keys` from the packaged skill's declared `Arguments:` blocks, so declaring `session_context` in the seed skill is now pinned in `test_skill_tool_mapping` and `test_capabilities` (`("session_id", "session_context")`). No registry or handler code changed.

Verified end to end over stdio against local source: an AC containing `&&` yields a Seed; `session_id="s1; rm -rf ..."` in the same call is still rejected.

PR #2392, opened 2026-09-12
fix: avoid PowerShell error for Microsoft Store Python alias

hackertaco:
Refs #2393

Skip Microsoft Store python.exe/python3.exe execution aliases under WindowsApps before invoking them, preventing the misleading NativeCommandError in Windows PowerShell 5.1. The existing uv-managed Python fallback remains unchanged.


Q00 (approved):
Reviewed: the WindowsApps Microsoft Store alias guard prevents the misleading PowerShell NativeCommandError while preserving the uv fallback.

